PR #32, opened 2022-12-16
dynamically allocate threads

valeriyprovalov:
Implement dynamic thread allocation similar to libFuse.
---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


jamesbornholt:
We ended up pulling the multi-threaded FUSE support into mountpoint rather than fuser (#135), so this is stale now. We might revisit in the future, but closing for now.

Issue #153, opened 2023-03-15
Mounted filesystem is read-only


jht-prog:
### Mountpoint for Amazon S3 version

mountpoint-s3 0.2.0-b8363a4

### AWS Region

us-east-1

### Describe the running environment

Running Ubuntu 20.04 on the headnode of a pcluster with fullaccesss3 privileges.

### What happened?

Was able to build client and mount s3 bucket but the file system is write only, reads contents just fine. Probably doing something simple wrong. Using file-mode doesn't seem to fix it.

dannycjones:
Thanks for trying out the project!

Mountpoint is currently an alpha release, and does not yet support read-write access. We are already working on sequential write access to files, tracked in #27. We have also shared a number of other features we hope to add in the [Mountpoint for Amazon S3 public roadmap](https://github.com/orgs/awslabs/projects/84).

jht-prog:
Thanks for the quick feedback!

Hoot

PR #155, opened 2023-03-15
README tweaks to highlight common questions earlier

### LINKED ISSUE #151: Add option --profile for a specific profile in the AWS credentials file
### Tell us more about this new feature.

It would be nice if we can have an option `--profile`, like in other AWS command-line tools, so that we can choose a specific profile to use from ~/.aws/credentials.  Right now it seems the only way to select a different profile is via environment variables, which is also not documented. 

Current behavior

```
./target/release/mount-s3 --profile
error: Found argument '--profile' which wasn't expected, or isn't valid in this context

        If you tried to supply `--profile` as a value rather than a flag, use `-- --profile`
```

jamesbornholt:
Fixes #151 and #153.

---


dannycjones (approved):
LGTM!

monthonk (approved):

PR #167, opened 2023-03-17
Split integration tests into a new CI workflow

### LINKED ISSUE #143: Split out CI tests that require S3 credentials
Right now our CI workflow assumes you have S3 credentials, but that won't work for PRs submitted from a public fork. We should split the tests up (off the `s3_tests` feature flag) so that forks don't try to run the S3 tests, and figure out a way to manually trigger those tests when we need them (see https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).

jamesbornholt:
Our integration tests are tricky to run because they require S3 credentials, which GitHub (quite rightly) makes difficult to access through a pull request from a fork repo. Even if we turn on GitHub's feature for requiring manual approval for a PR to run workflows, it still won't have access to the permissions it needs to get S3 credentials, because the `configure-aws-credentials` action needs `id_token: write` permissions, which fork PRs can never have.

This change makes our integration tests run using the `pull_request_target` event. This event runs in the context of the base repo rather than the PR, and so has access to the right permissions. That's dangerous, of course, so we want to manually approve runs of this workflow. But `pull_request_target` workflows are _automatically approved_, so the normal "require manual approval for a PR to run workflows" no longer works. Instead, we rig something up using deployment environments, which can be configured to require manual approval before deployments. The integration tests ask to be "deployed" to an environment with approval configured, and so won't run until someone with permission to approve those deployments clicks OK. Same effect as the normal PR approval mechanism, just a little more jury-rigged and with some confusing verbiage around "deployment".

Pushes to `main` also run the integration tests, but they ask for a null deployment, and so run as normal workflows. This is safe because pushes to `main` already require PR approvals.

Workflows on `pull_request_target` have a different security boundary [[1]], and in particular have two permissions normal `pull_request` workflows don't have:
1. Their GITHUB_TOKEN has write access to the repository. We explicitly downgrade the token to have only read access using the `permissions` field in integration test workflows.
2. They run in the context of the base repository, and so have access to secrets. We don't use any secrets. We also explicitly prevent the integration tests from storing cached artifacts if running from a PR, which could poison other users of the cache.

Since they run in the base context, `pull_request_target` workflows will always run with the workflow files from the base repo, ignoring any changes in the pull request. That means a PR can't change the workflow to undo these protections.

Fixes #143.

[1]: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

---

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,141 @@
+name: Integration tests
+
+# We use environments to require approval to run integration tests on PRs, but not on pushes to
+# `main` (which have been approved already since PRs are required for `main`).
+on:
+  workflow_call:
+    inputs:
+      environment:
+        type: string
+      ref:
+        required: true
+        type: string
+
+env:
+  RUST_BACKTRACE: 1
+  CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: 0
+  S3_BUCKET_NAME: s3-file-connector-github-test-bucket
+  S3_BUCKET_TEST_PREFIX: github-actions-tmp/run-${{ github.run_id }}/attempt-${{ github.run_attempt }}/
+  # A bucket our IAM role has no access to, but is in the right region, for permissions tests
+  S3_FORBIDDEN_BUCKET_NAME: s3-file-connector-github-test-bucket-forbidden
+  S3_REGION: us-east-1
+  RUST_FEATURES: fuse_tests,s3_tests
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  test:
+    name: Tests
+    runs-on: ubuntu-22.04
+
+    environment: ${{ inputs.environment }}
+
+    strategy:
+      matrix:
+        fuse: [fuse, fuse3]
+
+    steps:
+    - name: Configure AWS credentials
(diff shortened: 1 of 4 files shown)


jamesbornholt:
Workflow changes are a little annoying to test, but to show what's going on, I pushed this change to my fork.
* Commit to `main`: https://github.com/jamesbornholt/mountpoint-s3/commit/14c03f8e9625e1148e73144b49d4f07a873a693e -- all actions ran without any manual intervention (except benchmarks since I don't have the larger runner on my account)
* PR into `main` from the same repo: https://github.com/jamesbornholt/mountpoint-s3/pull/5 -- the non-integration tests ran without any intervention, the integration tests are stuck "waiting" until I click the "Approve deployment" button
* PR into `main` from a fork: https://github.com/jamesbornholt/mountpoint-s3/pull/6 -- no tests ran until I clicked the "Approve and run" button on the PR. After that, the integration tests are still stuck waiting to deploy, like the same-repo case.

monthonk (approved):
LGTM! it's definitely not a straightforward problem and also hard to test, thanks for figuring this out.

I think we also want to do the same with the benchmark CI, but that can be a different PR.

PR #171, opened 2023-03-20
Arrange CLI arguments into groups

jamesbornholt:
Our arguments are getting pretty unwieldly. This change updates to clap
v4 and then adds help headings so that `-h` is easier to parse.

With this change, the output looks like this:

```
[18:21:41] ~/c/mountpoint-s3 (clap-headings|✔) $ target/debug/mount-s3 -h
Mountpoint for Amazon S3

Usage: mount-s3 [OPTIONS] <BUCKET_NAME> <MOUNT_POINT>

Arguments:
  <BUCKET_NAME>  Name of bucket to mount
  <MOUNT_POINT>  Mount point for file system

Options:
  -l, --log-directory <LOG_DIRECTORY>  Log file directory [default: $HOME/.mountpoint-s3]
  -f, --foreground                     Run as foreground process
  -h, --help                           Print help
  -V, --version                        Print version

Bucket options:
      --prefix <PREFIX>              Prefix inside the bucket to mount [default: mount the entire bucket]
      --region <REGION>              AWS region of the bucket [default: auto-detect region]
      --endpoint-url <ENDPOINT_URL>  S3 endpoint URL [default: auto-detect endpoint]
      --virtual-addressing           Force virtual-host-style addressing
      --path-addressing              Force path-style addressing

Mount options:
      --auto-unmount           Automatically unmount on exit
      --allow-root             Allow root user to access file system
      --allow-other            Allow other non-root users to access file system
      --uid <UID>              Owner UID [default: current user's UID]
      --gid <GID>              Owner GID [default: current user's GID]
      --dir-mode <DIR_MODE>    Directory permissions [default: 0755]
      --file-mode <FILE_MODE>  File permissions [default: 0644]

Client options:
      --throughput-target-gbps <N>  Desired throughput in Gbps [default: 10]
      --thread-count <N>            Number of FUSE daemon threads [default: 1]
      --part-size <PART_SIZE>       Part size for multi-part GET and PUT [default: 8388608]
```

---

### DIFF
--- a/mountpoint-s3/src/main.rs
+++ b/mountpoint-s3/src/main.rs
@@ -5,7 +5,7 @@ use std::time::Duration;
 use std::{fs, fs::File};
 
 use anyhow::{anyhow, Context as _};
-use clap::{ArgGroup, Parser};
+use clap::{value_parser, ArgGroup, Parser};
 use fuser::{MountOption, Session};
 use mountpoint_s3::fs::S3FilesystemConfig;
 use mountpoint_s3::fuse::session::FuseSession;
@@ -78,7 +78,7 @@ fn init_tracing_subscriber(is_foreground: bool, log_directory: Option<&Path>) ->
 
 #[derive(Parser)]
 #[clap(about = "Mountpoint for Amazon S3", version = build_info::FULL_VERSION)]
-#[clap(group(ArgGroup::new("addressing-style").args(&["virtual-addressing", "path-addressing"])))]
+#[clap(group(ArgGroup::new("addressing-style").args(&["virtual_addressing", "path_addressing"])))]
 struct CliArgs {
     #[clap(help = "Name of bucket to mount")]
     pub bucket_name: String,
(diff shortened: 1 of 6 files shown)


mountpoint-s3/src/main.rs
```diff
@@ -78,63 +78,76 @@ fn init_tracing_subscriber(is_foreground: bool, log_directory: Option<&Path>) ->
...
+    #[clap(
+        long,
+        help = "Allow other non-root users to access file system",
+        help_heading = "Mount options"
+    )]
     pub allow_other: bool,
 
-    #[clap(long, help = "Desired throughput in Gbps", value_name = "N (Gbps)", value_parser = clap::value_parser!(u64).range(1..))]
+    #[clap(long, help = "Desired throughput in Gbps", value_name = "N", default_value = "10", value_parser = value_parser!(u64).range(1..), help_heading = "Client options")]
     pub throughput_target_gbps: Option<u64>,
```
dannycjones, line 133: If we have default values for these, should they still be `Option<>`? Or instead just the `u64`, etc.?

dannycjones (changes requested):
Two minor comments

mountpoint-s3/src/main.rs
jamesbornholt: I think `Option` is a little friendlier to someone trying to use `CliArgs` programmatically.

dannycjones (approved):
LGTM, nice improvement

Issue #170, opened 2023-03-20
Upgrade CI usage of aws-actions/configure-aws-credentials from v1 to v2


dannycjones:
The GitHub action for configuring AWS credentials was recently bumped from V1 to V2. We should update our CI to use V2. Updates will only be published to V2 in the future. https://github.com/aws-actions/configure-aws-credentials/blob/cb1361ade3c0ad3551e7d2be86bf97afcb51e4d6/README.md#recent-updates

There are no expected changes, the action is just updated to use a new version of Node LTS.

PR #172, opened 2023-03-20
Update benchmark CI to use `pull_request_target`

jamesbornholt:
This is the same as #167 but for the benchmark workflow, which also needs S3 permissions. I also took this chance to switch to v2 of the credentials actions (#170) and remove a redundant Rust installation since the toolchain action does that for us already.

---


monthonk (approved):
Looks good to me. I guess we can't test it until it's been merged to main because of the nature of `pull_request_target` event?

PR #177, opened 2023-03-22
Fix time dependency in mountpoint-s3-client

jamesbornholt:
Just a tiny thing: this crate's use of `time` requires some features.
This isn't a problem for us because mountpoint-s3 pulls those features
in, so compilation still works, but it's a problem if someone wants to
take a direct dependency on mountpoint-s3-client. Also, there's no need
for the direct dependency on the -sys crate here.

---

### DIFF
--- a/mountpoint-s3-client/Cargo.toml
+++ b/mountpoint-s3-client/Cargo.toml
@@ -7,7 +7,6 @@ publish = false
 
 [dependencies]
 mountpoint-s3-crt = { path = "../mountpoint-s3-crt" }
-mountpoint-s3-crt-sys = { path = "../mountpoint-s3-crt-sys" }
 
 async-trait = "0.1.57"
 auto_impl = "1.0.1"
@@ -21,7 +20,7 @@ pin-project = "1.0.12"
 regex = "1.7.1"
 static_assertions = "1.1.0"
 thiserror = "1.0.34"
-time = "0.3.14"
+time = { version = "0.3.17", features = ["formatting", "parsing"] }
 tracing = { version = "0.1.35", default-features = false, features = ["std", "log"] }
 xmltree = "0.10.3"
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1868,7 +1868,6 @@ dependencies = [
  "libc-stdhandle",
  "metrics",
  "mountpoint-s3-crt",
- "mountpoint-s3-crt-sys",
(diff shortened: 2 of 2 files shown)


jamesbornholt:
Yep!

mountpoint-s3-client/Cargo.toml
```diff
@@ -7,7 +7,6 @@ publish = false
 
 [dependencies]
 mountpoint-s3-crt = { path = "../mountpoint-s3-crt" }
-mountpoint-s3-crt-sys = { path = "../mountpoint-s3-crt-sys" }
```
sauraank, line 10: Why is mountpoint-s3-crt-sys removed as a dependency for s3-client. Is it because it only depends directly on mountpoint-s3-crt? Or there is some other reason?

dannycjones: Yeah, there is no reason for the `mountpoint-s3-client` to depend on the `mountpoint-s3-crt-sys` crate, which is providing the C/Rust bindings to the CRT.

Instead, the `mountpoint-s3-client` should only care about the Rust-ified interface provided by `mountpoint-s3-crt`.

dannycjones:
>This crate's use of time requires some features.
This isn't a problem for us because mountpoint-s3 pulls those features in, so compilation still works, but it's a problem if someone wants to take a direct dependency on mountpoint-s3-client.

@jamesbornholt Can you give a rough example of what the failure looks like? Which API they couldn't use, for example?

passaro:
> @jamesbornholt Can you give a rough example of what the failure looks like? Which API they couldn't use, for example?

I just gave it a try and this is an example of the errors you would get before @jamesbornholt change:
```
error[E0599]: no function or associated item named `parse` found for struct `OffsetDateTime` in the current scope
   --> src/s3_crt_client/list_objects.rs:114:45
    |
114 |         let last_modified = OffsetDateTime::parse(&last_modified, &Rfc3339)
    |                                             ^^^^^ function or associated item not found in `OffsetDateTime`
```

FYI: one quick way to see all the failures is to temporarily remove the root Cargo.toml and then try to build just mountpoint-s3-client, e.g.:
```
rm Cargo.toml
cd mountpoint-s3-client
cargo build
```

dannycjones (approved):

dannycjones:
Thanks, makes sense.

PR #184, opened 2023-03-28
Introduce crt imds client.

eslrahc-swa:
The main goal for using crt imds client is to query ec2 instance type mountpoint-s3 running on.

---

### DIFF
--- a/mountpoint-s3-client/src/imds_crt_client.rs
+++ b/mountpoint-s3-client/src/imds_crt_client.rs
@@ -1,0 +1,98 @@
+use std::pin::Pin;
+use std::task::Context;
+use std::task::Poll;
+
+use futures::channel::oneshot;
+use futures::Future;
+use mountpoint_s3_crt::auth::imds_client::{ImdsClient, ImdsClientConfig};
+use mountpoint_s3_crt::common::allocator::Allocator;
+use mountpoint_s3_crt::common::error;
+use mountpoint_s3_crt::io::channel_bootstrap::{ClientBootstrap, ClientBootstrapOptions};
+use mountpoint_s3_crt::io::event_loop::EventLoopGroup;
+use mountpoint_s3_crt::io::host_resolver::{HostResolver, HostResolverDefaultOptions};
+use pin_project::pin_project;
+use thiserror::Error;
+
+#[derive(Debug)]
+/// Instance Metadata Service (IMDS) client responsible for sending EC2 instance metadata query requests.
+pub struct ImdsCrtClient {
+    imds_client: ImdsClient,
+    #[allow(unused)]
+    event_loop_group: EventLoopGroup,
+    #[allow(unused)]
+    allocator: Allocator,
+}
+
+impl ImdsCrtClient {
+    pub fn new() -> Result<Self, error::Error> {
+        let allocator = Allocator::default();
+
+        let mut event_loop_group = EventLoopGroup::new_default(&allocator, None, || {}).unwrap();
+
+        let resolver_options = HostResolverDefaultOptions {
+            max_entries: 8,
+            event_loop_group: &mut event_loop_group,
+        };
+
+        let mut host_resolver = HostResolver::new_default(&allocator, &resolver_options).unwrap();
+
+        let bootstrap_options = ClientBootstrapOptions {
+            event_loop_group: &mut event_loop_group,
+            host_resolver: &mut host_resolver,
+        };
+
+        let client_bootstrap = ClientBootstrap::new(&allocator, &bootstrap_options).unwrap();
+
+        let mut client_config = ImdsClientConfig::new();
+        client_config.client_bootstrap(client_bootstrap);
+
+        let imds_client = ImdsClient::new(&allocator, client_config)?;
+
+        Ok(Self {
+            imds_client,
+            event_loop_group,
+            allocator,
+        })
+    }
+
+    /// Query the type of the EC2 instance this code is executed on.
+    pub fn make_instance_type_query(&self) -> Result<ImdsQueryRequest, ImdsQueryRequestError> {
+        let (tx, rx) = oneshot::channel();
+        self.imds_client.get_instance_type(move |result| {
+            let _ = tx.send(result.map_err(ImdsQueryRequestError::CrtError));
+        })?;
+
+        Ok(ImdsQueryRequest { receiver: rx })
+    }
+}
+
+#[derive(Debug)]
+#[pin_project]
+/// ImdsQueryRequest is a data encapsulation wrapper for asnychronous instance metadata query.
+pub struct ImdsQueryRequest {
+    #[pin]
+    receiver: oneshot::Receiver<Result<String, ImdsQueryRequestError>>,
+}
+
+impl Future for ImdsQueryRequest {
+    type Output = Result<String, ImdsQueryRequestError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        this.receiver
+            .poll(cx)
+            .map(|result| result.unwrap_or_else(|err| Err(ImdsQueryRequestError::InternalError(Box::new(err)))))
+    }
+}
+
+/// ImdsQueryRequestError is returned by an asynchronous query.
+#[derive(Error, Debug)]
+pub enum ImdsQueryRequestError {
+    /// An internal error from within the Imds client. The request may have been sent.
+    #[error("Internal Imds client error")]
+    InternalError(#[source] Box<dyn std::error::Error + Send + Sync>),
+
+    /// An internal error from within the AWS Common Runtime. The request may have been sent.
+    #[error("Unknown CRT error")]
+    CrtError(#[from] error::Error),
+}
--- a/mountpoint-s3-crt/src/auth/imds_client.rs
+++ b/mountpoint-s3-crt/src/auth/imds_client.rs
@@ -1,0 +1,117 @@
+//! A client for retrieving ec2 instance metadata.
+
+use crate::auth::auth_library_init;
+use crate::common::allocator::Allocator;
+use crate::common::error::Error;
+use crate::io::channel_bootstrap::ClientBootstrap;
+use crate::io::retry_strategy::RetryStrategy;
+use crate::CrtError;
+use mountpoint_s3_crt_sys::{
+    aws_byte_buf, aws_imds_client, aws_imds_client_get_instance_type, aws_imds_client_new, aws_imds_client_options,
+    aws_imds_client_release,
+};
+use std::ptr::NonNull;
+
+/// A client for instance metadata query.
+#[derive(Debug)]
+pub struct ImdsClient {
+    /// A pointer to the underlying `aws_imds_client`
+    inner: NonNull<aws_imds_client>,
+}
+
+/// Configurations for creating a new [ImdsClient].
+#[derive(Debug, Default)]
+pub struct ImdsClientConfig {
+    /// The structure that can pass into the CRT's functions.
+    inner: aws_imds_client_options,
+
+    /// The [ClientBootstrap] to use to create connection to IMDS.
+    client_bootstrap: Option<ClientBootstrap>,
+
+    /// The [RetryStrategy] to use to reschedule failed requests to IMDS. This is reference counted,
+    /// so we only need to hold onto it until this [ImdsClientConfig] is consumed, at which point the
+    /// client will take ownership.
+    retry_strategy: Option<RetryStrategy>,
+}
+
+impl ImdsClientConfig {
+    /// Create a new [ImdsClientConfig] with default options.
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Client bootstrap used for common staples such as event loop group, host resolver, etc.
+    pub fn client_bootstrap(&mut self, client_bootstrap: ClientBootstrap) -> &mut Self {
+        self.inner.bootstrap = client_bootstrap.inner.as_ptr();
+        self.client_bootstrap = Some(client_bootstrap);
+        self
+    }
+
+    /// Retry strategy used to reschedule failed requests
+    pub fn retry_strategy(&mut self, retry_strategy: RetryStrategy) -> &mut Self {
+        self.inner.retry_strategy = retry_strategy.inner.as_ptr();
+        self.retry_strategy = Some(retry_strategy);
+        self
+    }
+}
+
+impl ImdsClient {
+    /// Create a new [ImdsClient]
+    pub fn new(allocator: &Allocator, config: ImdsClientConfig) -> Result<Self, Error> {
+        auth_library_init(allocator);
+
+        // SAFETY: `aws_imds_client_new` copies what it needs out of `config`, so it's safe to pass only
+        // a reference and to drop it after this method completes.
+        let inner = unsafe { aws_imds_client_new(allocator.inner.as_ptr(), &config.inner).ok_or_last_error()? };
+        Ok(Self { inner })
+    }
+
+    /// Get the EC2 instance type.
+    pub fn get_instance_type<F>(&self, callback: F) -> Result<(), Error>
+    where
+        F: FnOnce(Result<String, Error>) + 'static,
+    {
+        let callback_wrapper = Box::new(ImdsClientGetResourceCallback(Box::new(callback)));
+        let callback_raw_ptr = Box::into_raw(callback_wrapper) as *mut libc::c_void;
+        let get_resource_callback_fn_ptr: Option<unsafe extern "C" fn(*const aws_byte_buf, i32, *mut libc::c_void)> =
+            Some(imds_client_get_resource_callback);
+
+        // SAFETY: `self.inner` is a valid `aws_imds_client`. `get_resource_callback_fn_ptr` is leaked by [Box::into_raw]
+        // and so will live until the `imds_client_get_resource_callback` function is invoked.
+        unsafe {
+            aws_imds_client_get_instance_type(self.inner.as_ptr(), get_resource_callback_fn_ptr, callback_raw_ptr)
+                .ok_or_last_error()
+        }
+    }
+}
+
+impl Drop for ImdsClient {
+    fn drop(&mut self) {
+        // SAFETY: `self.inner` always points to a valid underlying `aws_imds_client`.
+        unsafe { aws_imds_client_release(self.inner.as_ptr()) };
+    }
+}
+
+type OnGetResource = Box<dyn FnOnce(Result<String, Error>)>;
+struct ImdsClientGetResourceCallback(OnGetResource);
+
+/// Rust binding for CRT's callback function `aws_imds_client_on_get_resource_callback_fn`.
+unsafe extern "C" fn imds_client_get_resource_callback(
+    resource: *const aws_byte_buf,
+    error_code: i32,
+    user_data: *mut libc::c_void,
+) {
+    // SAFETY: `user_data` is a raw pointer to a `Box<ImdsClientGetResourceCallback>` created and leaked at query time.
+    // This function will be executed at most once, so the Box is still valid right now.
+    let callback = Box::from_raw(user_data as *mut ImdsClientGetResourceCallback).0;
+
+    let result = if 0 != error_code {
+        Err(error_code.into())
+    } else {
+        // SAFETY: The CRT guarantees `resource` is a valid `*const aws_byte_buf` if the IMDS query succeeded.
+        let resource = std::slice::from_raw_parts((*resource).buffer, (*resource).len).to_vec();
+        Ok(String::from_utf8(resource).expect("resource response should be encoded with utf8."))
+    };
+
+    callback(result)
+}
--- a/mountpoint-s3/src/main.rs
+++ b/mountpoint-s3/src/main.rs
@@ -12,6 +12,7 @@ use mountpoint_s3::fuse::session::FuseSession;
 use mountpoint_s3::fuse::S3FuseFilesystem;
 use mountpoint_s3::metrics::{metrics_tracing_span_layer, MetricsSink};
 use mountpoint_s3::prefix::Prefix;
+use mountpoint_s3_client::ImdsCrtClient;
 use mountpoint_s3_client::{
     AddressingStyle, Endpoint, HeadBucketError, ObjectClientError, S3ClientConfig, S3CrtClient,
 };
@@ -342,6 +343,14 @@ fn mount(args: CliArgs) -> anyhow::Result<FuseSession> {
         .transpose()
         .context("Failed to parse endpoint URL")?;
 
+    match retrieve_instance_type() {
+        Ok(ec2_instance_type) => {
+            // TODO: config throughtput_target_gbps here.
+            tracing::info!("The EC2 instance type is: {:?}", ec2_instance_type);
+        }
+        Err(e) => tracing::error!("failed to detect EC2 instance type. {:?}", e),
+    }
+
     let client_config = S3ClientConfig {
         throughput_target_gbps: args.throughput_target_gbps.map(|t| t as f64),
         part_size: args.part_size.map(|t| t as usize),
@@ -474,4 +483,15 @@ fn parse_perm_bits(perm_bit_str: &str) -> Result<u16, anyhow::Error> {
     } else {
         Ok(perm)
     }
+}
+
+pub fn retrieve_instance_type() -> anyhow::Result<String> {
+    let imds_crt_client = ImdsCrtClient::new().context("failed to create IMDS client")?;
+
+    let query = imds_crt_client
+        .make_instance_type_query()
+        .context("failed to send IMDS query")?;
+
+    let result = futures::executor::block_on(query).context("IMDS query failed")?;
+    Ok(result)
 }
--- a/mountpoint-s3-client/src/lib.rs
+++ b/mountpoint-s3-client/src/lib.rs
@@ -1,11 +1,13 @@
 mod endpoint;
 pub mod failure_client;
+mod imds_crt_client;
 pub mod mock_client;
 mod object_client;
 mod s3_crt_client;
 mod util;
 
 pub use endpoint::{AddressingStyle, Endpoint};
+pub use imds_crt_client::ImdsCrtClient;
 pub use object_client::*;
 pub use s3_crt_client::head_bucket::HeadBucketError;
 pub use s3_crt_client::{S3ClientConfig, S3CrtClient, S3RequestError};
--- a/mountpoint-s3-crt-sys/build.rs
+++ b/mountpoint-s3-crt-sys/build.rs
@@ -28,6 +28,7 @@ const CRT_LIBRARIES: &[&str] = &[
 /// bindings for the stuff we (transitively) need, rather than everything.
 const CRT_HEADERS: &[&str] = &[
     "auth/credentials.h",
+    "auth/aws_imds_client.h",
     "checksums/crc.h",
     "common/atomics.h",
     "common/log_channel.h",
--- a/mountpoint-s3-crt/src/auth.rs
+++ b/mountpoint-s3-crt/src/auth.rs
@@ -7,6 +7,7 @@ use mountpoint_s3_crt_sys::aws_auth_library_init;
 use crate::common::allocator::Allocator;
 
 pub mod credentials;
+pub mod imds_client;
 pub mod signing_config;
 
 static AUTH_LIBRARY_INIT: Once = Once::new();


mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,70 @@
...
+
+        Ok(Self {
+            imds_client,
+            event_loop_group,
+            allocator,
+        })
+    }
+
+    /// Make an HTTP request using this IMDS client that invokes the given callbacks.
+    pub fn make_instance_type_query(&self) -> oneshot::Receiver<Result<OsString, error::Error>> {
```
monthonk, line 63: Could we add some tests around this function?

monthonk (commented):
Great work! left a few comments that I think we should address.

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,126 @@
...
+    {
+        let callback_closure: OnGetResource = Box::new(callback);
+        let callback_wrapper: Box<ImdsClientGetResourceCallback> =
+            Box::new(ImdsClientGetResourceCallback(Box::new(callback_closure)));
+        let callback_wrapper: &mut ImdsClientGetResourceCallback = Box::leak(callback_wrapper);
+        // callback_wrapper is safe so the raw pointer referring to that is safe as well.
+        let callback_raw_ptr: *mut libc::c_void =
+            callback_wrapper as *mut ImdsClientGetResourceCallback as *mut libc::c_void;
+        let get_resource_callback_fn_ptr: Option<unsafe extern "C" fn(*const aws_byte_buf, i32, *mut libc::c_void)> =
+            Some(imds_client_get_resource_callback);
```
dannycjones, line 65: I think we can probably reorganise this a bit to make it a bit clearer.

Is the callback_closure double boxed? Does it need to be?

I think we can use `Box::into_raw(T) -> *mut T` instead to simplify here too. This SO answer describes our use case: https://stackoverflow.com/a/73330433

Dropping some of the types may also help make the code a bit clearer.

```suggestion
        let callback_closure: OnGetResource = Box::new(callback);
        let callback_wrapper: Box<ImdsClientGetResourceCallback> =
            Box::new(ImdsClientGetResourceCallback(callback_closure));
        // callback_wrapper is safe so the raw pointer referring to that is safe as well.
        let callback_raw_ptr = Box::into_raw(callback_wrapper) as *mut libc::c_void;
        let get_resource_callback_fn_ptr = Some(imds_client_get_resource_callback);
```

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,126 @@
...
+
+impl ImdsClient {
+    /// Create a new [ImdsClient]
+    pub fn new(allocator: &Allocator, config: ImdsClientConfig) -> Result<Self, Error> {
+        auth_library_init(allocator);
+
+        // SAFETY: `config` is moved into the [ImdsClient] on success, so `config.inner` (and the values
+        // inside) are guaranteed to live at least as long as this [ImdsClient] does. `allocator` is
+        // guaranteed to be a valid allocator because of the type-safe wrapper.
+        let inner = unsafe { aws_imds_client_new(allocator.inner.as_ptr(), &config.inner).ok_or_last_error()? };
```
dannycjones, line 47: It was a bit hard to follow at first the safety comment, and maybe we are not calling out the real reason it is safe.

Does this capture it well? (I did not verify the explanation about the allocator.)

```suggestion
        // SAFETY: Struct memory referenced by `config.inner` will be managed by CRT, so it is safe to drop `ImdsClientConfig` passed into this method.
        // It is safe to pass in `allocator` as a pointer as it is only used during initialization of the client.
        let inner = unsafe { aws_imds_client_new(allocator.inner.as_ptr(), &config.inner).ok_or_last_error()? };
```

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+}
+
+impl Drop for ImdsClient {
+    fn drop(&mut self) {
+        // SAFETY: `self.inner` points to a valid aws_imds_client, and when this [ImdsClient] is dropped.
+        unsafe { aws_imds_client_release(self.inner.as_ptr()) };
+    }
+}
```
dannycjones, line 83: Simplify safety comment?

```suggestion
impl Drop for ImdsClient {
    fn drop(&mut self) {
        // SAFETY: `self.inner` is reference counted by CRT, and valid for lifetime of `ImdsClient`.
        unsafe { aws_imds_client_release(self.inner.as_ptr()) };
    }
}
```

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,126 @@
...
+    let resource = OsStr::from_bytes(resource_slice).to_os_string();
+    let callback = Box::from_raw(user_data as *mut ImdsClientGetResourceCallback).0;
+    let result = if 0 != error_code {
+        Err(Error::from(error_code))
+    } else {
+        Ok(resource)
+    };
+
+    callback(result);
+}
```
dannycjones, line 126: Maybe we can improve the flow here a bit. Some stuff we don't need to do if we encounter the error case.

```suggestion
/// Rust binding for CRT's callback function `aws_imds_client_on_get_resource_callback_fn`.
unsafe extern "C" fn imds_client_get_resource_callback(
    resource: *const aws_byte_buf,
    error_code: i32,
    user_data: *mut libc::c_void,
) {
    let callback = Box::from_raw(user_data as *mut ImdsClientGetResourceCallback).0;
    
    let result = if 0 != error_code {
        Err(Error::from(error_code))
    } else {
        let resource_slice = std::slice::from_raw_parts((*resource).buffer, (*resource).len);
        let resource = OsStr::from_bytes(resource_slice).to_os_string();
        Ok(resource)
    };
    
    callback(result)
}
```

.gitignore
```diff
@@ -1,2 +1,3 @@
 /target
 .cargo/
+TAGS
```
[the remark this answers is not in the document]
eslrahc-swa: will move it to `.git.exclude`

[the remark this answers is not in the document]
eslrahc-swa: it's for emacs ctag plugin, not specifically for moutpont-s3 itself, will move it to .git/exclude.

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,70 @@
...
+use mountpoint_s3_crt::auth::imds_client::ImdsClientConfig;
+use mountpoint_s3_crt::common::allocator::Allocator;
+use mountpoint_s3_crt::common::error;
+use mountpoint_s3_crt::io::channel_bootstrap::{ClientBootstrap, ClientBootstrapOptions};
+use mountpoint_s3_crt::io::event_loop::EventLoopGroup;
+use mountpoint_s3_crt::io::host_resolver::{HostResolver, HostResolverDefaultOptions};
+use mountpoint_s3_crt::io::retry_strategy::{ExponentialBackoffJitterMode, RetryStrategy, StandardRetryOptions};
+
+#[derive(Debug)]
+pub struct ImdsCrtClient {
```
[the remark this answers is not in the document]
eslrahc-swa: will add

mountpoint-s3-client/src/imds_crt_client.rs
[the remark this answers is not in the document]
eslrahc-swa: I think it should be `OsSting` otherwise we will run into the string encoding/decoding issue.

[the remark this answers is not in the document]
eslrahc-swa: This function should not be `async`, just like [s3_crt_client.rs -> fn make_meta_request<T: Send + 'static, E: Send + 'static>](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-client/src/s3_crt_client.rs#L185)

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,126 @@
...
+use std::ptr::NonNull;
+
+/// A client for instance metadata query.
+#[derive(Debug)]
+pub struct ImdsClient {
+    /// A pointer to the underlying `aws_imds_client`
+    inner: NonNull<aws_imds_client>,
+}
+
+///Configurations for creating a new [ImdsClient].
```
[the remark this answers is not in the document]
eslrahc-swa: nice catch! will fix

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// client will take ownership.
+    retry_strategy: Option<RetryStrategy>,
+}
+
+impl ImdsClient {
+    /// Create a new [ImdsClient]
+    pub fn new(allocator: &Allocator, config: ImdsClientConfig) -> Result<Self, Error> {
+        auth_library_init(allocator);
+
+        // SAFETY: `config` is moved into the [ImdsClient] on success, so `config.inner` (and the values
```
eslrahc-swa, line 44: I think I misplaced the config safety comment here, will reword.

mountpoint-s3-crt/src/auth/imds_client.rs
eslrahc-swa: The modified comments make sense to me. However, the key words `unsafe` here is to guarantee the reference to two objects must have the same lifetime as `ImdsClient`

1. >   // SAFETY: Struct memory referenced by `config.inner` will be managed by CRT, so it is safe to drop `ImdsClientConfig` passed into this method.

    * `config.inner` will be actually used by CRT, the object(in memory) is owned by client side but not CRT. When we instantiate  a `ImdsClient` object, the object itself owns the `config` along with its inner object, and when the object of `ImdsClient` get dropped, `config` will be dropped together.


2, > // It is safe to pass in `allocator` as a pointer as it is only used during initialization of the client.
    * It's true, the adds up for this statement is CRT will use `allocator` in variety of places like allocate memory for internal `ListNodes` which act as a node of the request queue or allocates memory for internal `imds_user_data` etc. Rust client owns the memory of  `allocator` itself.

 the code inside an unsafe block will access memory in a valid way because there is no memory leak, and the reference passed in is implemented and managed by in Rust, the reference is live as long as the `ImdsClient`.

mountpoint-s3-crt/src/auth/imds_client.rs
eslrahc-swa: Will fix, and that will simplify type conversion

mountpoint-s3/src/main.rs
```diff
@@ -340,6 +340,9 @@ fn mount(args: CliArgs) -> anyhow::Result<FuseSession> {
         .transpose()
         .context("Failed to parse endpoint URL")?;
 
+    let ec2_instance_type = retrieve_instance_type()?;
```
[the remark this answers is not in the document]
eslrahc-swa: will fix

mountpoint-s3-crt/src/auth/imds_client.rs
eslrahc-swa: will fix

mountpoint-s3-client/src/imds_crt_client.rs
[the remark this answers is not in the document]
eslrahc-swa: > Also, the result should be String instead of OsString.
    * will fix that.
However, I do not see the benefit of mapping `mountpoint_s3_crt::common::error::Error` to other type of error given it's either success or failure with no other reason.

dannycjones:
Maybe we considered this already: should we be using CRT for this or should we use a Rust-based HTTP client?

CRT of course will make things easier in terms of how to use IMDS, but it does mean we have to put thought into the safety interacting with its C code.

mountpoint-s3-client/src/imds_crt_client.rs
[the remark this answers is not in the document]
dannycjones: Hmmmm, I see. In that case, the `Ok` case of the result is a future of some kind - you can use `.await`. I will take a look a bit more later too.

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,70 @@
...
+            allocator,
+        })
+    }
+
+    /// Make an HTTP request using this IMDS client that invokes the given callbacks.
+    pub fn make_instance_type_query(&self) -> oneshot::Receiver<Result<String, error::Error>> {
+        let (tx, rx) = oneshot::channel::<Result<String, error::Error>>();
+        let _ = self.imds_client.get_instance_type(move |result| {
+            let _ = tx.send(result);
+        });
```
dannycjones, line 67: We are not handling error cases here.

Later when we await a message - are we going to block forever waiting for a response before starting the file system?

mountpoint-s3-crt/src/auth/imds_client.rs
dannycjones: OK.

So for 1., we are saying that `aws_imds_client_new` will be using the config and then doesn't need it anymore, so its safe to drop all of it at the end of this `new` method. At the moment, the safety comment for number 1 isn't covering why its safe yet.

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,119 @@
...
+    user_data: *mut libc::c_void,
+) {
+    let callback = Box::from_raw(user_data as *mut ImdsClientGetResourceCallback).0;
+
+    callback(if 0 != error_code {
+        Err(Error::from(error_code))
+    } else {
+        let resource = std::slice::from_raw_parts((*resource).buffer, (*resource).len).to_vec();
+        Ok(String::from_utf8(resource).unwrap())
+    })
```
dannycjones, line 118: I wouldn't combine the callback with the return of the if statement, it is making it difficult to follow.

Also, I'd use `.expect(&str)` instead of `.unwrap()` so that we get a meaningful panic message if we do find ourselves in the situation - which hopefully we wouldn't :)

```suggestion
    let result = if 0 != error_code {
        Err(Error::from(error_code))
    } else {
        let resource = std::slice::from_raw_parts((*resource).buffer, (*resource).len).to_vec();
        let resource = String::from_utf8(resource).expect("resource response should be utf8");
        Ok(resource)
    }
    
    callback(result)
```

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub fn get_instance_type<F>(&self, callback: F) -> Result<i32, Error>
+    where
+        F: FnOnce(Result<String, Error>) + 'static,
+    {
+        let callback_closure = Box::new(callback);
+        let callback_wrapper = Box::new(ImdsClientGetResourceCallback(Box::new(callback_closure)));
+        let callback_raw_ptr = Box::into_raw(callback_wrapper) as *mut libc::c_void;
+        // must explicit declare the type as function pointer instead of function item if not .
+        let get_resource_callback_fn_ptr: Option<unsafe extern "C" fn(*const aws_byte_buf, i32, *mut libc::c_void)> =
+            Some(imds_client_get_resource_callback);
```
dannycjones, line 60: if not what?

dannycjones (changes requested):

mountpoint-s3-crt/src/auth/imds_client.rs
eslrahc-swa: if not, the type of `Some(imds_client_get_resource_callback)` would bi function itself called `item` something like 
```
error[E0308]: mismatched types
    --> mountpoint-s3-crt/src/auth/imds_client.rs:67:72
     |
67   |                 aws_imds_client_get_instance_type(self.inner.as_ptr(), get_resource_callback_fn_ptr, callback_raw_ptr);
     |                 ---------------------------------                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected fn pointer, found fn item
     |                 |
     |                 arguments to this function are incorrect
     |
     = note: expected enum `Option<unsafe extern "C" fn(_, _, _)>`
                found enum `Option<unsafe extern "C" fn(_, _, _) {imds_client_get_resource_callback}>`
note: function defined here

```
instead of

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        // Safety: move the reference where the data store in memory out of box potential leaks memory,
+        // however, the binding callback function will drop it once the corresponding crt function finished the execution.
+        // See `extern "C" fn imds_client_get_resource_callback`.
```
jamesbornholt, line 64: Memory leaks aren't a safety problem in Rust (leaking memory is safe by definition). The safety argument here is about why `callback_raw_ptr` and `self.inner.as_ptr()` will be valid.

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,119 @@
...
+type OnGetResource = Box<dyn FnOnce(Result<String, Error>)>;
+struct ImdsClientGetResourceCallback(OnGetResource);
+
+/// Rust binding for CRT's callback function `aws_imds_client_on_get_resource_callback_fn`.
+unsafe extern "C" fn imds_client_get_resource_callback(
+    resource: *const aws_byte_buf,
+    error_code: i32,
+    user_data: *mut libc::c_void,
+) {
+    let callback = Box::from_raw(user_data as *mut ImdsClientGetResourceCallback).0;
```
jamesbornholt, line 111: Need a comment about why this is safe; for example, can `imds_client_get_resource_callback` be called twice?

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    callback(if 0 != error_code {
+        Err(Error::from(error_code))
```
jamesbornholt, line 114:
```suggestion
        Err(error_code.into())
```

mountpoint-s3/src/main.rs
```diff
@@ -340,6 +340,13 @@ fn mount(args: CliArgs) -> anyhow::Result<FuseSession> {
         .transpose()
         .context("Failed to parse endpoint URL")?;
 
+    let ec2_instance_type = retrieve_instance_type().unwrap_or_else(|_| {
+        tracing::error!("Unable to detect instance type.");
+        String::from("unknown instance type.")
+    });
+
+    tracing::debug!("the ec2 instance type is: {}", &ec2_instance_type);
```
jamesbornholt, line 348: Seems like we'd prefer `ec2_instance_type` to just be `Option<String>` here to represent the failure case.

mountpoint-s3/src/main.rs
```diff
@@ -340,6 +340,13 @@ fn mount(args: CliArgs) -> anyhow::Result<FuseSession> {
         .transpose()
         .context("Failed to parse endpoint URL")?;
 
+    let ec2_instance_type = retrieve_instance_type().unwrap_or_else(|_| {
+        tracing::error!("Unable to detect instance type.");
```
jamesbornholt, line 344: Need to log the actual error we got back

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+impl Drop for ImdsClient {
+    fn drop(&mut self) {
+        // SAFETY: `self.inner` points to a valid aws_imds_client.
+        unsafe { aws_imds_client_release(self.inner.as_ptr()) };
+    }
+}
+
+impl ImdsClientConfig {
```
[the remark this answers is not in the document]
eslrahc-swa: done

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,119 @@
...
+        // SAFETY: `config` is moved into the [ImdsClient] on success, so `config.inner` (and the values
+        // inside) are guaranteed to live at least as long as this [ImdsClient] does. `allocator` is
+        // guaranteed to be a valid allocator because of the type-safe wrapper.
+        let inner = unsafe { aws_imds_client_new(allocator.inner.as_ptr(), &config.inner).ok_or_last_error()? };
+
+        Ok(Self { inner })
+    }
+
+    /// Make a get instance type request to ec2 hosts using [ImdsClient].
+    pub fn get_instance_type<F>(&self, callback: F) -> Result<i32, Error>
```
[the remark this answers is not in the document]
eslrahc-swa: Oh, I haven't aware of this piece of code exists.
```
/// Some CRT functions return an int that is either AWS_OP_SUCCESS or AWS_OP_ERR, and the caller           
/// should use last_error to find out what happened. This simplifies that pattern.                         
impl CrtError for i32 {                                                                                    
    type Return = ();                                                                                      
                                                                                                           
    unsafe fn ok_or_last_error(self) -> Result<Self::Return, Error> {                                      
        match self {                                                                                       
            AWS_OP_SUCCESS => Ok(()),                                                                      
            AWS_OP_ERR => Err(Error::last_error()),                                                        
            // This case shouldn't happen if used correctly since we should use this on functions          
            // that only return SUCCESS or ERR. But if it does happen, we can attempt to convert the       
            // error code directly, which may or may not work (but at least the Error won't be swallowed). 
            n => Err(common::error::Error::from(n)),                                                       
        }                                                                                                  
    }                                                                                                      
}                                                                                                          

```

mountpoint-s3/src/main.rs
eslrahc-swa: Changed.

mountpoint-s3/src/main.rs
eslrahc-swa: Changed

mountpoint-s3-crt/src/auth/imds_client.rs
eslrahc-swa: done

mountpoint-s3-crt/src/auth/imds_client.rs
eslrahc-swa: done

mountpoint-s3/src/main.rs
```diff
@@ -479,3 +484,38 @@ fn parse_perm_bits(perm_bit_str: &str) -> Result<u16, anyhow::Error> {
         Ok(perm)
     }
 }
+
+pub fn retrieve_instance_type() -> Option<String> {
```
jamesbornholt, line 488: This should return `anyhow::Result<String>`, and the caller should log the error.

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,109 @@
...
+            + 'static,
+    ) -> Result<ImdsQueryRequest, ImdsQueryRequestError> {
+        let (tx, rx) = oneshot::channel::<Result<Result<String, error::Error>, ImdsQueryRequestError>>();
+        let _ = self.imds_client.get_instance_type(move |result| {
+            let res = callback(result);
+            let _ = tx.send(res);
+        });
+
+        Ok(ImdsQueryRequest { receiver: rx })
+    }
```
jamesbornholt, line 77: Why does this method need a callback? The returned `ImdsQueryRequest` is a `Future` that resolves to the returned `String`, or just an error. So anything the callback could do can instead just be done by the caller.

Also, nested `Result` types are a hint that we haven't quite got the interface right.

So I think this method ends up being something like:
```suggestion
    pub fn make_instance_type_query(&self) -> Result<ImdsQueryRequest, ImdsQueryRequestError> {
        let (tx, rx) = oneshot::channel();
        let _ = self.imds_client.get_instance_type(move |result| {
            let _ = tx.send(result.map_err(ImdsQueryRequestError::CrtError));
        });

        Ok(ImdsQueryRequest { receiver: rx })
    }
```

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+#[derive(Debug)]
+#[pin_project]
+/// ImdsQueryRequest is a data encapsulation wrapper for asnychronous instance metadata query.
+pub struct ImdsQueryRequest {
+    #[pin]
+    receiver: oneshot::Receiver<Result<Result<String, error::Error>, ImdsQueryRequestError>>,
```
jamesbornholt, line 85:
```suggestion
    receiver: oneshot::Receiver<Result<String, ImdsQueryRequestError>>,
```

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+impl Future for ImdsQueryRequest {
+    type Output = Result<Result<String, error::Error>, ImdsQueryRequestError>;
```
jamesbornholt, line 89:
```suggestion
    type Output = Result<String, ImdsQueryRequestError>;
```

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        this.receiver
+            .poll(cx)
+            .map(|result| result.unwrap_or_else(|err| Err(ImdsQueryRequestError::InternalError(Box::new(err)))))
+    }
+}
+
+/// ImdsQueryRequestError is responsible for hide and unify underlying crt and async error.
```
jamesbornholt, line 99: nit: This comment isn't going to be meaningful for a customer of this client. The type is just "Errors returned by an asynchronous IMDS query".

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        let bootstrap_options = ClientBootstrapOptions {
+            event_loop_group: &mut event_loop_group,
+            host_resolver: &mut host_resolver,
+        };
+
+        let client_bootstrap = ClientBootstrap::new(&allocator, &bootstrap_options).unwrap();
+
+        let mut retry_strategy_options = StandardRetryOptions::default(&mut event_loop_group);
+        let retry_strategy = RetryStrategy::standard(&allocator, &retry_strategy_options).unwrap();
```
jamesbornholt, line 47: 1. Doesn't need to be `mut`.
2. What are the default options? Are they appropriate for the IMDS use case?

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,123 @@
...
+        auth_library_init(allocator);
+
+        // SAFETY: `config` is moved into the [ImdsClient] on success, so `config.inner` (and the values
+        // inside) are guaranteed to live at least as long as this [ImdsClient] does. `allocator` is
+        // guaranteed to be a valid allocator because of the type-safe wrapper.
+        let inner = unsafe { aws_imds_client_new(allocator.inner.as_ptr(), &config.inner).ok_or_last_error()? };
+        Ok(Self { inner })
+    }
+
+    /// Make a get instance type request to ec2 hosts using [ImdsClient].
```
jamesbornholt, line 70:
```suggestion
    /// Get the instance type of the EC2 instance
```

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,123 @@
...
+            aws_imds_client_get_instance_type(self.inner.as_ptr(), get_resource_callback_fn_ptr, callback_raw_ptr)
+                .ok_or_last_error()
+        }
+    }
+}
+
+impl Drop for ImdsClient {
+    fn drop(&mut self) {
+        // SAFETY: `self.inner` points to the memory area `aws_imds_client`,
+        // crt assures the safy of revoking memory.
```
jamesbornholt, line 95: Not sure what this means. Can you explain why you should this call is safe?

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,123 @@
...
+        F: FnOnce(Result<String, Error>) + 'static,
+    {
+        let callback_closure = Box::new(callback);
+        let callback_wrapper = Box::new(ImdsClientGetResourceCallback(Box::new(callback_closure)));
+        let callback_raw_ptr = Box::into_raw(callback_wrapper) as *mut libc::c_void;
+        let get_resource_callback_fn_ptr: Option<unsafe extern "C" fn(*const aws_byte_buf, i32, *mut libc::c_void)> =
+            Some(imds_client_get_resource_callback);
+
+        // SAFETY: `get_resource_callback_fn_ptr` is a guaranteed non-null callback function pointer pass to CRT.
+        // `callback_raw_ptr` is a pointer to a senquence of valid instructions in memory the binding function will
```
jamesbornholt, line 82: `callback_raw_ptr` does not point to instructions, it points to a closure.

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,123 @@
...
+) {
+    // SAFETY: CRT invokes the callback only once given it has `FnOnce` trait, the execution of callback
+    // instructions(rust closure) is in unsafe zone, however, the borrow checker assures that instuctions
+    // will not result memory safety issue.
+    let callback = Box::from_raw(user_data as *mut ImdsClientGetResourceCallback).0;
+
+    let result = if 0 != error_code {
+        Err(error_code.into())
+    } else {
+        let resource = std::slice::from_raw_parts((*resource).buffer, (*resource).len).to_vec();
```
jamesbornholt, line 118: Why is it safe to dereference the `resource` pointer? Deserves a safety comment.

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,109 @@
...
+            .map(|result| result.unwrap_or_else(|err| Err(ImdsQueryRequestError::InternalError(Box::new(err)))))
+    }
+}
+
+/// ImdsQueryRequestError is responsible for hide and unify underlying crt and async error.
+#[derive(Error, Debug)]
+pub enum ImdsQueryRequestError {
+    /// An internal error from within the Imds client. The request may have been sent.
+    #[error("Internal Imds client error")]
+    InternalError(#[source] Box<dyn std::error::Error>),
```
jamesbornholt, line 104: This will be really annoying to use without Send and Sync:
```suggestion
    InternalError(#[source] Box<dyn std::error::Error + Send + Sync>),
```

mountpoint-s3/src/main.rs
```diff
@@ -479,3 +484,38 @@ fn parse_perm_bits(perm_bit_str: &str) -> Result<u16, anyhow::Error> {
...
+        .map_err(|e| {
+            tracing::error!(
+                "failed to retrieve ec2 instance type by calling AWS Common Runtime(CRT) library {:?}",
+                e
+            );
+            e
+        })
+        .ok()?;
+
+    output.ok()
```
jamesbornholt, line 520: If we get the `Result` types fixed up, this can all just be something like
```rust
    let imds_crt_client = ImdsCrtClient::new().context("failed to create IMDS client")?;

    let query = imds_crt_client
        .make_instance_type_query()
        .context("failed to send IMDS query")?;

    Ok(futures::executor::block_on(query).context("IMDS query failed")?)
```

jamesbornholt:
Also, how do you think we should test this code?

mountpoint-s3/src/main.rs
eslrahc-swa: fixed

mountpoint-s3/src/main.rs
eslrahc-swa: fixed

mountpoint-s3-client/src/imds_crt_client.rs
eslrahc-swa: fixed

mountpoint-s3-client/src/imds_crt_client.rs
eslrahc-swa: fixed

mountpoint-s3-client/src/imds_crt_client.rs
eslrahc-swa: fixed

mountpoint-s3-client/src/imds_crt_client.rs
eslrahc-swa: fixed

mountpoint-s3/src/main.rs
eslrahc-swa: fixed

mountpoint-s3-client/src/imds_crt_client.rs
eslrahc-swa: 1.fixed

mountpoint-s3-crt/src/auth/imds_client.rs
eslrahc-swa: fixed

mountpoint-s3-client/src/imds_crt_client.rs
jamesbornholt: Oh, except we don't want to ignore the synchronous error case:
```suggestion
    pub fn make_instance_type_query(&self) -> Result<ImdsQueryRequest, ImdsQueryRequestError> {
        let (tx, rx) = oneshot::channel();
        self.imds_client.get_instance_type(move |result| {
            let _ = tx.send(result.map_err(ImdsQueryRequestError::CrtError));
        })?;

        Ok(ImdsQueryRequest { receiver: rx })
    }
```

mountpoint-s3-client/src/imds_crt_client.rs
eslrahc-swa: 2. fixed

mountpoint-s3-client/src/imds_crt_client.rs
eslrahc-swa: yeah, my mental model is there are two errors might happen: 
1. Error happened in CRT which is `moutpoint_s3::common::error::Error`
2. Error happened during the data transfer in `oneshot` channel.

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+impl Drop for ImdsClient {
+    fn drop(&mut self) {
+        // SAFETY: `self.inner` always points to a valid underlying `aws_imds_client`.
+        // when `aws_imdw_client` creation failed, crt frees the allocated memroy
+        // otherwise return the pointer to `aws_imds_client`.
+        unsafe { aws_imds_client_release(self.inner.as_ptr()) };
+    }
+}
```
dannycjones, line 99: typos. but do we even need the latter sentence? This is one of the more straightforward safety cases I think.

```suggestion
impl Drop for ImdsClient {
    fn drop(&mut self) {
        // SAFETY: `self.inner` always points to a valid underlying `aws_imds_client`.
        unsafe { aws_imds_client_release(self.inner.as_ptr()) };
    }
}
```

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,127 @@
...
+
+/// Callback trait for when finished getting ec2 instance type.
+type OnGetResource = Box<dyn FnOnce(Result<String, Error>)>;
+struct ImdsClientGetResourceCallback(OnGetResource);
```
dannycjones, line 103: Bit unclear here.

`OnGetResource` isn't a trait but a type alias.

I don't understand what the purpose of the second struct is? Can you add an explanation?

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,127 @@
...
+    error_code: i32,
+    user_data: *mut libc::c_void,
+) {
+    // SAFETY: CRT invokes the callback only once given it has `FnOnce` trait, the execution of callback
+    // rust closure is in unsafe zone, however, the borrow checker assures that instuctions
+    // will not result memory safety issue.
+    // Deferenece resource pointer only happenes when crt successful retrieves instance type, and since
+    // it is immutable pointer, no potential undefined behavior could occur to the actual data resource
+    // points to.
+    let callback = Box::from_raw(user_data as *mut ImdsClientGetResourceCallback).0;
```
dannycjones, line 117: There's a few typos in here.

I'm also not very confident in the explanation of the safety, can we elaborate? (Not necessarily in code, we can discuss safety here and updated comment if necessary.)

I think lines 114-116 belong with the `*resource` de-referencing on line 122.

mountpoint-s3/src/main.rs
```diff
@@ -13,8 +13,9 @@ use mountpoint_s3::fuse::S3FuseFilesystem;
 use mountpoint_s3::metrics::{metrics_tracing_span_layer, MetricsSink};
 use mountpoint_s3::prefix::Prefix;
 use mountpoint_s3_client::{
-    AddressingStyle, Endpoint, HeadBucketError, ObjectClientError, S3ClientConfig, S3CrtClient,
+    AddressingStyle, Endpoint, HeadBucketError, ImdsCrtClient, ObjectClientError, S3ClientConfig, S3CrtClient,
 };
+
```
dannycjones, line 18: drop this newline. not sure if there's a good way to enforce this with a style checker instead. :(

mountpoint-s3/src/main.rs
```diff
@@ -342,6 +342,14 @@ fn mount(args: CliArgs) -> anyhow::Result<FuseSession> {
         .transpose()
         .context("Failed to parse endpoint URL")?;
 
+    match retrieve_instance_type() {
+        Ok(ec2_instance_type) => {
+            // TODO: config throughtput_target_gbps here.
+            tracing::debug!("The ec2 instance type is: {:?}", ec2_instance_type);
+        }
+        Err(e) => tracing::error!("failed to retrieve ec2 instance type. {:?}", e),
```
dannycjones, line 350: Let's make sure AWS service names like EC2 are capitalized correctly where possible - especially in customer-facing bits like logging.

mountpoint-s3-crt/src/auth/imds_client.rs
dannycjones: This still isn't resolved for me. For 1., I don't understand why `config` is moved into `ImdsClient`. `new()` takes ownership, passes a pointer into unsafe CRT code. Why is it not dropped?

dannycjones (changes requested):

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,114 @@
...
+        retry_strategy_options.backoff_retry_options.jitter_mode = ExponentialBackoffJitterMode::Full;
+        let retry_strategy = RetryStrategy::standard(&allocator, &retry_strategy_options).unwrap();
+
+        let mut client_config = ImdsClientConfig::new();
+
+        client_config
+            .client_bootstrap(client_bootstrap)
+            .retry_strategy(retry_strategy);
+
+        let imds_client = ImdsClient::new(&allocator, client_config).unwrap();
```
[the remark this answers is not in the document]
eslrahc-swa: yeah, nice catch

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,114 @@
...
+        let mut retry_strategy_options = StandardRetryOptions::default(&mut event_loop_group);
+
+        // Match the Java SDK retry strategies:
+        // 1.max retries: https://tinyurl.com/3vbn5ty6
+        // 2.backoff strategy: https://tinyurl.com/ycmta4vw
+        // 3.backoff duration: https://tinyurl.com/4rfzsdva
+        retry_strategy_options.backoff_retry_options.max_retries = 3;
+        retry_strategy_options.backoff_retry_options.backoff_scale_factor = Duration::from_millis(20000);
+        retry_strategy_options.backoff_retry_options.jitter_mode = ExponentialBackoffJitterMode::Full;
+        let retry_strategy = RetryStrategy::standard(&allocator, &retry_strategy_options).unwrap();
```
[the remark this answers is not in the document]
jamesbornholt: This change isn't what I meant when I commented on this before. We don't want to match the generic SDK retry strategies here. The SDKs use a different retry strategy specifically for IMDS calls. That's what we should be trying to match.

mountpoint-s3-crt/src/auth/imds_client.rs
eslrahc-swa: I think the I put the comment to a wrong place, the point is unit-like struct implement a trait on some type but don’t have any data. 
Let me rephrase it a little bit.

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,114 @@
...
+        let client_bootstrap = ClientBootstrap::new(&allocator, &bootstrap_options).unwrap();
+
+        let mut retry_strategy_options = StandardRetryOptions::default(&mut event_loop_group);
+
+        // Match the SDK "legacy" retry strategies:
+        // 1.max retries: https://tinyurl.com/3vbn5ty6
+        // 2.backoff strategy: https://tinyurl.com/ycmta4vw
+        // 3.backoff duration: https://tinyurl.com/4rfzsdva
+        retry_strategy_options.backoff_retry_options.max_retries = 3;
+        retry_strategy_options.backoff_retry_options.backoff_scale_factor = Duration::from_millis(20000);
```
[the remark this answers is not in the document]
eslrahc-swa: so in general:
1. JDK [here](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/backoff/BackoffStrategy.java#L51-L56) uses [BackoffStrategy.defaultStrategy](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/backoff/BackoffStrategy.java#L51-L56)

2. [BackoffStrategy](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/backoff/BackoffStrategy.java#L42-L45) calculate the `backoff_scale_factor` via `calculateExponentialDelay`

3. Java JDK adopt `FullJitterBackoffStrategy` [here](https://github.com/aws/aws-sdk-java-v2/blob/8db0b7c562e98937a05771a6c87a239faa9e4184/core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataRetryPolicy.java#L43-L49) and the way `FullJitterBackoffStrategy` `calculateExponentialDelay` is use `baseDelay` 

4.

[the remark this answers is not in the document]
eslrahc-swa: Given there are 3 times of retry and the maximum retry backoff time will be 100*2^2 = 400 millis.

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Query the instnace type mountpoint-s3 running on.
+    pub fn make_instance_type_query(&self) -> Result<ImdsQueryRequest, ImdsQueryRequestError> {
+        let (tx, rx) = oneshot::channel();
+        let _ = self.imds_client.get_instance_type(move |result| {
+            let _ = tx.send(result.map_err(ImdsQueryRequestError::CrtError));
+        });
+
+        Ok(ImdsQueryRequest { receiver: rx })
+    }
+}
```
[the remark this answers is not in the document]
eslrahc-swa: Oh gee, this is very nice catch

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,114 @@
...
+
+        Ok(Self {
+            imds_client,
+            event_loop_group,
+            allocator,
+        })
+    }
+
+    /// Query the instnace type mountpoint-s3 running on.
+    pub fn make_instance_type_query(&self) -> Result<ImdsQueryRequest, ImdsQueryRequestError> {
```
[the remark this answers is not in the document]
eslrahc-swa: We haven't put thoughts to Lambda use case yet.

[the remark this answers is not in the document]
jamesbornholt: We need to at least test this code somewhere that isn't an EC2 instance.

mountpoint-s3-crt/src/auth/imds_client.rs
eslrahc-swa: the latter sentence is actually the reason of  `self.inner` is valid, basically the summarize of function [here](https://github.com/awslabs/aws-c-auth/blob/dd505b55fd46222834f35c6e54165d8cbebbfaaa/source/aws_imds_client.c#L111-L198)

jamesbornholt: Danny's point is that if we're dropping an `ImdsClient`, we know that the `aws_imds_client_new` function you linked to has already succeeded, so the "creation failed" case isn't relevant here.

mountpoint-s3-crt/src/auth/imds_client.rs
eslrahc-swa: I think think the comment needs to focus on explain why deference raw pointers are safe.
two  raw pointers `resource: *const aws_byte_buf` and `user_data: *mut libc::c_void`

1. `resource: *const aws_byte_buf`:  this raw pointer is the place where crt put the result bytes to, [if we crt won't return error code, then the instance type bytes are guaranteed in the memory it points to ](https://github.com/awslabs/aws-c-auth/blob/dd505b55fd46222834f35c6e54165d8cbebbfaaa/source/aws_imds_client.c#L610-L613) and thus deference  *const aws_byte_buf` is safe because a) it's a this raw pointer is imutable pointer and it won't now points to anywhere other than ec2 instance type bytes. b) when we deference this pointer, the pointer points to the valid data. 

2.  `user_data: *mut libc::c_void`: this is the actually rust closure passed to crt, acting as the C "function body". In rust, we define it type as `Box<dyn FnOnce(Result<String, Error>)>`  means whatever rust closure passed to CRT as callback function will be only execute once per request. Rust compiler assures the safety of closure. To be specific, we convert the closure outside the unsafe code block and we know the closure we wrapped a round is valid, hence within unsafe block, deference the raw pointer created outside the unsafe block is safe.

mountpoint-s3-client/src/imds_crt_client.rs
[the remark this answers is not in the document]
eslrahc-swa: self.imds_client.get_instance_type returns Result<(), mountpoint-s3-crt::common::error::Error> and `mountpoint-s3-crt::common::error::Error` is compatible with `ImdsQueryRequestError`, just put a `?` to the end of function to propagate up and let the caller to handle.

mountpoint-s3-client/src/imds_crt_client.rs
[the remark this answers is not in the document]
eslrahc-swa: Tested on my local laptop(ubuntu), the program can run as normal and it will shows the following errors when calling  `retrieve_instance_type`
```
2023-04-03T19:57:45.731861Z  WARN awscrt::Unknown: id=0x55f099dfa390: IMDS Client failed to acquire a connection, error code 1048(socket operation timed out.)    
2023-04-03T19:57:45.731911Z ERROR awscrt::Unknown: (id=0x55f099dfa390) IMDS client failed to update the token from IMDS. 
2023-04-03T19:57:48.732022Z  WARN awscrt::Unknown: id=0x55f099dfa390: IMDS Client failed to acquire a connection, error code 1048(socket operation timed out.)
```

mountpoint-s3-crt/src/auth/imds_client.rs
eslrahc-swa: Ohhhh, I know why you are confused about this comment, that was my bad. I keep thinking I have ImdsClient like this

```
/// A client for instance metadata query.
#[derive(Debug)]
pub struct ImdsClient {
    /// A pointer to the underlying `aws_imds_client`
    inner: NonNull<aws_imds_client>,
}
```
but actually defined as this

```
 /// A client for instance metadata query.                   
 #[derive(Debug)]                                            
 pub struct ImdsClient { 3 implementations                   
     /// A pointer to the underlying `aws_imds_client`       
     inner: NonNull<aws_imds_client>,                        
     config: ImdsClientConfig,                               
 }                                                           
```
and when we create `ImdsClient`, we pass in` ClientConfig` which return Self {inner, config} that is `config` moved in
```

Comment (jamesbornholt) [reply mountpoint-s3-crt/src/auth/imds_client.rs:66]: There shouldn't be any need to retain the `ImdsClientConfig` beyond the end of `ImdsClient::new`. The `aws_imds_client_new` bumps the reference count of everything it needs out of the config struct ([retry strategy](https://github.com/awslabs/aws-c-auth/blob/54f8d804120daab9e0d75a56a113a222b334d0f9/source/aws_imds_client.c#L179), [bootstrap](https://github.com/awslabs/aws-c-http/blob/0600662610aa871a11aebe6ed67a11997317cbef/source/connection_manager.c#L869)) and doesn't hold onto the pointer to the `aws_imds_client_options`. That's why this is safe.

File: mountpoint-s3-client/src/imds_crt_client.rs [hunk 43]
```diff
@@ -0,0 +1,97 @@
...
+use mountpoint_s3_crt::common::allocator::Allocator;
+use mountpoint_s3_crt::common::error;
+use mountpoint_s3_crt::io::channel_bootstrap::{ClientBootstrap, ClientBootstrapOptions};
+use mountpoint_s3_crt::io::event_loop::EventLoopGroup;
+use mountpoint_s3_crt::io::host_resolver::{HostResolver, HostResolverDefaultOptions};
+use pin_project::pin_project;
+use thiserror::Error;
+
+#[derive(Debug)]
+/// Instance Metadata Service (IMDS) client responsible for sending ec2 instance metadata query requests.
```
jamesbornholt, line 18:
```suggestion
/// Instance Metadata Service (IMDS) client responsible for sending EC2 instance metadata query requests.
```

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,119 @@
...
+        Ok(Self { inner })
+    }
+
+    /// Get the EC2 instance type.
+    pub fn get_instance_type<F>(&self, callback: F) -> Result<(), Error>
+    where
+        F: FnOnce(Result<String, Error>) + 'static,
+    {
+        let callback_closure = Box::new(callback);
+        let callback_wrapper = Box::new(ImdsClientGetResourceCallback(Box::new(callback_closure)));
```
jamesbornholt, line 75: One too many `Box`es:
```suggestion
        let callback_wrapper = Box::new(ImdsClientGetResourceCallback(Box::new(callback)));
```

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,97 @@
...
+pub struct ImdsCrtClient {
+    imds_client: ImdsClient,
+    event_loop_group: EventLoopGroup,
+    allocator: Allocator,
```
jamesbornholt, line 22:
```suggestion
    #[allow(unused)]
    event_loop_group: EventLoopGroup,
    #[allow(unused)]
    allocator: Allocator,
```

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,119 @@
...
+        let callback_raw_ptr = Box::into_raw(callback_wrapper) as *mut libc::c_void;
+        let get_resource_callback_fn_ptr: Option<unsafe extern "C" fn(*const aws_byte_buf, i32, *mut libc::c_void)> =
+            Some(imds_client_get_resource_callback);
+
+        // SAFETY: `get_resource_callback_fn_ptr` is a guaranteed non-null callback function pointer pass to CRT.
+        // `callback_raw_ptr` is a pointer to the valid rust closure.
+        // See `extern "C" fn imds_client_get_resource_callback`.
```
jamesbornholt, line 82:
```suggestion
        // SAFETY: `self.inner` is a valid `aws_imds_client`. `get_resource_callback_fn_ptr` is leaked by [Box::into_raw]
        // and so will live until the `imds_client_get_resource_callback` function is invoked.
```

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,119 @@
...
+type OnGetResource = Box<dyn FnOnce(Result<String, Error>)>;
+struct ImdsClientGetResourceCallback(OnGetResource);
+
+/// Rust binding for CRT's callback function `aws_imds_client_on_get_resource_callback_fn`.
+unsafe extern "C" fn imds_client_get_resource_callback(
+    resource: *const aws_byte_buf,
+    error_code: i32,
+    user_data: *mut libc::c_void,
+) {
+    // SAFETY: user_data is a raw pointer to rust callback closure, this callback will be exected at most once.
```
jamesbornholt, line 106:
```suggestion
    // SAFETY: `user_data` is a raw pointer to a `Box<ImdsClientGetResourceCallback>` created and leaked at query time.
    // This function will be executed at most once, so the Box is still valid right now.
```

mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -0,0 +1,119 @@
...
+    let callback = Box::from_raw(user_data as *mut ImdsClientGetResourceCallback).0;
+
+    let result = if 0 != error_code {
+        Err(error_code.into())
+    } else {
+        // SAFETY: The contnets point to by `resource` is managed by crt, deferenece
+        // `resource` occurs after crt successfully retrieved instance type.
```
jamesbornholt, line 113:
```suggestion
        // SAFETY: The CRT guarantees `resource` is a valid `*const aws_byte_buf` if the IMDS query succeeded.
```

mountpoint-s3/src/main.rs
```diff
@@ -229,7 +230,6 @@ fn main() -> anyhow::Result<()> {
     if args.foreground {
         init_tracing_subscriber(args.foreground, args.log_directory.as_deref())
             .context("failed to initialize logging")?;
-
```
jamesbornholt, line 232: Undo this deletion

mountpoint-s3/src/main.rs
```diff
@@ -342,6 +342,14 @@ fn mount(args: CliArgs) -> anyhow::Result<FuseSession> {
         .transpose()
         .context("Failed to parse endpoint URL")?;
 
+    match retrieve_instance_type() {
+        Ok(ec2_instance_type) => {
+            // TODO: config throughtput_target_gbps here.
+            tracing::debug!("The EC2 instance type is: {:?}", ec2_instance_type);
```
jamesbornholt, line 348:
```suggestion
            tracing::info!("detected EC2 instance type: {:?}", ec2_instance_type);
```

mountpoint-s3/src/main.rs
```diff
@@ -342,6 +342,14 @@ fn mount(args: CliArgs) -> anyhow::Result<FuseSession> {
...
+        }
+        Err(e) => tracing::error!("failed to retrieve EC2 instance type. {:?}", e),
```
jamesbornholt, line 350:
```suggestion
        Err(e) => tracing::error!("failed to detect EC2 instance type: {:?}", e),
```

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,97 @@
+use std::pin::Pin;
+use std::task::Context;
+use std::task::Poll;
+
+use futures::channel::oneshot;
+use futures::Future;
+use mountpoint_s3_crt::auth::imds_client::ImdsClient;
+use mountpoint_s3_crt::auth::imds_client::ImdsClientConfig;
```
jamesbornholt, line 8:
```suggestion
use mountpoint_s3_crt::auth::imds_client::{ImdsClient, ImdsClientConfig};
```

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -0,0 +1,97 @@
...
+#[derive(Debug)]
+/// Instance Metadata Service (IMDS) client responsible for sending ec2 instance metadata query requests.
+pub struct ImdsCrtClient {
+    imds_client: ImdsClient,
+    event_loop_group: EventLoopGroup,
+    allocator: Allocator,
+}
+
+impl ImdsCrtClient {
+    pub fn new() -> Result<Self, error::Error> {
```
jamesbornholt, line 26: The first thing this function needs to do is call `auth_library_init(allocator);`.

jamesbornholt: Oops, never mind, forgot which crate we're in.

jamesbornholt (approved):

dannycjones:
Nice work!

mountpoint-s3-client/src/imds_crt_client.rs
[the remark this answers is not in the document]
dannycjones: One concern here - we've added a delay during mount whenever we run on something that isn't an EC2 instance.

Customers can make it go away by specifying a value for throughput, but maybe we can be smarter and fail earlier. Maybe a file exists on an EC2 instance that we can check before making the IMDS calls?

PR #185, opened 2023-03-29
Add files and instructions for using Mountpoint with Docker

mheshwg:
---

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ For the alpha release, additional configuration is required to use Mountpoint fo
 
 ## Contributing
 
-We welcome contributions to Mountpoint for Amazon S3! Please see [CONTRIBUTING.md](doc/CONTRIBUTING.md) for more information on how to report bugs or submit pull requests, and [LOGGING.md](doc/LOGGING.md) for details on how to capture logging data for bug reports.
+We welcome contributions to Mountpoint for Amazon S3! Please see [CONTRIBUTING.md](doc/CONTRIBUTING.md) for more information on how to report bugs or submit pull requests, [LOGGING.md](doc/LOGGING.md) for details on how to capture logging data for bug reports, and [BENCHMARKING.md](doc/BENCHMARKING.md) for performance regression testing.
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -59,7 +80,7 @@ For the alpha release, additional configuration is required to use Mountpoint fo
 
 ## Contributing
 
-We welcome contributions to Mountpoint for Amazon S3! Please see [CONTRIBUTING.md](doc/CONTRIBUTING.md) for more information on how to report bugs or submit pull requests, [LOGGING.md](doc/LOGGING.md) for details on how to capture logging data for bug reports, and [BENCHMARKING.md](doc/BENCHMARKING.md) for performance regression testing.
```
monthonk, line 62: I don't think you intend to change this line.

monthonk (changes requested):
Only one minor comment and it should be good.

README.md
mheshwg: Done!. thank you.

README.md
```diff
@@ -47,6 +47,27 @@ Then run the client, specifying the directory in which your S3 bucket should be
 
 The client will run in the background by default, and the `~/mnt` directory now gives access to the objects in your S3 bucket.
 
+### Mountpoint for Amazon S3 using Docker
+
+First, build the docker image using the below command.
+
+    cd mountpoint-s3/docker
```
monthonk, line 54: It's `mountpoint-s3/docker` here but the actual Dockerfile is in `docker`. So, either we move that file to the right path or change the command here.

monthonk (commented):
Sorry, one more comment, I missed that one last time.

monthonk (dismissed):

README.md
monthonk: We have offline conversation that we're assuming user are running the command from outside project root directory. It's the same assumption on [how to compile section](https://github.com/awslabs/mountpoint-s3/blob/2f07b5ec976aad219e2d1445909eca3cf0b4976a/README.md?plain=1#L36-L37), so I think this is fine. This is a bit confusing that we also use the same name for sub project.

monthonk: But feel free to suggest the change if you still think we should update it,

mheshwg: Thank you Monthonk; I am aligned. Keeping the current version as is.

dannycjones (approved):
LGTM, tested instructions locally also.

dannycjones:
Merged, thanks for the contribution @mheshwg! 🚀

Issue #162, opened 2023-03-17
Validate `--prefix` argument and document it


jamesbornholt:
In https://github.com/awslabs/mountpoint-s3/issues/145#issuecomment-1473715299 we saw that `mount-s3` panics if you pass a prefix that doesn't end in `/`. We shouldn't do that—we should validate and return an error instead.

We should also document the `--prefix` argument since it's a common question.

passaro:
Just to confirm the current behavior: 
* we want users to explicitly pass a string ending in `/` (vs accepting any string and internally adding the `/`),
* `--prefix ""` (empty string) is valid and equivalent to no prefix provided,
* `--prefix "/"` is valid and mounts an empty folder (not sure if this is really desired behavior).

dannycjones:
Relating to how we want to solve this... We can return a result instead after validating, or should we validate once and pass around a type that indicates its a valid prefix?

passaro:
About this:
> > ```
> > * `--prefix "/"` is valid and mounts an empty folder (not sure if this is really desired behavior).
> > ```
> 
> I think that last one is a bit strange. I'd lean on rejecting that, but maybe others have thoughts there? (We can also relax it later if we change our mind.)

Just tried it out and `--prefix` accepts paths with empty "components" (e.g. "/", "foo//bar/", etc.). Mount will succeed and files under that prefix will be exposed correctly. The potentially confusing behavior is that none of those files are visible when mounting the entire bucket. 

For example: 
An object with key `/foo.txt` is **not** visible when mounting the entire bucket (no prefix), but it is visible when `--prefix "/"` is used.

passaro:
I've opened a draft PR, but also happy to look into using a validated type as @dannycjones suggested.

dannycjones:
#186 looks good and fixes the bug that created this issue.

- The behavior around `//` seems a bit strange and we should discuss some more.
- Adopting a validated type may be a nice way to eliminate the panics over returning results, if we still want to make that change.

passaro:
Had a go at the validated type approach in #187, but it is a larger change and may require some refinements. Shall we go with #186 for now and optionally add the new type later?

dannycjones:
> Had a go at the validated type approach in #187, but it is a larger change and may require some refinements. Shall we go with #186 for now and optionally add the new type later?

Happy to merge the basic fix and follow up with the improvement of adding the type.

PR #187, opened 2023-03-29
Introduce `Prefix` validated type

passaro:
Follow-up to #162. Introduces a new `Prefix` type which contains a valid string (either empty or ending in `/`) to use as a prefix argument in `S3Filesystem`, `S3FuseFilesystem`, and `Superblock`. Uses the new type for the `--prefix` argument and returns a meaningful error if not valid.

---


passaro:
I'm looking for feedback on a couple of points:
* I added `Prefix` to `inode.rs` since it's used by `Superblock`. Is it the right place?
* I've converted the argument from `Option<String>` to `Option<Prefix>` and propagated the latter downstream. This makes it clear that the prefix is optional at every level, but may be redundant: `None` and `Some(Prefix { "" })` are equivalent. Would it be preferable to have the empty string to be the default value for `Prefix` (and remove `Option<>`)?

mountpoint-s3/src/inode.rs
```diff
@@ -839,6 +837,57 @@ pub enum InodeError {
     InodeNotReadableWhileWriting(InodeNo),
 }
 
+mod prefix {
```
eslrahc-swa, line 840: Not sure if we should define prefix in a separate file and assign the related path in `lib.rs` ?

passaro:
>     * I've converted the argument from `Option<String>` to `Option<Prefix>` and propagated the latter downstream. This makes it clear that the prefix is optional at every level, but may be redundant: `None` and `Some(Prefix { "" })` are equivalent. Would it be preferable to have the empty string to be the default value for `Prefix` (and remove `Option<>`)?

I've now switched from Option<Prefix> to Prefix (and implemented Default).

dannycjones (commented):
Nice! I like this change, as we know now that a `Prefix` is always safe to use without any validation checks.

Only minor comments.

mountpoint-s3/src/inode.rs
```diff
@@ -839,6 +837,50 @@ pub enum InodeError {
...
+            } else {
+                Ok(Self {
+                    path: prefix.to_owned(),
+                })
+            }
+        }
+
+        pub fn is_valid(prefix: &str) -> bool {
+            prefix.is_empty() || prefix.ends_with('/')
+        }
```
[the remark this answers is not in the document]
dannycjones: We should add a few test cases for this too.

- `ashlbafshasfbhjla`
- ``
- `/`
- `//`
- `/hello/`
- `hello/`

Then this validation behavior is documented through tests.

mountpoint-s3/src/inode.rs
passaro: Moved to prefix.rs - but happy to change if there is a better location.

mountpoint-s3/src/inode.rs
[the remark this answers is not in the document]
passaro: About the `is_valid` method: my intent was to make it a public method to check whether a string is a valid prefix (and keep the logic in one place). One (the only?) example is in Superblock::lookup, which I have now modified.
Thoughts?

mountpoint-s3/src/prefix.rs
```diff
@@ -0,0 +1,65 @@
+use std::fmt::Display;
+use std::str::FromStr;
+
+use anyhow::anyhow;
+
+/// A prefix string ending in `/`, or the empty string
+#[derive(Debug, Clone, Default)]
+pub struct Prefix {
+    path: String,
+}
```
jamesbornholt, line 10: I think you're right that we should just treat `None` and `Prefix("")` the same way, so we should get rid of `Option<Prefix>` everywhere. That also simplifies a few other things.

mountpoint-s3/src/inode.rs
```diff
@@ -127,7 +126,7 @@ impl Superblock {
             return Err(InodeError::NotADirectory(parent_ino));
         }
         let mut full_path = parent.full_key().to_owned();
-        assert!(full_path.is_empty() || full_path.ends_with('/'));
+        assert!(Prefix::is_valid(&full_path));
```
jamesbornholt, line 129: I wouldn't change this—to me they feel similar only coincidentally; a `Prefix` is "a subset of the bucket" whereas this is more like "a query I'm about to send to S3". I can imagine worlds where we might change the behavior of `Prefix` in the future (for example, to no longer require ending in a `/`).

mountpoint-s3/src/prefix.rs
```diff
@@ -0,0 +1,65 @@
+use std::fmt::Display;
+use std::str::FromStr;
+
+use anyhow::anyhow;
```
jamesbornholt, line 4: I'd build a real error type here instead of using anyhow:
```rust
use thiserror::Error;

#[derive(Error, Debug)]
pub enum PrefixError {
    #[error("prefixes must end in '/'")]
    MissingFinalDelimiter,
}
```
and then
```rust
impl Prefix {
    pub fn new(prefix: &str) -> Result<Self, PrefixError> {
        ....
```

(Roughly speaking, our approach has been "real error types everywhere except `main.rs`", which means anyone using this crate as a library gets real error types)

mountpoint-s3/src/prefix.rs
```diff
@@ -0,0 +1,65 @@
...
+        } else {
+            Ok(Self {
+                path: prefix.to_owned(),
+            })
+        }
+    }
+
+    pub fn is_valid(prefix: &str) -> bool {
+        prefix.is_empty() || prefix.ends_with('/')
+    }
```
jamesbornholt, line 25: If you do the other comment in `lookup`, this doesn't need to be pub any more.

mountpoint-s3/src/prefix.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+impl Display for Prefix {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!("{}", self.path))
+    }
+}
+
+impl FromStr for Prefix {
+    type Err = anyhow::Error;
```
jamesbornholt, line 35: Same thing here about the error type

mountpoint-s3/src/prefix.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    pub fn is_valid(prefix: &str) -> bool {
+        prefix.is_empty() || prefix.ends_with('/')
+    }
+}
+
+impl Display for Prefix {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!("{}", self.path))
```
jamesbornholt, line 30:
```suggestion
        write!(f, "{}", self.path)
```

jamesbornholt (commented):
Neat. I like this approach a bunch!

mountpoint-s3/src/prefix.rs
passaro: ah, much simpler! :)

mountpoint-s3/src/prefix.rs
passaro: Removed last usage of Option<Prefix> in CliArgs (using default_value = "" instead)

mountpoint-s3/src/prefix.rs
```diff
@@ -0,0 +1,71 @@
...
+    #[test_case(""; "empty string")]
+    #[test_case("hello/"; "ending in slash")]
+    #[test_case("hello/world/"; "nested folder ending in slash")]
+    #[test_case(" /"; "whitespace ending in slash")]
+    #[test_case("/"; "single slash")]
+    #[test_case("//"; "double slash")]
+    #[test_case("/hello/"; "starting with slash")]
+    fn test_valid_prefix(prefix: &str) {
+        assert!(Prefix::is_valid(prefix))
+    }
```
jamesbornholt, line 70: One nitpick: for these tests we should test `Prefix::new` instead of `Prefix::is_valid`, which is now an implementation detail of `new`.

mountpoint-s3/src/prefix.rs
```diff
@@ -0,0 +1,71 @@
+use std::fmt::Display;
+use std::str::FromStr;
+
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum PrefixError {
+    #[error("prefixes must end in '/'")]
```
jamesbornholt, line 8: Fixing my own grammar mistake: this should be "prefix must end in '/'"

jamesbornholt (commented):
Two more tiny things I missed, sorry!

jamesbornholt (approved):

Issue #27, opened 2022-12-02
Create new files (sequentially written)


jamesbornholt:
>Creating new files with sequential writes only

- [x] Support `PutObject` (#22)
- [x] Implement open/write/close operations (#119)
- [x] Make opened files visible in `readdir`
- [x] Use MPU for large object uploads

PR #189, opened 2023-03-31
Make opened files visible in readdir

monthonk:
Currently, new files that are opened for write will not be visible in the file system until they are closed and their contents are uploaded to S3, but we want them to be visible as soon as we open the files.

To do this, we will store inodes of the opened files in its parent state and remove them once we finish writing. Then, on readdir, we can append these inodes to the results we get from LIST API to make the opened files visible in the file system.

Another problem is results ordering. Previously, we're using a double-ended queue to store the sorted LIST results but it doesn't work well with newly created files since we don't know their exact positions in paginated results. In this commit, we will use two queues, one for local results, and another one for remote results. Each queue is sorted by file names, then we compare the first element in both lists to decide which to return next.

A part of https://github.com/awslabs/mountpoint-s3/issues/27.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -277,8 +277,7 @@ where
                 return Err(libc::EINVAL);
             }
 
-            let parent_lookup = self.superblock.getattr(&self.client, lookup.inode.parent()).await?;
-            lookup.inode.start_writing(&parent_lookup.inode)?;
+            lookup.inode.start_writing()?;
             FileHandleType::Write {
                 parts: Default::default(),
             }
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -279,7 +279,6 @@ where
             }
 
             let inode_handle = self.superblock.write(&self.client, ino, lookup.inode.parent()).await?;
-            inode_handle.start_writing()?;
             FileHandleType::Write {
                 parts: Default::default(),
                 handle: inode_handle,
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -8,7 +8,7 @@ use tracing::{debug, error, trace};
 use fuser::{FileAttr, KernelConfig};
 use mountpoint_s3_client::{ObjectClient, PutObjectParams};
 
-use crate::inode::{Inode, InodeError, InodeKind, LookedUp, ReaddirHandle, Superblock};
+use crate::inode::{Inode, InodeError, InodeKind, LookedUp, ReaddirHandle, Superblock, WriteHandle};
 use crate::prefetch::{PrefetchGetObject, PrefetchReadError, Prefetcher, PrefetcherConfig};
 use crate::prefix::Prefix;
 use crate::sync::atomic::{AtomicI64, AtomicU64, Ordering};
@@ -51,6 +51,7 @@ enum FileHandleType<Client: ObjectClient, Runtime> {
     },
     Write {
         parts: AsyncMutex<Vec<Box<[u8]>>>,
+        handle: WriteHandle,
     },
 }
 
@@ -277,9 +278,11 @@ where
                 return Err(libc::EINVAL);
             }
 
-            lookup.inode.start_writing()?;
+            let inode_handle = self.superblock.write(&self.client, ino, lookup.inode.parent()).await?;
+            inode_handle.start_writing()?;
             FileHandleType::Write {
                 parts: Default::default(),
+                handle: inode_handle,
             }
         } else {
             lookup.inode.start_reading()?;
@@ -394,7 +397,7 @@ where
             return Err(libc::EBADF);
         };
         let mut request = match &handle.typ {
-            FileHandleType::Write { parts } => parts.lock().await,
+            FileHandleType::Write { parts, .. } => parts.lock().await,
             FileHandleType::Read { .. } => return Err(libc::EBADF),
         };
 
@@ -512,18 +515,18 @@ where
         _lock_owner: Option<u64>,
         _flush: bool,
     ) -> Result<(), libc::c_int> {
-        let handle = {
+        let file_handle = {
             let mut file_handles = self.file_handles.write().await;
             file_handles.remove(&fh).ok_or(libc::EBADF)?
         };
 
-        match handle.typ {
-            FileHandleType::Write { parts } => {
+        match file_handle.typ {
+            FileHandleType::Write { parts, handle } => {
                 // TODO how do we make sure we didn't already handle this via `flush`?
                 let parts = parts.into_inner();
                 let size = parts.iter().map(|part| part.len()).sum::<usize>();
                 let stream = futures::stream::iter(parts);
-                let key = handle.full_key;
+                let key = file_handle.full_key;
 
                 let put = self
                     .client
@@ -542,14 +545,13 @@ where
                     }
                 };
 
-                let parent_lookup = self.superblock.getattr(&self.client, handle.inode.parent()).await?;
-                handle.inode.finish_writing(size, &parent_lookup.inode)?;
+                handle.finish_writing(size)?;
 
                 result
             }
             FileHandleType::Read { request: _ } => {
                 // TODO make sure we cancel the inflight PrefetchingGetRequest. is just dropping enough?
-                handle.inode.finish_reading()?;
+                file_handle.inode.finish_reading()?;
                 Ok(())
             }
         }
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -277,7 +277,8 @@ where
                 return Err(libc::EINVAL);
             }
 
-            lookup.inode.start_writing()?;
+            let parent_lookup = self.superblock.getattr(&self.client, lookup.inode.parent()).await?;
+            lookup.inode.start_writing(&parent_lookup.inode)?;
             FileHandleType::Write {
                 parts: Default::default(),
             }
@@ -542,7 +543,8 @@ where
                     }
                 };
 
-                handle.inode.finish_writing(size)?;
+                let parent_lookup = self.superblock.getattr(&self.client, handle.inode.parent()).await?;
+                handle.inode.finish_writing(size, &parent_lookup.inode)?;
 
                 result
             }
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -20,7 +20,7 @@
 //! Some cached state is dependent on the inode kind; that state is hidden behind a [InodeStatKind]
 //! enum.
 
-use std::collections::{HashMap, VecDeque};
+use std::collections::{BinaryHeap, HashMap, HashSet};
 use std::ffi::{OsStr, OsString};
 use std::os::unix::prelude::OsStrExt;
 use std::time::Instant;
@@ -82,6 +82,7 @@ impl Superblock {
                 write_status: WriteStatus::Remote,
                 kind_data: InodeKindData::Directory {
                     children: Default::default(),
+                    writing_children: Default::default(),
                 },
             }),
         };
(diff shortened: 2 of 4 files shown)


mountpoint-s3/src/fs.rs
```diff
@@ -282,7 +282,8 @@ where
                 return Err(libc::EINVAL);
             }
 
-            lookup.inode.start_writing()?;
+            let parent_lookup = self.superblock.getattr(&self.client, lookup.inode.parent()).await?;
+            lookup.inode.start_writing(&parent_lookup.inode)?;
```
jamesbornholt, line 286: I think we probably want to do this little "connect it to the parent" dance in `mknod`? That's when the file is "created". This `start_writing` stuff is just to decide who wins the race to start actually writing into the file.

mountpoint-s3/src/inode.rs
```diff
@@ -312,7 +312,7 @@ impl Superblock {
         // Check again for the child now that the parent is locked, since we might have lost to a
         // racing lookup. (It would be nice to lock the parent and *then* lookup, but we'd have to
         // hold that lock across the remote API calls).
-        let InodeKindData::Directory { children } = &mut parent_state.kind_data else {
+        let InodeKindData::Directory { children, writing_children: _ } = &mut parent_state.kind_data else {
```
jamesbornholt, line 315: nit, if you prefer:
```suggestion
        let InodeKindData::Directory { children, .. } = &mut parent_state.kind_data else {
```

mountpoint-s3/src/inode.rs
```diff
@@ -487,7 +497,10 @@ impl SuperblockInner {
                 debug_assert!(false, "inodes never change kind");
                 return Err(InodeError::NotADirectory(parent.ino()));
             }
-            InodeKindData::Directory { children } => {
+            InodeKindData::Directory {
+                children,
+                writing_children: _,
+            } => {
```
jamesbornholt, line 503: Here and everywhere else:
```suggestion
            InodeKindData::Directory { children, .. } => {
```

mountpoint-s3/src/inode.rs
```diff
@@ -507,6 +520,31 @@ pub struct LookedUp {
...
+impl PartialEq for LookedUp {
+    fn eq(&self, other: &Self) -> bool {
+        self.inode.ino() == other.inode.ino()
+            && self.inode.parent() == other.inode.parent()
+            && self.inode.name() == other.inode.name()
+            && self.inode.full_key() == other.inode.full_key()
+            && self.inode.kind() == other.inode.kind()
+            && self.stat == other.stat
+    }
+}
```
jamesbornholt, line 546: `PartialOrd` and `PartialEq` [need to be consistent with each other](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html), which these aren't. In particular, the first rule is that `a == b` if and only if `partial_cmp(a, b) == Some(Equal)`, which isn't true here.

mountpoint-s3/src/inode.rs
```diff
@@ -515,7 +553,10 @@ pub struct ReaddirHandle {
     parent_ino: InodeNo,
     full_path: String,
     page_size: usize,
-    results: RwLock<VecDeque<LookedUp>>,
+    // Use binary heap as a priority queue
+    results: RwLock<BinaryHeap<LookedUp>>,
+    // A set of files that only exist locally, we use them for deduplication
+    local_results: RwLock<HashSet<InodeNo>>,
```
jamesbornholt, line 559: Just a suggestion of a different approach: instead of a priority queue plus a set, we just have two `VecDeque`s, one "remote" and one "local". Then `ReaddirHandle::next` effectively merge-sorts them—refills the "remote" queue if it's empty, and then compares the first element in both lists to decide which to return next, breaking the tie if necessary.

Not sure which of these approaches is clearer, though? My intuition here is that the priority queue isn't quite what we want because entries never get inserted anywhere except the bottom, but perhaps it's worth the trouble anyway.

mountpoint-s3/src/inode.rs
```diff
@@ -540,12 +581,45 @@ impl ReaddirStreamState {
...
+                if *next_token == ReaddirStreamState::NotStarted {
+                    let inode = self.inner.get(self.dir_ino)?;
+                    let kind_data = &inode.inner.sync.read().unwrap().kind_data;
+                    let local_files = match kind_data {
+                        InodeKindData::File { .. } => unreachable!("we know this is a directory"),
+                        InodeKindData::Directory {
+                            children: _,
+                            writing_children,
+                        } => writing_children.iter().map(|ino| {
+                            let inode = self.inner.get(*ino)?;
```
jamesbornholt, line 600: We'll have to think later about the right behavior here when files can be unlinked (== inodes can disappear).

mountpoint-s3/tests/fuse_tests/readdir_test.rs
```diff
@@ -30,38 +33,107 @@ fn prepare_fs(mut put_object_fn: PutObjectFn, map: &HashMap<String, File>) {
...
+}
+
+#[test]
+fn readdir_mock() {
+    readdir(crate::fuse_tests::mock_session::new, "");
+}
+
+#[test_case(vec!["aaa", "file150", "file151", "zzz"]; "new files at various positions")]
+fn readdir_while_writing_mock(new_files: Vec<&str>) {
+    readdir_while_writing(crate::fuse_tests::mock_session::new, "readdir_s3", new_files);
```
jamesbornholt, line 138: This behavior might be worth writing either a proptest or a randomized test for. The interaction between the ListObjects pagination and which objects are being written is pretty subtle with a few corner cases, so these testing approaches might give us more confidence we're covering the corner cases appropriately.

mountpoint-s3/src/fs.rs
```diff
@@ -547,7 +548,8 @@ where
                     }
                 };
 
-                handle.inode.finish_writing(size)?;
+                let parent_lookup = self.superblock.getattr(&self.client, handle.inode.parent()).await?;
+                handle.inode.finish_writing(size, &parent_lookup.inode)?;
```
jamesbornholt, line 552: It might be a little cleaner to push the parent lookup into `finish_writing`, maybe taking a ref to the superblock as input instead.

Or we could give every inode a pointer back to its superblock, but that creates an annoying cycle that means the superblock can never be freed (superblock points to inode, which points back to superblock).

Another option would be to move these state transition methods onto `Superblock`.

jamesbornholt (commented):
Cool! A few suggestions.

mountpoint-s3/tests/fuse_tests/readdir_test.rs
monthonk: Agree that we will need a proptest for this, and I found another problem reported in https://github.com/awslabs/mountpoint-s3/issues/190, it seems like our current pagination logic might not be correct?

mountpoint-s3/src/fs.rs
monthonk: What if we have a new type of handle to control state of the write, let's say `WriteHandle`, it can be used the same way with `ReaddirHandle` and could be extended when we support multipart upload. Does that make sense?

jamesbornholt: Well, we already have a write handle in the `S3Filesystem`, and I assume that's where the MPU stuff will live, since the inode doesn't care about the contents of the file. We could add a separate token type to encode the state machine that `start_writing`/`finish_writing` represents. But the problem I'm trying to solve here is that now those transitions need a reference to the superblock since they need to do a lookup. I'd rather the filesystem didn't do that lookup itself, because it leaks an implementation detail of the inode (that it needs to know the parent inode when making write transitions).

monthonk: By write the handle we already have, do you mean `FileHandle`? I think it's the same problem because we don't have a reference from file handles to the superblock. But, for directories, I see that we have `DirHandle` and then there is [`ReaddirHandle`](https://github.com/awslabs/mountpoint-s3/blob/286541e395d5dd774fd4de70644b5eaa20c5ec5f/mountpoint-s3/src/inode.rs#L509-L519) inside, which holds the reference to the superblock to do readdir stuff.

jamesbornholt: That DirHandle -> ReaddirHandle indirection is because the DirHandle (which is a filesystem concept) shouldn't know about the implementation details of the ReaddirHandle (which is an inode concept). It's kind of the same thing here: the FileHandleType (which is a filesystem concept) shouldn't know about the implementation details of the inode. So, yeah, I suppose a new WriteHandle that holds a reference to the superblock would make sense here.

mountpoint-s3/tests/fuse_tests/readdir_test.rs
monthonk: These tests are now using randomized file names.

mountpoint-s3/src/inode.rs
```diff
@@ -254,6 +253,22 @@ impl Superblock {
         Ok(LookedUp { inode, stat })
     }
 
+    /// Create a new write handle to be used for state transition
+    pub async fn write<OC: ObjectClient>(
```
jamesbornholt, line 257: Is there ever a case where we'd call this without immediately calling `start_writing` afterwards? If not we should just fold that transition into this constructor.

mountpoint-s3/tests/fuse_tests/readdir_test.rs
```diff
@@ -30,38 +33,112 @@ fn prepare_fs(mut put_object_fn: PutObjectFn, map: &HashMap<String, File>) {
...
+    let readdir_size = 5;
+    let filesystem_config = S3FilesystemConfig {
+        readdir_size,
+        ..Default::default()
+    };
+
+    let mut map = HashMap::new();
+    let mut expected_list = Vec::new();
+    for i in 0..readdir_size * 4 {
+        let random_str = Alphanumeric.sample_string(&mut rand::thread_rng(), 5);
```
jamesbornholt, line 83: We should use a fixed seed for the RNG here (and everywhere we use `rand::thread_rng()`) so test failures are reproducible. Then we should probably run this whole test in a loop, at least in the mock case, so we can get coverage of different random seeds.

mountpoint-s3/src/inode.rs
```diff
@@ -627,18 +742,41 @@ impl ReaddirHandle {
...
 
+    fn compare_and_get_next(&self) -> Option<LookedUp> {
+        let mut local_locked = self.local_results.write().unwrap();
+        let mut remote_locked = self.remote_results.write().unwrap();
+        match (local_locked.front(), remote_locked.front()) {
+            (None, None) => None,
+            (None, Some(_)) => remote_locked.pop_front(),
+            (Some(_), None) => local_locked.pop_front(),
+            (Some(local_lookup), Some(remote_lookup)) => {
+                let ordering = local_lookup.inode.name().cmp(remote_lookup.inode.name());
```
jamesbornholt, line 765: nit: lift the comment up here about how we need to consider the case where both inodes have the same name, and how we resolve that tie

mountpoint-s3/src/inode.rs
```diff
@@ -540,12 +629,42 @@ impl ReaddirStreamState {
...
+                            return Err(e);
+                        }
+                    }
+                }
+
                 if *next_token == ReaddirStreamState::Finished {
                     trace!(self=?self as *const _, "readdir finished");
-                    return Ok(None);
+                    return Ok(self.compare_and_get_next());
                 }
```
jamesbornholt, line 668: Can we write a unit test (probably in this file) specifically to cover the cases around this logic?
1. Remote directory is totally empty
2. All local keys come after all remote keys

monthonk: Added these tests.

mountpoint-s3/src/inode.rs
monthonk: Done.

mountpoint-s3/src/inode.rs
monthonk: Yeah, we probably don't have such case, already moved it.

mountpoint-s3/tests/fuse_tests/readdir_test.rs
monthonk: Updated these tests.

Jcc2488:
Stop

jamesbornholt (approved):

PR #192, opened 2023-04-04
Allow `mountpoint-s3-crt-sys` to link to an existing CRT installation

jamesbornholt:
Some customers may already have a build of the CRT they'd like to use with Mountpoint rather than building the one we include; for example, one installed by their package manager. This change lets us specify environment variables pointing to an existing CRT build rather than always building it ourselves.

There's not a great versioning or compatibility story for CRT shared libraries, so this change should only be used when the customer can guarantee a suitable CRT version will be linked.

---

### DIFF
--- a/mountpoint-s3-crt-sys/build.rs
+++ b/mountpoint-s3-crt-sys/build.rs
@@ -1,13 +1,13 @@
 use std::path::{Path, PathBuf};
 use std::{env, fs};
 
-use bindgen::{BindgenError, Bindings};
+use bindgen::BindgenError;
 use rustflags::Flag;
 
 /// Path to the CRT repos, relative to the crate root.
 const CRT_PATH: &str = "crt";
 
-/// On Linux, the CRT needs to build and link against Amazon's libcrypto. On Darwin it uses the
+/// On Linux, the CRT gets its TLS stack from AWS libcrypto and s2n. On Darwin it uses the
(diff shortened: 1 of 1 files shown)


passaro:
Should we document the usage of `MOUNTPOINT_CRT_LIB_DIR` and `MOUNTPOINT_CRT_INCLUDE_DIR` somewhere (other than in the comment on `main`)?

monthonk (commented):
I wonder how we can be sure that the build will be working correctly if we set these variables. Is it possible to add a new CI that has the CRT pre-installed then try to build and test Mountpoint with this approach?

jamesbornholt:
> Should we document the usage of `MOUNTPOINT_CRT_LIB_DIR` and `MOUNTPOINT_CRT_INCLUDE_DIR` somewhere (other than in the comment on `main`)?

I'd prefer not to document it for now, given the compatibility concerns around CRT versioning, and since (AFAIK) there are no public prebuilt distributions of the CRT beyond what's embedded in the SDKs.

> I wonder how we can be sure that the build will be working correctly if we set these variables. Is it possible to add a new CI that has the CRT pre-installed then try to build and test Mountpoint with this approach?

Yeah, good idea. Let me look into this in a separate PR, as I think it's going to end up involving duplicating some of the CRT repos' CI scripts.

monthonk (approved):

PR #200, opened 2023-04-11
Add error handling for FUSE worker threads that panic

dannycjones:
When encountering a panic in a FUSE operation, the filesystem was not exiting cleanly. We were not handling the panicking thread case before this change. With this change, we log at error and allow the unmount process to continue.

---

### DIFF
--- a/mountpoint-s3/src/fuse/session.rs
+++ b/mountpoint-s3/src/fuse/session.rs
@@ -52,10 +52,22 @@ impl FuseSession {
                 .name("fuse-worker-waiter".to_owned())
                 .spawn(move || {
                     for thd in workers {
-                        let result = thd.join().unwrap();
-                        if let Err(e) = result {
-                            error!("fuse worker thread failed: {e:?}");
-                        }
+                        let thread_name = thd
+                            .thread()
+                            .name()
+                            .expect("worker threads should have names")
+                            .to_owned();
+
+                        match thd.join() {
+                            Err(_panic_param) => {
+                                error!("thread {thread_name} panicked");
+                            }
+                            Ok(thd_result) => {
+                                if let Err(fuse_worker_error) = thd_result {
+                                    error!("thread {thread_name} failed: {fuse_worker_error:?}");
+                                }
+                            }
+                        };
                     }
 
                     let _ = tx.send(Message::WorkersExited);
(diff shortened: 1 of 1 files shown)


mountpoint-s3/src/fuse/session.rs
```diff
@@ -52,10 +52,22 @@ impl FuseSession {
...
                 .spawn(move || {
                     for thd in workers {
-                        let result = thd.join().unwrap();
-                        if let Err(e) = result {
-                            error!("fuse worker thread failed: {e:?}");
-                        }
+                        let thread_name = thd
+                            .thread()
+                            .name()
+                            .expect("worker threads should have names")
```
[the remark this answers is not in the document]
jamesbornholt: Yeah, let's use a default here—double-panics are a huge pain to debug.

mountpoint-s3/src/fuse/session.rs
```diff
@@ -52,10 +52,22 @@ impl FuseSession {
...
+                            .to_owned();
+
+                        match thd.join() {
+                            Err(_panic_param) => {
```
passaro, line 62: You could use `downcast_ref::<String>()` on the parameter to get the message to report.

dannycjones:
I'll provide the default rather than using expect on thread name.

I'll also take a look into trying to print some part of the panic message based on the info, thanks Alex.

Will work on these tomorrow.

mountpoint-s3/src/fuse/session.rs
dannycjones: Makes sense. I found that it won't work if the panic was called with a string literal only (i.e. it was called with `&str`), so its a little more complex but I think OK.

dannycjones:
Just pushed the latest code.

Here's the logs with the new changes:

```
thread 'fuse-worker-0' panicked at 'uh oh 2', /Users/djonesoa/devel/mountpoint-s3/mountpoint-s3/src/fuse.rs:126:9
2023-04-12T13:49:35.109854Z  WARN fuser::reply: Reply not sent for operation 5, replying with I/O error    
2023-04-12T13:49:35.109965Z  INFO fuser::session: Unmounted /Users/djonesoa/mnt/s3-file-connector/    
2023-04-12T13:49:35.110354Z ERROR mountpoint_s3::fuse::session: worker thread panicked thread_name="fuse-worker-0" panic_msg="uh oh 2"
```

If the thread name (or attempt at downcasting the panic message) returns none, the args won't appear.

Issue #77, opened 2023-02-09
Support for mkdir


dannycjones:
>Creation of new directories with Mountpoint for Amazon S3

This ticket tracks support for the `mkdir` operation.

Note that we need to define the semantics of the `mkdir` operation. For example, is it local inode changes only?

PR #202, opened 2023-04-11
Implement mkdir

passaro:
Address #77 by creating local-only directories:
* `mkdir` will create a new empty directory in the file system, but not affect the content of the bucket (i.e. no "directory marker" is created in the bucket).
* If a file is created under the new directory or a nested directory, the local directory will revert to a normal, implicit directory.

The new behavior is implemented using the same mechanism for writing files: a directory is created with a `WriteStatus::LocalUnopened` state and added to parent's `writing_files` set. When a new file is uploaded and its state transition to `Remote` in `finish_writing`, we also traverse its ancestors to transition local directories to `Remote`.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -373,7 +373,30 @@ where
             return Err(libc::EINVAL);
         }
 
-        let lookup = self.superblock.create(&self.client, parent, name).await?;
+        let lookup = self
+            .superblock
+            .create(&self.client, parent, name, InodeKind::File)
+            .await?;
+        let attr = self.make_attr(&lookup);
+
+        Ok(Entry {
+            ttl: self.config.stat_ttl,
+            attr,
+            generation: 0,
+        })
+    }
+
+    pub async fn mkdir(
+        &self,
+        parent: InodeNo,
+        name: &OsStr,
+        _mode: libc::mode_t,
+        _umask: u32,
+    ) -> Result<Entry, libc::c_int> {
+        let lookup = self
+            .superblock
+            .create(&self.client, parent, name, InodeKind::Directory)
+            .await?;
         let attr = self.make_attr(&lookup);
 
         Ok(Entry {
--- a/mountpoint-s3/src/fuse.rs
+++ b/mountpoint-s3/src/fuse.rs
@@ -223,6 +223,17 @@ where
         }
     }
 
+    #[instrument(level="debug", skip_all, fields(req=_req.unique(), parent=parent, name=?name))]
+    fn mkdir(&self, _req: &Request<'_>, parent: u64, name: &OsStr, mode: u32, umask: u32, reply: ReplyEntry) {
+        // mode_t is u32 on Linux but u16 on macOS, so cast it here
+        let mode = mode as libc::mode_t;
+
+        match block_on(self.fs.mkdir(parent, name, mode, umask).in_current_span()) {
+            Ok(entry) => reply.entry(&entry.ttl, &entry.attr, entry.generation),
+            Err(e) => reply.error(e),
+        }
+    }
+
     #[instrument(level="debug", skip_all, fields(req=_req.unique(), ino=ino, fh=fh, offset=offset, length=data.len()))]
     fn write(
         &self,
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -612,15 +612,18 @@ impl WriteHandle {
         }
     }
 
-    /// Update status of the inode and its parent
+    /// Update status of the inode and of containing "local" directories.
     pub fn finish_writing(self, object_size: usize) -> Result<(), InodeError> {
         let inode = self.inner.get(self.ino)?;
 
-        // Ancestor inodes, from parent to first remote ancestor
+        // Collect ancestor inodes that may need updating,
+        // from parent to first remote ancestor.
         let ancestors = {
             let mut ancestors = Vec::new();
             let mut ancestor_ino = self.parent_ino;
+            let mut visited = HashSet::new();
             loop {
+                assert!(visited.insert(ancestor_ino), "cycle detected in inode ancestors");
                 let ancestor = self.inner.get(ancestor_ino)?;
                 ancestors.push(ancestor.clone());
                 if ancestor.ino() == ROOT_INODE_NO
@@ -633,8 +636,7 @@ impl WriteHandle {
             ancestors
         };
 
-        // acquire locks on ancestors first
-        // ancestors_states goes from first remote ancestor to parent
+        // Acquire locks on ancestors in descending order to avoid deadlocks.
         let mut ancestors_states: Vec<_> = ancestors
             .iter()
             .rev()
@@ -647,7 +649,8 @@ impl WriteHandle {
                 state.write_status = WriteStatus::Remote;
                 state.stat.size = object_size;
 
-                // traverse ancestors from parent to first remote ancestor
+                // Walk up the ancestors from parent to first remote ancestor to transition
+                // the inode and all "local" containing directories to "remote".
                 let children_inos = std::iter::once(self.ino).chain(ancestors.iter().map(|ancestor| ancestor.ino()));
                 for (ancestor_state, child_ino) in ancestors_states.iter_mut().rev().zip(children_inos) {
                     match &mut ancestor_state.kind_data {
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -80,10 +80,7 @@ impl Superblock {
             sync: RwLock::new(InodeState {
                 stat: InodeStat::for_directory(mount_time, Instant::now()), // TODO expiry
                 write_status: WriteStatus::Remote,
-                kind_data: InodeKindData::Directory {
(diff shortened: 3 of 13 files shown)


mountpoint-s3/src/inode.rs
```diff
@@ -351,6 +351,59 @@ impl Superblock {
 
         Ok(LookedUp { inode, stat })
     }
+
+    /// Create a new directory inode
+    pub async fn createdir<OC: ObjectClient>(
```
passaro, line 356: This is copied from `create`, with the only changes in `kind`/`kind_data`. Planning to refactor if approach confirmed.

mountpoint-s3/src/fs.rs
```diff
@@ -370,6 +370,33 @@ where
         })
     }
 
+    pub async fn mkdir(
```
passaro, line 373: This is *very* similar to `mknod`, but I assume we want to keep 1-1 mapping between methods in `S3Filesystem` and `S3FuseFilesystem`.

dannycjones (commented):
Just some very high level comments. Would like @monthonk's feedback here since he's just explored the local-only changes with readdir.

mountpoint-s3/src/inode.rs
```diff
@@ -351,6 +351,59 @@ impl Superblock {
...
+        let expiry = Instant::now(); // TODO local inode stats never expire?
+        let stat = InodeStat::for_directory(OffsetDateTime::now_utc(), expiry);
+        let kind = InodeKind::Directory;
+        let state = InodeState {
+            stat: stat.clone(),
+            kind_data: InodeKindData::Directory {
+                children: Default::default(),
+                writing_children: Default::default(),
+            },
+            write_status: WriteStatus::LocalUnopened,
```
monthonk, line 398: Not sure what status we should be using, but we might want to update this to `WriteStatus::Remote` when the first file is created.

monthonk (commented):
Looks good, only have a few comments but I think you can continue this path.

jamesbornholt (commented):
Couple of things:
- How do we think `readdir` should work on local directories? Can we skip the `ListObjects` call in those cases?
- How to think about concurrent mutations -- what if the remote side creates a new object that overlaps the local directory?

mountpoint-s3/src/inode.rs
```diff
@@ -351,6 +351,59 @@ impl Superblock {
...
+        name: &OsStr,
+    ) -> Result<LookedUp, InodeError> {
+        trace!(parent=?dir, ?name, "createdir");
+
+        let existing = self.lookup(client, dir, name).await;
+        match existing {
+            Ok(lookup) => return Err(InodeError::FileAlreadyExists(lookup.inode.ino())),
+            Err(InodeError::FileDoesNotExist) => (),
+            Err(e) => return Err(e),
+        }
```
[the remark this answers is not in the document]
passaro: Not too sure about new errors: wouldn't `FileAlreadyExists' already cover directories as well? Even if we wanted to differentiate, I think they would map to the same error code anyway.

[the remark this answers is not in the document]
dannycjones: Agree it'll return the same error.

I'm just wondering how we can make the error type's meaning clearer. Maybe a Rustdoc on it, or renaming "FileAlreadyExists" as "PathAlreadyExists" and "FileDoesNotExist" as "EntryDoesNotExist".

mountpoint-s3/src/fuse.rs
```diff
@@ -223,6 +223,17 @@ where
         }
     }
 
+    #[instrument(level="debug", skip_all, fields(req=_req.unique(), parent=parent, name=?name))]
+    fn mkdir(&self, _req: &Request<'_>, parent: u64, name: &OsStr, mode: u32, umask: u32, reply: ReplyEntry) {
```
sauraank, line 227: Why mkdir does not have rdev as one of the arguments unlike mknod? I can see rdev is used to query/set image root device, RAM disk size, or video mode. I dont find this specific to file or directory.

sauraank (commented):
Just wanted to know, where are these shell commands like mkdir, touch etc mapped to these methods? Is it in the fuse?

mountpoint-s3/src/fuse.rs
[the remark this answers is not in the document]
passaro: `mknod` can be used to create various type of inodes. Some types like device special files use the device parameter, but we only support regular files, so we ignore `rdev` anyway.
`mkdir` is specific to directories, so it does not need the parameter.

In any case, here we are implementing the `FileSystem` trait from `fuser`, so just using the signature from there.

mountpoint-s3/src/inode.rs
passaro: Transition from local to remote implemented in `finish_writing` of nested files.

passaro:
Added integration tests for `mkdir`. The mock/s3 test session only offered a PutObject function, I've extended them with a few more functions and created a `TestClient` trait. Any thoughts on the approach?

mountpoint-s3/src/inode.rs
```diff
@@ -354,15 +352,16 @@ impl Superblock {
         }
 
         let expiry = Instant::now(); // TODO local inode stats never expire?
-                                     // Objects don't have an ETag until they are uploaded to S3
-        let stat = InodeStat::for_file(0, OffsetDateTime::now_utc(), expiry, None);
-        let kind = InodeKind::File;
+        let datetime = OffsetDateTime::now_utc();
+        let stat = match kind {
+            InodeKind::File => InodeStat::for_file(0, datetime, expiry, None), // Objects don't have an ETag until they are uploaded to S3
+            InodeKind::Directory => InodeStat::for_directory(datetime, expiry),
```
monthonk, line 358: lookup and readdir use mount time for directory stat, should we use the same value here?

mountpoint-s3/src/inode.rs
```diff
@@ -587,6 +578,18 @@ impl WriteHandle {
 
     /// Update status of the inode and its parent
     pub fn finish_writing(self, object_size: usize) -> Result<(), InodeError> {
+        fn remove_from_writing_children(parent_state_data: &mut InodeKindData, ino: &InodeNo) {
+            match parent_state_data {
+                InodeKindData::File { .. } => unreachable!("we know parent is a directory"),
+                InodeKindData::Directory {
+                    children: _,
+                    writing_children,
+                } => {
```
monthonk, line 587: nit: 
```suggestion
                InodeKindData::Directory { writing_children, .. } => {
```

mountpoint-s3/src/inode.rs
```diff
@@ -587,6 +578,18 @@ impl WriteHandle {
 
     /// Update status of the inode and its parent
     pub fn finish_writing(self, object_size: usize) -> Result<(), InodeError> {
+        fn remove_from_writing_children(parent_state_data: &mut InodeKindData, ino: &InodeNo) {
```
monthonk, line 581: I would prefer moving this function out.

mountpoint-s3/src/inode.rs
```diff
@@ -597,13 +600,22 @@ impl WriteHandle {
...
-                        writing_children.remove(&inode.ino());
+                remove_from_writing_children(&mut parent_state.kind_data, &inode.ino());
+                if parent_state.write_status != WriteStatus::Remote {
+                    parent_state.write_status = WriteStatus::Remote;
+
+                    let mut ancestor_ino = parent.inner.parent;
+                    let mut child_ino = parent.ino();
+                    loop {
+                        let ancestor = self.inner.get(ancestor_ino)?;
+                        let mut ancestor_state = ancestor.inner.sync.write().unwrap();
```
monthonk, line 611: I'm a bit worried here since we're still holding a lock on parent_state and then we try to lock its ancestors. Could deadlock happen from this situation?

mountpoint-s3/tests/fs.rs
```diff
@@ -445,3 +445,46 @@ async fn test_stat_block_size() {
...
+        .fh;
+
+    fs.release(file_ino, file_handle, 0, None, false).await.unwrap();
+
+    // Remove the new object from the client
+    client.remove_object(&format!("{prefix}{dirname}/{filename}"));
+
+    // Verify that the directory disappeared
+    let lookup = fs.lookup(FUSE_ROOT_INODE, dirname.as_ref()).await;
+    assert!(matches!(lookup, Err(libc::ENOENT)));
```
monthonk, line 489: This behavior should be documented somewhere, maybe in semantics doc.

monthonk (commented):
I like your approach on the test! added a few more comments.

mountpoint-s3/src/inode.rs
passaro: I think you are right. Updated to lock the ancestors from top to bottom (and removed helper function).

mountpoint-s3/tests/fs.rs
passaro: I'm adding a section for `mkdir` to the semantics doc. Should I add it to this PR or a new one?

monthonk: Either is fine.

monthonk (dismissed):
LGTM

jamesbornholt (approved):

PR #204, opened 2023-04-11
Introduced new cli options

eslrahc-swa:
Introduced Cli options.

* --no-sign-request: allow customer to omit credential.
* --profile: customer can specify credentials name.

---


mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -42,6 +54,25 @@ impl CredentialsProvider {
 
         Ok(Self { inner })
     }
+
+    /// Creates the profile credential provider.
+    pub fn new_profile(allocator: &Allocator, options: &CredentialsProviderProfileOptions) -> Result<Self, Error> {
+        auth_library_init(allocator);
+
+        // SAFETY: CredentialProvider does not modify contents of profile_name_override,
```
passaro, line 62: Not sure I understand the safety comment: isn't the question really whether `aws_credentials_provider_new_profile` makes a deep copy of the options (and thus of `profile_name_override`)?

eslrahc-swa: There are 2 things needs to point out, to my understanding, 
* Why doe  `options.profile_name_override.as_aws_byte_cursor()` is a valid convert? by valid it means `CredentialsProvider` would not modify the contents referred by `profile_name_override`. No modified profile contents should pass in CRT.
* Once converting `profile_name_override` to `aws_byte_cursor`, `CredentialsProvider` must live at least as long as CRT as CRT might continue use the byte cursor.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -49,6 +51,8 @@ pub(crate) mod put_object;
 
 #[derive(Debug, Clone, Default)]
 pub struct S3ClientConfig {
+    pub profile_name_override: Option<String>,
```
passaro, line 54: Aren't these options mutually exclusive? Shall we enforce it here and in CLI options?

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -89,12 +93,27 @@ impl S3CrtClient {
 
         let mut client_bootstrap = ClientBootstrap::new(&allocator, &bootstrap_options).unwrap();
 
-        let creds_options = CredentialsProviderChainDefaultOptions {
-            bootstrap: &mut client_bootstrap,
+        let mut client_config = ClientConfig::new();
+
+        let mut credentials_provider = match config.profile_name_override {
```
passaro, line 98: Should this be moved under the if at line 113?

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -89,12 +93,27 @@ impl S3CrtClient {
...
+            Some(profile_name_override) => {
+                let credentials_profile_options = CredentialsProviderProfileOptions {
+                    profile_name_override: &profile_name_override,
+                };
```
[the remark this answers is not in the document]
eslrahc-swa: `client_bootstrap` usage to profile based credential is for customer who uses assume role with STS. My understanding now is customer only uses `IAM Role` or `IAM User` as per `README.md` mentioned. 

However, I think we should add client_bootstrap to support any customer use assume role and sending http/https request to  STS

mountpoint-s3-client/src/s3_crt_client.rs
eslrahc-swa: It should not, we need to bump ref count of `CredentialProvider` to 1 all the way until `S3CrtClient` owns the client config.

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -15,6 +20,13 @@ pub struct CredentialsProviderChainDefaultOptions<'a> {
     pub bootstrap: &'a mut ClientBootstrap,
 }
 
+/// Options for creating a profile credentials provider
+#[derive(Debug)]
+pub struct CredentialsProviderProfileOptions<'a> {
+    /// The name of profile to use.
+    pub profile_name_override: &'a str,
```
jamesbornholt, line 27: This should be `Option<&'a str>`

mountpoint-s3-crt/src/auth/credentials.rs
jamesbornholt: The lifetime of `CredentialsProvider` doesn't help us here -- the `options.profile_name_override` &str has lifetime `'a` which isn't tied to the `CredentialsProvider` in any way. So for this to work we need to either know that `aws_credentials_provider_new_profile` copies the string out of the byte cursor, or arrange for `CredentialsProvider` to own a copy of the string.

mountpoint-s3-client/src/s3_crt_client.rs
eslrahc-swa: They are basically optional, if customer omit some of them, those configuration should never reflect to client

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -42,6 +54,25 @@ impl CredentialsProvider {
 
         Ok(Self { inner })
     }
+
+    /// Creates the profile credential provider.
+    pub fn new_profile(allocator: &Allocator, options: &CredentialsProviderProfileOptions) -> Result<Self, Error> {
```
jamesbornholt, line 59: We should take ownership of the options here.
```suggestion
    pub fn new_profile(allocator: &Allocator, options: CredentialsProviderProfileOptions) -> Result<Self, Error> {
```

mountpoint-s3-client/src/s3_crt_client.rs
jamesbornholt: They're not quite mutually exclusive -- profiles can contain things other than credentials, like CLI configuration. We don't use any of those today, but conceivably could in the future.

mountpoint-s3/src/main.rs
```diff
@@ -201,6 +201,16 @@ struct CliArgs {
 
     #[clap(short, long, help = "Run as foreground process")]
     pub foreground: bool,
+
+    #[clap(
+        short,
```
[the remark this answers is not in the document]
eslrahc-swa: short only reflect to `-n` I suppose but not `--no-sign-request`?

I followed what aws cli did for `--no-sign-request` which only include long?

mountpoint-s3-client/src/s3_crt_client.rs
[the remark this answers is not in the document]
jamesbornholt: Profiles can specify a role to assume, so we do need to include the bootstrap here.

jamesbornholt (commented):
We need to write tests for this, as they would have caught the missing bootstrap in the profile credentials provider.
- Test mounting an anonymous-readable bucket
- Test specifying profile provider with STS credentials

mountpoint-s3-crt/src/auth/credentials.rs
eslrahc-swa: CRT actually copies the bytes out of the byte cursor to a string: [aws_get_profile_nam](https://github.com/awslabs/aws-c-auth/blob/dd505b55fd46222834f35c6e54165d8cbebbfaaa/source/credentials_provider_profile.c#L377) --> [aws_string_new_from_array](https://github.com/awslabs/aws-c-sdkutils/blob/208a701fa01e99c7c8cc3dcebc8317da71362972/source/aws_profile.c#L1488) --> (aws_string_new_from_array)[https://github.com/awslabs/aws-c-common/blob/5e64b039356c72d17fcf1fb0bfc7637f8c2270f7/source/string.c#L189-L209]

mountpoint-s3-crt/src/auth/credentials.rs
jamesbornholt: that should too

mountpoint-s3-client/src/s3_crt_client.rs
passaro: I think I see what you mean, but maybe it's the signature of `init_default_signing_config` that's an issue: why is it taking a (mutable) reference instead of taking ownership of `CredentialProvider`? Is this a similar issue to what @jamesbornholt pointed out about the options parameter below? (or am I missing something?)

Issue #203, opened 2023-04-11
Credentials failure when attempting to mount bucket


tim-synergysuite:
### Mountpoint for Amazon S3 version

mountpoint-s3 0.2.0-9ad4854

### AWS Region

eu-west-1

### Describe the running environment

Running on Amazon Linux EC2 t4g.micro instance. IAM Role assigned to the instance. Have verified that I can access the bucket from the AWS CLI. 

### What happened?

When I try to mount a bucket I get
    Error: Failed to create S3 client
    
    Caused by:
        0: HeadBucket failed for bucket s3://<my bucket name>/ in region eu-west-1
        1: Service error
        2: Permission denied

I've listed the log file output below.

It also appears to ignore the ~/.aws/config file. This file exists with 

    region = eu-west-1

but if I omit '--region eu-west-1' on the mount-s3 command line, it reports that it failed to mount the bucket from 'us-east-1'. I've tried using environment variables instead and they also get ignored.

### Relevant log output

```shell
2023-04-11T15:12:40.656389Z ERROR awscrt::common-io: static: Failed to open file. path:'/home/ec2-user/.aws/credentials' mode:'rb' errno:2 aws-error:44(AWS_ERROR_FILE_INVALID_PATH)    
2023-04-11T15:12:40.656465Z ERROR awscrt::AuthCredentialsProvider: Failed to resolve role arn during sts web identity provider initialization.    
2023-04-11T15:12:40.673653Z ERROR awscrt::common-io: static: Failed to open file. path:'/home/ec2-user/.aws/credentials' mode:'rb' errno:2 aws-error:44(AWS_ERROR_FILE_INVALID_PATH)    
2023-04-11T15:12:40.683627Z ERROR awscrt::S3MetaRequest: id=0xaaaafb0715f0 Meta request cannot recover from error 14343 (Invalid response status from request). (request=0xffff78000dc0, response status=400)
```

tim-synergysuite:
That's it!. And this is actually documented in the README if I'd paid closer attention. I'm just so used to the URL style bucket reference.

Possibly this could use a bit of emphasis in the README at the moment. It would certainly be more consistent with the CLI if URL style bucket refs were supported. At the cost of parsing the URL, would remove the need for the ``--prefix`` argument.

Thanks for your help.

jamesbornholt:
We added some validation for this in #207. We're thinking more about the URL-style bucket references, but gonna close this out since the immediate issue is fixed.

PR #207, opened 2023-04-12
Add basic validation of bucket name

jamesbornholt:
This will help catch common CLI errors like #203. Customers coming from
the AWS CLI might expect to be able to pass an s3:// URI as a bucket
name, which we don't support. It's also possible to accidentally swap
the bucket name and mount point arguments -- mount points will often be
invalid bucket names, so some validation helps us catch some of these
cases more quickly.

Arguably we could support parsing s3:// URIs, including using the key as
a prefix for the mount. But that would be a bigger feature and I'm not
totally sure it's the right thing, so for now some validation is better
than nothing.

---

### DIFF
--- a/mountpoint-s3/src/main.rs
+++ b/mountpoint-s3/src/main.rs
@@ -504,7 +504,7 @@ fn parse_bucket_name(bucket_name: &str) -> anyhow::Result<String> {
 
     // Actual bucket names must start/end with a letter, but bucket aliases can end with numbers
     // (-s3), so let's just naively check for invalid characters.
-    let bucket_regex = Regex::new(r"^[0-9a-zA-Z\-\.]+$").unwrap();
+    let bucket_regex = Regex::new(r"^[0-9a-zA-Z\-\._]+$").unwrap();
     if !bucket_regex.is_match(bucket_name) {
         return Err(anyhow!("bucket names can only contain letters, numbers, . and -"));
     }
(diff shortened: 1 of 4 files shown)


mountpoint-s3/src/main.rs
```diff
@@ -490,6 +491,27 @@ fn parse_perm_bits(perm_bit_str: &str) -> Result<u16, anyhow::Error> {
...
+    if bucket_name.contains("s3://") {
+        return Err(anyhow!("bucket name should not be an s3:// URI (provide the bare bucket name instead; use --prefix for prefix mounts)"));
+    }
+
+    // Actual bucket names must start/end with a letter, but bucket aliases can end with numbers
+    // (-s3), so let's just naively check for invalid characters.
+    let bucket_regex = Regex::new(r"^[0-9a-zA-Z\-\.]+$").unwrap();
+    if !bucket_regex.is_match(bucket_name) {
+        return Err(anyhow!("bucket names can only contain letters, numbers, . and -"));
+    }
```
dannycjones, line 510: _us-east-1_ allowed underscores in bucket names until 2018, so we should allow that in the regular expression. https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html

(I assume that they can only be used when using path-style addressing.)

dannycjones (commented):
This is a nice improvement. I just have one doubt on the bucket name regex.

dannycjones (approved):
LGTM!

Issue #181, opened 2023-03-27
Add support for unsigned requests


pschmied:
### Tell us more about this new feature.

Allowing for unsigned requests would allow Mountpoint to work as a client for public, world-readable S3 buckets. This would function similarly to the AWS CLI's `--no-sign-request` flag (e.g. https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3/ls.html).

jamesbornholt:
Done in #209.

Issue #191, opened 2023-04-03
Tests for requester-pays support


jamesbornholt:
We should create a requester-pays bucket and test against it in CI so we can get test coverage of #168.

jamesbornholt:
We should also test unsigned requests and different profiles from #209. I think they're all similar so lumping that in here.

PR #209, opened 2023-04-14
Add support for AWS profile and unsigned requests as command line arguments

eslrahc-swa:
* Add option --profile for a specific profile in the AWS credentials file #151
* Add support for unsigned requests #181

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -8,7 +8,9 @@ use std::sync::{Arc, Mutex};
 use std::task::{Context, Poll};
 use std::time::{Duration, Instant};
 
-use mountpoint_s3_crt::auth::credentials::{CredentialsProvider, CredentialsProviderChainDefaultOptions};
+use mountpoint_s3_crt::auth::credentials::{
+    CredentialsProvider, CredentialsProviderChainDefaultOptions, CredentialsProviderProfileOptions,
+};
 use mountpoint_s3_crt::common::allocator::Allocator;
 use mountpoint_s3_crt::common::uri::Uri;
 use mountpoint_s3_crt::http::request_response::{Header, Headers, Message};
@@ -43,12 +45,15 @@ pub(crate) mod delete_object;
 pub(crate) mod get_object;
 pub(crate) mod get_object_attributes;
 pub(crate) mod head_bucket;
+
 pub(crate) mod head_object;
 pub(crate) mod list_objects;
 pub(crate) mod put_object;
 
 #[derive(Debug, Clone, Default)]
 pub struct S3ClientConfig {
+    pub profile_name_override: Option<String>,
+    pub no_sign_request: bool,
     pub throughput_target_gbps: Option<f64>,
     pub part_size: Option<usize>,
     pub endpoint: Option<Endpoint>,
@@ -89,12 +94,7 @@ impl S3CrtClient {
 
         let mut client_bootstrap = ClientBootstrap::new(&allocator, &bootstrap_options).unwrap();
 
-        let creds_options = CredentialsProviderChainDefaultOptions {
-            bootstrap: &mut client_bootstrap,
-        };
-
-        let mut creds_provider = CredentialsProvider::new_chain_default(&allocator, &creds_options).unwrap();
-        let signing_config = init_default_signing_config(region, &mut creds_provider);
+        let mut client_config = ClientConfig::new();
 
         let mut retry_strategy_options = StandardRetryOptions::default(&mut event_loop_group);
         // Match the SDK "legacy" retry strategies
@@ -103,11 +103,29 @@ impl S3CrtClient {
         retry_strategy_options.backoff_retry_options.jitter_mode = ExponentialBackoffJitterMode::Full;
         let retry_strategy = RetryStrategy::standard(&allocator, &retry_strategy_options).unwrap();
 
-        let mut client_config = ClientConfig::new();
+        if !config.no_sign_request {
+            let credentials_provider = match config.profile_name_override {
+                Some(profile_name_override) => {
+                    let credentials_profile_options = CredentialsProviderProfileOptions {
+                        bootstrap: &mut client_bootstrap,
+                        profile_name_override: &profile_name_override,
+                    };
+                    CredentialsProvider::new_profile(&allocator, credentials_profile_options)
+                }
+                None => {
+                    let credentials_chain_default_options = CredentialsProviderChainDefaultOptions {
+                        bootstrap: &mut client_bootstrap,
+                    };
+                    CredentialsProvider::new_chain_default(&allocator, credentials_chain_default_options)
+                }
+            }
+            .map_err(NewClientError::ProviderFailure)?;
+            let signing_config = init_default_signing_config(region, credentials_provider);
+            client_config.signing_config(signing_config);
+        }
 
         client_config
             .client_bootstrap(client_bootstrap)
-            .signing_config(signing_config)
             .retry_strategy(retry_strategy);
 
         if let Some(throughput_target_gbps) = config.throughput_target_gbps {
@@ -465,6 +483,9 @@ pub enum NewClientError {
     /// Invalid S3 endpoint
     #[error("invalid S3 endpoint")]
     InvalidEndpoint(#[from] EndpointError),
+    /// Invalid AWS credentials
+    #[error("invalid AWS credentials")]
+    ProviderFailure(#[from] mountpoint_s3_crt::common::error::Error),
 }
 
 /// Failed S3 request results
--- a/mountpoint-s3-crt/src/auth/credentials.rs
+++ b/mountpoint-s3-crt/src/auth/credentials.rs
@@ -1,11 +1,16 @@
 //! AWS credentials providers
 
+use mountpoint_s3_crt_sys::{
+    aws_credentials_provider, aws_credentials_provider_chain_default_options,
+    aws_credentials_provider_new_chain_default, aws_credentials_provider_new_profile,
+    aws_credentials_provider_profile_options, aws_credentials_provider_release,
+};
+
 use crate::auth::auth_library_init;
 use crate::common::allocator::Allocator;
 use crate::common::error::Error;
 use crate::io::channel_bootstrap::ClientBootstrap;
-use crate::CrtError as _;
-use mountpoint_s3_crt_sys::*;
+use crate::{CrtError as _, StringExt};
 use std::ptr::NonNull;
 
 /// Options for creating a default credentials provider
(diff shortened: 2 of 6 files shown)


mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -42,6 +57,28 @@ impl CredentialsProvider {
...
         Ok(Self { inner })
     }
+
+    /// Creates the profile credential provider.
+    pub fn new_profile(allocator: &Allocator, options: CredentialsProviderProfileOptions) -> Result<Self, Error> {
+        auth_library_init(allocator);
+
+        let profile_name = options
+            .profile_name_override
+            .ok_or(Error::from(AWS_AUTH_CREDENTIALS_PROVIDER_PROFILE_SOURCE_FAILURE as i32))?;
```
passaro, line 67: if we error on missing profile name, why make the field an Option?

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -42,6 +57,28 @@ impl CredentialsProvider {
...
+
+        // SAFETY: CredentialProvider does not modify contents of profile_name_override,
+        // aws_credentials_provider_new_profile copies the contents of profile_name_override.
+        let inner = unsafe {
+            let inner_options = aws_credentials_provider_profile_options {
+                profile_name_override: profile_name.as_aws_byte_cursor(),
+                ..Default::default()
```
passaro, line 74: this seems to be missing the bootstrap

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -42,6 +57,28 @@ impl CredentialsProvider {
...
+
+    /// Creates the profile credential provider.
+    pub fn new_profile(allocator: &Allocator, options: CredentialsProviderProfileOptions) -> Result<Self, Error> {
+        auth_library_init(allocator);
+
+        let profile_name = options
+            .profile_name_override
+            .ok_or(Error::from(AWS_AUTH_CREDENTIALS_PROVIDER_PROFILE_SOURCE_FAILURE as i32))?;
+
+        // SAFETY: CredentialProvider does not modify contents of profile_name_override,
```
passaro, line 69: > CredentialProvider does not modify contents of profile_name_override,

I think this could be removed.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -89,12 +93,30 @@ impl S3CrtClient {
...
+                    let credentials_profile_options = CredentialsProviderProfileOptions {
+                        bootstrap: &mut client_bootstrap,
+                        profile_name_override: Some(&profile_name_override),
+                    };
+                    CredentialsProvider::new_profile(&allocator, credentials_profile_options)
+                        .map_err(|_| CredentialsError::ProfileNotFound)
+                }
+                None => {
+                    let credentials_chain_default_options = CredentialsProviderChainDefaultOptions {
+                        bootstrap: &mut client_bootstrap,
```
passaro, line 110: (existing code, so it may be a wider question, but you also used this on line 102)
Why the exclusive ref here? And why client_bootstrap needs to be mut in the first place?
Shouldn't we either:
1. pass by shared ref and rely on the crt to acquire what it needs to hold on to when we call `aws_credentials_provider_new_chain_default` inside `new_chain_default`?
2. or take ownership and clone here, which is what we are doing e.g. on line 129?

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -639,7 +640,7 @@ pub fn init_default_signing_config(region: &str, credentials_provider: &mut Cred
         aws_s3_init_default_signing_config(
             &mut signing_config.inner,
             region_cursor,
-            credentials_provider.inner.as_ptr(),
+            signing_config.credentials_provider.inner.as_ptr(),
```
passaro, line 643: I think this is correct, but we should mention it in the SAFETY comment above.

passaro (changes requested):
Raised a couple of points.

mountpoint-s3-crt/src/auth/credentials.rs
eslrahc-swa: The is address the comment from james, which I don't think passed in profile_name reference should be `Option<&'a &str>` because in `s3_crt_client.rs` we have take the string out of `Option`

mountpoint-s3-crt/src/auth/credentials.rs
eslrahc-swa: it is, nice catch

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -1,11 +1,17 @@
...
+};
+
 use crate::auth::auth_library_init;
 use crate::common::allocator::Allocator;
 use crate::common::error::Error;
 use crate::io::channel_bootstrap::ClientBootstrap;
-use crate::CrtError as _;
-use mountpoint_s3_crt_sys::*;
+use crate::{CrtError as _, StringExt};
+use mountpoint_s3_crt_sys::aws_auth_errors::AWS_AUTH_CREDENTIALS_PROVIDER_PROFILE_SOURCE_FAILURE;
```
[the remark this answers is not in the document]
eslrahc-swa: I don't think we need explicitly raise error here, so I will remove this.

mountpoint-s3-client/src/s3_crt_client.rs
jamesbornholt: Two parts to this:
1. We pass by reference instead of cloning for two reasons:
  a. to avoid the refcount increment/decrement overhead (very premature optimization)
  b. to avoid baking the fact that these things are cheaply cloneable into our API, in case we wanted to change that for some reason in the future.
3. Given that we're passing by reference, we pass with `&mut` because the CRT API expects a `*mut` pointer. Technically we could cast the `&` reference into a `*mut` pointer, and it would be fine here, but it's sketchy in general.

I think probably what we should do is reverse our choice in (1) and making all this stuff pass by value with a ref count increment when cloning. But we can do that in a separate PR.

mountpoint-s3-crt/src/auth/credentials.rs
jamesbornholt: I was wrong about this, actually -- the CRT requires `profile_name_override`, so we shouldn't make it an `Option`. Sorry!

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -89,12 +93,30 @@ impl S3CrtClient {
...
-        let signing_config = init_default_signing_config(region, &mut creds_provider);
+        if !config.no_sign_request {
+            let credentials_provider = match config.profile_name_override {
+                Some(profile_name_override) => {
+                    let credentials_profile_options = CredentialsProviderProfileOptions {
+                        bootstrap: &mut client_bootstrap,
+                        profile_name_override: Some(&profile_name_override),
+                    };
+                    CredentialsProvider::new_profile(&allocator, credentials_profile_options)
+                        .map_err(|_| CredentialsError::ProfileNotFound)
```
jamesbornholt, line 106: I don't think `ProfileNotFound` is an accurate description of this error. Looking at https://github.com/awslabs/aws-c-auth/blob/main/source/credentials_provider_profile.c#L366, it could be:
* an invalid path to a profile file
* an invalid path to a credentials file
* the profile wasn't specified
* the profile doesn't exist
* or an allocation failure

So maybe I'd call this a more generic `CredentialsError::ProviderFailure`.

We also want it to not throw away the underlying error, so the enum variant should have a `#[source] mountpoint_s3::common::error::Error` field.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -89,12 +93,30 @@ impl S3CrtClient {
...
+                }
+                None => {
+                    let credentials_chain_default_options = CredentialsProviderChainDefaultOptions {
+                        bootstrap: &mut client_bootstrap,
+                    };
+                    CredentialsProvider::new_chain_default(&allocator, credentials_chain_default_options)
+                        .map_err(|_| CredentialsError::NoValidCredentials)
```
jamesbornholt, line 113: Same thing here about the error:
1. `CredentialsError::ProviderFailure`
2. don't throw away the underlying error

eslrahc-swa:
We need to write tests for this, as they would have caught the missing bootstrap in the profile credentials provider.

    Test mounting an anonymous-readable bucket
    Test specifying profile provider with STS credentials

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -103,11 +103,30 @@ impl S3CrtClient {
         retry_strategy_options.backoff_retry_options.jitter_mode = ExponentialBackoffJitterMode::Full;
         let retry_strategy = RetryStrategy::standard(&allocator, &retry_strategy_options).unwrap();
 
-        let mut client_config = ClientConfig::new();
+        let credentials_provider = match config.profile_name_override {
```
jamesbornholt, line 106: In the previous revision I thought we didn’t construct the credentials provider at all if `no_sign_request` was set — why’d we change that?

eslrahc-swa: `credentials_provider` will be dropped if put it into
```
if !config.no_sign_request {
.....
}
```

and I thought we don't want to [singing_config](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-crt/src/auth/signing_config.rs#L11-L20)owns it and let crt copy/increase counts

jamesbornholt: Not sure what you mean. `init_default_signing_config` takes a `&CredentialsProvider` as input, so we know that it's OK to drop the `CredentialsProvider` after the call to `init_default_signing_config`. Otherwise Rust wouldn't let this code compile.

jamesbornholt: Actually, unless I'm missing something, I think the existing interface is just wrong. It calls `aws_s3_init_default_signing_config`, which doesn't seem to increment the reference count on the credentials provider it's given.

I think we need to change `init_default_signing_config` to take ownership of the `CredentialsProvider`, and then add that as a field to `SigningConfigInner` so that it lives as long as the underlying signing config does.

eslrahc-swa: > I think we need to change init_default_signing_config to take ownership of the CredentialsProvider, and then add that as a field to SigningConfigInner so that it lives as long as the underlying signing config does.

This is the last revision, seems like my understanding of your last comment was wrong.

eslrahc-swa: > so we know that it's OK to drop the CredentialsProvider after the call to init_default_signing_config.
rust would compile, but when it will result segmentation fault.

eslrahc-swa: side notes, let's reach consensus on Pull Request next time to get rid of these back and forth

mountpoint-s3/src/main.rs
```diff
@@ -202,6 +203,15 @@ struct CliArgs {
 
     #[clap(short, long, help = "Run as foreground process")]
     pub foreground: bool,
+
+    #[clap(
+        long,
+        help = "Do not sign requests. Credentials will not be loaded if this argument is provided."
```
passaro, line 209: Should this also have the same `help_heading`?

jamesbornholt (approved):

jamesbornholt:
#191 for the followup on testing.

ooboyle-rs10:
Thank you, all, for your work on this. I rebuilt the latest version and --profile works great so far.

PR #210, opened 2023-04-14
Make opened file visible in lookup

monthonk:
We have implemented this in readdir but missed it in lookup. We want the new files to be visible there as well. There might be two options here, 1/ We can lookup locally first and return immediately when the item we want is an uncommitted inode and we don't have to call any S3 APIs since it should not exist in S3 yet, or 2/ we can lookup remotely first then fall back to local. In this commit, we go with the second option because there is a possibility that someone could make a change remotely like creating a directory marker that shadow the uncommitted file.

---

### DIFF
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -238,7 +238,27 @@ impl Superblock {
                 .update_or_insert(parent_ino, name, stat.clone(), InodeKind::File)?;
             Ok(LookedUp { inode, stat })
         } else {
-            Err(InodeError::FileDoesNotExist)
+            // if object with the given name doesn't exist we will look it up locally since it could be an uncommitted inode
+            let parent_state = parent.inner.sync.read().unwrap();
+            match &parent_state.kind_data {
+                InodeKindData::File { .. } => unreachable!("we know this is a directory"),
+                InodeKindData::Directory {
+                    children,
+                    writing_children,
+                } => {
+                    if let Some(inode) = children.get(name) {
+                        if writing_children.contains(&inode.ino()) {
+                            let inode = inode.clone();
+                            let stat = inode.inner.sync.read().unwrap().stat.clone();
+                            Ok(LookedUp { inode, stat })
+                        } else {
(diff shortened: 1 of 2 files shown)


mountpoint-s3/src/inode.rs
```diff
@@ -238,7 +238,27 @@ impl Superblock {
...
+                InodeKindData::File { .. } => unreachable!("we know this is a directory"),
+                InodeKindData::Directory {
+                    children,
+                    writing_children,
+                } => {
+                    if let Some(inode) = children.get(name) {
+                        if writing_children.contains(&inode.ino()) {
+                            let inode = inode.clone();
+                            let stat = inode.inner.sync.read().unwrap().stat.clone();
+                            Ok(LookedUp { inode, stat })
```
passaro, line 253: nit: maybe return here and just fallback to Err otherwise, avoiding the multiple else branches below.

passaro (approved):
With option 2, this is consistent with readdir, right? 
While writing to a new file:
* it will be visible in lookup and readdir,
* but if a file with the same key is added remotely, it will shadow the local one (at least `finish_writing`)
Do we have a way to test this behaviour?

mountpoint-s3/src/inode.rs
```diff
@@ -238,7 +238,27 @@ impl Superblock {
                 .update_or_insert(parent_ino, name, stat.clone(), InodeKind::File)?;
             Ok(LookedUp { inode, stat })
         } else {
-            Err(InodeError::FileDoesNotExist)
```
sauraank, line 241: Should we also add this case in the comments at line 233, otherwise comment can be misleading?

monthonk: Updated, I think we can say on line 233 that the object doesn't exist remotely.

monthonk:
@passaro yes, I think it's consistent with readdir and it's possible to write that kind of test maybe in `write_test.rs`. But I'm not sure if we want to, since we will just return remote result first, whatever it might be.

passaro (approved):
LGTM

PR #213, opened 2023-04-17
Add benchmark for readdir

monthonk:
In this commit, we add a new benchmark script for readdir operation. The benchmark itself is pretty simple, we're just doing `ls` against pre-generated directories with different size. However, comparing the results is a bit tricky because we're now measuring latency instead of throughput which is what we do in the existing benchmark.

Our current CI workflows don't support this scenario, so we have to introduce another workflow job just for this kind of benchmark. This new job will run the latency benchmark script, compare the new result with the previous one, and store them at dev/latency_bench path.

Max number of data points for the chart is also added in this commit to keep the chart clean.

---


monthonk:
You can see what the result page looks like in my forked repo [here](https://monthonk.github.io/mountpoint-s3/dev/latency_bench/).

mountpoint-s3/scripts/fs_latency_bench.sh
```diff
@@ -0,0 +1,99 @@
...
+
+	# verify that the target directory exists before running the benchmark
+	if [ ! -d "${target_dir}" ]; then
+		echo "Target directory ${target_dir} does not exist."
+		exit 1
+	fi
+
+    sleep $startdelay
+    # run each case for 10 iterations
+    for i in {1..10};
```
jamesbornholt, line 66: nit: maybe lift `10` out into a variable so we don’t miss that it needs to change in the average as well as here

mountpoint-s3/scripts/fs_latency_bench.sh
```diff
@@ -0,0 +1,99 @@
...
+mkdir -p ${results_dir}
+
+# start readdir benchmarking
+dir_size=100
+while [ $dir_size -le 100000 ]
+do
+    sum=0
+    job_name="readdir_${dir_size}"
+    mount_dir=$(mktemp -d /tmp/mount-s3-XXXXXXXXXXXX)
+    target_dir="${mount_dir}/bench_dir_${dir_size}"
```
jamesbornholt, line 42: we should update the benchmarking docs to describe how this directory structure is created

jamesbornholt (commented):
Neat! There’s a lot of commonality between the two benchmark scripts, so it feels like in future we might want to do something about that (maybe rewrite into a common Python script or something). But that’s a problem for another day.

mountpoint-s3/scripts/fs_latency_bench.sh
monthonk: Added a readdir section in the doc to explain this.

jamesbornholt (approved):

PR #216, opened 2023-04-20
Remove children not found on lookup and refactor update methods

passaro:
Address the issue with lookup only updating/inserting new inodes but never removing children deleted on the client (see added test).

Refactor the update methods to group all operations which modify the parent inode state:
* pulled out the client calls in `lookup` into a new `remote_lookup` method,
* `remote_lookup` returns a (result of an optional) `RemoteLookup` (kind + stat),
* `update_from_remote` (formerly `update_or_insert`) processes `RemoteLookup` and can update/insert/delete,
* in turn, it calls `try_update_child` (formerly `update_if_present`), which looks for existing inodes, updates them if required, and returns an `UpdateStatus` which informs the caller of the actions to perform on the parent directory (insert or delete). 

---

### DIFF
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -388,8 +388,10 @@ impl SuperblockInner {
         // Fast path: try with only a read lock on the directory first.
         {
             let parent_state = parent.inner.sync.read().unwrap();
-            if let UpdateStatus::Updated(lookedup) = Self::try_update_child(&parent_state, name, &remote)? {
-                return Ok(lookedup);
+            match Self::try_update_child(&parent_state, name, &remote)? {
+                UpdateStatus::Neither => return Err(InodeError::FileDoesNotExist),
+                UpdateStatus::Updated(lookedup) => return Ok(lookedup),
+                _ => {} // Fallback, we need a write lock to update the parent.
             }
         }
 
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -119,9 +119,18 @@ impl Superblock {
             return Err(InodeError::InvalidFileName(name.into()));
         }
 
-        // TODO use caches. if we already know about this name, we just need to revalidate the stat
-        // cache and then read it.
+        let remote = self.remote_lookup(client, parent_ino, name).await?;
+        self.inner.update_from_remote(parent_ino, name, remote)
+    }
 
+    /// Lookup an inode in the parent directory with the given name
+    /// on the remote client.
+    async fn remote_lookup<OC: ObjectClient>(
+        &self,
+        client: &OC,
+        parent_ino: InodeNo,
+        name: &str,
+    ) -> Result<Option<RemoteLookup>, InodeError> {
         let parent = self.inner.get(parent_ino)?;
         if parent.kind() != InodeKind::Directory {
             return Err(InodeError::NotADirectory(parent_ino));
@@ -221,9 +230,7 @@ impl Superblock {
                     if found_directory {
                         trace!(?parent, ?name, "lookup ListObjects found a directory");
                         let stat = InodeStat::for_directory(self.inner.mount_time, Instant::now());
-                        let inode =
-                            self.inner.update_or_insert(parent_ino, name, stat.clone(), InodeKind::Directory)?;
-                        return Ok(LookedUp { inode, stat });
+                        return Ok(Some(RemoteLookup { kind: InodeKind::Directory, stat }));
                     }
                 }
             }
@@ -231,32 +238,13 @@ impl Superblock {
 
         // If we reach here, the ListObjects didn't find a shadowing directory, so we know we either
         // have a valid file, or both requests failed to find the object so the file must not exist remotely
-        if let Some(stat) = file_state {
+        Ok(file_state.map(|stat| {
             trace!(?parent, ?name, "found a regular file");
-            let inode = self
-                .inner
-                .update_or_insert(parent_ino, name, stat.clone(), InodeKind::File)?;
-            Ok(LookedUp { inode, stat })
-        } else {
-            // if object with the given name doesn't exist we will look it up locally since it could be an uncommitted inode
-            let parent_state = parent.inner.sync.read().unwrap();
-            match &parent_state.kind_data {
-                InodeKindData::File { .. } => unreachable!("we know this is a directory"),
-                InodeKindData::Directory {
-                    children,
-                    writing_children,
-                } => {
-                    if let Some(inode) = children.get(name) {
-                        if writing_children.contains(&inode.ino()) {
-                            let inode = inode.clone();
-                            let stat = inode.inner.sync.read().unwrap().stat.clone();
-                            return Ok(LookedUp { inode, stat });
-                        }
-                    }
-                    Err(InodeError::FileDoesNotExist)
-                }
+            RemoteLookup {
+                kind: InodeKind::File,
+                stat,
             }
-        }
+        }))
     }
 
     /// Retrieve the attributes for an inode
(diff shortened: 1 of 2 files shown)


mountpoint-s3/src/inode.rs
```diff
@@ -382,77 +370,95 @@ impl SuperblockInner {
...
-        // Fast path: try with only a read lock on the directory first
+        // Fast path: try with only a read lock on the directory first.
         {
             let parent_state = parent.inner.sync.read().unwrap();
-            if let Some(inode) = Self::update_if_present(parent_ino, &parent_state, name, kind, &stat)? {
-                return Ok(inode);
+            if let UpdateStatus::Updated(lookedup) = Self::try_update_child(&parent_state, name, &remote)? {
+                return Ok(lookedup);
             }
         }
```
monthonk, line 394: We can skip this path if the remote lookup is None since the result will never bee `UpdateStatus::Updated`.

mountpoint-s3/src/inode.rs
```diff
@@ -548,6 +557,29 @@ impl SuperblockInner {
...
+/// Data from a remote object.
+#[derive(Debug)]
+pub struct RemoteLookup {
+    kind: InodeKind,
+    stat: InodeStat,
+}
+
+/// Result of a call to [SuperblockInner::try_update_child].
+#[derive(Debug)]
+enum UpdateStatus<'n, 'r> {
```
monthonk, line 569: The lifetime for this struct is a bit complicated and I noticed that we still have to clone the data when using. Can we simplify this by taking ownership of these variable?

mountpoint-s3/src/inode.rs
```diff
@@ -382,77 +370,95 @@ impl SuperblockInner {
...
-    ) -> Result<Inode, InodeError> {
-        let parent = {
-            let inodes = self.inodes.read().unwrap();
-            inodes
-                .get(&parent_ino)
-                .cloned()
-                .ok_or(InodeError::InodeDoesNotExist(parent_ino))?
-        };
+        remote: Option<RemoteLookup>,
+    ) -> Result<LookedUp, InodeError> {
```
monthonk, line 380: I still have mixed feelings about this function, the name is `update_from_remote` but we're still updating it even if remote data is None and then we return `LookedUp` as a result. So, I'm not sure if we should have another layer to call this function when there is something to update, and also control when to remove (remove itself could be another function).

monthonk (commented):
Have some high level comments that we probably want to address.

mountpoint-s3/src/inode.rs
passaro: Well spotted. But actually we could also have `UpdateStatus::Neither` here and return an error, so probably still worth checking.

mountpoint-s3/src/inode.rs
passaro: I see your point. I would still prefer to keep all the logic in this function, but we definitely need a better name/description for it!

In my mind, this is where we take what we got from the client (`Option<RemoteLookup>`) and update our local state:
* if both inode and key exist - or neither do! - we use the fast path,
* otherwise, we need to write-lock the parent to insert or delete an inode.

Moving the deletion outside would mean duplicating the locking logic somewhere else.

monthonk: Yeah, I understand that you are trying to avoid multiple locking and cloning but it's not really straightforward when we read function/struct definition (e.g. why include `LookedUp` in `UpdateStatus::Updated` instead of `Inode` like we do in `UpdateStatus::LocalOnly`).

Let's keep the logic in one function and update the name/description to make it clearer.

mountpoint-s3/src/inode.rs
```diff
@@ -382,77 +370,93 @@ impl SuperblockInner {
...
+                            // Return the local inode.
+                            let stat = inode.inner.sync.read().unwrap().stat.clone();
+                            return Ok(LookedUp { inode, stat });
+                        }
+                        // Remove from children.
+                        // TODO: also handle inode in [Self::inodes].
+                        children.remove(name);
+                    }
+                }
+                Err(InodeError::FileDoesNotExist)
```
jamesbornholt, line 421: Maybe a little clearer with less fallthrough:

```suggestion
                match &mut parent_state.kind_data {
                    InodeKindData::File {} => unreachable!("we know parent is a directory"),
                    InodeKindData::Directory {
                        children,
                        writing_children,
                    } => {
                        if writing_children.contains(&inode.ino()) {
                            // Return the local inode.
                            let stat = inode.inner.sync.read().unwrap().stat.clone();
                            Ok(LookedUp { inode, stat })
                        } else {
                            // Remote no longer exists, so remove from local children
                            // TODO: also handle inode in [Self::inodes].
                            children.remove(name);
                            Err(InodeError::FileDoesNotExist)
                        }
                    }
                }
```

mountpoint-s3/src/inode.rs
```diff
@@ -221,42 +230,21 @@ impl Superblock {
...
                     }
                 }
             }
         }
 
         // If we reach here, the ListObjects didn't find a shadowing directory, so we know we either
         // have a valid file, or both requests failed to find the object so the file must not exist remotely
-        if let Some(stat) = file_state {
+        Ok(file_state.map(|stat| {
             trace!(?parent, ?name, "found a regular file");
```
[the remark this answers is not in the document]
jamesbornholt: Also, I think we'd like a trace in both cases now that we're relying on the negative case for some logic.

mountpoint-s3/src/inode.rs
```diff
@@ -119,9 +119,18 @@ impl Superblock {
...
 
-        // TODO use caches. if we already know about this name, we just need to revalidate the stat
-        // cache and then read it.
+        let remote = self.remote_lookup(client, parent_ino, name).await?;
+        self.inner.update_from_remote(parent_ino, name, remote)
+    }
 
+    /// Lookup an inode in the parent directory with the given name
+    /// on the remote client.
+    async fn remote_lookup<OC: ObjectClient>(
```
jamesbornholt, line 128: Will there ever be a case where you don't immediately call `self.inner.update_from_remote` with the result of `remote_lookup`? If not, maybe we should push the `update_from_remote` call into it so (hypothetical future) callers can't forget.

mountpoint-s3/tests/fs.rs
```diff
@@ -445,3 +445,38 @@ async fn test_stat_block_size() {
     assert_eq!(lookup.attr.blocks, 9);
     assert_eq!(lookup.attr.blksize, 4096);
 }
+
+#[test_case("foo"; "remove file")]
+#[test_case("bar/foo"; "remove directory")]
+#[tokio::test]
+async fn test_lookup_removes_old_children(key: &str) {
```
jamesbornholt, line 452: Can we also test that `readdir` has the same effect?

mountpoint-s3/src/inode.rs
passaro: My preference would still be to keep them separate:
* `remote_lookup` wraps all the (double) request logic and returns `Option<RemoteLookup>`,
* `update_from_remote` takes it and updates the local "cache".

mountpoint-s3/tests/fs.rs
passaro: We have not changed `readdir` behavior: it only updates keys/prefixes as they are returned, but it does not remove any "missing" children. In order to do it, I think it would need to keep track of all the "names" returned across all the `list_objects` calls in order to work out what to remove at the end.

I'm not sure though: do we need `readdir` to do it? Or can we just rely on lookup?

jamesbornholt: Oh, right, we'll never test the negative case through `readdir`.

PR #214, opened 2023-04-18
Implement `unlink` operation

dannycjones:
This PR implements `unlink` in Mountpoint S3.

## Current Status

Partially implemented, not ready for proper review although early feedback is welcome.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

Issue #78, opened 2023-02-09
Support for unlink/rm


dannycjones:
>Deleting objects with Mountpoint for Amazon S3

- [x] #83
- [x] Implement unlink in the filesystem

dannycjones:
My thinking here is to break down the FS changes into two steps:

- Implement link counting to the Inode in one PR, such that we are confident in the changes to operations such as `mknod`, `open`, `release`. 
- Implement `unlink` and the deletion of the object in S3 (by both `unlink` and `release`) in a second PR, after we are confident in the link counting changes.

dannycjones:
Revised break down:

My thinking here is to break down the FS changes into two steps:

- Implement `unlink` and the deletion of the object in S3 in one PR, and open file handles will return IO error if trying to continue to read the deleted file (simple case)
- Implement link counting to the Inode in a follow-up PR, if we feel that open file handles need to be able to continue reading. (more POSIX compliant, although adds complexity around when the delete is performed)

jamesbornholt:
We can't just drop the inode on the floor in response to an `unlink`, as the kernel expects the inode to still exist until the lookup count for the inode reaches zero (via `forget` calls). So it might be easier to do those two things the other way around.

jamesbornholt:
Done in #232.

PR #220, opened 2023-04-24
Add description of file delete (`unlink`) semantics

dannycjones:
This describes the semantics of file deletes (`unlink`, #78).

If we are happy with these semantics, we can move forward with the implementation of `unlink` and related operations such as `forget`.

Note, I have not described how the lookup count and `forget` operation will work. I am considering this to be better suited in RustDoc and more of an 'implementation detail'.

A rough implementation is already being drafted in #214.

---

### DIFF
--- a/doc/SEMANTICS.md
+++ b/doc/SEMANTICS.md
@@ -79,16 +79,14 @@ Space allocation (`fallocate`, `posix_fallocate`) are not supported.
 File deletion (`unlink`) is planned but not yet supported (see [#78](https://github.com/awslabs/mountpoint-s3/issues/78)).
 When `unlink` is implemented, the following semantics are proposed.
 
-For files not yet committed to S3, the client will make a best effort to stop the file being uploaded.
+For files not yet committed to S3, the client will make a best effort to stop any inflight uploads,
+and immediately remove the file from its directory.
+If there are any open file handles to the file, future writes to them will fail.
 
-- If the file was not yet opened for writing, the file entry will be deleted.
-- If a file is already open and being written to, the multipart upload will be cancelled and any future bytes written to the file handle will be discarded.
-
-Where the file is already uploaded to S3,
-an unlink operation will _immediately_ perform the delete on S3 and unlink the Inode from its parent.
-
-- If a file was open for reading locally, future reads to the file handle will fail with an IO error.
-- If a file was open for reading on another Mountpoint S3 process / host, future reads will fail with an IO error.
+For files already committed to S3, the client will _immediately_ delete the corresponding object from S3,
+and remove the file from its directory.
+If there are still open file handles to the file, future reads to them will fail.
+Because the object is immediately deleted from S3, future reads from other hosts will also fail.
 
 ### Directory operations
 
@@ -96,7 +94,7 @@ Basic read-only directory operations (`opendir`, `readdir`, `closedir`) are supp
 
 Renaming files and directories (`rename`, `renameat`) is not currently supported.
 
-File deletion (`unlink`) semantics are described under the file operations section.
+File deletion (`unlink`) semantics are described in the [Deletes](#Deletes) section.
 
 Empty directory removal (`rmdir`) is not supported.
 
--- a/doc/SEMANTICS.md
+++ b/doc/SEMANTICS.md
@@ -74,13 +74,29 @@ Synchronization operations (`fsync`, `fdatasync`) are currently no-ops because w
 
 Space allocation (`fallocate`, `posix_fallocate`) are not supported.
 
+#### Deletes
+
+File deletion (`unlink`) is planned but not yet supported (see [#78](https://github.com/awslabs/mountpoint-s3/issues/78)).
+When `unlink` is implemented, the following semantics are proposed.
+
+For files not yet committed to S3, the client will make a best effort to stop the file being uploaded.
+
+- If the file was not yet opened for writing, the file entry will be deleted.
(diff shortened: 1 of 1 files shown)


doc/SEMANTICS.md
```diff
@@ -74,13 +74,29 @@ Synchronization operations (`fsync`, `fdatasync`) are currently no-ops because w
...
 
+#### Deletes
+
+File deletion (`unlink`) is planned but not yet supported (see [#78](https://github.com/awslabs/mountpoint-s3/issues/78)).
+When `unlink` is implemented, the following semantics are proposed.
+
+For files not yet committed to S3, the client will make a best effort to stop the file being uploaded.
+
+- If the file was not yet opened for writing, the file entry will be deleted.
+- If a file is already open and being written to, the multipart upload will be cancelled and any future bytes written to the file handle will be discarded.
```
jamesbornholt, line 85: I think this is a bit too in the weeds (it requires you to know the difference between `mknod` and `open`). I'd just say something like:
> For files not yet committed to S3, the client will make a best effort to stop any inflight uploads, and immediately remove the file from its directory. If there are still open file handles to the file, future writes to them will fail.

doc/SEMANTICS.md
```diff
@@ -74,13 +74,29 @@ Synchronization operations (`fsync`, `fdatasync`) are currently no-ops because w
...
+
+Where the file is already uploaded to S3,
+an unlink operation will _immediately_ perform the delete on S3 and unlink the Inode from its parent.
+
+- If a file was open for reading locally, future reads to the file handle will fail with an IO error.
+- If a file was open for reading on another Mountpoint S3 process / host, future reads will fail with an IO error.
```
jamesbornholt, line 91:
```suggestion
For files already committed to S3, the client will _immediately_ delete the corresponding object from S3, and remove the file from its directory. If there are still open file handles to the file, future reads to them will fail. Because the object is immediately deleted from S3, future reads from other hosts will also fail.
```

doc/SEMANTICS.md
```diff
@@ -74,13 +74,29 @@ Synchronization operations (`fsync`, `fdatasync`) are currently no-ops because w
...
+
 ### Directory operations
 
 Basic read-only directory operations (`opendir`, `readdir`, `closedir`) are supported.
 
 Renaming files and directories (`rename`, `renameat`) is not currently supported.
 
-File deletion (`unlink`) is not currently supported, but will be [in the future](https://github.com/awslabs/mountpoint-s3/issues/78).
+File deletion (`unlink`) semantics are described under the file operations section.
```
jamesbornholt, line 99:
```suggestion
File deletion (`unlink`) semantics are described in the [Deletes](#Deletes) section.
```

doc/SEMANTICS.md
dannycjones: Thanks, that's much nicer for users of higher-level file interfaces.

dannycjones: To clarify on behavior, I think we describe two different things - although I prefer your proposed behavior.

>future writes to [file handles] will fail.

In this case, we are saying after unlink that `libc::EIO` will be returned. I am happy with that, and if there's a use case we can update the logic in future to discard bytes with no error.

doc/SEMANTICS.md
```diff
@@ -74,13 +74,27 @@ Synchronization operations (`fsync`, `fdatasync`) are currently no-ops because w
...
 Space allocation (`fallocate`, `posix_fallocate`) are not supported.
 
+#### Deletes
+
+File deletion (`unlink`) is planned but not yet supported (see [#78](https://github.com/awslabs/mountpoint-s3/issues/78)).
+When `unlink` is implemented, the following semantics are proposed.
+
+For files not yet committed to S3, the client will make a best effort to stop any inflight uploads,
+and immediately remove the file from its directory.
+If there are any open file handles to the file, future writes to them will fail.
```
monthonk, line 84: Alternatively, we can fail the unlink. It would fail faster since we don't have to cancel inflight uploads and probably less surprise?

dannycjones: I hadn't considered this one.

Considering strict POSIX compliance, we should allow writing to continue (presumably also reading to continue if we supported that, with file handles able to see new content) after the `unlink`, and once these file handles are closed, the data is deleted from 'disk'.

We do want to try and be close to POSIX where feasible, so allowing `unlink` during write makes sense to me. Since we don't support reading files during writing, perhaps the better behavior to avoid surprise would be to allow writing to continue, with the bytes being discarded rather than the `write` operation returning EIO. However, I was concerned about one-way doors with this option and simply returning EIO is the simplest option for now. (Yes, allowing `unlink` to succeed is also a one-way door - but an intentional one.)

dannycjones: FWIW, I am imagining this use case to be a very infrequent one. Thus, for customer applications that may try it (delete during write) we should fail in the way least likely to break applications and easy to understand what went wrong.

dannycjones: Discussed offline - failing `unlink` for local files leaves open the path for implementing it closer to POSIX in future. We can implement this behavior as the initial case, and if there is demand we can extend the behavior.

Trying to handle the local files also presents complexity for now, so we can deliver `unlink` for remote files now and tackle local files later.

Thoughts @jamesbornholt?

doc/SEMANTICS.md
jamesbornholt: Yeah, we should definitely fail writes that we know won't be persisted.

dannycjones:
Closing in favour of #232.

PR #221, opened 2023-04-25
Implement 'lookup count' tracking for `Inode`

dannycjones:
Early draft of tracking `lookup_count`, so that we know when we can drop from the superblock.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -305,7 +305,6 @@ where
 
         let full_key = lookup.inode.full_key().to_owned();
 
-        lookup.inode.inc_lookup_count();
         let fh = self.next_handle();
         let handle = FileHandle {
             inode: lookup.inode,
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -240,6 +240,8 @@ where
         let lookup = self.superblock.lookup(&self.client, parent, name).await?;
         let attr = self.make_attr(&lookup);
 
+        lookup.inode.inc_lookup_count();
+
         Ok(Entry {
             ttl: self.config.stat_ttl,
             attr,
@@ -257,6 +259,11 @@ where
             ttl: self.config.stat_ttl,
             attr,
         })
+    }
+
+    pub async fn forget(&self, ino: InodeNo, n: u64) {
+        trace!("fs:forget with ino {:?} n {:?}", ino, n);
+        self.superblock.forget(ino, n).await;
     }
 
     pub async fn open(&self, ino: InodeNo, flags: i32) -> Result<Opened, libc::c_int> {
@@ -298,6 +305,7 @@ where
 
         let full_key = lookup.inode.full_key().to_owned();
 
+        lookup.inode.inc_lookup_count();
         let fh = self.next_handle();
         let handle = FileHandle {
             inode: lookup.inode,
@@ -379,6 +387,8 @@ where
             .await?;
         let attr = self.make_attr(&lookup);
 
+        lookup.inode.inc_lookup_count();
+
         Ok(Entry {
             ttl: self.config.stat_ttl,
             attr,
@@ -398,6 +408,8 @@ where
             .create(&self.client, parent, name, InodeKind::Directory)
             .await?;
         let attr = self.make_attr(&lookup);
+
+        lookup.inode.inc_lookup_count();
 
         Ok(Entry {
             ttl: self.config.stat_ttl,
@@ -544,12 +556,13 @@ where
 
     pub async fn release(
         &self,
-        _ino: InodeNo,
+        ino: InodeNo,
         fh: u64,
         _flags: i32,
         _lock_owner: Option<u64>,
         _flush: bool,
     ) -> Result<(), libc::c_int> {
+        trace!("fs:release with ino {:?} fh {:?}", ino, fh);
         let file_handle = {
             let mut file_handles = self.file_handles.write().await;
             file_handles.remove(&fh).ok_or(libc::EBADF)?
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -578,10 +578,8 @@ where
             ) {
                 handle.handle.readd(next);
                 return Ok(reply);
-            } else {
-                if plus {
-                    next.inode.inc_lookup_count();
-                }
+            } else if plus {
+                next.inode.inc_lookup_count();
             }
             handle.next_offset();
         }
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -483,11 +483,13 @@ where
         Ok(Opened { fh, flags: 0 })
     }
 
+    /// Implementation for both FUSE `readdir` and `readdirplus`.
     pub async fn readdir<R: DirectoryReplier>(
         &self,
         parent: InodeNo,
         fh: u64,
         offset: i64,
+        plus: bool,
         mut reply: R,
     ) -> Result<R, libc::c_int> {
         trace!("fs:readdir with ino {:?} fh {:?} offset {:?}", parent, fh, offset);
@@ -548,6 +550,10 @@ where
             ) {
                 handle.handle.readd(next);
                 return Ok(reply);
+            } else {
+                if plus {
+                    next.inode.inc_lookup_count();
+                }
             }
             handle.next_offset();
         }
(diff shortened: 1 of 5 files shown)


dannycjones:
Some open questions for me:
- Are there ways to better test this? (I don't think we can force the Kernel to perform `forget(ino, nlookup)`).
- Should we be doing saturating operations, particularly on subtraction, to avoid underflow errors?

dannycjones:
I think the `lookup_count` is not accurate. It is not capturing file handles correctly, or if it is - it needs more clear explanation.

jamesbornholt:
> I think the `lookup_count` is not accurate. It is not capturing file handles correctly, or if it is - it needs more clear explanation.

There shouldn't be any direct any interaction between file (or directory) handles and lookup count. The trick is that the only way to open a file (directory) handle is to call `open` (`opendir`), and the only way to call those is to already know the inode number, which the kernel must have discovered from an earlier call to `lookup` or `readdirplus` or `mknod`.

Another way to say this is that the lookup count is a mechanism for remember which inodes the kernel knows about. Opening a file/directory handle doesn't affect which inodes the kernel knows about, since it already knew about the one it opened.

mountpoint-s3/src/fs.rs
```diff
@@ -298,6 +305,7 @@ where
 
         let full_key = lookup.inode.full_key().to_owned();
 
+        lookup.inode.inc_lookup_count();
```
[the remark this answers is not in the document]
dannycjones: Good catch. I only intended to add where we return an "entry" or "create" message to FUSE driver.

mountpoint-s3/src/inode.rs
```diff
@@ -100,6 +101,30 @@ impl Superblock {
...
+                    "forget should not be called on inode already removed from superblock"
+                );
+                error!("forget called on inode {ino} already removed from the superblock");
+            }
+            Some(inode) => {
+                let lookup_count = inode.dec_lookup_count(n);
+                if lookup_count < 1 {
+                    // Safe to remove, kernel no longer has a reference to it.
+                    trace!(ino, "removing inode from superblock");
+                    inodes.remove(&ino);
```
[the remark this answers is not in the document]
dannycjones: Was there somewhere you had in mind for making this check? I'm not against such a check, but I'm also not clear where we'd make such a check.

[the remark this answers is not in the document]
dannycjones: Aside - I think removing the `Ino -> Inode` mapping at this point is wrong. There's no reason I have right now to believe other parts of our filesystem implementation won't rely on the Inode being present in this map.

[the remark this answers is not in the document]
jamesbornholt: Those implementations would be wrong, which is kinda what I mean about having another way to check. I was thinking something like we'd `assert(!inode.forgotten())` any time we're about to return an inode out of `inode.rs`.

mountpoint-s3/src/fs.rs
```diff
@@ -1,3 +1,31 @@
+//! File system types and operations, not relying on our FUSE library.
```
jamesbornholt, line 1: general nit on this comment: we've been roughly trying to wrap at at 120 chars wherever we can (which is what our rustfmt config is set to, but it doesn't affect comments).

mountpoint-s3/src/fs.rs
```diff
@@ -1,3 +1,31 @@
...
+//!
+//! **Note:** this abstraction is not intended to abstract FUSE itself,
+//! only allow us to run tests without FUSE in the loop.
```
jamesbornholt, line 4:
```suggestion
//! only allow us to run tests without a real FUSE device in the loop.
```

mountpoint-s3/src/fs.rs
```diff
@@ -1,3 +1,31 @@
...
+//! If and when we add a new filesystem view that isn't FUSE related,
+//! we can consider how to ensure this module is generic.
+//!
+//! #### Counting references to [Inode]
+//!
+//! One of the responsibilities of [S3Filesystem] is to track the _lookup count_.
+//! Lookup count is a FUSE concept and tracks the number of times we provide a reference to an [InodeNo] to the FUSE driver,
```
jamesbornholt, line 11:
```suggestion
//! Lookup count is a FUSE concept and tracks the number of times we provide a reference to an [InodeNo] to the kernel,
```

mountpoint-s3/src/fs.rs
```diff
@@ -1,3 +1,31 @@
...
+//! ensuring that we never delete an [Inode] that may be references in the future by the FUSE driver.
```
jamesbornholt, line 12:
```suggestion
//! ensuring that we never forget about an [InodeNo] that may be referenced in the future by the kernel.
```

mountpoint-s3/src/fs.rs
```diff
@@ -1,3 +1,31 @@
...
+//! Note - this should not be confused with any counting of references to an [Inode] from its parent or a [FileHandle].
```
jamesbornholt, line 13:
```suggestion
//! This should not be confused with any counting of references to an [Inode] from its parent or a [FileHandle].
```

mountpoint-s3/src/fs.rs
```diff
@@ -1,3 +1,31 @@
...
+//!
+//! Some operations will not return a new [InodeNo] but rather one the FUSE driver already knows about.
+//! `open` is one such case since FUSE already provides the [InodeNo] as an argument when it calls open.
```
jamesbornholt, line 16: This isn't quite the intent -- any operation that takes an InodeNo as _input_ doesn't need to change the lookup count of that `InodeNo`.

mountpoint-s3/src/fs.rs
```diff
@@ -472,11 +511,13 @@ where
         Ok(Opened { fh, flags: 0 })
     }
 
+    /// Implementation for both FUSE `readdir` and `readdirplus`.
     pub async fn readdir<R: DirectoryReplier>(
         &self,
         parent: InodeNo,
         fh: u64,
         offset: i64,
+        plus: bool,
```
jamesbornholt, line 520:
```suggestion
        inc_lookup_count: bool,
```

mountpoint-s3/src/fuse.rs
```diff
@@ -1,3 +1,8 @@
+//! `fuser`-specific bindings, mostly wrapping the generic filesystem code in [crate::fs].
+//!
+//! Note that the purpose of this module is not necessarily to abstract away FUSE from a FileSystem,
+//! but instead to provide a thin wrapper between the `fuser` library and our filesystem handlers.
```
jamesbornholt, line 4: for testing purposes

mountpoint-s3/src/inode.rs
```diff
@@ -39,6 +39,8 @@ use crate::sync::{Arc, Mutex, RwLock};
 pub type InodeNo = u64;
 
 pub const ROOT_INODE_NO: InodeNo = 1;
+/// Starts at zero, as it tracks references Kernel has to the [Inode]. It's zero until returned in a FUSE reply.
+const DEFAULT_LOOKUP_COUNT: u64 = 0;
```
jamesbornholt, line 43:
```suggestion
const INITIAL_LOOKUP_COUNT: u64 = 0;
```

mountpoint-s3/src/inode.rs
```diff
@@ -97,6 +100,30 @@ impl Superblock {
...
+            None => {
+                debug_assert!(
+                    false,
+                    "forget should not be called on inode {ino} missing from superblock"
+                );
+                error!(ino, "forget called on inode missing from the superblock");
+            }
+            Some(inode) => {
+                let lookup_count = inode.dec_lookup_count(n);
+                if lookup_count < 1 {
```
jamesbornholt, line 118: it's unsigned, so
```suggestion
                if lookup_count == 0 {
```

mountpoint-s3/src/inode.rs
```diff
@@ -97,6 +100,30 @@ impl Superblock {
         Self { inner: Arc::new(inner) }
     }
 
+    /// The kernel tells us when it removes a reference to an [InodeNo] from its internal caches via a forget call.
```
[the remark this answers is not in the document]
dannycjones: It wasn't a priority for us - during alpha, it did not feel essential to cleanup the superblock list of Inodes.

Caching doesn't come into why we're implementing forget I don't think, but it may (IDK) influence _when_ the kernel is telling us its forgotten about a particular Inode.

passaro:
Superseded by #359

PR #225, opened 2023-04-27
Fuse Test Etag prefetcher

sauraank:
I have added the fuse test for checking E-tag behaviour, i.e, it is not prefetching the request from object if the object is mutated in between. In mock session it is always creating same E-Tag because `client = Arc::new(MockClient::new(client_config));`. ( new method for MockClient is adding default resulting in same e-tag always). So, I will be making that change further.

### DIFF
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -93,7 +93,7 @@ pub struct MockObject {
     // TODO Set storage class from [MockClient::put_object]
     storage_class: String,
     pub last_modified: OffsetDateTime,
-    pub etag: Option<ETag>,
+    pub etag: ETag,
 }
 
 impl MockObject {
@@ -109,7 +109,7 @@ impl MockObject {
             generator: Box::new(move |offset, size| bytes[offset as usize..offset as usize + size].into()),
             storage_class: "STANDARD".to_owned(),
             last_modified: OffsetDateTime::now_utc(),
-            etag: Some(etag),
+            etag,
         }
     }
 
@@ -119,7 +119,7 @@ impl MockObject {
             size,
             storage_class: "STANDARD".to_owned(),
             last_modified: OffsetDateTime::now_utc(),
-            etag: Some(etag),
+            etag,
         }
     }
 
@@ -139,7 +139,7 @@ impl MockObject {
             size,
             storage_class: "STANDARD".to_owned(),
             last_modified: OffsetDateTime::now_utc(),
-            etag: Some(etag),
+            etag,
         }
     }
 
@@ -267,7 +267,7 @@ impl ObjectClient for MockClient {
                 .etag
                 .clone();
             if let Some(etag_match) = if_match {
-                if etag_match != etag.expect("E-Tag should be set") {
+                if etag_match != etag {
                     return Err(ObjectClientError::ServiceError(GetObjectError::PreconditionFailed));
                 }
             }
@@ -305,7 +305,7 @@ impl ObjectClient for MockClient {
 
         let objects = self.objects.read().unwrap();
         if let Some(object) = objects.get(key) {
-            let etag = object.etag.clone().expect("E-Tag should be set");
+            let etag = object.etag.clone();
             Ok(HeadObjectResult {
                 bucket: bucket.to_string(),
                 object: ObjectInfo {
@@ -408,7 +408,7 @@ impl ObjectClient for MockClient {
                     key: key.to_string(),
                     size: object.len() as u64,
                     last_modified: object.last_modified,
-                    etag: object.etag.clone().expect("E-Tag should be set").as_str().to_string(),
+                    etag: object.etag.clone().as_str().to_string(),
                     storage_class: None,
                 });
             }
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -74,6 +74,17 @@ impl MockClient {
     pub fn remove_object(&self, key: &str) {
         self.objects.write().unwrap().remove(key);
     }
+
+    /// Returns `true` if this mock client's bucket contains the specified key
+    pub fn contains_key(&self, key: &str) -> bool {
+        self.objects.read().unwrap().contains_key(key)
+    }
+
+    /// Returns `true` if this mock client's bucket contains the specified common prefix
+    pub fn contains_prefix(&self, prefix: &str) -> bool {
+        let prefix = format!("{prefix}/");
+        self.objects.read().unwrap().keys().any(|k| k.starts_with(&prefix))
+    }
 }
 
 pub struct MockObject {
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -257,16 +257,10 @@ impl ObjectClient for MockClient {
         }
 
         let objects = self.objects.read().unwrap();
+
         if let Some(object) = objects.get(key) {
-            let etag = self
-                .objects
-                .read()
-                .unwrap()
-                .get(key)
-                .expect("object key should be set")
-                .etag
-                .clone();
             if let Some(etag_match) = if_match {
+                let etag = objects.get(key).expect("object key should be set").etag.clone();
                 if etag_match != etag {
                     return Err(ObjectClientError::ServiceError(GetObjectError::PreconditionFailed));
                 }
@@ -312,7 +306,7 @@ impl ObjectClient for MockClient {
                     key: key.to_string(),
                     size: object.size as u64,
                     last_modified: object.last_modified,
-                    etag: ETag::as_str(&etag).to_string(),
+                    etag: etag.as_str().to_string(),
                     storage_class: None,
                 },
             })
@@ -408,7 +402,7 @@ impl ObjectClient for MockClient {
                     key: key.to_string(),
                     size: object.len() as u64,
                     last_modified: object.last_modified,
-                    etag: object.etag.clone().as_str().to_string(),
+                    etag: object.etag.as_str().to_string(),
                     storage_class: None,
                 });
             }
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -85,23 +85,6 @@ impl MockClient {
         let prefix = format!("{prefix}/");
         self.objects.read().unwrap().keys().any(|k| k.starts_with(&prefix))
     }
-
-    /// Returns E-Tag for the object in the mock client with specified key. Return default if etag not set.
-    pub fn get_etag(&self, key: &str) -> ETag {
-        if let Some(etag) = self
-            .objects
-            .read()
-            .unwrap()
-            .get(key)
-            .expect("object key should be set")
-            .etag
-            .clone()
-        {
-            etag
-        } else {
-            ETag::for_tests()
-        }
-    }
 }
 
 pub struct MockObject {
@@ -175,7 +158,7 @@ impl MockObject {
 
 impl<T: AsRef<[u8]>> From<T> for MockObject {
     fn from(bytes: T) -> Self {
-        MockObject::from_bytes(bytes.as_ref(), ETag::for_tests())
+        MockObject::from_bytes(bytes.as_ref(), ETag::from_object_bytes(bytes.as_ref()))
     }
 }
 
@@ -269,18 +252,26 @@ impl ObjectClient for MockClient {
     ) -> ObjectClientResult<Self::GetObjectResult, GetObjectError, Self::ClientError> {
         trace!(bucket, key, ?range, ?if_match, "GetObject");
 
-        if let Some(etag) = if_match {
-            if etag != self.get_etag(key) {
-                return Err(ObjectClientError::ServiceError(GetObjectError::PreconditionFailed));
-            }
-        }
-
         if bucket != self.config.bucket {
             return Err(ObjectClientError::ServiceError(GetObjectError::NoSuchBucket));
         }
 
         let objects = self.objects.read().unwrap();
         if let Some(object) = objects.get(key) {
+            let etag = self
+                .objects
+                .read()
+                .unwrap()
+                .get(key)
+                .expect("object key should be set")
+                .etag
+                .clone();
+            if let Some(etag_match) = if_match {
+                if etag_match != etag.expect("E-Tag should be set") {
+                    return Err(ObjectClientError::ServiceError(GetObjectError::PreconditionFailed));
+                }
+            }
+
             let (next_offset, length) = if let Some(range) = range {
                 if range.start >= object.len() as u64 || range.end > object.len() as u64 {
                     return mock_client_error(format!("invalid range, length={}", object.len()));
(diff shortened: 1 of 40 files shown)


sauraank:
After resolving the conflicts, few of the things have changed which I need to work on to get the tests passing.

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -0,0 +1,101 @@
...
+use mountpoint_s3::S3FilesystemConfig;
+use std::fs::OpenOptions;
+use std::io::Read;
+use std::io::Seek;
+use tempfile::TempDir;
+use test_case::test_case;
+
+use crate::fuse_tests::TestClientBox;
+
+// test for checking either prefetching fails or read original object when object is mutated during read.
```
jamesbornholt, line 12: nit: this is a comment documenting the function, so it should be a doc comment (`///`) so that it shows up in your IDE, in documentation we generate, etc

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    let test_config = S3FilesystemConfig {
+        prefetcher_config,
+        ..Default::default()
+    };
+
+    let (mount_point, _session, mut test_client) = creator_fn(prefix, test_config);
+    let mut read_buf = vec![0u8; OBJECT_SIZE];
+
+    // Make sure there's an existing directory
```
jamesbornholt, line 33: probably just delete this comment (we don't care about the directory in this test)

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -0,0 +1,101 @@
...
+    test_client.put_object("dir/hello.txt", &read_buf).unwrap();
+
+    let mut path = mount_point.path().join("dir/hello.txt");
+
+    // Opening the file to read and write
```
jamesbornholt, line 38: it's open read-only here, though. probably delete this comment too -- it's just repeating the code rather than explaining anything else.

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -0,0 +1,101 @@
...
+    let mut f = OpenOptions::new()
+        .read(true)
+        .open(path)
+        .expect("should be able to open the file");
+
+    let mut buf = vec![0u8; read_size];
+    f.read_exact(&mut buf).expect("Should be able to read file to buf");
+
+    read_buf = vec![255u8; OBJECT_SIZE];
+    test_client.put_object("dir/hello.txt", &read_buf).unwrap();
```
jamesbornholt, line 48: explain in a comment why we're doing this (why 255? why write now?)

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    // In order to read from a position where prefetching next block is necessary
+    f.seek(std::io::SeekFrom::Start(SEEK_POS))
+        .expect("Seek position should be less than object size");
+    let mut dest_buf = vec![0u8; read_size];
+    match f.read_exact(&mut dest_buf) {
+        Ok(()) => assert_eq!(buf, dest_buf),
+        Err(err) => assert_eq!(err.raw_os_error(), Some(libc::EIO)),
+    };
```
jamesbornholt, line 57: rather than seeking, could we just keep reading until we get an error or reach end-of-file? then we can validate that no byte we successfully read is 255. that will make this test a little more resilient to the size of the prefetch window.

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    drop(f);
+
+    // File wont be mutated during read now to check if prefetcher reads correctly otherwise.
```
jamesbornholt, line 61: I think the actual explanation here is that re-opening the file will see the new file contents.

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -0,0 +1,101 @@
...
+#[test_case(64 * 1024, 1; "first request size smaller than seek position reading 1 byte")]
+#[test_case(512 * 1024, 1; "first request size greater than seek position reading 1 byte")]
+#[test_case(256 * 1024, 200 * 1024; "default first request size reading till seek")]
+fn prefetch_test_etag_mock(request_size: usize, read_size: usize) {
+    // prefetch_test_etag(
+    //     crate::fuse_tests::mock_session::new,
+    //     "prefetch_etag_test",
+    //     request_size,
+    //     read_size,
+    // );
```
jamesbornholt, line 93: could we implement the etag check in the mock client too? shouldn't be very hard, and then we'll be able to have a version of this test that doesn't require S3 creds, which we should aspire to do whenever we can

jamesbornholt (commented):
Neat! Just one suggestion about doing this test slightly differently, and a few small things.

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
sauraank: Yes. We will have to generate a new e-tag when we create a mock client. Thanks.

sauraank: Oh! It is failing with different etags as well in mock client. I need dive deeper on how mock client is reading.

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
sauraank: Like starting to read from start and keep reading the full object size? In that case I should read the object full in the first read as well. So, all the blocks will be prefetched and we will always get one of the case, i.e., data in `dest_buf `same as `buf`. We will never get the case of prefetching failing due to etag.

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -252,6 +269,12 @@ impl ObjectClient for MockClient {
     ) -> ObjectClientResult<Self::GetObjectResult, GetObjectError, Self::ClientError> {
         trace!(bucket, key, ?range, ?if_match, "GetObject");
 
+        if let Some(etag) = if_match {
```
passaro, line 272: Shouldn't the etag be checked after the object is found? 

I think that also means we do not need the `get_etag` method.

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -291,13 +314,14 @@ impl ObjectClient for MockClient {
 
         let objects = self.objects.read().unwrap();
         if let Some(object) = objects.get(key) {
+            let etag = object.etag.clone().expect("E-Tag should be set");
```
passaro, line 317: Why is it an Option then?

mountpoint-s3/tests/fuse_tests/mod.rs
```diff
@@ -84,7 +87,8 @@ mod mock_session {
     impl TestClient for MockTestClient {
         fn put_object(&mut self, key: &str, value: &[u8]) -> Result<(), Box<dyn std::error::Error>> {
             let full_key = format!("{}{}", self.prefix, key);
-            self.client.add_object(&full_key, value.into());
```
passaro, line 86: Would it be better to do this in the `From` implementation for MockObject? (mock_client.rs:161)

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -85,16 +84,16 @@ where
...
-    //     read_size,
-    // );
+    prefetch_test_etag(
+        crate::fuse_tests::mock_session::new,
+        "prefetch_etag_test_mock",
+        request_size,
+        read_size,
+    );
 
     prefetch_test_etag(
```
passaro, line 94: I think the pattern we are using elsewhere is to have 2 functions, see for example write_tests.rs. 

In this case, they would be: 
* `fn prefetch_test_etag_mock` 
* `fn prefetch_test_etag_s3` with `#[cfg(feature = "s3_tests")]`

That means there can be some duplication in the `test_case` attributes, but in general they do not need to be the same.

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -30,12 +30,10 @@ where
     let (mount_point, _session, mut test_client) = creator_fn(prefix, test_config);
     let mut read_buf = vec![0u8; OBJECT_SIZE];
```
passaro, line 31: Could we use a more descriptive name here? Maybe: 
```suggestion
    let original_data_buf = vec![0u8; OBJECT_SIZE];
```

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -44,6 +42,7 @@ where
     let mut buf = vec![0u8; read_size];
     f.read_exact(&mut buf).expect("Should be able to read file to buf");
 
+    // changed the value of read_buf to distinguish it from previous data of the object.
     read_buf = vec![255u8; OBJECT_SIZE];
```
passaro, line 46:
```suggestion
    let new_data_buf = vec![255u8; OBJECT_SIZE];
```

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
jamesbornholt: No, not quite, I mean something like this:
```
read next 100KiB
do the new put
while !EOF {
    read next 100KiB
    assert!(read failed || read saw the old bytes);
    if read failed {
        break;
    }
}
```
Basically, do a small read from the file first (to start the prefetcher), then do the PUT, then read the rest of the file. We expect the reads to fail eventually (once we get outside the already-prefetched data), but in the meantime, they should still return the old values. If they ever return the new bytes, that's a bug.

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
passaro: Also, the assert on line 69 could compare `dest_buf` with a slice from this buffer.

mountpoint-s3-client/src/object_client.rs
```diff
@@ -20,11 +22,22 @@ impl ETag {
...
             etag: "test_etag".to_string(),
         }
     }
+
+    // Creating unique etag from bytes
+    pub fn from_object_bytes(data: &[u8]) -> Self {
+        let mut hasher = Md5::new();
+        hasher.update(data);
+
+        let hash = hasher.finalize();
```
passaro, line 37: nit:  have you considered something like this:
```suggestion
        let hash = Md5::new()
            .update(data)
            .finalize();
```
may be subjective though :)

sauraank:
I have made test to keep prefetching until it fails or reach end of file. Also, added checking E-Tag during Mock session by adding E-Tag in Mock Client's head_object and list_object. Added the if_match parameter in get_object() of Mock Client. Please review it.

passaro (changes requested):

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    f.read_to_end(&mut new_dest_buf)
+        .expect("Should be able to read file to buf");
+    // Now 'new_dest_buf' should have new data.
+    assert_eq!(new_dest_buf, final_data_buf);
+
+    drop(f);
+}
+
+// Prefetching of next request occurs when more than half of the current request is being read.
```
passaro, line 90: Are these comments for the test or the specific cases?

Ideally, each case description should give enough details (or they could have a specific comment if really needed).

If the comment is for the whole test, we could expand the description of `prefetch_test_etag` instead.

passaro (dismissed):

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -0,0 +1,121 @@
...
+        };
+
+        match f.read_exact(&mut dest_buf) {
+            // since all bytes are same, we can assert every slice will be equal.
+            Ok(()) => assert_eq!(reader_buf, dest_buf),
+            Err(err) => {
+                if let Some(error) = err.raw_os_error() {
+                    assert_eq!(error, libc::EIO);
+                }
+                break;
```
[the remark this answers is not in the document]
sauraank: Getting a case where the error is None. Looking into it.

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
```diff
@@ -0,0 +1,121 @@
...
+    // changed the value of data buf to distinguish it from previous data of the object.
+    let final_data_buf = vec![255u8; OBJECT_SIZE];
+    test_client.put_object("dir/hello.txt", &final_data_buf).unwrap();
+
+    let mut seek_pos: u64 = 0;
+    let mut dest_buf = vec![0u8; read_size];
+
+    // Reading the file untill we keep getting the prefetched data or we get an IO error where E-Tag did not match.
+    loop {
+        seek_pos = match f.seek(std::io::SeekFrom::Start(seek_pos + read_size as u64)) {
```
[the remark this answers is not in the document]
sauraank: I dont think so. It is seeking to the next block of read_size to keep reading untill we are not able to seek (which should be EOF)?

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -291,13 +305,14 @@ impl ObjectClient for MockClient {
 
         let objects = self.objects.read().unwrap();
         if let Some(object) = objects.get(key) {
+            let etag = object.etag.clone();
```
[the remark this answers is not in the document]
sauraank: Yes, It is. Because `object` type is `&Arc<MockObject>`. And we cannot move etag out of an Arc without cloning.

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
sauraank: Like, it is generally adding why the test should behave the way they are. I can move it to the description of `prefetch_test_etag` instead.

mountpoint-s3/tests/fuse_tests/prefetch_test.rs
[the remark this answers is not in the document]
sauraank: Not getting the None error now.

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -93,7 +93,7 @@ pub struct MockObject {
     // TODO Set storage class from [MockClient::put_object]
     storage_class: String,
     pub last_modified: OffsetDateTime,
-    pub etag: Option<ETag>,
+    pub etag: ETag,
```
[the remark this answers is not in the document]
sauraank: Right. Making the change.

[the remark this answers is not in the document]
sauraank: Done. Also, made changes for other recommendations. Thanks.

jamesbornholt (approved):

Issue #218, opened 2023-04-24
Mountpoint crashes on reading 1 byte file


sauraank:
### Mountpoint for Amazon S3 version

mountpoint-s3 0.2.0-70b2b68

### AWS Region

eu-west-1

### Describe the running environment

Running in EC2 on Amazon Linux 2 using instance profile credentials against an S3 bucket in the same account.
Also same happened on on my Mac using instance profile credentials against an S3 bucket.

### What happened?

Mountpoint is constantly crashing when trying to read a 1 byte file. Tried it on several such objects. When they are  wriiten to a file using getObject s3api CLI command, output the following result -
`.%` (where % is highlighted).
I am attaching the trace logs. Prefetcher is able to find the correct range and also finishing the rpefetching task. But it stops working after that.

### Relevant log output

```shell
2023-04-21T16:40:51.019337Z TRACE getattr{req=5 ino=5}: mountpoint_s3::fs: fs:getattr with ino 5
2023-04-21T16:40:51.019471Z TRACE open{req=2 ino=5}: mountpoint_s3::fs: fs:open with ino 5 flags 0
2023-04-21T16:40:51.019742Z TRACE read{req=6 ino=5 fh=1 offset=0 size=1}: mountpoint_s3::fs: fs:read with ino 5 fh 1 offset 0 size 1
2023-04-21T16:40:51.019770Z TRACE read{req=6 ino=5 fh=1 offset=0 size=1}: mountpoint_s3::prefetch: read offset=0 length=1 next_seq_offset=0
2023-04-21T16:40:51.019789Z TRACE read{req=6 ino=5 fh=1 offset=0 size=1}: mountpoint_s3::prefetch: to_read=1
2023-04-21T16:40:51.019812Z TRACE read{req=6 ino=5 fh=1 offset=0 size=1}: mountpoint_s3::prefetch: spawning request range=0..1 size=1
2023-04-21T16:40:51.020002Z DEBUG read{req=6 ino=5 fh=1 offset=0 size=1}:prefetch{range=0..1}:get_object{id=21}: mountpoint_s3_client::s3_crt_client::get_object: new request bucket="plutotestankit" key="test/test_sdk_create_object/14899866888963238197/hello" range=Some(0..1) if_match=Some(ETag { etag: "\"5058f1af8388633f609cadb75a75dc9d\"" }) size=Some(1)
[...]

PR #226, opened 2023-04-28
Add get_object test with 1 byte range

passaro:
Demonstrate issue #218.

This test reproduces the underlying cause of the issue reported in #218, isolated on `mountpoint-s3-client`. 

Reproduction
* Make a GetObject request with a range of `bytes=0-0`

Expected
* response with body of size 1

Actual 
* response with empty body (but `Content-Length: 1`)

Investigation still in progress. Notes so far:

**mountpoint** 

Although the tests shows the issue on GetObject, we should also review the calling code in the prefetcher: can we detect the inconsistency and report an error rather than get stuck?

**CRT**

It looks like the issue is in CRT code:
* A similar problem was recently fixed here: [PR 272](https://github.com/awslabs/aws-c-s3/pull/272)
* We have updated `aws-c-s3` to `v0.2.8`, which includes the previous fix.
* Unfortunately, this does not seem to have fixed our issue - is it specific to range of `bytes=0-0`?

---

### DIFF
--- a/mountpoint-s3-client/tests/get_object.rs
+++ b/mountpoint-s3-client/tests/get_object.rs
@@ -76,6 +76,32 @@ async fn test_get_object_large() {
 }
 
 #[tokio::test]
+async fn test_get_object_one_byte_range() {
+    let sdk_client = get_test_sdk_client().await;
+    let (bucket, prefix) = get_test_bucket_and_prefix("test_get_object_one_byte_range");
+
+    // Create one object named "onebyte"
+    let key = format!("{prefix}/onebyte");
+    let body = b".";
+    sdk_client
+        .put_object()
+        .bucket(&bucket)
+        .key(&key)
+        .body(ByteStream::from(Bytes::from_static(body)))
+        .send()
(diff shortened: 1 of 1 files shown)


passaro:
Just a draft: not to be merged for now.

Once the issue is fixed, we could review the test (maybe convert to `test_case`?) and merge.

mountpoint-s3-client/tests/get_object.rs
```diff
@@ -75,6 +75,32 @@ async fn test_get_object_large() {
...
+        .bucket(&bucket)
+        .key(&key)
+        .body(ByteStream::from(Bytes::from_static(body)))
+        .send()
+        .await
+        .unwrap();
+
+    let client: S3CrtClient = get_test_client();
+
+    let result = client
```
sauraank, line 97: As I can see we are not checking for E-Tag here in get_object. Just wanted to know that while testing a feature, we should enter every parameter for it for it being more robust or just the ones relevant for our use case? (In general scenarios)

PR #229, opened 2023-05-01
Basic property-based tests for writes

jamesbornholt:
This is the first step in property-based testing for the write path.
Right now, it tests scenarios where the entire write sequence happens at
once: mknod, open, write once, release. It generates a vector of such
operations, with random paths and random file contents/sizes. After each
operations, the test checks that the file system has the expected
contents.

I'm stopping this PR here because it's enough to get the testing off the
ground, but doesn't yet cover a lot of the interesting cases we'd like to
check. The focus here is getting the infrastructure in place to start
adding more complex operations to the `Op` structure in future PRs.

---

### DIFF
--- a/mountpoint-s3/tests/reftests/harness.rs
+++ b/mountpoint-s3/tests/reftests/harness.rs
@@ -1,6 +1,6 @@
 use crate::common::{make_test_filesystem, DirectoryReply, ReadReply};
-use crate::reftests::gen_tree::{flatten_tree, gen_tree, Content, FileSize, Name, TreeNode};
-use crate::reftests::reference::{build_reference, Node, Reference};
+use crate::reftests::generators::{flatten_tree, gen_tree, valid_name_strategy, FileContent, FileSize, Name, TreeNode};
+use crate::reftests::reference::{build_reference, File, Node, Reference};
 use fuser::FileType;
 use futures::executor::ThreadPool;
 use futures::future::{BoxFuture, FutureExt};
@@ -11,9 +11,41 @@ use mountpoint_s3::{
 };
 use mountpoint_s3_client::mock_client::{MockClient, MockObject};
 use proptest::prelude::*;
+use proptest_derive::Arbitrary;
 use std::collections::{BTreeMap, HashSet};
 use std::fmt::Debug;
+use std::path::{Component, Path};
 use std::sync::Arc;
+use tracing::debug;
+
+/// Operations that the mutating proptests can perform on the file system.
+// TODO: mkdir, unlink
+// TODO: "reboot" (forget all the local inodes and re-bootstrap)
+// TODO: incremental writes (test partially written files)
+#[derive(Debug, Arbitrary)]
+pub enum Op {
+    WriteFile(
+        #[proptest(strategy = "valid_name_strategy()")] String,
+        DirectoryIndex,
+        FileContent,
+    ),
+}
+
+/// An index into the reference model's list of directories. We use this to randomly select an
+/// existing directory to operate on (for put, rmdir, etc).
+#[derive(Debug, Arbitrary)]
+pub struct DirectoryIndex(usize);
+
+impl DirectoryIndex {
+    /// Get the full path to the directory at the given index in the reference (wrapping around if
+    /// the index is larger than the number of directories)
+    fn get<'a>(&self, reference: &'a Reference) -> &'a str {
+        let directories = reference.directories();
+        assert!(!directories.is_empty(), "directories can never be empty");
+        let idx = self.0 % directories.len();
+        &directories[idx]
+    }
+}
 
 #[derive(Debug)]
 pub struct Harness {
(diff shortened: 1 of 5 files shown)


mountpoint-s3/tests/reftests/harness.rs
```diff
@@ -143,189 +280,249 @@ impl Harness {
...
+}
 
-    /// Walk the filesystem tree and check that at each level, contents match the reference
-    async fn compare_contents(&self) {
-        let root = self.reference.root();
-        self.compare_contents_recursive(FUSE_ROOT_INODE, FUSE_ROOT_INODE, root)
-            .await;
+/// Read-only reftests that generate random S3 buckets and check the mapping from S3 keys to file
+/// paths is correct.
+mod read_only {
```
jamesbornholt, line 287: GitHub doesn't do a good job rendering this part of the diff, but it's just taking all the existing proptest code and sticking it into a `read_only` module, so we can separate it from the new tests.

mountpoint-s3/tests/reftests/reference.rs
```diff
@@ -110,6 +125,31 @@ impl Reference {
...
+                    children.get(&dir)?
+                }
+                _ => panic!("unexpected internal file node"),
+            };
+        }
+
+        Some(node)
+    }
+
+    pub fn directories(&self) -> &[String] {
```
passaro, line 150: Maybe document here that this is never empty?

mountpoint-s3/tests/reftests/reference.rs
```diff
@@ -110,6 +125,31 @@ impl Reference {
...
+
+    /// Get a node from a full path, if it exists
+    pub fn lookup(&self, path: &str) -> Option<&Node> {
+        let path = Path::new(path);
+
+        let mut components = path.components().peekable();
+        assert_eq!(components.next(), Some(Component::RootDir));
+
+        let mut node = &self.root;
+        for dir in components {
```
passaro, line 137: `dir` misled me initially: the last one can be a file.

mountpoint-s3/tests/reftests/reference.rs
```diff
@@ -110,6 +125,31 @@ impl Reference {
...
+            node = match node {
+                Node::Directory(children) => {
+                    let dir = dir.as_os_str().to_str().unwrap().to_string();
+                    children.get(&dir)?
+                }
+                _ => panic!("unexpected internal file node"),
```
passaro, line 143: If I understand correctly, `lookup("/a/b")` panics if `a` is a file, but returns `None` if `a` is a directory (and `b` does not exist). 

Is that correct? Why not return `None` in both cases?

mountpoint-s3/tests/reftests/reference.rs
```diff
@@ -110,6 +125,31 @@ impl Reference {
             };
         }
     }
+
+    /// Get a node from a full path, if it exists
+    pub fn lookup(&self, path: &str) -> Option<&Node> {
+        let path = Path::new(path);
+
+        let mut components = path.components().peekable();
```
passaro, line 133: Why `peekable()`?

mountpoint-s3/tests/reftests/reference.rs
passaro: The only usage is to lookup a file in an existing directory, so it works out there. Still maybe this is not a general method?

mountpoint-s3/tests/reftests/harness.rs
```diff
@@ -23,14 +55,122 @@ pub struct Harness {
...
 
+    /// Run a sequence of mutation operations on the test harness, checking equivalence between the
+    /// reference model and file system after each operation.
+    pub async fn run(&mut self, ops: Vec<Op>) {
+        for op in ops {
+            debug!(?op, "executing operation");
+            match &op {
+                Op::WriteFile(name, directory_index, contents) => {
+                    let dir = Path::new(directory_index.get(&self.reference));
+                    let full_path = format!("{}/{name}", dir.display());
```
passaro, line 75: Is there a reason to prefer this to `dir.join(name)...`?

passaro (commented):
LGTM, just a couple of minor comments

mountpoint-s3/tests/reftests/reference.rs
jamesbornholt: Yeah, it should just return None here, I think. In the future we might want to be able to distinguish "a directory along the way didn't exist" from "the leaf didn't exist", but not right now.

mountpoint-s3/tests/reftests/harness.rs
jamesbornholt: Ended up doing this and then changing `add_file` and `lookup` to take `impl AsRef<Path>` instead of `&str`, so that we're not doing manual path manipulation anywhere.

mountpoint-s3/tests/reftests/harness.rs
```diff
@@ -143,189 +287,248 @@ impl Harness {
...
+                ),
+                Op::WriteFile(
+                    "b".to_string(),
+                    DirectoryIndex(1),
+                    FileContent(0x0b, FileSize::Small(10)),
+                ),
+            ],
+            0,
+        );
+    }
```
dannycjones, line 521: Just to confirm understanding on this one (and I see we already had something similar for RO), we're just checking that the following filesystem behaves as expected so that we can be sure we didn't break some basic case (without needing to worry about the trees that proptest will generate)?

```
/ (dir)
/-/ (dir)
/-/- (file)
```

becomes...


```
/ (dir)
/-/ (dir)
/a (file)
/-/- (file)
/-/b (file)
```

dannycjones (approved):
Also reviewed, LGTM.

mountpoint-s3/tests/reftests/harness.rs
passaro: Nice, I like it

passaro (approved):

mountpoint-s3/tests/reftests/harness.rs
jamesbornholt: Yeah exactly, it's just a useful sanity check. It's most helpful when working on the tests themselves, because it's a lot easier to debug a small test like this than to wait for proptest to do its thing.

PR #230, opened 2023-05-02
Extra supports for cutomer specified profile.

eslrahc-swa:
---


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -648,4 +648,52 @@ mod tests {
 
         assert_eq!(expected_user_agent, user_agent_header_value);
     }
+
+    #[test]
+    #[ignore]
+    // Simple test to ensure the user agent header is correct even when given profile contains calid credentials.
+    fn test_client_with_valid_credentials() {
+        let expected_user_agent = "mountpoint-s3-client";
+        let profile_name_override = String::from("VALID_CREDENTIALS");
```
eslrahc-swa, line 657: I am not sure if there are good ways to test remote and automatically:
* configure assume role and sts token on remote testing machine.
* configure `VALIDE_PROFILE` on remote testing machine.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -648,4 +648,50 @@ mod tests {
 
         assert_eq!(expected_user_agent, user_agent_header_value);
     }
+
+    #[test]
+    #[ignore]
+    // Simple test to ensure the user agent header is correct even when given profile contains calid credentials.
```
dannycjones, line 654: typo: calid, valid

Also, may as well use RustDoc when commenting on the method.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -648,4 +648,50 @@ mod tests {
...
+
+        let config = S3ClientConfig {
+            profile_name_override: Some(profile_name_override),
+            ..Default::default()
+        };
+
+        let client = S3CrtClient::new("eu-west-1", config).expect("Create test client");
+
+        let mut message = client
+            .new_request_template("GET", "plutotestankit")
```
dannycjones, line 667: for bucket name, let's use something generic. maybe `some-bucket` or review another example.

put this in a constant in this test file. (probably better if we define this somewhere common TBH, but we can do that later.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -648,4 +648,50 @@ mod tests {
...
+            .expect("new request template expected");
+
+        let headers = message.inner.get_headers().expect("Expected a block of HTTP headers");
+
+        let user_agent_header = headers
+            .get("User-Agent")
+            .expect("User Agent Header expected with given prefix");
```
dannycjones, line 674: That's not what we are expecting, only that "user agent header should be set".

```suggestion
        let user_agent_header = headers
            .get("User-Agent")
            .expect("user agent header should be set");
```

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -648,4 +648,50 @@ mod tests {
...
+            profile_name_override: Some(profile_name_override),
+            ..Default::default()
+        };
+
+        let client = S3CrtClient::new("eu-west-1", config).expect("Create test client");
+
+        let mut message = client
+            .new_request_template("GET", "plutotestankit")
+            .expect("new request template expected");
+    }
```
dannycjones, line 696: When should this panic? During client creation?

We can use `expect_err` if we want to test if something is `Err` instead of `Ok`, rather than using `should_panic`.

Side note - the profile doesn't exist, right? Let's describe it that way in the string, I was wondering what credentials were invalid.

```suggestion
    #[test]
    // Simple test with invalid credentials when creating S3CrtClient.
    fn test_client_with_invalid_credentials() {
        let profile_name_override = String::from("NONEXISTENT_PROFILE");

        let config = S3ClientConfig {
            profile_name_override: Some(profile_name_override),
            ..Default::default()
        };

        let client = S3CrtClient::new("eu-west-1", config).expect("create test client");

        let mut message = client
            .new_request_template("GET", "some-bucket")
            .expect_err("request construction should fail");
    }
```

dannycjones (changes requested):
Some minor comments ignoring how we can make these pass in CI, given dependence on `.aws/config` right now.

We should drop the CRT changes from this PR. I think you just need a rebase or something.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -648,4 +651,78 @@ mod tests {
...
+        delete_credentials();
+
+        let profile_name_override = String::from("NONEXISTENT_CREDENTIALS");
+        let config = S3ClientConfig {
+            profile_name_override: Some(profile_name_override),
+            ..Default::default()
+        };
+        let error = S3CrtClient::new("eu-west-1", config).unwrap_err();
+
+        assert_eq!(error.to_string(), "invalid AWS credentials");
```
jamesbornholt, line 691: Check the error directly, don't convert it to a string.

dannycjones (changes requested):

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -648,4 +662,92 @@ mod tests {
...
+            ..Default::default()
+        };
+        let error = S3CrtClient::new("eu-west-1", config).unwrap_err();
+        let expected_error = NewClientError::ProviderFailure(CrtErr::from(AWS_ERROR_FILE_INVALID_PATH));
+
+        assert_eq!(error, expected_error);
+    }
+
+    #[test]
+    fn test_create_client_with_invalid_credentials() {
```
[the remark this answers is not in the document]
eslrahc-swa: Making request is the only way for detecting  invalid credentials. However, if we move this to integration test, we have to make `s3_crt_client module` and `makeSimpleHttp` request public. So I think it would be fine to keep this type of test here.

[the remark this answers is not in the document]
jamesbornholt: I don't really understand this test anyway -- it doesn't check that the error is right. Seems like we could just move the test to `tests/` where we have a real S3 bucket to test against, so we can verify that the failure is due to credentials rather than the bucket not existing or something.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -746,69 +721,6 @@ mod tests {
         assert_eq!(error, expected_error);
     }
 
-    #[test]
```
eslrahc-swa: delete this test for this pr and separate it as an integration test in a new PR

PR #232, opened 2023-05-03
Implement `unlink` operation for remote files

dannycjones:
PR focusing on `unlink`, `unlinkat` only.

- This implementation does not attempt to tackle the unbounded growth of the superblock. This should be tackled at a later date.
- This implementation does not allow a file that is being written locally to be unlinked. The file must be closed (and written to S3) first.

---

### DIFF
--- a/doc/SEMANTICS.md
+++ b/doc/SEMANTICS.md
@@ -74,6 +74,19 @@ Synchronization operations (`fsync`, `fdatasync`) are currently no-ops because w
 
 Space allocation (`fallocate`, `posix_fallocate`) are not supported.
 
+#### Deletes
+
+File deletion (`unlink`) is planned but not yet supported (see [#78](https://github.com/awslabs/mountpoint-s3/issues/78)).
+When `unlink` is implemented, the following semantics are proposed.
+
+For files not yet committed to S3, the client will not permit `unlink` operations.
+The file should be closed and thus committed to S3, at which point an `unlink` can be performed on the remote file.
+
+For files already committed to S3, the client will _immediately_ delete the corresponding object from S3,
+and remove the file from its directory.
+If there are still open file handles to the file, future reads to them will fail.
+Because the object is immediately deleted from S3, future reads from other hosts will also fail.
+
 ### Directory operations
 
 Basic read-only directory operations (`opendir`, `readdir`, `closedir`) are supported.
@@ -86,7 +99,7 @@ Creating directories (`mkdir`) is not currently supported, but will be [in the f
 
 Renaming files and directories (`rename`, `renameat`) is not currently supported.
 
-File deletion (`unlink`) is not currently supported, but will be [in the future](https://github.com/awslabs/mountpoint-s3/issues/78).
+File deletion (`unlink`) semantics are described in the [Deletes](#Deletes) section.
 
 Empty directory removal (`rmdir`) is not supported.
 
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -635,7 +635,7 @@ impl From<InodeError> for i32 {
             InodeError::ShadowedByDirectory(_, _) => libc::ENOENT,
             InodeError::FileAlreadyExists(_) => libc::EEXIST,
             // Not obvious what InodeNotWritable, InodeNotReadableWhileWriting should be.
-            // EINVAL or EROFS would also be reasonable -- but we'll treat them like a sealed file for now.
+            // EINVAL or EROFS would also be reasonable -- but we'll treat them like sealed files for now.
             InodeError::InodeNotWritable(_) => libc::EPERM,
             InodeError::InodeNotReadableWhileWriting(_) => libc::EPERM,
             InodeError::UnlinkNotPermittedWhileWriting(_) => libc::EPERM,
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -1,4 +1,5 @@
 use futures::task::Spawn;
+use futures::TryFutureExt;
 use nix::unistd::{getgid, getuid};
 use std::collections::HashMap;
 use std::ffi::OsStr;
@@ -616,6 +617,13 @@ where
         let mut dir_handles = self.dir_handles.write().await;
         dir_handles.remove(&fh).map(|_| ()).ok_or(libc::EBADF)
     }
+
+    pub async fn unlink(&self, parent_ino: InodeNo, name: &OsStr) -> Result<(), libc::c_int> {
+        self.superblock
+            .unlink(&self.client, parent_ino, name)
+            .map_err(Into::into)
+            .await
+    }
 }
 
 impl From<InodeError> for i32 {
@@ -626,12 +634,14 @@ impl From<InodeError> for i32 {
             InodeError::InodeDoesNotExist(_) => libc::ENOENT,
             InodeError::InvalidFileName(_) => libc::EINVAL,
             InodeError::NotADirectory(_) => libc::ENOTDIR,
+            InodeError::IsDirectory(_) => libc::EPERM,
             InodeError::ShadowedByDirectory(_, _) => libc::ENOENT,
             InodeError::FileAlreadyExists(_) => libc::EEXIST,
-            // Not obvious what these two cases should be -- EINVAL would also be reasonable, or
-            // EROFS for not-writable -- but we'll treat it like a sealed file
+            // Not obvious what InodeNotWritable, InodeNotReadableWhileWriting should be.
+            // EINVAL or EROFS would also be reasonable -- but we'll treat them like a sealed file for now.
             InodeError::InodeNotWritable(_) => libc::EPERM,
             InodeError::InodeNotReadableWhileWriting(_) => libc::EPERM,
+            InodeError::UnlinkNotPermittedWhileWriting(_) => libc::EPERM,
         }
     }
 }
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -1,5 +1,4 @@
 use futures::task::Spawn;
-use futures::TryFutureExt;
 use nix::unistd::{getgid, getuid};
 use std::collections::HashMap;
 use std::ffi::OsStr;
@@ -619,10 +618,8 @@ where
     }
 
     pub async fn unlink(&self, parent_ino: InodeNo, name: &OsStr) -> Result<(), libc::c_int> {
-        self.superblock
-            .unlink(&self.client, parent_ino, name)
-            .map_err(Into::into)
-            .await
+        self.superblock.unlink(&self.client, parent_ino, name).await?;
+        Ok(())
     }
 }
 
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -358,6 +358,10 @@ impl Superblock {
     /// Unlink the entry described by `parent_ino` and `name`.
     ///
     /// If the entry exists, delete it from S3 and the superblock.
+    ///
+    /// We know that the Linux Kernel's VFS will lock both the parent and child,
+    /// so we can safely ignore concurrent operations within the same Mountpoint process to the file and its parent.
+    /// See: https://www.kernel.org/doc/html/next/filesystems/directory-locking.html
     pub async fn unlink<OC: ObjectClient>(
         &self,
         client: &OC,
@@ -413,14 +417,19 @@ impl Superblock {
                 debug_assert!(false, "inodes never change kind");
                 return Err(InodeError::NotADirectory(parent.ino()));
             }
-            InodeKindData::Directory {
-                children,
-                writing_children,
-            } => {
-                writing_children.remove(&inode.ino());
-                if children.remove(inode.name()).is_none() {
-                    debug!("child was not present, another operation may have occurred in parallel");
-                }
+            InodeKindData::Directory { children, .. } => {
+                // We want to remove the original child.
+                // We assume that the VFS will hold a lock on the parent and child.
+                // However, we don't hold this lock over remote calls as we don't want to move to async locks right now.
+                // Instead, we will panic when our assumption appears broken.
+                let removed_inode = children
+                    .remove(inode.name())
+                    .expect("parent should contain child assuming VFS does not permit concurrent op on parent");
+                assert_eq!(
+                    removed_inode.ino(),
+                    inode.ino(),
+                    "child ino number shouldn't change assuming VFS does not permit concurrent op on parent",
+                );
             }
         };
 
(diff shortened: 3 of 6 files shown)


mountpoint-s3/src/fs.rs
```diff
@@ -591,6 +592,13 @@ where
             }
         }
     }
+
+    pub async fn unlink(&self, parent_ino: InodeNo, name: &OsStr) -> Result<(), libc::c_int> {
+        self.superblock
+            .unlink(&self.client, parent_ino, name)
+            .map_err(Into::into)
+            .await
```
passaro, line 600: Is the `map_err` needed?

We could do something like this instead (and remove the `use` above):
```suggestion
        self.superblock.unlink(&self.client, parent_ino, name).await?;
        Ok(())
```

mountpoint-s3/src/inode.rs
```diff
@@ -360,6 +360,68 @@ impl Superblock {
...
+                debug_assert!(false, "inodes never change kind");
+                return Err(InodeError::NotADirectory(parent.ino()));
+            }
+            InodeKindData::Directory {
+                children,
+                writing_children,
+            } => {
+                writing_children.remove(&inode.ino());
+                if children.remove(inode.name()).is_none() {
+                    debug!("child was not present, another operation may have occurred in parallel");
```
jamesbornholt, line 418: I think we want to arrange the locking/state such that this state isn't possible. Maybe what we need is to record in the inode itself that a DeleteObject is in flight for the inode.

doc/SEMANTICS.md
```diff
@@ -74,6 +74,19 @@ Synchronization operations (`fsync`, `fdatasync`) are currently no-ops because w
...
+File deletion (`unlink`) is planned but not yet supported (see [#78](https://github.com/awslabs/mountpoint-s3/issues/78)).
+When `unlink` is implemented, the following semantics are proposed.
+
+For files not yet committed to S3, the client will not permit `unlink` operations.
+The file should be closed and thus committed to S3, at which point an `unlink` can be performed on the remote file.
+
+For files already committed to S3, the client will _immediately_ delete the corresponding object from S3,
+and remove the file from its directory.
+If there are still open file handles to the file, future reads to them will fail.
+Because the object is immediately deleted from S3, future reads from other hosts will also fail.
```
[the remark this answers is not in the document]
dannycjones: I don't think we will want to delay deleting the file until there are no open handles, even in the future. My reasoning being that an immediate delete on S3 is much easier for customers to reason about compared to a delayed delete.

Do you foresee there being use cases blocked by ruling this out now?

mountpoint-s3/src/fs.rs
dannycjones: I see, yeah we can do either. I'll use `?` just to be consistent.

mountpoint-s3/src/inode.rs
```diff
@@ -360,6 +360,68 @@ impl Superblock {
...
+        match &mut parent_state.kind_data {
+            InodeKindData::File { .. } => {
+                debug_assert!(false, "inodes never change kind");
+                return Err(InodeError::NotADirectory(parent.ino()));
+            }
+            InodeKindData::Directory {
+                children,
+                writing_children,
+            } => {
+                writing_children.remove(&inode.ino());
```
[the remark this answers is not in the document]
sauraank: We can probably check `if writing_children.contains(&inode.ino())`, and if there is new local file with same name being created, we may abort our operation, as new local file will overwrite the remote object anyway?

mountpoint-s3/src/inode.rs
```diff
@@ -1042,12 +1105,16 @@ pub enum InodeError {
     ShadowedByDirectory(String, InodeNo),
     #[error("inode {0} is not a directory")]
     NotADirectory(InodeNo),
+    #[error("inode {0} is a directory")]
```
sauraank, line 1108: What is meant by inode {0} here?

mountpoint-s3/tests/fuse_tests/unlink_test.rs
```diff
@@ -0,0 +1,121 @@
...
+    );
+}
+
+#[cfg(feature = "s3_tests")]
+#[test]
+fn simple_unlink_test_s3() {
+    simple_unlink_tests(crate::fuse_tests::s3_session::new, "simple_unlink_tests");
+}
+
+#[test_case(""; "no prefix")]
```
sauraank, line 63: Is there any specific reason we perform the test cases with no prefix and with prefix? I have seen it in multiple other places.

mountpoint-s3/tests/fuse_tests/unlink_test.rs
```diff
@@ -0,0 +1,121 @@
...
+        vec!["other.txt"]
+    );
+
+    let _new_pos = f.seek(SeekFrom::Start((B_IN_MB * 120) as u64)).unwrap(); // Seek far ahead in file, to exceed prefetcher's progress
+    let err = f
+        .read_exact(&mut [0u8; 1])
+        .expect_err("fresh read using open file handle should fail as object no longer exists");
+    let raw_os_err = err.raw_os_error().expect("err should be OS-level err");
+    assert_eq!(raw_os_err, libc::EIO);
+}
```
sauraank, line 109: Is there a way we can test concurrent cases like when a file is being written locally, when we are trying to unlink.

mountpoint-s3/tests/fuse_tests/unlink_test.rs
```diff
@@ -0,0 +1,121 @@
...
+    test_client.put_object("dir/hello.txt", b"hello world").unwrap();
+    test_client.put_object("dir/foo.txt", b"bar").unwrap();
+
+    let main_dir = mount_point.path().join("dir");
+    let path = mount_point.path().join("dir/hello.txt");
+    let nonexistent_path = mount_point.path().join("dir/not-here.txt");
+
+    // files should be visible in readdir
+    let dir = fs::read_dir(&main_dir).unwrap();
+    let dirs: Vec<_> = dir.map(|f| f.unwrap()).collect();
```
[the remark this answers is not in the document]
dannycjones: I like that idea - it was annoying me too :)

Put the changes into #237 to keep this PR as simple as possible - I want to merge that one first though.

mountpoint-s3/src/inode.rs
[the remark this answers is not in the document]
dannycjones: Agree that would be the best solution if we don't change how we lock / serialize operations for unlink.

Based on James' feedback, I'm thinking that tracking 'delete status' is the way to move forward.

mountpoint-s3/src/inode.rs
dannycjones: I had similar thoughts, but I'm not sure what we should expect of `unlink` operations. 

- Should we try to block subsequent `unlink` until the first is complete? My thinking there is that we'd have to add something like `unlink_mutex: Mutex<()>` which can let us serialize `unlink` operations.
- Should we return an error... but what error?
- Should we do something like @passaro suggests which is to continue with the `unlink`, and only remove the child if the `Inode` or `InodeNo` is equal to the one we saw at the beginning of the `unlink` operation?

mountpoint-s3/src/inode.rs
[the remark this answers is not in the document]
dannycjones: I think we need to tackle the concurrency of `unlink` head on - this might be a solution, or maybe there's an alternative. Let's discuss in the other thread (https://github.com/awslabs/mountpoint-s3/pull/232/files#r1186043700).

mountpoint-s3/src/inode.rs
dannycjones: I'm leaning towards serializing the unlinks, since `unlink` would be atomic for POSIX FS (right?). Thus, we could then return `ENOENT` ("error, no entry found") for subsequent `unlink` requests.

Thoughts?

mountpoint-s3/src/inode.rs
dannycjones: It is performing string formatting / interpolation.

The `0` is referring to the index of the value in the `IsDirectory` enum variant - in this case, the `InodeNo` on the line below.

This attribute is used by macros in the crate `thiserror`. You can see a quick summary of what this turns into under the "A Display impl is generated for your error" bullet point. https://docs.rs/thiserror/latest/thiserror/#details

mountpoint-s3/tests/fuse_tests/unlink_test.rs
dannycjones: I'm not super sure on this one. I think generally we just want to make sure everything works when using and not using a prefix.

dannycjones:
Pushed latest code - it addresses some feedback, but not the locking/concurrency open questions.

mountpoint-s3/tests/fuse_tests/unlink_test.rs
sauraank: Oh! I was asking because I needed it for my rmdir implementation as well. Like to test the case where there is mknod/mkdir call concurrently with rmdir, and its executed after rmdir.

mountpoint-s3/src/inode.rs
```diff
@@ -360,6 +360,68 @@ impl Superblock {
 
         Ok(LookedUp { inode, stat })
     }
+
+    /// Unlink the entry described by `parent_ino` and `name`.
+    ///
+    /// If the entry exists, delete it from S3 and the superblock.
+    pub async fn unlink<OC: ObjectClient>(
```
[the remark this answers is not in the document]
sauraank: In my implementation of `rmdir`, I am adding a `Deleted` state in `WriteStatus` in order to mark the inode as deleted so that any other concurrent operation on the given inode shall fail. Maybe it can work here as well? (Still not sure though if thats right approach for rmdir as well)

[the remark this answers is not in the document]
dannycjones: I had a similar idea - I'm not sure if its the right approach right now, but we can always change it.

There's two different issues I suppose:
- Trying to perform a concurrent operations _during_ the `unlink`/`rmdir`
- Trying to perform a concurrent operation which is occuring after the `unlink`/`rmdir` is complete - i.e. the operation started and has the `Inode`, but now the other concurrent operations already completed.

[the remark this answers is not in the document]
dannycjones: After reviewing how the VFS is handling concurrent `unlink`, I've moved the locking to take write/exclusive locks on both the parent and the child in line with the VFS implementation.

At the moment, we gain nothing but complexity by trying to reduce the time we hold locks inside Mountpoint.

Discussed a bit more in the `rmdir` PR too: https://github.com/awslabs/mountpoint-s3/pull/247#discussion_r1199074008

mountpoint-s3/src/inode.rs
```diff
@@ -354,6 +354,76 @@ impl Superblock {
...
+        let LookedUp { inode, .. } = self.lookup(client, parent_ino, name).await?;
+
+        if inode.kind() == InodeKind::Directory {
+            return Err(InodeError::IsDirectory(inode.ino()));
+        }
+
+        // Take exclusive write locks early.
+        // We know that the VFS will already have locked its equivalent representations.
+        let mut parent_state = parent.inner.sync.write().unwrap();
+        // Take a write lock even if we don't mutate, since we would not want to permit any concurrent operations.
```
sauraank, line 381: Why are we taking write lock on child? Even if we take a read lock, it ensures there wont be any mutations. Others can only take read lock over it concurrently and not write lock.

dannycjones: I agree - the main reason I took the write lock was just to be consistent with the VFS.

I've since removed this now.

dannycjones:
I've returned the locks back as they were - i.e. not being held over the whole unlink call. Instead, we now panic if we see anything that breaks the VFS locking assumptions we make when removing children at the end.

The risk with holding the sync lock over `unlink` is that I believe one or more operations (`getattr` is one) will not hold any kind of lock in the VFS. Meanwhile, we are taking a read lock for `InodeState`. If there was an unlink and getattr concurrently, with `unlink` taking the Mountpoint lock first - **deadlock**.

Due to this, I'll move forward without holding the locks and panic if assumptions broken.

Ideally, we might move to using `AsyncRwLock<..>` however it adds more complexity to our code and possibly a (negligible?) performance hit. `unlink` is the only operation we wanted to take a lock over an `await` point, and today we are happy with the VFS locking to cover this case.

dannycjones:
I'd like to look into adding some reference testing for this change, but I'd do it in a follow-up PR. Not planning to do that too soon though.

mountpoint-s3/src/inode.rs
```diff
@@ -439,6 +439,87 @@ impl Superblock {
...
+            WriteStatus::LocalUnopened | WriteStatus::LocalOpen => {
+                // In the future, we may permit `unlink` and cancel any in-flight uploads.
+                error!(
+                    parent = parent_ino,
+                    ?name,
+                    "unlink called on local file, unlink not supported until write is complete",
+                );
+                return Err(InodeError::UnlinkNotPermittedWhileWriting(inode.ino()));
+            }
+            WriteStatus::Remote => {
```
sauraank, line 478: What should be the paraemters like WriteStatus for `InodeState` once the Inode is unlinked? Or we dont really care? Like I get that lookup will not show up these files while create. But if a new object is created by any other client remotely, and `update_from_remote()` is called, it will still see the InodeStat as a `InodeKind::File` and `WriteStatus::Remote`.

sauraank (commented):
Are unit tests to test to see if lookup, readdir etc are functioning as intended relevant here to be added?

mountpoint-s3/tests/fuse_tests/unlink_test.rs
```diff
@@ -0,0 +1,104 @@
...
+    simple_unlink_tests(crate::fuse_tests::s3_session::new, "simple_unlink_tests");
+}
+
+#[test_case(""; "no prefix")]
+#[test_case("simple_unlink_test"; "prefix")]
+fn simple_unlink_test_mock(prefix: &str) {
+    simple_unlink_tests(crate::fuse_tests::mock_session::new, prefix);
+}
+
+/// Testing behavior when a file is unlinked in the middle of reading
```
passaro, line 57: Should we also test attempts to unlink while writing?

dannycjones: Yes we should, I'll add this

mountpoint-s3/src/inode.rs
dannycjones: For the `WriteStatus`, I didn't end up updating this as the risk felt negligible.

If the Kernel still has the Inode number, it can still do things like `getattr` or `open`, however it feels OK for now to allow those to return. Maybe we should block.

What I can do is look into adding unit tests for those, as you suggested, to document how it works today and note the assumptions we're making / what we're OK with.

passaro (dismissed):
LGTM

dannycjones:
@passaro sorry, had to make a small fix for prefix case on one of the tests.

passaro (approved):

PR #228, opened 2023-05-01
Bump aws-c-s3 to v0.2.9

jamesbornholt:
This picks up the fix for #218.

Signed-off-by: James Bornholt <bornholt@amazon.com>

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

PR #233, opened 2023-05-03
Handle channel closure in PartQueue

passaro:
Improve PartQueue to report an error (rather than blocking indeterminately) when trying to read beyond the body returned by the GetObject request. 

The previous behavior, compounded with a bug in the CRT (now fixed: #228) where GetObject unexpectedly returned an empty body, manifested in #218.

---

### DIFF
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -24,7 +24,7 @@ use thiserror::Error;
 use tracing::{debug_span, error, trace, Instrument};
 
 use crate::prefetch::part::Part;
-use crate::prefetch::part_queue::PartQueue;
+use crate::prefetch::part_queue::{unbounded_part_queue, PartQueue};
 use crate::sync::{Arc, RwLock};
 
 type TaskError<Client> = ObjectClientError<GetObjectError, <Client as ObjectClient>::ClientError>;
@@ -244,13 +244,13 @@ where
 
         let size = end - start;
         let range = start..end;
-        let part_queue = Arc::new(PartQueue::new());
+
+        let (part_queue, part_queue_producer) = unbounded_part_queue();
 
         trace!(?range, size, "spawning request");
 
         let request_task = {
             let client = Arc::clone(&self.inner.client);
-            let part_queue = Arc::clone(&part_queue);
             let bucket = self.bucket.to_owned();
             let key = self.key.to_owned();
             let etag = self.etag.clone();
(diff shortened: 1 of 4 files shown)


mountpoint-s3/src/prefetch.rs
```diff
@@ -347,6 +347,9 @@ impl<E: std::error::Error + Send + Sync> RequestTask<E> {
 pub enum PrefetchReadError<E: std::error::Error> {
     #[error("get request failed")]
     GetRequestFailed(#[from] E),
+
+    #[error("get request terminate unexpectedly")]
```
jamesbornholt, line 351:
```suggestion
    #[error("get request terminated unexpectedly")]
```

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -1,5 +1,7 @@
 use std::time::Instant;
 
+use tracing::log::trace;
```
jamesbornholt, line 3:
```suggestion
use tracing::trace;
```

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -47,19 +55,22 @@ impl<E: std::error::Error + Send + Sync> PartQueue<E> {
...
-                part
+                part.map_err(|e| e.into())
             } else {
                 let start = Instant::now();
-                let part = self.receiver.recv().await.expect("channel should not close");
+                let part = self.receiver.recv().await;
                 metrics::histogram!("prefetch.part_queue_starved_us", start.elapsed().as_micros() as f64);
-                part
+                match part {
+                    Err(_) => Err(PrefetchReadError::GetRequestTerminatedUnexpectedly),
```
jamesbornholt, line 64: let's write the error explicitly on the LHS, in the very unlikely event `async_channel` ever gets more error cases

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -73,11 +84,15 @@ impl<E: std::error::Error + Send + Sync> PartQueue<E> {
...
+}
 
+impl<E: std::error::Error + Send + Sync> PartQueueProducer<E> {
     /// Push a new [Part] onto the back of the queue
     pub fn push(&self, part: Result<Part, E>) {
         // Unbounded channel will never actually block
-        self.sender.send_blocking(part).unwrap();
+        if self.sender.send_blocking(part).is_err() {
+            trace!("closed channel");
+        }
```
jamesbornholt, line 95: i think we don't expect this to ever happen, but it's not fatal, so maybe `warn!` instead. also style preference/nit, not putting side effects in conditions
```suggestion
        let result = self.sender.send_blocking(part);
        if result.is_err() {
            warn!("closed channel");
        }
```

PR #234, opened 2023-05-03
Implement checksum validation for prefetcher

monthonk:
We want to ensure data integrity while object data is within Mountpoint's custody. So, we will calculate CRC checksum for the data as soon as we receive it from S3 and validate them every time we move it around our part queue, and before returning them to FUSE.

The validation would definitely impact Mountpoint's performance. But we can improve it a bit if we assume that most of the sequential read would have the same read size. In that case, we can pre-split the parts in part queues to a preferred size and avoid splitting them during read.

---

### DIFF
--- a/mountpoint-s3-crt/src/checksums/crc.rs
+++ b/mountpoint-s3-crt/src/checksums/crc.rs
@@ -1,0 +1,34 @@
+use mountpoint_s3_crt_sys::*;
+
+/// Calculate CRC32 checksum of the data in the given buffer, append to the previous checksum if provided.
+pub fn checksums_crc32(buf: &[u8], previous_checksum: Option<u32>) -> u32 {
+    // SAFETY: we pass a valid buffer to the CRT, and trust
+    // the CRT function to only read from the buffer's boundary.
+    unsafe { aws_checksums_crc32(buf.as_ptr(), buf.len() as i32, previous_checksum.unwrap_or(0)) }
+}
+
+/// Calculate CRC32C checksum of the data in the given buffer, append to the previous checksum if provided.
+pub fn checksums_crc32c(buf: &[u8], previous_checksum: Option<u32>) -> u32 {
+    // SAFETY: we pass a valid buffer to the CRT, and trust
+    // the CRT function to only read from the buffer's boundary.
+    unsafe { aws_checksums_crc32c(buf.as_ptr(), buf.len() as i32, previous_checksum.unwrap_or(0)) }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn crc32_works() {
+        let buf: &[u8] = b"123456789";
+        let crc = checksums_crc32(buf, None);
+        assert_eq!(crc, 0xcbf43926);
+    }
+
+    #[test]
+    fn crc32c_works() {
+        let buf: &[u8] = b"123456789";
+        let crc = checksums_crc32c(buf, None);
+        assert_eq!(crc, 0xe3069283);
+    }
+}
--- a/mountpoint-s3/proptest-regressions/prefetch.txt
+++ b/mountpoint-s3/proptest-regressions/prefetch.txt
@@ -1,0 +1,9 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc 1cefc6aa7160d95c9497b4f4407115c749a2f97e74c5dd3a5e0f68f33edc5a41 # shrinks to size = 3, read_factor = 2, config = TestConfig { first_request_size: 16, max_request_size: 16, sequential_prefetch_multiplier: 1, client_part_size: 16 }
+cc 5766b08dab3255bc3f26382330b9cb738e51542bf25f0e86bf2c8cfaa8ff1f88 # shrinks to size = 241849, read_size = 1, config = TestConfig { first_request_size: 16, max_request_size: 16, sequential_prefetch_multiplier: 1, client_part_size: 16 }
+cc 3e1688fb734cb931736cf60002a8bd2ec417e67676d2b08fd7bc9ace755e6a38 # shrinks to size = 275240, read_size = 2, config = TestConfig { first_request_size: 16, max_request_size: 16, sequential_prefetch_multiplier: 1, client_part_size: 57639 }
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -20,6 +20,7 @@ use futures::stream::StreamExt;
 use futures::task::{Spawn, SpawnExt};
 use metrics::counter;
 use mountpoint_s3_client::{ETag, GetObjectError, ObjectClient, ObjectClientError};
+use mountpoint_s3_crt::checksums::crc::checksums_crc32c;
 use thiserror::Error;
 use tracing::{debug_span, error, trace, Instrument};
 
@@ -27,7 +28,12 @@ use crate::prefetch::part::Part;
 use crate::prefetch::part_queue::PartQueue;
 use crate::sync::{Arc, RwLock};
 
-type TaskError<Client> = ObjectClientError<GetObjectError, <Client as ObjectClient>::ClientError>;
+use self::part::PartMismatchError;
+
+/// CRC32C checksum
+pub type ChecksumCrc = u32;
+
+type TaskError<Client> = PrefetchTaskError<ObjectClientError<GetObjectError, <Client as ObjectClient>::ClientError>>;
 
 #[derive(Debug, Clone, Copy)]
 pub struct PrefetcherConfig {
@@ -100,6 +106,8 @@ pub struct PrefetchGetObject<Client: ObjectClient, Runtime> {
     future_tasks: Arc<RwLock<VecDeque<RequestTask<TaskError<Client>>>>>,
     bucket: String,
     key: String,
+    // preferred part size in the prefetcher's part queue, not the object part
+    preferred_part_size: usize,
     next_sequential_read_offset: u64,
     next_request_size: usize,
     next_request_offset: u64,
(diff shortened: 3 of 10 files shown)


monthonk:
We probably want to put the validation under a feature flag because of the performance reasons then add an option to enable it later, but I want to run the benchmark on it first. Please let me know if we want to disable it by default.

mountpoint-s3/src/prefetch.rs
```diff
@@ -20,14 +20,20 @@ use futures::stream::StreamExt;
...
 
 use crate::prefetch::part::Part;
 use crate::prefetch::part_queue::PartQueue;
 use crate::sync::{Arc, RwLock};
 
-type TaskError<Client> = ObjectClientError<GetObjectError, <Client as ObjectClient>::ClientError>;
+use self::part::PartMismatchError;
+
+/// CRC32C checksum
+pub type ChecksumCrc = u32;
```
jamesbornholt, line 34: I wonder if we should make a version of this typed interface part of the CRT bindings. It prevents us from confusing checksums with other values, or confusing checksums of different types. It would just be a `struct Crc32c(u32)` and the same for `Crc32`.

Actually, I wonder if we should implement a [`Hasher`](https://doc.rust-lang.org/std/hash/trait.Hasher.html) for the checksums in the CRT bindings too. That might be the more Rust idiomatic way to do it.

mountpoint-s3/src/prefetch.rs
```diff
@@ -250,29 +277,53 @@ where
...
+                                            Bytes::copy_from_slice(&body[body_offset..to]),
+                                        );
+                                        let part_size = part.len();
+                                        parts_checksum = checksums_crc32c(part.bytes_ref(), Some(parts_checksum));
+                                        trace!(preferred_part_size, offset, actual_size = part.len(), "create_part");
+                                        part_queue.push(Ok(part));
+
+                                        body_offset += part_size;
+                                        remaining -= part_size;
+                                    }
```
jamesbornholt, line 322: Is this loop easier to write by using `body.chunks(preferred_part_size)`?

mountpoint-s3/src/prefetch.rs
```diff
@@ -250,29 +277,53 @@ where
...
                     Err(e) => {
                         error!(error=?e, "RequestTask get object failed");
-                        part_queue.push(Err(e));
+                        part_queue.push(Err(e.into()));
                     }
                     Ok(request) => {
                         pin_mut!(request);
                         loop {
                             match request.next().await {
                                 Some(Ok((offset, body))) => {
```
jamesbornholt, line 301: I wonder if we should instead do `let body = Bytes::from(body);` here, so that the splitting into ranges stuff doesn't actually have to copy the underlying bytes.

mountpoint-s3/src/prefetch/part.rs
```diff
@@ -11,17 +15,28 @@ pub struct Part {
     key: String,
     offset: u64,
     bytes: Bytes,
+    checksum: ChecksumCrc,
 }
 
 impl Part {
     pub fn new(key: &str, offset: u64, bytes: Bytes) -> Self {
+        let checksum = checksums_crc32c(&bytes, None);
```
jamesbornholt, line 23: Maybe make the caller pass in a checksum rather than computing it here.

jamesbornholt: Actually I can't tell if you're intentionally computing it both here and in the caller, but that doesn't seem necessary.

mountpoint-s3/src/prefetch/part.rs
```diff
@@ -11,17 +15,28 @@ pub struct Part {
...
         }
     }
 
+    pub fn checksum(&self) -> ChecksumCrc {
+        self.checksum
+    }
+
+    pub fn bytes_ref(&self) -> &[u8] {
+        self.bytes.as_ref()
+    }
```
jamesbornholt, line 38: This method shouldn't exist, it defeats the purpose of this interface (which prevents you from retrieving the bytes unless you know the key/offset).

mountpoint-s3/src/prefetch.rs
```diff
@@ -139,6 +148,12 @@ where
             "read"
         );
 
+        // Currently, preferred part size is the current read size.
+        // This is a naive assumption that the read size will be the same for most sequential
+        // read and it can be aligned to the size of prefetched data. Even so, it's worth to
+        // try since the performance will be a lot better when the conditions are met.
+        self.preferred_part_size = length;
```
jamesbornholt, line 155: We probably want to put some bounds on this value (e.g. if someone does a 1 byte read, that's probably not the part size we want).

jamesbornholt (commented):
I'm having trouble following the read path here and exactly where and when checksums are validated. Part of the problem is that we do the splitting and re-checksumming in too many different places. I think we should try to break this PR down into smaller parts and put it behind a feature flag for now.

The first thing I'd do here is design how we want the read path to work. I'd probably even do this in an empty crate just to focus on it in isolation. What we want to get to is a design where the fast path `read` only ever needs to compute one checksum, right before it returns the bytes to the file system. I think this probably means rethinking how `Part`s work.

Maybe what we're missing is a version of the `Bytes` crate that is checksum-aware. Every one of these new `ChecksummedBytes` combines both some bytes and a checksum for those bytes. When you call `split_off`, it doesn't copy the bytes (just like `Bytes`), but it remembers that the split-off `ChecksummedBytes` comes from the same backing buffer. Then the only way to get a slice out of this new `ChecksummedBytes` thing is to first validate the entire backing buffer's checksum. Just spitballing here though, I haven't tried to implement it.

Once we get that straightened out, then (in later PRs) we can focus on making it more efficient with the preferred part size stuff. But for now, we should just checksum the entire part when it arrives, and design a way for that checksum to be carried around everywhere automatically (via `ChecksummedBytes`).

monthonk:
Thanks for reviewing the PR.

About where and when to validate the checksum, my initial idea is that we will validate them whenever we copy the bytes into a new buffer, with your suggestions I think this is how we should do the validation
1. When we receive data from get_object and put them into part queues. We will calculate the checksum for each part and validating that their combined checksum equals to the original checksum
2. When we read data from part queues which includes
    - fast path: the data to read is exactly one part, we validate the checksum for that part before returning it to the file system
    - the data to read is smaller than a part size, we split the part into two and validate the checksum for both parts + another validation for the part we return
    - the data to read is bigger than a part size, we validate the checksum for every part we read, then combine them into a single buffer and do the final validation with combined checksum

For what to do next
1. I will put the checksum validation under a feature flag
2. Implement `ChecksummedBytes` and refactor the validation logic with this new struct (in another PR)
3. Optimize the performance (in another PR)

mountpoint-s3/src/prefetch/part.rs
monthonk: I intentionally computing it so that the caller can validate its own checksum with the one in the part struct, but it's probably not necessary.

mountpoint-s3-crt/src/checksums/crc.rs
```diff
@@ -0,0 +1,34 @@
...
+
+/// Calculate CRC32C checksum of the data in the given buffer, append to the previous checksum if provided.
+pub fn checksums_crc32c(buf: &[u8], previous_checksum: Option<u32>) -> u32 {
+    // SAFETY: we pass a valid buffer to the CRT, and trust
+    // the CRT function to only read from the buffer's boundary.
+    unsafe { aws_checksums_crc32c(buf.as_ptr(), buf.len() as i32, previous_checksum.unwrap_or(0)) }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
monthonk: Sounds good, will add the tests

mountpoint-s3/src/prefetch.rs
monthonk: Sounds good, I will move the type to `mountpoint-s3-crt`.

I was thinking about implementing the hasher as well, just not sure if we want to deviate from CRT's interface. But yeah, I would prefer the hasher implementation.

mountpoint-s3-crt/src/checksums/crc.rs
```diff
@@ -0,0 +1,34 @@
+use mountpoint_s3_crt_sys::*;
+
+/// Calculate CRC32 checksum of the data in the given buffer, append to the previous checksum if provided.
+pub fn checksums_crc32(buf: &[u8], previous_checksum: Option<u32>) -> u32 {
```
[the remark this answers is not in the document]
eslrahc-swa: +1

mountpoint-s3-crt/src/checksums/crc.rs
```diff
@@ -0,0 +1,34 @@
...
+    // SAFETY: we pass a valid buffer to the CRT, and trust
```
eslrahc-swa, line 5: If CRT is actually conduct read-only to the given buffer, we can be more affirmative here.

Something like ""CRT reads from the buffer's boundary only"

mountpoint-s3-crt/src/checksums/crc.rs
```diff
@@ -0,0 +1,34 @@
...
+    // the CRT function to only read from the buffer's boundary.
+    unsafe { aws_checksums_crc32(buf.as_ptr(), buf.len() as i32, previous_checksum.unwrap_or(0)) }
+}
+
+/// Calculate CRC32C checksum of the data in the given buffer, append to the previous checksum if provided.
+pub fn checksums_crc32c(buf: &[u8], previous_checksum: Option<u32>) -> u32 {
+    // SAFETY: we pass a valid buffer to the CRT, and trust
```
eslrahc-swa, line 12: same here.

mountpoint-s3-crt/src/checksums/crc.rs
```diff
@@ -0,0 +1,34 @@
...
+    // the CRT function to only read from the buffer's boundary.
+    unsafe { aws_checksums_crc32c(buf.as_ptr(), buf.len() as i32, previous_checksum.unwrap_or(0)) }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
```
eslrahc-swa, line 21: As jamesbornholt@ mentioned here, probably add unit test cases for case `Some`

mountpoint-s3/proptest-regressions/prefetch.txt
```diff
@@ -0,0 +1,9 @@
+# Seeds for failure cases proptest has generated in the past. It is
```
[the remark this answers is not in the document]
eslrahc-swa: what is the purpose for this file? I am confused. maybe lack of some backgrounds here

PR #237, opened 2023-05-05
Extract `fn read_dir_to_string_vec` for FUSE tests

dannycjones:
Extracting some common logic repeated all over the FUSE tests. Feedback from #232.

---

### DIFF
--- a/mountpoint-s3/tests/fuse_tests/mod.rs
+++ b/mountpoint-s3/tests/fuse_tests/mod.rs
@@ -7,6 +7,9 @@ mod prefetch_test;
 mod readdir_test;
 mod write_test;
 
+use std::ffi::OsStr;
+use std::fs::ReadDir;
+
 use fuser::{BackgroundSession, MountOption, Session};
 use mountpoint_s3::fuse::S3FuseFilesystem;
 use mountpoint_s3::prefix::Prefix;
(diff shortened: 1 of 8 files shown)


dannycjones:
CI is failing ASAN because of something breaking with Nightly Rust. I'd propose we ignore - though we may need to investigate why our Nightly Rust ASAN tests are failing.

mountpoint-s3/tests/fuse_tests/mod.rs
```diff
@@ -222,3 +225,17 @@ mod s3_session {
         }
     }
 }
+
+/// Take a `read_dir` iterator and return the entry names
+pub fn read_dir_to_entry_names(read_dir_iter: ReadDir) -> Vec<String> {
+    read_dir_iter
+        .map(|entry_result| entry_result.expect("no io err during readdir"))
```
passaro, line 232: nit: is there a reason to have multiple `map`s here?

passaro (commented):
LGTM

mountpoint-s3/tests/fuse_tests/mod.rs
dannycjones: It doesn't matter either way - only to break into logical steps which we can do with local variables as well.

Happy to change if you would find that more readable.

dannycjones: I'll push it anyway as a single map

passaro (approved):

Issue #190, opened 2023-04-03
Pass readdir size config to readdir handle


monthonk:
Our readdir handle page size is currently [hard-coded to 1000](https://github.com/awslabs/mountpoint-s3/blob/286541e395d5dd774fd4de70644b5eaa20c5ec5f/mountpoint-s3/src/fs.rs#L422), this value should be passed down from our [file system config](https://github.com/awslabs/mountpoint-s3/blob/286541e395d5dd774fd4de70644b5eaa20c5ec5f/mountpoint-s3/src/fs.rs#L62). I thought this is an easy fix, but our proptest is broken as soon as I change the value. So, there might be something wrong with our readdir pagination logic and we need to investigate further into the problem.

jamesbornholt:
Started poking at this here: https://github.com/jamesbornholt/mountpoint-s3/tree/readdir-shadow-bug

Looks like an annoying edge case where a file gets shadowed by a directory from a later ListObjectsV2 request.

PR #238, opened 2023-05-07
Refactor ReaddirHandle to lazily create inodes 

jamesbornholt:
The motivation here is that `readdir` calls from FUSE generally only
retrieve a small number of entries, but ListObjectsV2 calls to S3
retrieve many entries. In the current code, we instantiate inodes for
those entries as soon as the ListObjectsV2 call completes, but because
of the small batch size of `readdir`, most of these inodes will not be
returned to FUSE until some time later. Because the TTL of an inode's
attributes is chosen at inode creation time, it will have expired by
this point, and so the kernel will immediately re-`lookup` these entries
when they're returned to `readdir`. That makes `readdir` do O(n)
`lookup` calls back to S3, defeating the purpose of ListObjectsV2.

To solve this problem, this change refactors the `ReaddirHandle`
implementation to defer creating inodes until an entry is actually
returned to `readdir`. The big idea is that we now just create a new
[ReaddirEntry] struct for each entry returned by ListObjectsV2, and all
the `readdir` logic works in terms of those things rather than directly
on inodes. The conversion from [ReaddirEntry] to an actual inode then
happens right before returning the entry to `readdir`.

I took this opportunity to refactor this code (hopefully for the
better!) since this change was adding yet another layer of complexity
into the logic for iterating directory entries. I also tried to capture
my understanding of this code in a new module comment. There is still a
bug here around page sizes (#190) that I think this comment and refactor
will make it easier to fix, but let's do that separately to avoid making
this change bigger.

---

### DIFF
--- a/mountpoint-s3/src/inode/readdir.rs
+++ b/mountpoint-s3/src/inode/readdir.rs
@@ -1,0 +1,233 @@
+use std::collections::VecDeque;
+use std::time::Instant;
+
+use mountpoint_s3_client::ObjectClient;
+use tracing::{error, trace, warn};
+
+use crate::sync::{Arc, Mutex, RwLock};
+
+use super::{
+    valid_inode_name, InodeError, InodeKind, InodeKindData, InodeNo, InodeStat, LookedUp, RemoteLookup, SuperblockInner,
+};
+
+/// Handle for an inflight directory listing
+#[derive(Debug)]
+pub struct ReaddirHandle {
+    inner: Arc<SuperblockInner>,
+    dir_ino: InodeNo,
+    parent_ino: InodeNo,
+    full_path: String,
+    page_size: usize,
+    remote_results: RwLock<VecDeque<LookedUp>>,
+    local_results: RwLock<VecDeque<LookedUp>>,
+    next_continuation_token: Mutex<ReaddirStreamState>,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+enum ReaddirStreamState {
+    NotStarted,
+    /// Continuation token for the next call
+    Continued(String),
+    Finished,
+}
+
+impl ReaddirStreamState {
+    fn take(&mut self) -> Option<String> {
+        let old_state = std::mem::replace(self, ReaddirStreamState::Finished);
+        if let ReaddirStreamState::Continued(s) = old_state {
+            Some(s)
+        } else {
+            None
+        }
+    }
+}
+
+impl ReaddirHandle {
+    pub(super) fn new(
+        inner: Arc<SuperblockInner>,
+        dir_ino: InodeNo,
+        parent_ino: InodeNo,
+        full_path: String,
+        page_size: usize,
+    ) -> Self {
+        Self {
+            inner,
+            dir_ino,
+            parent_ino,
+            full_path,
+            page_size,
+            remote_results: Default::default(),
+            local_results: Default::default(),
+            next_continuation_token: Mutex::new(ReaddirStreamState::NotStarted),
+        }
+    }
+
+    pub async fn next<OC: ObjectClient>(&self, client: &OC) -> Result<Option<LookedUp>, InodeError> {
+        // We will start fetching new results when number of items in the remote results queue is empty
+        while self.remote_results.read().unwrap().is_empty() {
+            let continuation_token = {
+                let mut next_token = self.next_continuation_token.lock().unwrap();
+
+                // populate local results before the first stream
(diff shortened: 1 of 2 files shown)


jamesbornholt:
The TTL duration hasn't changed. It's just that the inode itself is constructed later, and since the TTL is computed at inode construction time (when we create the `InodeStat`), it now happens right before returning the inode rather than as soon as it was received from ListObjectsV2. So it's `inode_construction_time + ttl_duration` rather than `list_objects_time + ttl_duration`.

eslrahc-swa:
My high level understanding here, correct me if I am wrong:

#### Current status
* FUSE call on `readdir` retrieve x entries whereas FUSE Damon implementation(Mountpoint-s3)`ListObjectsV2` retrieves, let's say 10x entries. 
* Inode creation happens after `ListObjectsV2` completed, where `readdir` can only deal with x entries as a batch at a time, so FUSE experiences some delay when it received rest 9x entries because of TTL expired. (The assumption here is we are not going to change TTL for some reason?)
* If TTL expired, linux kernel conducts `lookup` on the rest of 9x entries.

#### Problem
* The nuance of capability of handling the number of directory entries between `readdir` and `ListOjbectsV2`.
* More specifically, the size of`readdir` entry buffer cannot hold the number of directory at a time before TTL expired.

#### Solution:
* Another level of indirection
* Introduced `ReaddirEntry` so `readdir` don't have to deal with inode immediately, instead, convert `ReaddirEntry` to inode on demand basis, so it prevent re-lookup during `readdir`

jamesbornholt:
Nope, you're right! This is presupposing a future change that actually uses the expiry field correctly.

mountpoint-s3/src/inode/readdir.rs
```diff
@@ -0,0 +1,454 @@
...
+            return Ok(None);
+        }
+
+        let result = self
+            .inner
+            .update_from_remote(self.dir_ino, entry.name(), Some(remote_lookup));
+        match result {
+            Err(InodeError::ShadowedByDirectory(_, _)) => {
+                warn!(
+                    "{} is shadowed by another directory entry with the same name and will be unavailable",
```
sauraank, line 183: I am a bit confused here. In the Semantics doc its written Directories will be shadowed by files of the same name.

mountpoint-s3/src/inode/readdir.rs
```diff
@@ -0,0 +1,454 @@
...
+//!
+//! 1. It's possible for a directory to contain both a subdirectory and a file of the same name, in
+//!    which case we need to implement shadowing in the right way.
+//! 2. A directory can also contain local files/subdirectories, which we want to have appear in the
+//!    readdir stream, but be shadowed by remote files/subdirectories.
+//! 3. ListObjectsV2 returns common prefixes including the trailing '/', which messes up ordering
+//!    of entries in the readdir stream. For example, `a-/` < `a/`, but `a-` > `a` in lexicographic
+//!    order, so we need to re-sort the common prefixes rather than just streaming them directly.
+//! 4. We want to do large ListObjectsV2 calls (i.e. with a large page size), but `readdir` calls
+//!    typically are much smaller, so we need to hold onto ListObjectsV2 results for a while. But we
```
sauraank, line 14: What do we mean by holding on to the results of ListObjectsV2 calls? Is it having more results than a readdir calls can return? And are we handling them using multiple iterations or something? It was not clear for me from the description of ReadDirHandleIterator. Maybe I need to understand the code a bit better.

mountpoint-s3/src/inode/readdir.rs
```diff
@@ -0,0 +1,454 @@
+//! Utilities for implementing the `readdir` operation on a directory inode.
+//!
+//! `readdir` is conceptually simple ("just call ListObjectsV2") but has some subtleties that make
+//! it complicated:
+//!
+//! 1. It's possible for a directory to contain both a subdirectory and a file of the same name, in
+//!    which case we need to implement shadowing in the right way.
+//! 2. A directory can also contain local files/subdirectories, which we want to have appear in the
```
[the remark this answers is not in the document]
jamesbornholt: Shadowing means when there's a file and directory of the same name, we need to make a choice about which one to show, because file systems aren't allowed to do that. We choose to show directories over files if both have the same name.

mountpoint-s3/src/inode/readdir.rs
jamesbornholt: Huh, that documentation is just wrong, we actually do the opposite -- directories shadow files.

PR #242, opened 2023-05-11
Fix shadowing semantics doc and add tests

jamesbornholt:
The doc was backwards from what we actually do and what we intended.
Directories will shadow files of the same name, in the interest of
making as many objects visible as possible.

I also added a new integration test that just quotes the documentation
and implements it directly.

---

### DIFF
--- a/mountpoint-s3/tests/fuse_tests/semantics_doc_test.rs
+++ b/mountpoint-s3/tests/fuse_tests/semantics_doc_test.rs
@@ -1,0 +1,158 @@
+use std::path::Path;
+
+use fuser::BackgroundSession;
+use mountpoint_s3::S3FilesystemConfig;
+use tempfile::TempDir;
+use walkdir::WalkDir;
+
+use crate::fuse_tests::TestClientBox;
+
+/// Recursively list the contents of a directory and return the paths of all entries, with the
+/// initial `path` stripped. If `files` is true, the list contains only files; if false, it contains
+/// only directories.
+fn list_dir_recursive(path: impl AsRef<Path>, files: bool) -> Result<Vec<String>, walkdir::Error> {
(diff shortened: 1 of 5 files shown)


mountpoint-s3/tests/fuse_tests/semantics_doc_test.rs
```diff
@@ -0,0 +1,158 @@
...
+/// Recursively list the contents of a directory and return the paths of all entries, with the
+/// initial `path` stripped. If `files` is true, the list contains only files; if false, it contains
+/// only directories.
+fn list_dir_recursive(path: impl AsRef<Path>, files: bool) -> Result<Vec<String>, walkdir::Error> {
+    let path_prefix = format!("{}/", path.as_ref().to_string_lossy());
+    let contents_iter = WalkDir::new(path)
+        .sort_by_file_name()
+        .into_iter()
+        .collect::<Result<Vec<_>, _>>()?;
+    let contents_filtered = contents_iter.iter().filter(|d| {
```
sauraank, line 19: Is not this filtering the obvious? What are we trying to filter out here?

sauraank (approved):
Should we also add test for other points in semantics later? Like Object keys that contain null bytes.

mountpoint-s3/tests/fuse_tests/semantics_doc_test.rs
jamesbornholt: We want to only include files, or only include directories, because there's no way to tell which is which just from their name.

jamesbornholt:
> Should we also add test for other points in semantics later? Like Object keys that contain null bytes.

I thought about it, but I'd like to do it separately because I think it's a little more than a docs test like these ones are.

PR #245, opened 2023-05-15
Fix default config for logging

jamesbornholt:
We'd like to capture info-level logs by default -- they report useful information including metrics and failed requests. I was also wrong about the "brutal hack" for providing a default value for RUST_LOG -- `EnvFilter::new` has always existed to parse a string, so let's use that instead.

---


dannycjones (approved):
Happy with the change - but it isn't changing anything (other than dropping the FUSE worker info to debug, right?

PR #246, opened 2023-05-15
Add benchmarks for CRT checksum bindings

jamesbornholt:
Just a few simple benchmarks to make sure our bindings aren't affecting performance. This should also make it easy to play with performance on Graviton.

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -110,6 +110,35 @@ jobs:
     - name: Check all targets
       run: cargo check --locked --all-targets --all-features
 
+  bench:
+    name: Cargo benchmarks
+    runs-on: ubuntu-22.04
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+      with:
+        submodules: true
+    - name: Set up stable Rust
+      uses: actions-rs/toolchain@v1
+      with:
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -110,6 +110,35 @@ jobs:
     - name: Check all targets
       run: cargo check --locked --all-targets --all-features
 
+  bench:
+    name: Cargo benchmarks
+    runs-on: ubuntu-22.04
+
```
monthonk, line 116: I wonder what is the motivation to add this `cargo bench` to the CI. From what I see, there is no results being stored and each run could produce different results since it partially depends on the runner's performance (which is up to GitHub to decide).

monthonk (commented):
LGTM, only have a question about the CI.

.github/workflows/tests.yml
jamesbornholt: No strong motivation really. I don't think it does us any harm to run them, and it at least checks that they still run without panicing or something, but happy to remove them if you think they're going to confuse us in the future.

monthonk: We can keep it, I'm just curious.

monthonk (approved):

PR #247, opened 2023-05-15
Implementation of rmdir()

sauraank:
Implemented basic rmdir which delete only the empty directories locally. I still need to add the tests to it.
UPDATE: Added Fuse tests(also for remote directories), unit tests (also to test concurrent create). Also checked manually, working alright.
---


mountpoint-s3/src/inode.rs
```diff
@@ -354,6 +354,71 @@ impl Superblock {
...
+                children,
+                writing_children,
+            } => {
+                if children.contains_key(inode.full_key()) {
+                    error!(parent = parent_ino, ?name, "rmdir called on a remote directory",);
+                    return Err(InodeError::RemoteDirectory(inode.ino()));
+                }
+
+                writing_children.remove(&inode.ino());
+                parent_state.write_status = WriteStatus::Deleted;
```
sauraank, line 417: Just noticed I have marked Parent as Deleted. Correcting it.

passaro (changes requested):

mountpoint-s3/src/inode.rs
```diff
@@ -354,6 +354,70 @@ impl Superblock {
...
+        name: &OsStr,
+    ) -> Result<(), InodeError> {
+        let parent = self.inner.get(parent_ino)?;
+        let LookedUp { inode, .. } = self.lookup(client, parent_ino, name).await?;
+
+        if inode.kind() == InodeKind::File {
+            return Err(InodeError::NotADirectory(inode.ino()));
+        }
+
+        let mut inode_state = inode.inner.sync.write().unwrap();
```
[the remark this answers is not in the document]
sauraank: Why though? I can think of the reason that children in parent's state can be mutated while it is locked and can cause inconsistency. Are there other reasons as well?

mountpoint-s3/src/inode.rs
```diff
@@ -354,6 +354,70 @@ impl Superblock {
...
+                error!(parent = parent_ino, ?name, "rmdir called on a non existent directory",);
+                return Err(InodeError::InodeDoesNotExist(inode.ino()));
+            }
+            // Although Local Open is not possible for directories, added it to be on safe side.
+            WriteStatus::LocalOpen | WriteStatus::LocalUnopened => {
+                let inode_state = inode.inner.sync.read().unwrap();
+                let inode_kind_data = &inode_state.kind_data;
+                // Since it cannot be a remote directory, it only can have writing children which are not commited remotely.
+                // In such cases the direcotry is not empty locally.
+                if let InodeKindData::Directory { writing_children, .. } = inode_kind_data {
```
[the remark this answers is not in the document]
sauraank: `unreachable` for `File` case has already been added. So, I did not want to repeat it.

mountpoint-s3/src/inode.rs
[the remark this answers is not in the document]
passaro: That is the reason you need to acquire both the parent's and the child's locks. When that's the case, you need to be careful to always be consistent with the order in which multiple locks are acquired. For inodes we follow the hierarchical order from parent to child.
In general, the problem is that if different threads need to lock mutexes A and B, and they do it in different order, you may end up in a deadlock. For example,
Thread 1: locks A,
Thread 2: locks B,
Thread 1: wants to lock B, waits for Thread 1 to release
Thread 2: wants to lock A, waits for Thread 2 to release

Deadlock! Each thread is waiting on the other.

mountpoint-s3/src/inode.rs
```diff
@@ -354,6 +354,70 @@ impl Superblock {
...
+        match &mut parent_state.kind_data {
+            InodeKindData::File {} => {
+                debug_assert!(false, "inodes never change kind");
+                return Err(InodeError::NotADirectory(parent.ino()));
+            }
+            InodeKindData::Directory {
+                children,
+                writing_children,
+            } => {
+                if children.contains_key(inode.full_key()) {
```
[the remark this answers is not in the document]
passaro: Local directories (and files) are both in `children` and in `writing_children`.

mountpoint-s3/src/inode.rs
```diff
@@ -603,6 +667,10 @@ impl WriteHandle {
                 error!(inode=?self.ino, "inode already exists");
                 Err(InodeError::InodeNotWritable(self.ino))
             }
+            WriteStatus::Deleted => {
```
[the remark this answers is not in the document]
sauraank: Not doing anything on it right now. Will take action, once the mechanism to represent deleted directory/inode has been finalised.

sauraank:
Fuse tests are passing. Also tested manually with making Folders using console and other cases. Working fine. Now, I just need to add a few more tests. 
And list the cases where we actually need the Deleted Status of Inode so that we can finalise a way to represent it. (Currently doing it using `WriteStatus`).

mountpoint-s3/src/inode.rs
```diff
@@ -354,6 +354,65 @@ impl Superblock {
...
+            InodeKindData::File {} => {
+                debug_assert!(false, "inodes never change kind");
+                return Err(InodeError::NotADirectory(parent.ino()));
+            }
+            InodeKindData::Directory {
+                children,
+                writing_children,
+            } => {
+                writing_children.remove(&inode.ino());
+                children.remove(inode.name());
```
jamesbornholt, line 408: Should we assert something about these? If it's a local directory, it must have been in both these hashmaps, right?

mountpoint-s3/src/inode.rs
```diff
@@ -354,6 +354,70 @@ impl Superblock {
...
+            } => {
+                if children.contains_key(inode.full_key()) {
+                    error!(parent = parent_ino, ?name, "rmdir called on a remote directory",);
+                    return Err(InodeError::RemoteDirectory(inode.ino()));
+                }
+
+                writing_children.remove(&inode.ino());
+            }
+        }
+        // Changing the write status of the inode to deleted so that no further operations could be performed on it.
```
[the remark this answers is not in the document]
jamesbornholt: Is this a temporary thing that goes away once we correctly garbage collect inodes?

mountpoint-s3/src/inode.rs
```diff
@@ -603,6 +662,10 @@ impl WriteHandle {
                 error!(inode=?self.ino, "inode already exists");
                 Err(InodeError::InodeNotWritable(self.ino))
             }
+            WriteStatus::Deleted => {
+                error!(inode=?self.ino, "inode is deleted");
+                Err(InodeError::InodeDeleted(self.ino))
```
jamesbornholt, line 667: Same thing here about `InodeDeleted` being the same thing as `InodeDoesNotExist`?

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
```diff
@@ -0,0 +1,66 @@
...
+    assert_eq!(dir_entry_names, vec![empty_dirname, non_empty_dirname]);
+
+    // Write an object into the directory
+    let filename = "nested_file";
+    let filepath = non_empty_dirpath.join(filename);
+    fs::write(filepath, "").unwrap();
+
+    // remove the directories
+    fs::remove_dir(&empty_dirpath).expect("should be able to remove empty directory");
+    //should not remove non empty directory
```
jamesbornholt, line 41: nit: just delete this comment (the `expect_err` says the same thing)

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
```diff
@@ -0,0 +1,66 @@
...
+    let err = fs::remove_dir(&non_empty_dirpath).expect_err("removing non-empty directory should fail");
+    assert_eq!(err.kind(), ErrorKind::PermissionDenied);
```
jamesbornholt, line 43: I think we should check that `err.raw_os_error() == Some(libc::ENOTEMPTY)` instead of this

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
```diff
@@ -0,0 +1,66 @@
...
+use std::{
+    fs::{self, DirBuilder},
+    io::ErrorKind,
+};
+use tempfile::TempDir;
+use test_case::test_case;
+
+use crate::fuse_tests::{read_dir_to_entry_names, TestClientBox};
+
+fn rmdir_test<F>(creator_fn: F, prefix: &str)
```
jamesbornholt, line 12: Also need to test that an existing remote directory can't be removed

jamesbornholt (commented):
Looks pretty good! Just a few clarifications for me.

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
jamesbornholt: And maybe also test that we can re-create a directory that has been removed

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
sauraank: Sure. Just wanted to know the reason. Is it because `ErrorKind` is not consistent across different OS?

mountpoint-s3/src/inode.rs
```diff
@@ -354,6 +354,65 @@ impl Superblock {
...
+        let mut inode_state = inode.inner.sync.write().unwrap();
+
+        match inode_state.write_status {
+            WriteStatus::Remote => {
+                error!(parent = parent_ino, ?name, "rmdir called on a remote directory",);
+                return Err(InodeError::RemoteDirectory(inode.ino()));
+            }
+            WriteStatus::Deleted => {
+                error!(parent = parent_ino, ?name, "rmdir called on a non existent directory",);
+                return Err(InodeError::InodeDeleted(inode.ino()));
```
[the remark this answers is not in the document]
sauraank: Like technically, Inode still exist in the HashMap.

mountpoint-s3/src/inode.rs
```diff
@@ -354,6 +354,65 @@ impl Superblock {
...
+            }
+            // Although Local Open is not possible for directories, added it to be on safe side.
```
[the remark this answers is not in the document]
sauraank: Yes, I think I can add the `LocalOpen` arm as unreachable.

mountpoint-s3/src/inode.rs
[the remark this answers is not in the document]
jamesbornholt: I think from the user perspective they're the same though, right? The inode doesn't exist. The fact that it's still in the hashmap is an implementation detail (and arguably a bug).

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
jamesbornholt: `ErrorKind` is consistent across OSes, which is kind of the problem—it needs to map error codes from many different platforms into a single enum. We want to test more directly that we're returning the `ENOTEMPTY` we expect, and not `EPERM` or some other error that also gets mapped to `PermissionDenied`.

There's a proposed Rust feature to make these errors more precise, and it includes [`DirectoryNotEmpty`](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.DirectoryNotEmpty), but those haven't been stabilized yet so we can't use them.

sauraank: Oh! So, multiple os_error can be mapped to single ErrorKind. And we want to be correct about os error.

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
```diff
@@ -0,0 +1,83 @@
...
+    DirBuilder::new().recursive(true).create(&empty_dirpath).unwrap();
+    read_dir_iter = fs::read_dir(&main_path).unwrap();
+    dir_entry_names = read_dir_to_entry_names(read_dir_iter);
+    assert_eq!(dir_entry_names, vec![empty_dirname, non_empty_dirname]);
+
+    // explicitly testing remote directories not getting removed
+    test_client
+        .put_object("test_dir/remote_dir/hello.txt", b"hello world")
+        .unwrap();
+    let remote_dirname = "remote_dir";
```
jamesbornholt, line 61: maybe something like
```suggestion
    let remote_dirname = "remote_dir";
    test_client
        .put_object(&format!("{main_dirname}/{remote_dirname}/hello.txt"), b"hello world")
        .unwrap();
```

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
```diff
@@ -0,0 +1,83 @@
...
+    let remote_path = main_path.join(remote_dirname);
+    let err = fs::remove_dir(remote_path).expect_err("removing remote direcotry should fail");
```
jamesbornholt, line 63:
```suggestion
    let err = fs::remove_dir(remote_path).expect_err("removing remote directory should fail");
```

jamesbornholt (commented):
awesome, just a bunch of easy tiny things

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
```diff
@@ -0,0 +1,83 @@
...
+    // Write an object into the directory
+    let filename = "nested_file";
+    let filepath = non_empty_dirpath.join(filename);
+    fs::write(filepath, "").unwrap();
+
+    // remove the directories
+    fs::remove_dir(&empty_dirpath).expect("should be able to remove empty directory");
+
+    let err = fs::remove_dir(&non_empty_dirpath).expect_err("removing non-empty directory should fail");
+    assert_eq!(err.raw_os_error(), Some(libc::EPERM));
```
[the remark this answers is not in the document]
sauraank: Let me check. It was giving out EPERM.

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
```diff
@@ -0,0 +1,83 @@
...
+    // testing to re-create the removed directory
+    DirBuilder::new().recursive(true).create(&empty_dirpath).unwrap();
+    read_dir_iter = fs::read_dir(&main_path).unwrap();
+    dir_entry_names = read_dir_to_entry_names(read_dir_iter);
+    assert_eq!(dir_entry_names, vec![empty_dirname, non_empty_dirname]);
+
+    // explicitly testing remote directories not getting removed
+    test_client
+        .put_object("test_dir/remote_dir/hello.txt", b"hello world")
+        .unwrap();
```
dannycjones: Can we do this in a separate test `test_rmdir_remote_dir`? And this existing test becomes `test_rmdir_local_dir`?

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    // verify that the directories exist before testing `rmdir`
+    let read_dir_iter = fs::read_dir(&main_path).unwrap();
+    let dir_entry_names = read_dir_to_entry_names(read_dir_iter);
+    assert_eq!(dir_entry_names, vec![empty_dirname, non_empty_dirname]);
+
+    // Write an object into the directory
+    let filename = "nested_file";
+    let filepath = non_empty_dirpath.join(filename);
+    fs::write(filepath, "").unwrap();
```
dannycjones, line 34: Should we test that you still can't remove this directory after closing the file too?

dannycjones (changes requested):

mountpoint-s3/src/inode.rs
```diff
@@ -354,6 +358,79 @@ impl Superblock {
...
+                match inode_kind_data {
+                    InodeKindData::File {} => unreachable!("Already checked that inode is a directory"),
+                    InodeKindData::Directory {
+                        writing_children,
+                        deleted,
+                        ..
+                    } => {
+                        if *deleted {
+                            return Err(InodeError::InodeDoesNotExist(inode.ino()));
+                        }
```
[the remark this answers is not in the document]
sauraank: @jamesbornholt was also discussing on it. When we mark an Inode as deleted, it essentially meant to be removed and still having it in the HashMap maybe considered as a bug. I am not very sure, which option to take.

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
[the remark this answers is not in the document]
passaro: `EPERM` is expected in this case. That's because the directory at `non_empty_dirpath` is remote at this point and this test is actually verifying that it cannot be removed.

The `write` at line 34 is not just creating a nested file, but also writing to it and closing it, triggering a put on the client and thus converting the directory from local to remote.

mountpoint-s3/src/inode.rs
```diff
@@ -1189,6 +1273,130 @@ mod tests {
...
+            .rmdir(&client, FUSE_ROOT_INODE, dirname.as_ref())
+            .await
+            .expect("rmdir on empty local directory should succeed");
+
+        superblock
+            .lookup(&client, FUSE_ROOT_INODE, dirname.as_ref())
+            .await
+            .expect_err("should not do lookup on removed directory");
+
+        // currently readdir and getattr is working, because inode no still exist in superblock hashmap
```
passaro, line 1344: Do we think this is the correct behavior? I would have expected both to fail.

mountpoint-s3/src/inode.rs
```diff
@@ -1189,6 +1273,130 @@ mod tests {
...
+
+        superblock
+            .rmdir(&client, FUSE_ROOT_INODE, dirname.as_ref())
+            .await
+            .expect("rmdir on empty local directory should succeed");
+
+        superblock
+            .lookup(&client, FUSE_ROOT_INODE, dirname.as_ref())
+            .await
+            .expect_err("should not do lookup on removed directory");
```
passaro, line 1342: Shall we also test that `readdir` on the parent does not contain `dirname` anymore? (I think that's what was mentioned in other comments)

mountpoint-s3/src/inode.rs
sauraank: No, none of them are taking parent as input and testing children of the parent. They are only looking at the hashMap where the inode do still exist.

mountpoint-s3/src/inode.rs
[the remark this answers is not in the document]
dannycjones: For me, it was weird - we are saying Inode doesn't exist and passing in the Inode that doesn't exist.

The distinct error would mean that the error cause is clear from reading (we found a directory now marked as deleted), and in any error cause output it would also be clear.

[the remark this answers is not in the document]
dannycjones: Agree that keeping it around in the `HashMap` feels like a bug anyway. We aren't checking everywhere if the `Inode` was deleted.

Maybe we should be asserting/checking that the `Inode` does not have this flag set as part of `superblock.get(ino)`, but that involves another lock... :(

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
```diff
@@ -53,7 +58,29 @@ where
...
+
+fn rmdir_remote_dir_test<F>(creator_fn: F, prefix: &str)
+where
+    F: FnOnce(&str, S3FilesystemConfig) -> (TempDir, BackgroundSession, TestClientBox),
+{
+    let (mount_point, _session, mut test_client) = creator_fn(prefix, Default::default());
+
+    let main_dirname = "test_dir";
+    let main_path = mount_point.path().join(main_dirname);
     // explicitly testing remote directories not getting removed
```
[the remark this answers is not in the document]
dannycjones: We just need to add a zero-byte object at key `<path>/`, no?

mountpoint-s3/src/inode.rs
```diff
@@ -1189,6 +1273,132 @@ mod tests {
...
+
+        superblock
+            .rmdir(&client, FUSE_ROOT_INODE, dirname.as_ref())
+            .await
+            .expect("rmdir on empty local directory should succeed");
+
+        let parent = superblock.inner.get(FUSE_ROOT_INODE).unwrap();
+        let parent_state = parent.inner.sync.read().unwrap();
+        let parent_kind = &parent_state.kind_data;
+        match parent_kind {
```
dannycjones, line 1383: please move inline, these test methods end up getting quite long

```suggestion
        let parent_state = parent.inner.sync.read().unwrap();
        match &parent_state.kind_data {
```

mountpoint-s3/src/inode.rs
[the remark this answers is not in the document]
passaro: I see. But aren't all these implementation details? A user of `Superblock` will call `rmdir` (or `lookup`) and get `InodeError::InodeDoesNotExist`. What would it mean to them if they got `InodeError::InodeMarkedAsDeleted` sometimes? Would they get any access to such a node?

mountpoint-s3/src/inode.rs
```diff
@@ -1189,6 +1273,132 @@ mod tests {
...
+                assert!(writing_children.get(&inode.ino()).is_none());
+                assert!(children.get(inode.name()).is_none());
+            }
+        }
+
+        let inode_state = inode.inner.sync.read().unwrap();
+        let inode_kind = &inode_state.kind_data;
+        if let InodeKindData::Directory { deleted, .. } = inode_kind {
+            assert!(deleted);
+        }
```
dannycjones, line 1399:
```suggestion
        let inode_state = inode.inner.sync.read().unwrap();
        if let InodeKindData::Directory { deleted, .. } = &inode_state.kind_data {
            assert!(deleted);
        }
```

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
dannycjones: Ah I see, its this one. https://doc.rust-lang.org/std/fs/fn.write.html

I was thinking it was the lower-level write on a file object. Ignore this comment then :)

mountpoint-s3/src/inode.rs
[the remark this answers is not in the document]
jamesbornholt: Yeah my point was that the distinction between `deleted` and just removing it from the hashmap is an internal implementation detail, not worth surfacing as a separate error since the caller both doesn't know about the detail and can't do anything with it.

mountpoint-s3/src/inode.rs
jamesbornholt: This is just a straight up bug, right? It's why the `deleted` thing is so sketchy — the intent is to hide it from future callers, so `get` should be checking the flag.

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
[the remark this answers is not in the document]
jamesbornholt: Oh I missed the write. We should test the two cases separately then—open a file in the local directory but don't finish writing to it (`rmdir` should fail with ENOTEMPTY), versus finish writing to it (`rmdir` should fail with EPERM).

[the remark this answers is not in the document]
sauraank: What if we remove the inode from HashMap itself? Is there any issue with that? Currently I will check this deleted flag in `superblockInner.get()` as well.

mountpoint-s3/src/inode.rs
sauraank: What if we remove the inode from HashMap itself? Is there any issue with that? Currently I will check this deleted flag in `superblockInner.get()` as well.

sauraank:
I am thinking to add a unit test to add an inode as writing_children and children to parent inode ( not calling create as it may make child remote) to check if rmdir fails for local non empty directory.

mountpoint-s3/src/inode.rs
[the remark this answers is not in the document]
dannycjones: I'm OK with it being left as is then.

mountpoint-s3/src/inode.rs
```diff
@@ -354,17 +358,100 @@ impl Superblock {
...
+                        if *deleted {
+                            Err(InodeError::InodeDoesNotExist(ino))
+                        } else {
+                            Ok(inode.clone())
+                        }
+                    }
+                }
+            }
+            None => Err(InodeError::InodeDoesNotExist(ino)),
+        }
```
jamesbornholt, line 454: a little less indent-y with `?` and match-if:
```suggestion
        let inode = self
            .inodes
            .read()
            .unwrap()
            .get(&ino)
            .cloned()
            .ok_or(InodeError::InodeDoesNotExist(ino))?;
        let inode_state = inode.inner.sync.read().unwrap();
        match &inode_state.kind_data {
            // TODO: Include removal check for files
            InodeKindData::Directory { deleted, .. } if *deleted => Err(InodeError::InodeDoesNotExist(ino)),
            _ => Ok(inode.clone()),
        }
```

mountpoint-s3/src/inode.rs
```diff
@@ -1189,6 +1286,174 @@ mod tests {
...
+        // Marking deleted status as true to denote concurrent rmdir
+        {
+            let mut inode_state = inode.inner.sync.write().unwrap();
+
+            match &mut inode_state.kind_data {
+                InodeKindData::File {} => unreachable!("Created a directory"),
+                InodeKindData::Directory { deleted, .. } => *deleted = true,
+            }
+        }
+        // ending the lifetime of rwlock
```
jamesbornholt, line 1318: nit: don't think this comment is needed

jamesbornholt (approved):
Two optional things but LGTM!

mountpoint-s3/tests/fuse_tests/rmdir_test.rs
```diff
@@ -0,0 +1,185 @@
...
+    assert_eq!(err.raw_os_error(), Some(libc::ENOTEMPTY));
+
+    file.write_all(b"Hello World").unwrap();
+    drop(file);
+    let err =
+        fs::remove_dir(&non_empty_dirpath).expect_err("removing non-empty directory should fail even after closing");
+    assert_eq!(
+        err.raw_os_error(),
+        Some(libc::EPERM),
+        "It is wrting to remote after closing"
```
passaro, line 50: Typo. Also not really clear: see comment above.

mountpoint-s3/src/inode.rs
sauraank: Done

mountpoint-s3/src/inode.rs
sauraank: Done

passaro (changes requested):

mountpoint-s3/src/inode.rs
```diff
@@ -354,17 +358,100 @@ impl Superblock {
...
+            }
+            InodeKindData::Directory {
+                children,
+                writing_children,
+                ..
+            } => {
+                assert!(writing_children.remove(&inode.ino()));
+                children.remove(inode.name());
+
+                match &mut inode_state.kind_data {
```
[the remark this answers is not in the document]
sauraank: Yes. I thought it to add in the `LocalUnopened` WriteStatus. But, I felt `deleted` status should be updated after the inode has been removed from `children` and `writing_children`.

mountpoint-s3/src/inode.rs
```diff
@@ -354,17 +358,100 @@ impl Superblock {
...
+        let parent = self.inner.get(parent_ino)?;
+        let mut parent_state = parent.inner.sync.write().unwrap();
+
+        let mut inode_state = inode.inner.sync.write().unwrap();
+
+        match inode_state.write_status {
+            WriteStatus::LocalOpen => unreachable!("A directory cannot be in Local open state"),
+
+            WriteStatus::Remote => {
+                error!(parent = parent_ino, ?name, "rmdir called on a remote directory",);
```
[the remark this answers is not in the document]
sauraank: Yes. Its redundant. Removing it.

mountpoint-s3/src/inode.rs
```diff
@@ -1189,6 +1286,174 @@ mod tests {
...
+            .await
+            .unwrap();
+
+        // Marking deleted status as true to denote concurrent rmdir
+        {
+            let mut inode_state = inode.inner.sync.write().unwrap();
+
+            match &mut inode_state.kind_data {
+                InodeKindData::File {} => unreachable!("Created a directory"),
+                InodeKindData::Directory { deleted, .. } => *deleted = true,
```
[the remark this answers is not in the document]
sauraank: i am testing the case where when we are trying to create child directory within a directory which is concurrently getting removed. During `create`, it is able to `get()` the parent inode. But before applying the lock on the parent, the parent inode is deleted. In that case, we will have` deleted` status stored in `parent_state`. And that can be checked before creating child and fail the create for child.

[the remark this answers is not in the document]
passaro: I am still unconvinced this is a good test. 

We are artificially setting up an inconsistent state that just happens to match exactly what we know `create` is going to test  after acquiring the lock on this inode.

mountpoint-s3/src/inode.rs
```diff
@@ -1189,6 +1286,174 @@ mod tests {
         assert!(!client.contains_prefix(&prefix));
     }
 
+    // testing the case of concurrent deletion of parent before acquiring the lock on parent in create
```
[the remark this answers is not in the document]
sauraank: Yes. I think I should remove it from the children of its parent as well.

mountpoint-s3/src/inode.rs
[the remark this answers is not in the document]
sauraank: Yes. I think I should remove it from the children of its parent as well.

mountpoint-s3/src/inode.rs
```diff
@@ -354,17 +358,93 @@ impl Superblock {
...
             .get(&ino)
             .cloned()
-            .ok_or(InodeError::InodeDoesNotExist(ino))
+            .ok_or(InodeError::InodeDoesNotExist(ino))?;
+        let inode_state = inode.inner.sync.read().unwrap();
+        match &inode_state.kind_data {
+            // TODO: Include removal check for files
+            InodeKindData::Directory { deleted, .. } if *deleted => Err(InodeError::InodeDoesNotExist(ino)),
+            _ => Ok(inode.clone()),
+        }
```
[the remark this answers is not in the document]
sauraank: deleted status is added for that case itself. So, the operation on the directory must fail immediately after rmdir have succeeded.

dannycjones (commented):
Nice, I left a few comments. I haven't reviewed tests.

mountpoint-s3/src/inode.rs
```diff
@@ -354,6 +356,66 @@ impl Superblock {
...
+                debug_assert!(false, "inodes never change kind");
+                return Err(InodeError::NotADirectory(parent.ino()));
+            }
+            InodeKindData::Directory {
+                children,
+                writing_children,
+                ..
+            } => {
+                debug_assert!(
+                    writing_children.remove(&inode.ino()),
```
passaro, line 410: nit: I would move the action out of the assert to make it more visible. And just store the result in a bool, e.g.: `let removed = ...`

mountpoint-s3/src/inode.rs
```diff
@@ -1427,7 +1429,11 @@ mod tests {
                     .unwrap();
 
                 assert_eq!(
-                    dir_lookedup.inode.get_inode_state().unwrap().write_status,
+                    dir_lookedup
+                        .inode
+                        .get_inode_state()
+                        .expect("shouldd get inode state with read lock")
+                        .write_status,
```
dannycjones, line 1436: typo:

```suggestion
                    dir_lookedup
                        .inode
                        .get_inode_state()
                        .expect("should get inode state with read lock")
                        .write_status,
```

mountpoint-s3/src/inode.rs
```diff
@@ -634,10 +707,10 @@ impl WriteHandle {
         let mut ancestors_states: Vec<_> = ancestors
             .iter()
             .rev()
-            .map(|inode| inode.inner.sync.write().unwrap())
+            .map(|inode| inode.get_mut_inode_state().unwrap())
```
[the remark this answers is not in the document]
sauraank: I think it is fine to throw using expect() due to above logic. There is no case of getting InodeDoesNotExist. So, I should not return error.

mountpoint-s3/src/inode.rs
```diff
@@ -619,10 +683,9 @@ impl WriteHandle {
             loop {
                 assert!(visited.insert(ancestor_ino), "cycle detected in inode ancestors");
                 let ancestor = self.inner.get(ancestor_ino)?;
+                let ancestor_state = ancestor.get_inode_state()?;
```
passaro, line 687: can you move this back to inline?

passaro (commented):
Just a couple of comments

mountpoint-s3/src/inode.rs
```diff
@@ -271,8 +273,9 @@ impl Superblock {
     ) -> Result<LookedUp, InodeError> {
         let inode = self.inner.get(ino)?;
 
+        // TODO revalidate if expired
```
passaro, line 276: Remove (leftover from a merge?)

mountpoint-s3/src/inode.rs
```diff
@@ -663,13 +726,13 @@ impl WriteHandle {
         };
 
         // Acquire locks on ancestors in descending order to avoid deadlocks.
-        let mut ancestors_states: Vec<_> = ancestors
+        let ancestors_states: Result<Vec<_>, InodeError> = ancestors
             .iter()
             .rev()
-            .map(|inode| inode.inner.sync.write().unwrap())
+            .map(|inode| inode.get_mut_inode_state())
             .collect();
```
passaro, line 733: Almost there. But we want to return an error here, rather than after acquiring the inode lock.

Maybe this is a case for the turbofish?

```
let mut ancestors_states = ancestors
    .iter()
    .rev()
    .map(|inode| inode.get_mut_inode_state())
    .collect::<Result<Vec<_>, _>>()?;
```

mountpoint-s3/src/inode/readdir.rs
```diff
@@ -67,15 +67,13 @@ impl ReaddirHandle {
     ) -> Result<Self, InodeError> {
         let local_entries = {
             let inode = inner.get(dir_ino)?;
-            let kind_data = &inode.inner.sync.read().unwrap().kind_data;
+            let inode_state = &inode.get_inode_state()?;
```
passaro, line 70: inline. Note that this is the state of the parent. The ones on line 76 are for the children!

passaro (changes requested):

mountpoint-s3/src/inode.rs
sauraank: Thanks. It was good to learn about turbofish. Doing the change.

passaro (approved):
LGTM

PR #248, opened 2023-05-16
Use more specific labels for self-hosted runners

monthonk:
Just a small change to add specific labels when requesting self-hosted runners for workflows, we want to support more than one type of self-hosted runners in the future.

---

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -23,7 +23,7 @@ env:
 jobs:
   bench:
     name: Benchmark (Throughput)
-    runs-on: self-hosted
+    runs-on: [self-hosted, linux, x64, high-performance]
 
     environment: ${{ inputs.environment }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/bench.yml
```diff
@@ -23,7 +23,7 @@ env:
 jobs:
   bench:
     name: Benchmark (Throughput)
-    runs-on: self-hosted
+    runs-on: [self-hosted, linux, x64, high-performance]
```
dannycjones, line 26: I think I need a bit more understanding on how these would be defined to review the PR.

monthonk: Basically, I think we should follow the convention in the GitHub docs here https://github.com/philips-labs/terraform-aws-github-runner, so it should be `[self-hosted, operating-system, hardware-architecture, other-custom-labels]`.

dannycjones: Sounds good. Plus we will be labeling our existing runners as `high-performance` today.

dannycjones (approved):

PR #250, opened 2023-05-17
Add a CI job to validate Graviton instance support

monthonk:
We want to make sure that Mountpoint can be built and run on aarch64 and Graviton instances, but GitHub currently doesn't support this type of runner. What we can do for now is using self-hosted runners, which we maintain internally, until GitHub provides official ARM-based images.

The current configuration for our runners is [m7g.medium](https://aws.amazon.com/ec2/instance-types/m7g/) instance type with Amazon Linux 2 operating system.

---

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -83,6 +83,62 @@ jobs:
           target/
         key: ${{ steps.restore-cargo-cache.outputs.cache-primary-key }}
 
+  arm64-test:
+    name: Tests (arm64)
+    runs-on: [self-hosted, linux, arm64]
+
+    environment: ${{ inputs.environment }}
+
+    steps:
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v2
+      with:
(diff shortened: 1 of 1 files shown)


jamesbornholt (approved):
LGTM once the runner infrastructure is deployed.

.github/workflows/integration.yml
```diff
@@ -83,6 +83,62 @@ jobs:
           target/
         key: ${{ steps.restore-cargo-cache.outputs.cache-primary-key }}
 
+  arm64-test:
+    name: Tests (arm64)
+    runs-on: [self-hosted, linux, arm64]
```
dannycjones, line 88: If we also have `x86_64` self-hosted runners, can we just add ARM and x86 to the strategy matrix and reuse the workflow?

monthonk: I'm not entirely sure, but from examples in [github docs](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-using-a-multi-dimension-matrix) I believe we can.

PR #253, opened 2023-05-22
Improve file handles locking

monthonk:
Currently, we're holding the lock on the file handles hash table for the whole read and write operations. It blocks other operations from interacting with the hash table, especially at high concurrency. Instead, we can hold the read lock for only a short period and use the file handles via atomic reference.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -106,7 +106,7 @@ pub struct S3Filesystem<Client: ObjectClient, Runtime> {
     prefix: Prefix,
     next_handle: AtomicU64,
     dir_handles: AsyncRwLock<HashMap<u64, Arc<DirHandle>>>,
-    file_handles: AsyncRwLock<HashMap<u64, FileHandle<Client, Runtime>>>,
+    file_handles: AsyncRwLock<HashMap<u64, Arc<FileHandle<Client, Runtime>>>>,
 }
 
 impl<Client, Runtime> S3Filesystem<Client, Runtime>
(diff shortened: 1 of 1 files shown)


dannycjones (approved):
LGTM.

Only note is that it diverges a little to how `Inode` works (with an inner `Arc<>`). I don't see the need for that right now.

dannycjones (commented):
+1 to James' feedback, on logging if we fail to release.

jamesbornholt (approved):

dannycjones (approved):

mountpoint-s3/src/fs.rs
```diff
@@ -329,9 +329,12 @@ where
             size
         );
 
-        let file_handles = self.file_handles.read().await;
-        let Some(handle) = file_handles.get(&fh) else {
-            return reply.error(libc::EBADF);
+        let handle = {
```
sauraank, line 332: How is this different from previous implementation?

dannycjones: With the previous implementation, `file_handles` was in the scope of the whole `read` function. Only at the end of the function would it be dropped, and the read lock released.

With this new code, we introduce a new scope (due to the use of `{ ... }`). We acquire the read lock, grab a reference to the file handle we want and return it from the block denoted by `{ ... }`. In doing so, we drop `file_handles` inside the block, and only the cloned file handle itself is kept in scope.

monthonk: We're moving the file handles locking (`self.file_handles.read()`) into a new block, which mean as soon as this block is out of scope the lock will also be release.

PR #258, opened 2023-05-24
Implement checksummed bytes

monthonk:
This change introduces ChecksummedBytes struct which is a buffer containing both the bytes and its checksum. There will be follow-up changes that make use of this struct, as we planned to implement some integrity check along the read and write path, the ChecksummedBytes will make those changes simpler and more efficient.

---

### DIFF
--- a/mountpoint-s3/Cargo.toml
+++ b/mountpoint-s3/Cargo.toml
@@ -56,6 +56,7 @@ fuse_tests = []
 s3_tests = []
 shuttle = []
 put = []
+# A temporary feature flag to enable read-path checksums while we are working to improve its performance
 checksum = []
 
 [build-dependencies]
--- a/mountpoint-s3/Cargo.toml
+++ b/mountpoint-s3/Cargo.toml
@@ -56,6 +56,7 @@ fuse_tests = []
 s3_tests = []
 shuttle = []
 put = []
+checksum = []
 
 [build-dependencies]
 built = { version = "0.6.0", features = ["git2"] }
--- a/mountpoint-s3/src/prefetch/checksummed_bytes.rs
+++ b/mountpoint-s3/src/prefetch/checksummed_bytes.rs
@@ -1,0 +1,250 @@
+use bytes::{Bytes, BytesMut};
+use mountpoint_s3_crt::checksums::crc32c::{self, Crc32c};
+use thiserror::Error;
+
+/// A `ChecksummedBytes` is a bytes buffer that carries its checksum.
+/// The implementation guarantees that its integrity will be validated when data transformation occurs.
+#[derive(Debug)]
+pub struct ChecksummedBytes {
+    orig_bytes: Bytes,
+    curr_slice: Bytes,
+    checksum: Crc32c,
+}
+
+impl ChecksummedBytes {
+    pub fn new(bytes: Bytes, checksum: Crc32c) -> Self {
+        let curr_slice = bytes.clone();
+        Self {
+            orig_bytes: bytes,
+            curr_slice,
+            checksum,
+        }
+    }
+
+    /// Convert the `ChecksummedBytes` into `Bytes`, data integrity will be validated before converting.
+    ///
+    /// Return `IntegrityError` on data corruption.
+    pub fn into_bytes(self) -> Result<Bytes, IntegrityError> {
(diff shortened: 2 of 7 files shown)


mountpoint-s3/src/prefetch/checksummed_bytes.rs
```diff
@@ -0,0 +1,250 @@
...
+
+            // 2. compare the checksum between the two transformations.
+            if new_checksum != new_checksum_dup {
+                return Err(IntegrityError::ChecksumMismatch(new_checksum, new_checksum_dup));
+            }
+
+            // 3. validate original buffers to make sure that the data we have copied are still valid.
+            self.validate()?;
+            extend.validate()?;
+        }
```
jamesbornholt, line 92: The "do it twice" thing is required for more general transformations, but copying with CRC is a special case that shouldn't require it, because the checksums of the inputs and outputs are related. This code should be able to just do this:
1. Create C = A ++ B by copying from A and B
2. Compute CRC32(C) from C
3. Compute CRC32(A ++ B) = CRC32(A) # CRC32(B), where # is the combination function for CRC. We can do this cheaply since A and B are both ChecksummedBytes.
4. Fail if CRC32(C) != CRC32(A ++ B).

If A was corrupted into A' before copying into C, we'll find out when we compute CRC32(C) = CRC32(A' ++ B) and find that it doesn't match CRC32(A ++ B) = CRC32(A)  # CRC32(B). Same for B/B'. If C was corrupted after copying, it's the same thing -- CRC32(C') won't match CRC32(A)  # CRC32(B).

Notice this only works because of step (3), where we can compute the expected checksum of the output from the expected checksums of the inputs. For most transformations, and most checksum algorithms, that isn't possible. But this is a special case.

mountpoint-s3/src/prefetch/checksummed_bytes.rs
```diff
@@ -0,0 +1,250 @@
+use bytes::{Bytes, BytesMut};
```
jamesbornholt, line 1: We need to add `pub mod checksummed_bytes;` to `prefetch.rs` for this file to be compiled/tests to run.

mountpoint-s3/src/prefetch/checksummed_bytes.rs
```diff
@@ -0,0 +1,250 @@
...
+            orig_bytes,
+            curr_slice,
+            checksum,
+        }
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum IntegrityError {
+    #[error("Checksum mismatch. left: {0:?}, right: {1:?}")]
```
jamesbornholt, line 125: "expected" and "actual" might be clearer

mountpoint-s3/Cargo.toml
```diff
@@ -56,6 +56,7 @@ fuse_tests = []
 s3_tests = []
 shuttle = []
 put = []
+checksum = []
```
jamesbornholt, line 59: Could you add a comment for this feature? It's a temporary feature flag to enable read-path checksums while we iron out performance. (We should have done the same for `put`).

mountpoint-s3/src/prefetch/checksummed_bytes.rs
monthonk: I think that's true only when we copy the whole buffer. The problem here is that A and B could be slices of bigger buffers. For example, if A is a slice of X and B is a slice of Y then we can't compute CRC32(A ++ B) because we only have checksums of X and Y.

We can optimize it a bit here though, by checking that if A and B are original buffers we can skip the "do it twice" part.

mountpoint-s3/src/prefetch/checksummed_bytes.rs
monthonk: Ah, my bad, I branched it off from bigger changes and missed that line.

mountpoint-s3/src/prefetch/checksummed_bytes.rs
jamesbornholt: Hmm. I wonder how much we care about `split_off` performance versus making this operation faster, then. I guess we're happy because `extend` is likely to be the corner case (the slow path for `read`)?

PR #260, opened 2023-05-25
Prepare to publish client and CRT crates

jamesbornholt:
This change adds READMEs and CHANGELOGs for the client, crt, and crt-sys
crates, and bumps their version numbers just for clarity. It changes the
path dependencies to also include version numbers, so that local builds
of Mountpoint will use the path dependency but published releases will
use the published dependencies.

---

### DIFF
--- a/mountpoint-s3/Cargo.toml
+++ b/mountpoint-s3/Cargo.toml
@@ -7,8 +7,8 @@ publish = false
 
 [dependencies]
 fuser = { path = "../vendor/fuser", features = ["abi-7-28"] }
-mountpoint-s3-client = { path = "../mountpoint-s3-client" }
-mountpoint-s3-crt = { path = "../mountpoint-s3-crt" }
+mountpoint-s3-client = { path = "../mountpoint-s3-client", version = "0.2.1" }
+mountpoint-s3-crt = { path = "../mountpoint-s3-crt", version = "0.2.1" }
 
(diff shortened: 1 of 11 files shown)


mountpoint-s3/Cargo.toml
```diff
@@ -7,8 +7,8 @@ publish = false
 
 [dependencies]
 fuser = { path = "../vendor/fuser", features = ["abi-7-28"] }
-mountpoint-s3-client = { path = "../mountpoint-s3-client" }
-mountpoint-s3-crt = { path = "../mountpoint-s3-crt" }
+mountpoint-s3-client = { path = "../mountpoint-s3-client", version = "0.2.1" }
+mountpoint-s3-crt = { path = "../mountpoint-s3-crt", version = "0.2.1" }
```
dannycjones, line 11: not to change in this PR, but we were wondering why Mountpoint is depending directly on the CRT crate.

dannycjones (approved):
LGTM! Thanks for adding the instructions for bumping the version also.

monthonk (approved):

mountpoint-s3/Cargo.toml
jamesbornholt: Looks like it's currently just for `RustLogAdapter`. I guess we could re-export that from `mountpoint-s3-client` but that feels a little weird.

PR #263, opened 2023-05-26
Implement checksum validation for the prefetcher

monthonk:
We want to ensure data integrity while it is within Mountpoint's custody but S3 API doesn't support ranged-get checksum at the moment.

As a temporary measure, we will calculate CRC checksum for the data as soon as we receive it from S3 and do the validation every time we pass it around our part queue, and right before returning them to FUSE. To do this, we will use ChecksummedBytes from previous change instead of Bytes in the prefetcher.

---

### DIFF
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -206,6 +206,9 @@ where
 
             let part_len = part_bytes.len() as u64;
             if response.extend(part_bytes).is_err() {
+                // cancel inflight tasks
+                self.current_task = None;
+                self.future_tasks.write().unwrap().drain(..);
                 return Err(PrefetchReadError::Integrity);
             }
             to_read -= part_len;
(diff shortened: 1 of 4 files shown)


jamesbornholt:
https://github.com/awslabs/aws-checksums/pull/65 is what's causing the test segfaults.

monthonk:
@jamesbornholt thanks for looking into the segfaults problem, that is an interesting finding.

mountpoint-s3/src/prefetch.rs
```diff
@@ -198,8 +204,11 @@ where
                 return Ok(part_bytes);
             }
 
-            response.extend_from_slice(&part_bytes[..]);
-            to_read -= part_bytes.len() as u64;
+            let part_len = part_bytes.len() as u64;
+            if response.extend(part_bytes).is_err() {
+                return Err(PrefetchReadError::Integrity);
```
jamesbornholt, line 209: Do we need to do the same cancel inflight tasks work here as we do when `current_task.read()` above fails?

jamesbornholt (approved):

PR #266, opened 2023-05-30
Add crate version to mountpoint-s3-client user agent

jamesbornholt:
We already have this in mountpoint-s3, but for direct users of the
client crate there's currently no version information. This change just
duplicates our existing mountpoint-s3 logic into the client crate. I
also removed the "UNKNOWN" suffix for builds without git history (i.e.,
every crates.io build), since it seems spammy -- we can distinguish
them by the absence of the `-`.

---


passaro (approved):
LGTM

PR #273, opened 2023-06-02
Add HeadBucket NoSuchBucket error to S3CrtClient

dannycjones:
This change recognizes 404 as a real service error rather than falling back on our client error code.

This changes the following failure mode when trying to connect to a bucket that doesn't exist.

### Before

```

Caused by:
    0: HeadBucket failed for bucket dannycjones-use1 in region us-east-1
    1: Client error
    2: Unknown response error: MetaRequestResult { response_status: 404, crt_error: Error(14343, "aws-c-s3: AWS_ERROR_S3_INVALID_RESPONSE_STATUS, Invalid response status from request"), error_response_headers: Some(Headers { inner: 0x7f7d49f0afa0 }), error_response_body: None }
```

### After

```

Caused by:
    0: HeadBucket failed for bucket dannycjones-use1 in region us-east-1
    1: Service error
    2: The bucket does not exist
```

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client/head_bucket.rs
+++ b/mountpoint-s3-client/src/s3_crt_client/head_bucket.rs
@@ -11,7 +11,7 @@ pub enum HeadBucketError {
     IncorrectRegion(String),
 
     #[error("The bucket did not exist")]
-    NoSuchBucket(),
+    NoSuchBucket,
 
     #[error("Permission denied")]
     PermissionDenied(MetaRequestResult),
@@ -40,7 +40,7 @@ impl S3CrtClient {
                         ))),
                     // S3 returns 400 for invalid or expired STS tokens
                     400 | 403 => ObjectClientError::ServiceError(HeadBucketError::PermissionDenied(request_result)),
-                    404 => ObjectClientError::ServiceError(HeadBucketError::NoSuchBucket()),
+                    404 => ObjectClientError::ServiceError(HeadBucketError::NoSuchBucket),
                     _ => ObjectClientError::ClientError(S3RequestError::ResponseError(request_result)),
                 }
             })?
--- a/mountpoint-s3-client/src/s3_crt_client/head_bucket.rs
+++ b/mountpoint-s3-client/src/s3_crt_client/head_bucket.rs
@@ -10,6 +10,9 @@ pub enum HeadBucketError {
     #[error("Wrong region: should be {0}")]
     IncorrectRegion(String),
 
+    #[error("The bucket did not exist")]
+    NoSuchBucket(),
+
     #[error("Permission denied")]
     PermissionDenied(MetaRequestResult),
 }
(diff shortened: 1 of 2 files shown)


mountpoint-s3-client/tests/head_bucket.rs
```diff
@@ -41,3 +41,16 @@ async fn test_head_bucket_forbidden() {
         Err(ObjectClientError::ServiceError(HeadBucketError::PermissionDenied(_)))
     ));
 }
+
+#[tokio::test]
+async fn test_head_bucket_not_found() {
+    let client = get_test_client();
+    let bucket = "test-nosuchbucket-s3alias";
```
dannycjones, line 48: Tried to pick a bucket name that won't ever exist. `-s3alias` is a reserved suffix for Access Point bucket-based aliasing.

mountpoint-s3-client/src/s3_crt_client/head_bucket.rs
```diff
@@ -10,6 +10,9 @@ pub enum HeadBucketError {
     #[error("Wrong region: should be {0}")]
     IncorrectRegion(String),
 
+    #[error("The bucket did not exist")]
```
sauraank, line 13: Can we include the bucket name in error?

mountpoint-s3-client/tests/head_bucket.rs
```diff
@@ -41,3 +41,16 @@ async fn test_head_bucket_forbidden() {
         Err(ObjectClientError::ServiceError(HeadBucketError::PermissionDenied(_)))
     ));
 }
+
+#[tokio::test]
+async fn test_head_bucket_not_found() {
+    let client = get_test_client();
+    let bucket = "nosuch..bucket"; // Invalid bucket name, so cannot be created by anyone.
```
jamesbornholt, line 48: Think I'd rather use a bucket name that is explicitly documented to not exist, like `DOC-EXAMPLE-BUCKET`, rather than inventing our own.

mountpoint-s3-client/src/s3_crt_client/head_bucket.rs
jamesbornholt: probably not necessary since the caller knows what bucket name they passed in

mountpoint-s3-client/src/s3_crt_client/head_bucket.rs
```diff
@@ -10,6 +10,9 @@ pub enum HeadBucketError {
...
+    NoSuchBucket(),
```
jamesbornholt, line 14:
```suggestion
    NoSuchBucket,
```

mountpoint-s3-client/src/s3_crt_client/head_bucket.rs
dannycjones: It's good thinking. I was on the fence with this one.

>probably not necessary since the caller knows what bucket name they passed in

Yeah, I agree. Besides, I gave it a quick go and it wasn't trivial to get the bucket name in. (Lifetime nonsense.)

mountpoint-s3-client/tests/head_bucket.rs
dannycjones: I thought the same. The issue is that these buckets actually exist - they are just owned by AWS to prevent them being grabbed by others.

jamesbornholt: The uppercase versions don't exist, and so this is actually also (surprisingly) a good test of whether we're preserving case in bucket names or not.

dannycjones: Ah, interesting. I've updated the test and it does pass using that casing. It switches over to path-based addressing, so it is case sensitive with this behavior.

I'll push an update soon.

jamesbornholt (approved):

PR #282, opened 2023-06-07
Introduce uploader and use async streaming

passaro:
Adopt the new async streaming API offered by the CRT to implement streaming PUT on file uploads. Bottom-up list of changes:

### CRT Wrapper

Expose the new async streaming API by introducing `AsyncInputStream` and `AsyncStreamWriter`. The former internally implements an `async_input_stream` that can be passed to a `put_object` request, the latter exposes `write` and `complete` methods that feed data into the stream. These objects are created in pairs, internally connected through an `async_channel`: when the CRT invokes a `read` on the stream, a `ReadRequest` is sent through the channel with a pointer to the buffer to fill and a future to set when done. `AsyncStreamWriter::write(slice)` receives a read request, copies data from `slice` into its buffer, and set its future to signal the CRT. If `slice` is not completely consumed, it repeats the process with the next read request. `AsyncStreamWriter::complete` receives the next read request and set its future to `true` to signal eof.

### S3 Client

Modify `put_object` to return a `PutObjectRequest` which provides async `write` and `complete` methods to stream the body of the request. Note that `PutObjectRequest` is a new trait and each `ObjectClient` implementation provides its own implementation. The `S3CrtClient` implementation uses the new CRT async streaming API, while the `MockClient` simply records that an upload is in progress and adds the whole object on `complete`. 

### Mountpoint

Introduce a new `Uploader` object to handle file uploads, analogous to the `Prefetcher` for downloads. On `open` (for write), the `Uploader` is used to create an `UploadRequest`, which is stored in the `FileHandle` and starts a `PutObjectRequest`. Subsequent `write` calls push data slices to the `UploadRequest`, which writes them to the put request. Finally, the `UploadRequest` (and its internal `PutObjectRequest`) is completed when the handle is released and `finish_writing` is called.

---


mountpoint-s3-crt/src/io/async_stream.rs
```diff
@@ -0,0 +1,246 @@
...
+    fn new(allocator: &Allocator, sender: Sender<ReadRequest>) -> Self {
+        let ptr = Box::new(AsyncInputStreamImpl {
+            inner: Default::default(),
+            sender,
+        });
+
+        // Turn out Box pointer into a raw pointer (effectively leaking it).
+        let ptr = Box::into_raw(ptr) as *mut AsyncInputStreamImpl;
+
+        // SAFETY: We know ptr isn't null because we just made it from Box.
```
jamesbornholt, line 36: `AsyncInputStreamImpl` becomes self-referential here (`ptr.inner.impl == ptr`). Usually that means we'd have to `Pin` the `Box<AsyncInputStreamImpl>`, but because that thing never escapes this module, we can just document it in this safety comment instead.

Actually, `async_input_stream_to_impl` should maybe return `Pin<&mut AsyncInputStreamImpl>` just to make this expectation clear.

To see why `Pin` is necessary, suppose we did this in `read_impl`, which compiles fine:
```rust
let b = Box::from_raw(async_input_stream_to_impl(stream));
let b = Box::new(*b);
let stream = Box::leak(b);
```

Now we've moved the `AsyncInputStreamImpl` to a different address, and freed the old one, so `ptr.inner.impl` is bogus and we'll probably crash the next time we try to use this. Pinning prevents us from making the move:
```rust
let b = std::pin::Pin::new_unchecked(Box::from_raw(async_input_stream_to_impl(stream)));
let b = Box::new(*b);  // <-- no longer compiles -- cannot move out of dereference of Pin 
let stream = Box::leak(b);
```

mountpoint-s3/src/upload.rs
```diff
@@ -0,0 +1,101 @@
...
+use mountpoint_s3_client::{
+    ObjectClient, ObjectClientError, ObjectClientResult, PutObjectError, PutObjectParams, PutObjectRequest,
+    PutObjectResult,
+};
+
+use thiserror::Error;
+
+type PutRequestError<Client> = ObjectClientError<PutObjectError, <Client as ObjectClient>::ClientError>;
+
+/// An [Uploader] creates and manages streaming PutObject requests.
```
jamesbornholt, line 12: Can you explain a little more what you picture the role of this thing being beyond a `PutObjectRequest`? If I'm reading it right, the big thing it does is enforce sequential writes (which the current write code inlines into the `write` FUSE handler). Will it do other things too, like checksums? Does it do its own buffering?

mountpoint-s3/src/upload.rs
```diff
@@ -0,0 +1,101 @@
...
+        self.next_request_offset
+    }
+
+    pub async fn write(&mut self, offset: i64, data: &[u8]) -> Result<(), UploadWriteError<PutRequestError<Client>>> {
+        let next_offset = self.next_request_offset;
+        if offset != next_offset as i64 {
+            return Err(UploadWriteError::OutOfOrderWrite(offset as u64, next_offset));
+        }
+
+        self.request.write(data).await?;
```
jamesbornholt, line 93: It's worth thinking through (and maybe documenting) how this call creates backpressure -- if the upload is going slowly, does that backpressure reach here? Would we consider inserting a buffer somewhere to absorb some transient slowness in uploads, or is the CRT already providing that buffering (and is it unbounded)?

mountpoint-s3-crt/src/io/async_stream.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+}
+
+impl Clone for BoolPromise {
+    fn clone(&self) -> Self {
+        // SAFETY: inner contains a valid aws_future_bool.
+        let inner = unsafe { NonNull::new_unchecked(aws_future_bool_acquire(self.inner.as_ptr())) };
+        Self { inner }
+    }
+}
```
jamesbornholt, line 246: Think it's worth trying to write a few tests specifically for this code. Maybe spawn two threads and have one read while the other writes, or something. Try to exercise the various failure cases (what if the reader gives up? what if the writer gives up?).

jamesbornholt (commented):
This is awesome! I have a bunch of detailed comments, but overall I really like the approach, it's very clean.

mountpoint-s3-client/src/object_client.rs
```diff
@@ -245,6 +249,19 @@ pub enum GetObjectAttributesError {
...
+/// the body of the request.
+#[async_trait]
+pub trait PutObjectRequest: Debug {
+    type ClientError: std::error::Error + Send + Sync + 'static;
+
+    /// Write the given slice to the put request body.
+    async fn write(&mut self, slice: &[u8]) -> ObjectClientResult<(), PutObjectError, Self::ClientError>;
+
+    /// Complete the put request and return a [PutObjectResult].
+    async fn complete(self) -> ObjectClientResult<PutObjectResult, PutObjectError, Self::ClientError>;
```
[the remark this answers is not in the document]
passaro: Good point. The intended behavior is that the MPU should be aborted, but there is an issue right now where we may fail to notify the CRT. Will fix and add tests.

mountpoint-s3-crt/src/io/async_stream.rs
```diff
@@ -0,0 +1,246 @@
...
+    receiver: async_channel::Receiver<ReadRequest>,
+}
+
+impl AsyncStreamWriter {
+    /// Write a slice to the stream.
+    pub async fn write(&mut self, slice: &[u8]) -> Result<(), AsyncStreamWriterError> {
+        let mut remaining = slice;
+        while !remaining.is_empty() {
+            let request = self.receiver.recv().await?;
+            request.consume(&mut remaining);
```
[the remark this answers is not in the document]
passaro: Yes

mountpoint-s3/src/upload.rs
passaro: Yes, that's all it is now. It could be a good place to put additional logic that is specific to mountpoint (vs generic for PutObjectRequest).
Checksums, buffering: both are likely candidates, but I do not have a clear picture yet. For example, if we decide to support some form of local staging (e.g. implement fallocate), I would put all that logic here.

mountpoint-s3/src/upload.rs
passaro: We are not doing any buffering here, completely relying on the CRT. But I'll need to dig deeper to fully understand what the CRT strategy is.

mountpoint-s3-crt/src/io/async_stream.rs
```diff
@@ -168,25 +234,30 @@ unsafe fn async_input_stream_to_impl(stream: *mut aws_async_input_stream) -> *mu
...
+    let stream = async_input_stream_to_impl(stream);
 
     let promise = BoolPromise::new(stream.inner.alloc);
     let request = ReadRequest::new(dest, promise.clone());
-    if let Err(e) = stream.sender.try_send(request) {
-        debug!("channel closed: {:?}", e);
-        e.into_inner().complete();
+    if let Err(_e) = stream.sender.try_send(request) {
+        // _e implicitly drops the [ReadRequest].
     }
```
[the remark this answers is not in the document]
jamesbornholt: Oh, I see that contract in the CRT headers now: https://github.com/awslabs/aws-c-io/blob/c3f95ebf3d6d1608e1656bb08eca05e303cc4fd3/include/aws/io/async_stream.h#L94

mountpoint-s3-crt/src/io/async_stream.rs
```diff
@@ -20,20 +40,22 @@ pub fn new_stream(allocator: &Allocator) -> (AsyncInputStream, AsyncStreamWriter
...
+        // make it self-referential.
+        let ptr = Box::pin(AsyncInputStreamImpl {
             inner: Default::default(),
             sender,
         });
 
-        // Turn out Box pointer into a raw pointer (effectively leaking it).
-        let ptr = Box::into_raw(ptr) as *mut AsyncInputStreamImpl;
+        // Turn the pinned Box pointer into a raw pointer (effectively leaking it).
+        let ptr = Box::into_raw(Pin::into_inner(ptr));
```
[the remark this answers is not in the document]
passaro: It was my (clumsy) attempt to document the intent for `AsyncInputStreamImpl` to be pinned. But yes, 'PhantomPinned` makes much more sense.

mountpoint-s3-crt/src/io/async_stream.rs
```diff
@@ -112,22 +154,44 @@ unsafe impl Send for ReadRequest {}
...
+}
 
-    fn complete(self) {
-        self.promise.fulfill(Ok(true));
+impl Drop for ReadRequest {
+    fn drop(&mut self) {
+        if let Some(promise) = self.promise.take() {
+            // If the [ReadRequest] was not consumed/completed,
+            // notify the reader of the error.
+            promise.fulfill(Err((-1).into()));
```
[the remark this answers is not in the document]
passaro: It works as an unknown error. This is what we get in the logs:
```
Failed reading request body, error -1 (Unknown Error Code)   
```

But I'll try and see if there is anything more appropriate.

mountpoint-s3-client/tests/put_object.rs
```diff
@@ -95,3 +98,37 @@ async fn test_put_object_large(client: &impl ObjectClient, bucket: &str, prefix:
 }
 
 object_client_test!(test_put_object_large);
+
+// Test for dropped PUT object. Checks that the GET fails.
+async fn test_put_object_dropped(client: &impl ObjectClient, bucket: &str, prefix: &str) {
```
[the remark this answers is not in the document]
passaro: Yes, I've added that to the tests in `mountpoint`, where we already use the SDK client, but not here in `mountpoint-s3-client`. I'll try, but it may require a bit of reshuffling.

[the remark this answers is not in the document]
passaro: Actually, in `mountpoint` I'm only testing that the mpu is in progress. We don't exercise the abort there (yet?). 

So I still need to the SDK test dependency to s3-client, but also figure out how to test if the abort happens asynchronously...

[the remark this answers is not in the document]
passaro: I've added this to a new test, just for S3 (no mock). It calls `sleep` to wait for the abort, not sure what better options we have...

(deleted a couple of previous comments: we are already using the SDK client in `s3-client` tests)

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -53,7 +53,12 @@ pub struct MockClientConfig {
 #[derive(Debug)]
 pub struct MockClient {
     config: MockClientConfig,
-    objects: RwLock<BTreeMap<String, Arc<MockObject>>>,
+    objects: Arc<RwLock<BTreeMap<String, Arc<MockObject>>>>,
```
sauraank, line 56: Why have we added Arc here? Like where is the shared ownership between threads used?

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -437,13 +437,13 @@ impl S3CrtClient {
 /// virtual-hosted-style addresses. The `path_prefix` is appended to the front of all paths, and
 /// need not be terminated with a `/`.
 #[derive(Debug)]
-struct S3Message<'a> {
-    inner: Message<'a>,
+struct S3Message {
```
sauraank, line 440: Why was lifetime required earlier and removed now? It might be trivial but I am failing to understand it.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -630,9 +630,46 @@ fn extract_range_header(headers: &Headers) -> Option<Range<u64>> {
...
+    type ClientError = S3RequestError;
+
+    async fn write(&mut self, slice: &[u8]) -> ObjectClientResult<(), PutObjectError, Self::ClientError> {
+        self.writer
+            .write(slice)
+            .await
+            .map_err(|e| S3RequestError::InternalError(Box::new(e)).into())
+    }
+
+    async fn complete(mut self) -> ObjectClientResult<PutObjectResult, PutObjectError, Self::ClientError> {
```
sauraank, line 656: Should we add rustdoc on methods like complete here because it can be difficult what AsyncStreamWriter is doing (might be easier with VSCode though)?

mountpoint-s3-crt/src/io/async_stream.rs
```diff
@@ -15,16 +15,17 @@
 //!
 //! The following behavior is assumed from the reader (CRT):
 //! * the buffer passed on read remains valid at least until the future is fulfilled,
-//! * read calls on an `aws_async_input_stream` are never concurrent,
+//! * read calls on an `aws_async_input_stream` are never concurrent: the reader will only call
+//!   the next read after the future is fulfilled,
```
jamesbornholt, line 19:
```suggestion
//!   the next read after the future from the previous read is fulfilled,
```

mountpoint-s3-crt/src/io/async_stream.rs
```diff
@@ -203,6 +204,9 @@ struct AsyncInputStreamImpl {
 
     /// The sending side of a channel.
     sender: Sender<ReadRequest>,
+
+    /// The [aws_async_input_stream] is self-refencing, so mark as pinned.
```
jamesbornholt, line 208:
```suggestion
    /// The [aws_async_input_stream] is self-referencing, so mark as pinned.
```

mountpoint-s3-crt/src/io/async_stream.rs
```diff
@@ -347,25 +366,31 @@ mod test {
...
 
-        assert_eq!(read_result, Ok(false));
         if expect_write_error {
             assert!(write_result.is_err());
         } else {
             assert!(write_result.is_ok());
         }
 
-        let min_size = std::cmp::min(write_size, read_buffer_size);
+        let expected_bytes_read = std::cmp::min(write_size, read_buffer_size);
```
jamesbornholt, line 383:
```suggestion
        let expected_bytes_read = write_size.min(read_buffer_size);
```

mountpoint-s3-crt/src/io/async_stream.rs
```diff
@@ -397,25 +422,31 @@ mod test {
...
 
-        assert_eq!(read_result, Ok(expect_eof));
         if expect_write_error {
             assert!(write_result.is_err());
         } else {
             assert!(write_result.is_ok());
         }
 
-        let min_size = std::cmp::min(write_size, read_buffer_size);
+        let expected_bytes_read = std::cmp::min(write_size, read_buffer_size);
```
jamesbornholt, line 439:
```suggestion
        let expected_bytes_read = write_size.min(read_buffer_size);
```

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -493,6 +493,46 @@ impl ObjectClient for MockClient {
...
+    in_progress_uploads: Arc<RwLock<BTreeSet<String>>>,
+}
+
+impl MockPutObjectRequest {
+    fn new(
+        key: &str,
+        objects: &Arc<RwLock<BTreeMap<String, Arc<MockObject>>>>,
+        in_progress_uploads: &Arc<RwLock<BTreeSet<String>>>,
+    ) -> Self {
+        in_progress_uploads.write().unwrap().insert(key.to_owned());
```
jamesbornholt, line 510: We don't rely on it anywhere, but this need a `Drop` impl to clean up if the caller didn't `complete`.

mountpoint-s3/src/fs.rs
```diff
@@ -304,10 +308,15 @@ where
                 return Err(libc::EINVAL);
             }
 
-            let inode_handle = self.superblock.write(&self.client, ino, lookup.inode.parent()).await?;
+            let handle = self.superblock.write(&self.client, ino, lookup.inode.parent()).await?;
+            let request = self
+                .uploader
+                .put(&self.bucket, lookup.inode.full_key())
+                .await
+                .map_err(|_| libc::EIO)?;
```
jamesbornholt, line 316: We probably want to log the error here before returning.

mountpoint-s3/src/fs.rs
```diff
@@ -464,26 +471,20 @@ where
...
-        }
-
-        // If we'd go over the size limit, fail the entire write rather than short-writing
-        if next_offset + data.len() > MAX_OBJECT_SIZE {
-            error!("object too large");
-            return Err(libc::EFBIG);
+        match request.write(offset, data).await {
+            Ok(()) => {
+                let len = data.len();
+                Ok(len as u32)
```
jamesbornholt, line 481:
```suggestion
                Ok(data.len() as u32)
```

mountpoint-s3/src/fs.rs
```diff
@@ -602,19 +603,15 @@ where
...
-                let size = parts.iter().map(|part| part.len()).sum::<usize>();
-                let stream = futures::stream::iter(parts);
+            FileHandleType::Write { request, handle } => {
                 let key = file_handle.full_key;
 
-                let put = self
-                    .client
-                    .put_object(&self.bucket, &key, &PutObjectParams::default(), stream)
-                    .await;
+                // TODO how do we make sure we didn't already handle this via `flush`?
```
jamesbornholt, line 609: Is this really a TODO? We don't implement flush right now.

mountpoint-s3/src/upload.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum UploadWriteError<E: std::error::Error> {
+    #[error("put request failed")]
+    PutRequestFailed(#[from] E),
+
+    #[error("out of order write; expected offset {1} but got {0}")]
+    OutOfOrderWrite(u64, u64),
```
jamesbornholt, line 49: We should probably name these fields.

mountpoint-s3/src/fs.rs
```diff
@@ -464,26 +471,20 @@ where
...
-            error!("object too large");
-            return Err(libc::EFBIG);
+        match request.write(offset, data).await {
+            Ok(()) => {
+                let len = data.len();
+                Ok(len as u32)
+            }
+            Err(e) => {
+                error!("write failed: {:?}", e);
+                Err(e.into())
```
jamesbornholt, line 485: Hmm, I'm wondering what we should do with the upload if a write fails like this. Is it better to persist the object, knowing that we've dropped some writes and so it's not the object the customer expected? Or should we abandon the upload completely?

mountpoint-s3-crt/src/http/request_response.rs
```diff
@@ -243,15 +243,15 @@ impl<'a> Iterator for HeadersIterator<'a> {
...
 /// A single HTTP message, initialized to be empty (i.e., no headers, no body).
 #[derive(Debug)]
-pub struct Message<'a> {
+pub struct Message {
     /// The pointer to the inner `aws_http_message`.
     pub(crate) inner: NonNull<aws_http_message>,
 
     /// Input stream for the body of the http message, if present.
-    body_input_stream: Option<InputStream<'a>>,
+    body_input_stream: Option<AsyncInputStream>,
```
jamesbornholt, line 251: Can we just delete the old InputStream implementation now? We're not using it anywhere, right?

jamesbornholt (commented):
Couple of actual questions, plus a bunch of minor stuff. Almost finished, thanks for pushing on this!

mountpoint-s3-client/src/mock_client.rs
passaro: The reference to the objects map is cloned by a `MockPutObjectRequest` so it can insert the object when the put is completed.

mountpoint-s3-client/src/s3_crt_client.rs
passaro: That's because it previously owned (through `Message`) an `InputStream<'a>`, which in turn could have its lifetime bound by that of a non-owned buffer.
Now `Message` owns an `AsyncInputStream` which does not have lifetime bounds.

mountpoint-s3/src/fs.rs
passaro: Good question. My first answer would be that we should abort, but I have not thought it through:
* we can abort the upload, but what about the file handle? just keep failing on subsequent writes?
* is there a difference between writes erroring out immediately (e.g. out-of-order) and partial writes?
* where do write checksums fit in? 

Worth raising a separate issue?

jamesbornholt: Yeah, let's track it separately.

PR #283, opened 2023-06-08
Adjust prefetcher to spawn requests earlier

jamesbornholt:
This change makes three adjustments to our prefetcher:
1. Currently we only try spawning requests once per read, at the start.
   This means we miss the opportunity to spawn requests after the read,
   so if the read caused us to cross the prefetch boundary, we don't
   trigger the request until the next read starts. This gives up a
   little bit of latency. Tweaking this also makes the code a little
   simpler/less branchy.
2. The prefetch threshold is a strict inequality, which is a bad fit for
   our power-of-two-oriented configs -- FUSE will read in 128k pieces,
   so if the data remaining is exactly half the size (which it often
   will be because of powers of two), we again have to wait an extra
   request.
3. Our default prefetcher config is mostly made up, and in benchmarking
   we've seen it's a little too pessimistic about random reads. I picked
   a new, still mostly-made-up first request size that allows us to
   service up to 1MiB reads in one request. I also adjusted it a little
   to account for Linux readahead (see the comment about that).

Also one thing I noticed along the way: `max_readhead` can't be set to
zero (fuser returns an error that we're ignoring), so I dropped that.

---

### DIFF
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -25,7 +25,7 @@ use mountpoint_s3_crt::checksums::crc32c;
 use thiserror::Error;
 use tracing::{debug_span, error, trace, Instrument};
 
-use crate::prefetch::checksummed_bytes::ChecksummedBytes;
+use crate::prefetch::checksummed_bytes::{ChecksummedBytes, IntegrityError};
 use crate::prefetch::part::Part;
 use crate::prefetch::part_queue::{unbounded_part_queue, PartQueue};
 use crate::sync::{Arc, RwLock};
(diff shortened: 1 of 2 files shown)


jamesbornholt:
The ARM CI failure is a timeout fixed in https://github.com/awslabs/mountpoint-s3/pull/284.

mountpoint-s3/src/prefetch.rs
```diff
@@ -48,8 +48,17 @@ pub struct PrefetcherConfig {
...
-            first_request_size: 256 * 1024,
+            // This is a weird looking number! We really want our first request size to be 1MiB,
+            // which is a common IO size. But Linux's readahead will try to read an extra 128k on on
+            // top of a 1MiB read, which we'd have to wait for a second request to service. Because
+            // FUSE doesn't know the difference between regular reads and readahead reads, it will
+            // send us a READ request for that 128k, so we'll have to block waiting for it even if
+            // the application doesn't want it. This is all in the noise for sequential IO, but
+            // waiting for the readahead hurts random IO. So we add 128k to the first request size
+            // to avoid the latency hit of the second request.
+            first_request_size: 1 * 1024 * 1024 + 128 * 1024,
```
monthonk, line 61: This readahead value, is it the same across linux distributions?

jamesbornholt: No, it's baked into the kernel: https://elixir.bootlin.com/linux/latest/source/include/linux/pagemap.h#L1185 It's been 128k basically forever, although FUSE didn't always support > 4k reads.

monthonk (approved):

PR #285, opened 2023-06-09
Avoid CRT auto-ranged-get for small requests

jamesbornholt:
The CRT's auto-ranged-get will always do a HeadObject request when a
range is specified. For small requests, that's a fairly large latency
hit. Let's avoid using the auto-ranged-get for requests small enough
that they would be sent as a single part anyway.

It would probably be nicer to have a way to turn off this behavior in
the CRT, but this is a reasonable change for now. We'll have to be
careful if we ever start relying on any extra behavior of the
auto-ranged-get machinery, like checksum validation, but we're not
currently planning anything like that.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -100,6 +100,7 @@ pub struct S3CrtClient {
     /// Here it will add the user agent prefix and s3 client information.
     user_agent_header: String,
     request_payer: Option<String>,
+    part_size: Option<usize>,
 }
 
 impl S3CrtClient {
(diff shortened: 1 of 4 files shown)


passaro (approved):
Change looks good to me. 

Follow up questions:
1. Should we try and push down this logic into the CRT at some point?
2. Do we expect to see the lower latency in our existing benchmarks?

jamesbornholt:
>     1. Should we try and push down this logic into the CRT at some point?

Yeah, I think so. I actually originally tried making this change in the CRT as a "trust the range header" type thing, but there's a bunch of corner cases (checksums, empty objects, pause/resume, ...) that made it easier to just do it here quickly.

>     2. Do we expect to see the lower latency in our existing benchmarks?

Yep, and the TTFB benchmarks look like they improved with this change.

PR #287, opened 2023-06-12
Add `fuser` version to `mountpoint-s3` dependency

dannycjones:
We did not specify the version of our vendored fuser crate.

Adding this will ensure we're aware of the major or minor version changes when pulling in new fuser versions.

---


jamesbornholt (approved):
LGTM, but we'll have to rename that crate if we want to publish `mountpoint-s3`, or just _really_ vendor it by putting it inside the `mountpoint-s3` crate.

PR #289, opened 2023-06-13
Fix rmdir & unlink fuse tests by adding wait condition following release operation

sauraank:
Reproduced the failure by running fuse tests on 10 different terminals. Root cause looks like `release()` called after `drop() `get placed lower in queue when it is asked to perform multiple times. Hence, corresponding `unlink()` or `rmdir() `is being called earlier than release() making the file not uploaded to S3 as object.
Resolved the error by adding sleep before remove directory/file so that previous upload of files to S3 is completed.

---


mountpoint-s3/tests/fuse_tests/rmdir_test.rs
```diff
@@ -42,6 +44,9 @@ where
 
     file.write_all(b"Hello World").unwrap();
     drop(file);
+    // when there are multiple tests called release method can be called later than unlink or rmdir.
+    // So, adding the sleep to ensure release being called earlier.
+    sleep(Duration::from_secs(1));
     let err =
         fs::remove_dir(&non_empty_dirpath).expect_err("removing non-empty directory should fail even after closing");
     assert_eq!(
```
jamesbornholt, line 52: I think we'll be sad if we don't add a few retries to these sleeps, since sometimes the test runner might be very slow. Probably we want some kind of general function that retries a few times and then fails if none of the retries passed.

sauraank: Yes. I can add a macro which calls a methods  with a certain number of retries for it to succeed as was discussed yesterday.

mountpoint-s3/tests/fuse_tests/mod.rs
```diff
@@ -263,3 +287,22 @@ pub fn read_dir_to_entry_names(read_dir_iter: ReadDir) -> Vec<String> {
...
+                    } else {
+                        sleep(Duration::from_millis(10));
+                    }
+                }
+                Err(_) => panic!("The provided method could not provide result"),
+            }
+            assert!(i < 1000, "The provided method could not provide success result");
+        }
+    };
+}
```
dannycjones, line 308: I'd try and make this more generic if possible.

Can we have this macro take a closure and a failure message?

I want to be able to write something like...

```rust
let test_condition = || test_client.contains_key(&format!("{}/{}", non_empty_dirname, filename)).unwrap();
wait_until!(test_condition, "timeout waiting for s3 to contain new object");
```

dannycjones, line 308: I think the values and sleeps we have is a bit extreme here.

Each time, we're going to be making a call to S3 (when not using mock client), which will add a bunch of latency to any test if the test fails. In the worst case where the condition never holds, that would be `(1000 * avg_latency) + 10s`.

Maybe we retry five times with a 200ms delay each time? Better yet, we could grow the sleep exponentially, doubling with each retry (up to a small number like five retries).

mountpoint-s3/tests/fuse_tests/unlink_test.rs
```diff
@@ -143,7 +146,10 @@ where
     );
 
     drop(f); // close file
-
+             // when there are multiple tests called release method can be called later than unlink or rmdir.
+             // So, adding the sleep to ensure release being called earlier.
+    let result = test_client.contains_key("dir/writing.txt");
+    sleep_till_retry_succeed!(result);
```
dannycjones, line 152: Fix indentation here. Also the message can be a little clearer.

```suggestion
    // `release` operation triggered by `drop(File)` is asynchronous.
    // Add sleep to ensure condition is checked after `release` completes.
    let result = test_client.contains_key("dir/writing.txt");
    sleep_till_retry_succeed!(result);
```

dannycjones (commented):
Thanks Ankit, left some feedback

mountpoint-s3/tests/fuse_tests/mod.rs
sauraank: Will need to learn how to do it, but yes I will do it.

sauraank: Working on it

mountpoint-s3/tests/fuse_tests/mod.rs
```diff
@@ -263,3 +287,22 @@ pub fn read_dir_to_entry_names(read_dir_iter: ReadDir) -> Vec<String> {
...
+        for i in 1..5 {
+            match $method_result {
+                Ok(bool_value) => {
+                    if bool_value {
+                        break;
+                    } else {
+                        sleep(Duration::from_millis(100 * i));
+                    }
+                }
+                Err(_) => panic!("The provided method could not provide result"),
```
dannycjones, line 303: I'd like to see us be able to provide the message here on what we're waiting for

jamesbornholt (approved):

PR #300, opened 2023-06-19
Move GitHub workflow dependency installation to its own action file

dannycjones:
This change collects all the dependency installation into one script, such that it is consistent across operating systems.

This should allow us to start building a matrix up for different operating systems, architectures, AWS regions, etc..

---

### DIFF
--- a/doc/BENCHMARKING.md
+++ b/doc/BENCHMARKING.md
@@ -22,12 +22,11 @@ We keep the records of benchmarking results in `gh-pages` branch and the perform
 While our benchmark script is written for CI testing only, it is possible to run manually.
 You can use the following steps.
 
-1. Install dependencies by running below commands for Ubuntu.
+1. Install dependencies and configure FUSE by running the following script in the repository:
 
-        sudo apt-get update
-        sudo apt-get -y install cmake libclang-dev libunwind-dev pkg-config jq fio
-        sudo apt-get -y install fuse libfuse-dev
-        echo 'user_allow_other' | sudo tee -a /etc/fuse.conf
+        bash .github/actions/install-dependencies/install.sh \
+                --fuse-version 2 \
+                --with-fio --with-libunwind
 
 2. Set environment variables related to the benchmark. There are four required environment variables you need to set in order to run the benchmark.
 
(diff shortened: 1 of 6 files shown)


dannycjones:
Note, CI isn't going to actually test these changes since it won't pull in the updated workflow spec.

sauraank:
> Note, CI isn't going to actually test these changes since it won't pull in the updated workflow spec.
> 
> Here's a push to my fork which shows the Ubuntu runner working - I can't test self-hosted runner right now. https://github.com/dannycjones/mountpoint-s3/actions/runs/5315330797/jobs/9623572410

How do you test the runner?

doc/BENCHMARKING.md
```diff
@@ -22,12 +22,11 @@ We keep the records of benchmarking results in `gh-pages` branch and the perform
 While our benchmark script is written for CI testing only, it is possible to run manually.
 You can use the following steps.
 
-1. Install dependencies by running below commands for Ubuntu.
+1. Install dependencies and configure FUSE by running the following script in the repository:
```
sauraank, line 25: apt-get does not work outside Linux, right?

dannycjones:
> How do you test the runner?

For the GitHub managed runner, I've just pushed to my fork's main branch and seen all is OK.

For self-hosted runners, we'd have to push and find out later 😄 That being said, I did launch an EC2 Instance and run the commands by hand and check tests pass as expected.

sauraank (commented):
How were Github Workflow dependencies installed earler?

doc/BENCHMARKING.md
dannycjones: It won't work outside of Debian-based operating systems (so Ubuntu, Debian).

dannycjones:
>How were Github Workflow dependencies installed earler?

They were installed just using `apt-get` or `yum` hardcoded for each job. This takes those steps and puts them in a script, which figures out which package manager to use and which package names to use if they are different.

doc/BENCHMARKING.md
dannycjones: Also, the script only works on distros using apt/apt-get and rpm/yum/dnf.

sauraank (approved):

PR #304, opened 2023-06-21
Add clang and pkg-config dependencies to getting started instructions

dannycjones:
We were missing these two packages when using a fresh OS installation and following the getting started instructions.

`pkg-config` is already included on AL, but name it as a dependency anyway to be consistent with the Ubuntu instructions.

Tested by creating new EC2 Instances for:
- Ubuntu
- AL2
- AL2023

---


passaro (approved):
LGTM

PR #306, opened 2023-06-22
Remove AL2 + fuse3 pair from test matrix

dannycjones:
`fuse3` is not available on Amazon Linux 2. We should remove it from the test matrix.

I believe when we have Amazon Linux 2023 runners then we can run fine with fuse3. You can see this on the new package list: https://docs.aws.amazon.com/linux/al2023/release-notes/new-packages.html

---


sauraank (approved):
LGTM.

PR #308, opened 2023-06-22
Implement fsync and handle write errors

passaro:
Implement `fsync` to allow users to complete a put request and receive confirmation that it succeeded or failed. If a file handle is released without a call to `fsync`, `release` will still complete the upload as before. 

This change modifies `UploadRequest` to keep track of the state of a `PutObjectRequest`, so that
* `release` can detect whether the request had been already completed by an `fsync` call,
* `write` is invoked after an `fsync`,
* `write` (or `fsync`) is invoked after a previous call failed.

**Note**: Creating as draft:
* missing additional tests, especially around failures,
* should we add a flag to enable the new `fsync` behavior?

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -6,7 +6,7 @@ use std::collections::HashMap;
 use std::ffi::OsStr;
 use std::str::FromStr;
 use std::time::{Duration, UNIX_EPOCH};
-use tracing::{error, trace};
+use tracing::{debug, error, trace};
 
 use fuser::{FileAttr, KernelConfig};
 use mountpoint_s3_client::{ETag, ObjectClient};
@@ -479,8 +479,13 @@ where
             FileHandleType::Write(request) => request.lock().await,
             FileHandleType::Read { .. } => return Err(libc::EBADF),
         };
-        let len = request.write(offset, data).await?;
-        Ok(len as u32)
+        match request.write(offset, data).await {
+            Ok(len) => Ok(len as u32),
+            Err(e) => {
+                error!("write failed: {:?}", e);
+                Err(e.into())
+            }
+        }
     }
 
     pub async fn opendir(&self, parent: InodeNo, _flags: i32) -> Result<Opened, libc::c_int> {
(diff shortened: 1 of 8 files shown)


mountpoint-s3/src/fs.rs
```diff
@@ -55,10 +55,7 @@ enum FileHandleType<Client: ObjectClient, Runtime> {
         request: AsyncMutex<Option<PrefetchGetObject<Client, Runtime>>>,
         etag: ETag,
     },
-    Write {
-        request: AsyncMutex<UploadRequest<Client>>,
-        handle: WriteHandle,
-    },
+    Write(AsyncMutex<UploadRequest<Client, WriteHandle>>),
```
jamesbornholt, line 58: I think the `is_in_progress` logic belongs in the file system (it's ultimately about file handle management rather than the "stream"). And moving it here lets us keep using the type system more strongly. Make this `AsyncMutex<Option<UploadRequest<Client, WriteHandle>>>`, and then `complete` can still consume `self` by `take`ing it out of the `Option`. That way the uploader knows it's impossible to call `write` on an already finished upload, so we don't need all the tests for the state management there.

mountpoint-s3/src/upload.rs
```diff
@@ -31,88 +32,234 @@ where
...
             });
         }
 
-        self.request.write(data).await?;
-        self.next_request_offset += data.len() as u64;
-        Ok(())
+        let request = match &mut self.state {
+            UploadRequestState::InProgress { request, .. } => request,
+            UploadRequestState::Completed => {
+                error!(key = self.key, "object already uploaded");
```
jamesbornholt, line 122: See #310 -- for internal stuff (everything below the `fs.rs`), we should try to keep `error` logging out of cases where it's an "expected failure".

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -225,3 +225,57 @@ fn sequential_write_streaming_test_s3(object_size: usize, write_chunk_size: usiz
...
+
+    let mut rng = ChaCha20Rng::seed_from_u64(0x12345678 + OBJECT_SIZE as u64);
+    let mut body = vec![0u8; OBJECT_SIZE];
+    rng.fill(&mut body[..]);
+
+    f.write_all(&body).unwrap();
+
+    assert!(test_client.is_upload_in_progress(KEY).unwrap());
+
+    f.sync_all().unwrap();
```
jamesbornholt, line 254: Not in this PR, but we need to get around to adding `put` to the `FailureClient` so we can test the failure modes here.

mountpoint-s3/src/fs.rs
passaro: The counterpoint is that we do need to handle the state in `fs`. But I gave it a try and it is probably for the best: see #313 (we can then close this draft).

passaro:
Replaced by #313

PR #310, opened 2023-06-23
Move some inode-related error logging

jamesbornholt:
These failures in `inode.rs` aren't really errors -- in general, if the
user asked us to do something, it's not an internal error if that's not
possible. These are really "expected failures" from Mountpoint's
perspective. So this change shifts the actual `error!` log message into
the filesystem, and changes the `inode.rs` messages to warnings. This
makes our CI and test output less spammy, especially the reftests that
expect to test these cases repeatedly.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -308,7 +308,13 @@ where
                 return Err(libc::EINVAL);
             }
 
-            let handle = self.superblock.write(&self.client, ino, lookup.inode.parent()).await?;
+            let handle = match self.superblock.write(&self.client, ino, lookup.inode.parent()).await {
+                Ok(handle) => handle,
+                Err(e) => {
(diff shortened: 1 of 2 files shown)


mountpoint-s3/src/fs.rs
```diff
@@ -308,7 +308,13 @@ where
                 return Err(libc::EINVAL);
             }
 
-            let handle = self.superblock.write(&self.client, ino, lookup.inode.parent()).await?;
+            let handle = match self.superblock.write(&self.client, ino, lookup.inode.parent()).await {
+                Ok(handle) => handle,
+                Err(e) => {
+                    error!("open failed: {e:?}");
```
passaro, line 314: (subjective) nit: this is a lot of ceremony just for a log. I'd be tempted to use `map_err` or look for/add something like `on_err`.

passaro (approved):
LGTM

PR #311, opened 2023-06-23
reftests: always materialize reference model from bucket contents

jamesbornholt:
Today, the reference model for the reftests is just a tree structure. When we make changes to the expected state of the file system, we update this tree structure in place.

This needs to change to prepare us to extend the reftests with mutations to the *bucket*, not just the file system (i.e., concurrent mutations). When one of these mutations happens, we need to be able to compute what we expect the state of the file system to be after the change. For example, if my bucket contains keys `a` and `a/b`, and then I delete `a/b` from the bucket, what's the expected state of the file system? The current implementation asks us to compute this state by making in-place modifications to the file system tree, which is error prone -- we'd essentially be re-implementing the same inode management logic we use in the actual file system.

To avoid this complexity, this change instead "materializes" the file system structure from a reference bucket at every step of the reftest. The reference model is now just a "reference bucket" plus lists of local files and directories. In the example above, we can simulate deleting an object by just removing it from both the mock bucket (the one Mountpoint is accessing through the mock client) and the reference bucket (the one introduced in this change). Then, when we rematerialize the reference file system, we'll have the new directory structure, without having to manually compute the changes to make to that tree.

This is nice because it means the only "trusted" code in the reference model is `build_reference`, which specifies how to convert *any* bucket into a filesystem. One potential downside is that rematerializing the filesystem after every operation might be expensive, but in my testing the overhead hasn't seemed problematic.

This commit is not expected to change any semantics of the reftests, and doesn't add any new operations. Those will come later, building on this refactoring.

---

### DIFF
--- a/mountpoint-s3/tests/reftests/reference.rs
+++ b/mountpoint-s3/tests/reftests/reference.rs
@@ -1,4 +1,3 @@
-use fuser::FileType;
 use mountpoint_s3_client::mock_client::MockObject;
 use std::cell::RefCell;
 use std::collections::BTreeMap;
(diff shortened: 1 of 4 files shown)


mountpoint-s3/tests/reftests/reference.rs
```diff
@@ -15,62 +14,141 @@ pub enum File {
...
+                    children: new_children,
+                    is_local: new_is_local,
+                } = &new_node
+                {
+                    if let Some(Node::Directory {
+                        is_local: curr_is_local,
+                        ..
+                    }) = children.get_mut(dir)
+                    {
+                        assert!(new_children.is_empty(), "local directories are always empty");
```
passaro, line 94: Is there any advantage in passing in the new_node? I suspect just passing in the type (file vs dir) would make this method simpler.

passaro (approved):
Looks good. Just a couple of suggestions, but none blocking.

mountpoint-s3/tests/reftests/reference.rs
jamesbornholt: Yeah, you're right, no idea why I tried to do it this way. I added a new `NodeType` enum instead (could have just used `fuser::FileType` directory but it has extra variants we don't allow).

passaro (approved):

PR #313, opened 2023-06-23
Implement fsync and handle write errors

passaro:
Implement `fsync` to allow users to complete a put request and receive confirmation that it succeeded or failed. If a file handle is released without a call to `fsync`, `release` will still complete the upload as before. 

Wrap the `UploadRequest` in the file handle in a new `UploadState` enum, in order to detect:
* on `release`, whether the request had been already completed by an `fsync` call,
* `write` is invoked after an `fsync`,
* `write` (or `fsync`) is invoked after a previous call failed.

Also adds support for put failures to FailureClient.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -6,7 +6,7 @@ use std::collections::HashMap;
 use std::ffi::OsStr;
 use std::str::FromStr;
 use std::time::{Duration, UNIX_EPOCH};
-use tracing::{debug, error, trace};
+use tracing::{debug, error, trace, warn};
 
 use fuser::{FileAttr, KernelConfig};
 use mountpoint_s3_client::{ETag, ObjectClient};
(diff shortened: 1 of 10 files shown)


passaro:
Note that this PR replaces #308

mountpoint-s3/src/fs.rs
```diff
@@ -585,6 +589,37 @@ where
...
+                Some(handle) => handle.clone(),
+                None => return Err(libc::EBADF),
+            }
+        };
+        let mut request = match &file_handle.typ {
+            FileHandleType::Write(request) => request.lock().await,
+            FileHandleType::Read { .. } => return Ok(()),
+        };
+        let Some(in_progress_request) = request.take() else {
+            error!(key=?file_handle.full_key, "upload already completed or aborted");
```
jamesbornholt, line 605: I think we talked about tracking a little more state in the handle so we could distinguish these two cases. Is that worth doing?

jamesbornholt: Also, should be `warn!`

mountpoint-s3/src/fs.rs
```diff
@@ -611,26 +646,20 @@ where
...
-                    Ok(_) => {
-                        debug!(key, size, "put succeeded");
-                        Ok(())
-                    }
-                    Err(e) => {
-                        error!(key, size, "put failed, object was not uploaded: {e:?}");
-                        // This won't actually be seen by the user because `release` is async, but
-                        // it's the right thing to do.
-                        Err(libc::EIO)
+                if let Some(in_progress_request) = request {
```
jamesbornholt, line 651: This can just be
```rust
if let Some(request) = request.into_inner() {
    // ...
```

mountpoint-s3/src/fs.rs
```diff
@@ -611,26 +646,20 @@ where
...
+                    let key = &file_handle.full_key;
+                    let size = in_progress_request.size();
+                    match in_progress_request.complete().await {
+                        Ok(_) => debug!(key, size, "put succeeded"),
+                        Err(e) => {
+                            error!(key, size, "put failed, object was not uploaded: {e:?}");
```
jamesbornholt, line 657: We should keep the comment about the error not being user-visible.

mountpoint-s3/src/upload.rs
```diff
@@ -52,21 +53,23 @@ pub enum UploadWriteError<E: std::error::Error> {
 /// Manages the upload of an object to S3.
 ///
 /// Wraps a PutObject request and enforces sequential writes.
-pub struct UploadRequest<Client: ObjectClient> {
+pub struct UploadRequest<Client: ObjectClient, Handle> {
     bucket: String,
     key: String,
     next_request_offset: u64,
     request: Client::PutObjectRequest,
+    _handle: Handle,
```
jamesbornholt, line 61: I'm not sure the handle should be here -- it doesn't really have anything to do with S3 uploads (it's a filesystem concept). It should probably be back in `fs`. If that gets too messy, we should factor out a `WriteFileHandle` type and change the enum to `FileHandleType::Write(WriteFileHandle)` so that we can put the transition logic onto that struct instead.

So something like
```rust
struct WriteFileHandle {
    upload: UploadRequest<Client, WriteHandle>,
    token: WriteHandle,
}

impl WriteFileHandle {
    fn complete(mut self) -> Result<(), libc::c_int> {
        // ...the common logic that lives in both `fsync` and `release` right now...

        // and explicitly complete the write handle
        self.token.finish_writing()?;
    }
}
```

mountpoint-s3/src/inode.rs
```diff
@@ -846,6 +840,14 @@ impl WriteHandle {
...
 }
 
+impl Drop for WriteHandle {
+    fn drop(&mut self) {
+        if let Err(e) = self.finish_writing() {
+            error!(error=?e, ino=self.ino, "error closing file handle");
+        }
+    }
+}
+
```
jamesbornholt, line 850: See my other comment -- I don't like moving fallible stuff like this into `Drop` impls, it robs you of any chance to expose the error.

mountpoint-s3/src/fs.rs
passaro: I've added `UploadState` to keep track of failure vs completion.

mountpoint-s3/src/fs.rs
```diff
@@ -55,7 +55,91 @@ enum FileHandleType<Client: ObjectClient, Runtime> {
...
+
+    async fn complete_if_in_progress(self, key: &str) -> Result<(), libc::c_int> {
+        let upload = match self {
+            Self::InProgress(upload) => upload,
+            Self::Failed | Self::Completed => {
+                // Nothing to do.
+                return Ok(());
+            }
+        };
+        Self::complete_upload(upload, key).await
```
jamesbornholt, line 127: nit:
```suggestion
        match self {
            Self::InProgress(upload) => Self::complete_upload(upload, key).await,
            Self::Failed | Self::Completed => Ok(()),
        }
```

mountpoint-s3/src/fs.rs
```diff
@@ -55,7 +55,91 @@ enum FileHandleType<Client: ObjectClient, Runtime> {
...
+                    }
+                    UploadWriteError::OutOfOrderWrite { .. } => {}
+                };
+                Err(e.into())
+            }
+        }
+    }
+
+    async fn complete(&mut self, key: &str) -> Result<(), libc::c_int> {
+        let upload = match std::mem::replace(self, Self::Completed) {
```
jamesbornholt, line 101: This is kinda weird because it can replace `Self::Failed` with `Self::Completed`. I think that only messes up error messages in some edge cases, but probably we should try to get it right anyway.

jamesbornholt (approved):

PR #314, opened 2023-06-23
Fix dependency installation in GitHub workflows

dannycjones:
The action wasn't quite hooked up properly to the script, and the script wasn't parsing arguments properly if there were unexpected positional args. That's now fixed in the script too.

(I have no idea why the inputs aren't working properly as booleans. I'd rather we didn't have the `fromJSON(...)` but here we are.)

I've verified that this is now working in my fork: https://github.com/dannycjones/mountpoint-s3/actions/runs/5360459957

---


jamesbornholt (commented):
One day I'll convince people to write stuff like this in Python instead :-)

jamesbornholt (approved):

PR #320, opened 2023-06-28
Enable PUT checksums

passaro:
## Description of change

Enable Crc32c checksums when uploading objects. PUT requests now have an option to enable trailing checksums. 

N/A

---

### DIFF
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -247,7 +247,23 @@ pub enum GetObjectAttributesError {
 /// TODO: Populate this struct with parameters from the S3 API, e.g., storage class, encryption.
 #[derive(Debug, Default, Clone)]
 #[non_exhaustive]
-pub struct PutObjectParams {}
+pub struct PutObjectParams {
+    /// Enable Crc32c trailing checksums.
+    pub trailing_checksums: bool,
+}
+
+impl PutObjectParams {
+    /// Create a default [PutObjectParams].
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Set Crc32c trailing checksums.
+    pub fn trailing_checksums(mut self, value: bool) -> Self {
+        self.trailing_checksums = value;
+        self
+    }
+}
 
 /// A streaming put request which allows callers to asynchronously write
 /// the body of the request.
--- a/mountpoint-s3-client/tests/put_object.rs
+++ b/mountpoint-s3-client/tests/put_object.rs
@@ -2,12 +2,18 @@
 
 pub mod common;
 
+use base64ct::Base64;
+use base64ct::Encoding;
 use common::*;
 use futures::{pin_mut, StreamExt};
 use mountpoint_s3_client::GetObjectError;
 use mountpoint_s3_client::ObjectClient;
 use mountpoint_s3_client::ObjectClientResult;
+use mountpoint_s3_client::PutObjectParams;
 use mountpoint_s3_client::PutObjectRequest;
+use mountpoint_s3_client::S3ClientConfig;
+use mountpoint_s3_client::S3CrtClient;
+use mountpoint_s3_crt::checksums::crc32c;
 use rand::Rng;
 
 // Simple test for PUT object. Puts a single, small object as a single part and checks that the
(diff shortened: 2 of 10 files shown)


mountpoint-s3-client/tests/put_object.rs
```diff
@@ -184,3 +189,47 @@ async fn test_put_object_abort() {
     let uploads_in_progress = get_mpu_count_for_key(&sdk_client, &bucket, &key).await.unwrap();
     assert_eq!(uploads_in_progress, 0);
 }
+
+#[tokio::test]
+async fn test_put_checksums() {
+    const PART_SIZE: usize = 5 * 1024 * 1024;
+    let (bucket, prefix) = get_test_bucket_and_prefix("test_put_object_abort");
```
jamesbornholt, line 196:
```suggestion
    let (bucket, prefix) = get_test_bucket_and_prefix("test_put_checksums");
```

mountpoint-s3-client/src/object_client.rs
```diff
@@ -245,9 +245,19 @@ pub enum GetObjectAttributesError {
...
+    pub trailing_checksums: bool,
+}
+
+impl Default for PutObjectParams {
+    fn default() -> Self {
+        Self {
+            trailing_checksums: true,
+        }
+    }
+}
```
jamesbornholt, line 260: I don't think this should be on by default here -- as a generic S3 client it should be opt-in. But we should turn it on by default in `mountpoint_s3::Uploader`.

mountpoint-s3-client/src/object_client.rs
```diff
@@ -245,9 +245,19 @@ pub enum GetObjectAttributesError {
 
 /// Parameters to a [ObjectClient::put_object] request
 /// TODO: Populate this struct with parameters from the S3 API, e.g., storage class, encryption.
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Clone)]
 #[non_exhaustive]
-pub struct PutObjectParams {}
+pub struct PutObjectParams {
+    pub trailing_checksums: bool,
```
jamesbornholt, line 251: Needs a doc comment (specifically that it's crc32c).

jamesbornholt: Actually, maybe this should be `pub checksum_config: Option<ChecksumConfig>` and we re-export `mountpoint_s3_crt::ChecksumConfig` from this crate, so that we don't end up duplicating stuff if we add more options here.

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -829,3 +843,29 @@ pub fn init_default_signing_config(region: &str, credentials_provider: Credentia
...
+                location: aws_s3_checksum_location::AWS_SCL_TRAILER,
+                checksum_algorithm: aws_s3_checksum_algorithm::AWS_SCA_CRC32C,
+                validate_response_checksum: false,
+                validate_checksum_algorithms: std::ptr::null_mut(),
+            },
+        }
+    }
+
+    /// Get out the inner pointer to the checksum config
+    pub fn to_inner_ptr(&self) -> *const aws_s3_checksum_config {
```
jamesbornholt, line 868:
```suggestion
    pub(crate) fn to_inner_ptr(&self) -> *const aws_s3_checksum_config {
```

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -829,3 +843,29 @@ pub fn init_default_signing_config(region: &str, credentials_provider: Credentia
...
+
+impl ChecksumConfig {
+    /// Create a new [ChecksumConfig] with Crc32c traling checksums.
+    pub fn trailing_crc32c() -> Self {
+        Self {
+            inner: aws_s3_checksum_config {
+                location: aws_s3_checksum_location::AWS_SCL_TRAILER,
+                checksum_algorithm: aws_s3_checksum_algorithm::AWS_SCA_CRC32C,
+                validate_response_checksum: false,
+                validate_checksum_algorithms: std::ptr::null_mut(),
```
jamesbornholt, line 862: I think?
```suggestion
                location: aws_s3_checksum_location::AWS_SCL_TRAILER,
                checksum_algorithm: aws_s3_checksum_algorithm::AWS_SCA_CRC32C,
                ..Default::default()
```

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -829,3 +843,29 @@ pub fn init_default_signing_config(region: &str, credentials_provider: Credentia
...
+
+/// The checksum configuration.
+#[derive(Debug, Clone, Default)]
+pub struct ChecksumConfig {
+    /// The struct we can pass into the CRT's functions.
+    inner: aws_s3_checksum_config,
+}
+
+impl ChecksumConfig {
+    /// Create a new [ChecksumConfig] with Crc32c traling checksums.
```
jamesbornholt, line 855:
```suggestion
    /// Create a new [ChecksumConfig] with Crc32c trailing checksums.
```

jamesbornholt: Also, note that this is only for PUT. It's weird (but understandable) that the CRT combines checksums for PUT and GET into this one struct, but here we are.

mountpoint-s3-client/src/object_client.rs
passaro: I thought about that, but it's unclear to me whether we want to expose ChecksumConfig in the long run (vs something tailored to specific requests: PutObject, MPU, GetObject).
I'd rather keep it as a simple bool for now and think about a more complete API when/if we'll introduce more options.

dannycjones: I'd probably second having an enum/struct over boolean. Something like `Option<ChecksumAlgorithm>`, even if `ChecksumAlgorithm` will only be `CRC32C` today?

I'm thinking the API we're putting here matches up to this one in Java SDK: https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/model/PutObjectRequest.Builder.html#checksumAlgorithm(software.amazon.awssdk.services.s3.model.ChecksumAlgorithm)

We don't _have_ to re-export the CRT config struct, we could just use our own in `mountpoint-s3-client` if it makes more sense.

passaro: @dannycjones, we could start adding a `ChecksumAlgorithm` enum, but what about the location (trailer vs header)? That's part of `ChecksumConfig` but (I believe) only trailer is supported on multi-part uploads (which is what `put_object` uses). Do we want to limit the API to what a specific request allows, or just error on misconfiguration? Related: will we have a separate `put_object` for non-mpu? Or maybe that would also be configurable?

I don't think we **need** to answer any of these questions now, so my preference is for the simplest possible option, knowing we will replace it later.

monthonk: I probably prefer to keep it simple for now, so I'm ok with using boolean. Maybe we open a new issue to review this configuration before we publish a new version of mountpoint-s3-client? I think that's the point where we say there will be no breaking change after this and you can safely use this new config.

PR #321, opened 2023-06-29
Restrict log files access by default

monthonk:
## Description of change

Mountpoint log directories and files are currently created with operating system default permissions, which allows other users in the system to read log files. In this commit, we set the log directories and files to 750 and 640 respectively to limit access to the logs.

No, logging behavior is still the same, only log directories/files permissions changed.

---

### DIFF
--- a/mountpoint-s3/src/main.rs
+++ b/mountpoint-s3/src/main.rs
@@ -1,8 +1,10 @@
+use std::fs::File;
+use std::fs::{DirBuilder, OpenOptions};
 use std::io::{Read, Write};
-use std::os::unix::prelude::FromRawFd;
+use std::os::unix::fs::DirBuilderExt;
+use std::os::unix::prelude::{FromRawFd, OpenOptionsExt};
 use std::path::{Path, PathBuf};
 use std::time::Duration;
-use std::{fs, fs::File};
 
 use anyhow::{anyhow, Context as _};
 use clap::{value_parser, ArgGroup, Parser};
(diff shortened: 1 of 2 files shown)


mountpoint-s3/tests/cli.rs
```diff
@@ -170,3 +170,65 @@ fn invalid_profile() -> Result<(), Box<dyn std::error::Error>> {
...
 }
+
+#[test]
+fn validate_default_log_files_permissions() -> Result<(), Box<dyn std::error::Error>> {
+    let mut cmd = Command::cargo_bin("mount-s3")?;
+    let dir = assert_fs::TempDir::new()?;
+    let default_log_dir = home::home_dir().unwrap().join(".mountpoint-s3");
+
+    cmd.arg("test-bucket").arg(dir.path());
+    cmd.assert().failure();
```
dannycjones, line 181: Why failure?

I'm wondering if we're better using a known bucket like `doc-example-bucket` or the forbidden bucket?

dannycjones (commented):
Just one comment, otherwise LGTM

mountpoint-s3/tests/cli.rs
monthonk: Actually, we don't care what Mountpoint returns here as long as the logs are printed out to the file, so it could be either success or failure. In this case, it's just easier to make Mountpoint fail and not having to cleanup the mount after test.

monthonk: Also, the tests here are unit tests so we don't really want to mount to a real bucket, but I can change the bucket name to `doc-example-bucket` if we think that's better.

mountpoint-s3/src/main.rs
```diff
@@ -53,15 +55,28 @@ fn init_tracing_subscriber(is_foreground: bool, log_directory: Option<&Path>) ->
         .format(LOG_FILE_NAME_FORMAT)
         .context("couldn't format log file name")?;
 
+    // log directories and files should not be accessible by other users
+    let mut file_options = OpenOptions::new();
+    file_options.mode(0o640).write(true).create(true);
```
sauraank, line 60: Why are we allowing write access to log files?

monthonk: Only the owner would have write access, I think it makes sense, for example, they might want to move or delete the log files.

mountpoint-s3/tests/cli.rs
jamesbornholt: Yeah, let's use a bucket we know can't exist, which is `DOC-EXAMPLE-BUCKET`.

jamesbornholt (approved):

PR #328, opened 2023-06-29
Avoid a copy when optimizing part size for checksums

jamesbornholt:
## Description of change

The `chunks` iterator returns slices, so creating the `Bytes` for each
chunk needs to do a copy. We can instead just do the `Bytes` conversion
upfront from the `Box<[u8]>`, and then do O(1) splits of that `Bytes`
object for each chunk.

Before:

```
$ time dd if=$HOME/mnt/test100GiB-STD of=/dev/null bs=1M iflag=direct status=progress
107374182400 bytes (107 GB, 100 GiB) copied, 38.7547 s, 2.8 GB/s
```

After:
```
$ time dd if=$HOME/mnt/test100GiB-STD of=/dev/null bs=1M iflag=direct status=progress
107374182400 bytes (107 GB, 100 GiB) copied, 33.9263 s, 3.2 GB/s
```

No, this is an internal performance improvement.

---

### DIFF
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -301,14 +301,19 @@ where
                                 Some(Ok((offset, body))) => {
                                     // pre-split the body into multiple parts as suggested by preferred part size
                                     // in order to avoid validating checksum on large parts at read.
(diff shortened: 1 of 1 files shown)


monthonk (approved):
nice improvement!

PR #329, opened 2023-06-29
Enforce maximum upload size

### LINKED ISSUE #299: Report an error when objects become too big to upload
We can't upload an object of size larger than `part_size * 10000` to S3. Right now that failure only happens asynchronously when the CRT tries to actually upload the 10001th part, so the application has a hard time finding out about it. We should track the object size on our side and fail writes with `libc::EFBIG` when an object gets too big.

jamesbornholt:
This turned out to be a lot of refactoring to expose the client's part size and allow our FUSE tests to choose a part size. So it might be easier to read the three commits individually.

## Description of change

Uploads can fail if they require more than 10,000 parts. Right now we
only find out about that failure asynchronously when the CRT actually
constructs the 10,0001th part. This is bad because unlike other upload
errors, this one is a deterministic failure that we should be able to
report up front.

One trick here is that the kernel appears to retry failed writes from
page cache when returning EFBIG, so we need to remember the reason the
write failed.

Relevant issues: Fixes #299

Yes, uploads will fail earlier than they did before, and with a different error code.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -67,14 +67,75 @@ macro_rules! event {
     }
 }
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub struct S3ClientConfig {
-    pub auth_config: S3ClientAuthConfig,
-    pub throughput_target_gbps: Option<f64>,
-    pub part_size: Option<usize>,
-    pub endpoint: Option<Endpoint>,
-    pub user_agent_prefix: Option<String>,
-    pub request_payer: Option<String>,
+    auth_config: S3ClientAuthConfig,
+    throughput_target_gbps: f64,
+    part_size: usize,
+    endpoint: Option<Endpoint>,
+    user_agent_prefix: Option<String>,
+    request_payer: Option<String>,
+}
+
+impl Default for S3ClientConfig {
+    fn default() -> Self {
+        Self {
+            auth_config: Default::default(),
+            throughput_target_gbps: 10.0,
+            part_size: 8 * 1024 * 1024,
+            endpoint: None,
+            user_agent_prefix: None,
+            request_payer: None,
+        }
+    }
+}
+
+impl S3ClientConfig {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Set the configuration for authenticating to S3
+    #[must_use = "S3ClientConfig follows a builder pattern"]
+    pub fn auth_config(mut self, auth_config: S3ClientAuthConfig) -> Self {
+        self.auth_config = auth_config;
+        self
+    }
+
+    /// Set the part size for multi-part operations to S3 (both PUT and GET)
+    #[must_use = "S3ClientConfig follows a builder pattern"]
+    pub fn part_size(mut self, part_size: usize) -> Self {
+        self.part_size = part_size;
+        self
+    }
+
+    /// Set the target throughput in Gbps for the S3 client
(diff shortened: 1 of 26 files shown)


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -67,14 +67,75 @@ macro_rules! event {
...
+    user_agent_prefix: Option<String>,
+    request_payer: Option<String>,
+}
+
+impl Default for S3ClientConfig {
+    fn default() -> Self {
+        Self {
+            auth_config: Default::default(),
+            throughput_target_gbps: 10.0,
+            part_size: 8 * 1024 * 1024,
```
passaro, line 85: Default values are duplicated here and in mountpoint main. Not sure what to do about it, except maybe a comment here or there.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -662,6 +718,10 @@ impl ObjectClient for S3CrtClient {
     type PutObjectRequest = S3PutObjectRequest;
     type ClientError = S3RequestError;
 
+    fn part_size(&self) -> Option<usize> {
+        Some(self.inner.part_size)
```
passaro, line 722: Is this the actual part size or just what we set in ClientConfig? AFAIK the CRT will do some clamping.

mountpoint-s3/src/fs.rs
```diff
@@ -76,13 +77,14 @@ impl<Client: ObjectClient> UploadState<Client> {
             Err(e) => {
                 error!("write failed: {e}");
                 match e {
-                    UploadWriteError::PutRequestFailed(_) => {
+                    UploadWriteError::PutRequestFailed(_) | UploadWriteError::ObjectTooBig { .. } => {
```
passaro, line 80: Arguably, we could treat `ObjectTooBig` like `OutOfOrderWrite`: do not set the upload to `Failed` and allow for retries with a size that fits and/or subsequent fsync/release.
Is this a conscious choice not to do that?

I may also not fully understand the issue with kernel retries: wouldn't they also fail just with the size check in `UploadRequest::write`?

passaro: But I do like recording the error in `Failed` in general!

passaro (commented):
Just a few questions, mostly non-blocking. The only one that concerns me a bit is whether we are getting the actual part size from the CRT.

mountpoint-s3-client/src/s3_crt_client.rs
jamesbornholt: Yeah, the CRT doesn't let us interrogate the actual part size. But right now the only clamping it does is to the 5GiB part size limit, so I added the same clamping to our constructor.

mountpoint-s3/src/fs.rs
jamesbornholt: It was a conscious choice because I'm worried about uploading truncated objects. On reflection I think we should do the same thing for out-of-order writes -- we'd rather fail loudly in case applications don't actually check the results of `write`. Let's do that in a separate change.

> I may also not fully understand the issue with kernel retries: wouldn't they also fail just with the size check in UploadRequest::write?

No, because `UploadRequest::write` checks whether the upload has failed before calling `upload.write`, which is where the size check is.

passaro: > No, because `UploadRequest::write` checks whether the upload has failed before calling `upload.write`, which is where the size check is.

Sure, but it wouldn't if we treated it like `OutOfOrderWrite`. Anyway, happy to change the behavior of the latter.

passaro (approved):

PR #341, opened 2023-06-30
Abort upload on out-of-order writes

passaro:
## Description of change

Out-of-order writes previously failed but would not abort the upload and would even allow to resume writing sequentially.   

This change aborts the upload and mark the file handle has failed, so that subsequent operations (e.g. write, fsync) will return an error.

Yes, the behavior of write when trying to write out-of-order has changed.

---

### DIFF
--- a/mountpoint-s3/tests/fuse_tests/write_test.rs
+++ b/mountpoint-s3/tests/fuse_tests/write_test.rs
@@ -1,5 +1,5 @@
 use std::fs::{metadata, read, read_dir, File};
-use std::io::{ErrorKind, Read, Write};
+use std::io::{ErrorKind, Read, Seek, Write};
 use std::os::unix::prelude::OpenOptionsExt;
 use std::path::Path;
 use std::time::Duration;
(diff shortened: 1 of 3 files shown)


mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -328,3 +328,62 @@ where
...
+    assert_eq!(err.raw_os_error(), Some(libc::EINVAL));
+
+    drop(f);
+
+    // The kernel doesn't guarantee to flush the data as soon as the file is closed. Currently,
+    // the file won't be visible on the file system until it's flushed to S3, and so trying to stat
+    // the file will fail.
+    // TODO we can remove this when we implement fsync, or change it when we make files visible
+    // during writes
+    std::thread::sleep(Duration::from_secs(5));
```
monthonk, line 372: I think we can now use fsync?

passaro: Not really. A successful fsync guarantees that the file has been uploaded, but not currently that the inode has been updated.

jamesbornholt (approved):

PR #327, opened 2023-06-29
Turn write support on

passaro:
## Description of change

Enable write support by removing the `put` feature. Delete (`unlink`) is still not officially supported and only enabled under the `delete` feature.

Update SEMANTICS.md docs to reflect the change.

## Does this change impact existing behavior?

Write operations (not including delete) are now supported. Release version increased to `0.3.0`.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

PR #348, opened 2023-06-30
Explicitly prevent mounting over an existing mount

jamesbornholt:
## Description of change

It's valid to mount multiple file systems to the same directory, but
doing so requires the first one to be mounted read-write, which is why
this wasn't a problem for us until #327. It seems libfuse2's version of
fusermount explicitly checked this(?), but libfuse3 no longer rejects
it.

In principle this might be something we'd want to allow, but I think the
less surprising/error-prone customer experience is to refuse to do it,
so let's explicitly forbid it.

No.

---

### DIFF
--- a/mountpoint-s3/Cargo.toml
+++ b/mountpoint-s3/Cargo.toml
@@ -32,6 +32,9 @@ const_format = "0.2.30"
 home = "0.5.4"
 serde_json = "1.0.95"
 
+[target.'cfg(target_os = "linux")'.dependencies]
+procfs = { version = "0.15.1", default-features = false }
+
 [dev-dependencies]
 assert_cmd = "2.0.6"
 assert_fs = "1.0.9"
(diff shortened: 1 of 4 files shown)


mountpoint-s3/src/main.rs
```diff
@@ -251,14 +251,6 @@ impl CliArgs {
 fn main() -> anyhow::Result<()> {
     let args = CliArgs::parse();
 
-    // validate mount point
```
passaro, line 254: Are we okay moving this validation to the background process?

jamesbornholt: We should be -- it was the only validation we were doing in the foreground process previously, and we shouldn't be doing any work at all before the fork.

passaro: Makes sense, but we need to update the tests.

mountpoint-s3/Cargo.toml
```diff
@@ -31,6 +31,7 @@ time = { version = "0.3.17", features = ["macros", "formatting"] }
 const_format = "0.2.30"
 home = "0.5.4"
 serde_json = "1.0.95"
+procfs = { version = "0.15.1", default-features = false }
```
passaro, line 34: Isn't this platform-specific?

passaro (approved):

PR #355, opened 2023-07-05
Adds in support to configure storage class

ahmarsuhail:
## Description of change

Adds in new CLI option to configure storage class at mount time. Users can now do `mount --storage-class = "STANDARD"`, default storage class is set to intelligent tiering. 

No

---

### DIFF
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -99,6 +99,10 @@ impl MockClient {
     pub fn is_upload_in_progress(&self, key: &str) -> bool {
         self.in_progress_uploads.read().unwrap().contains(key)
     }
+
+    pub fn get_object_storage_class(&self, key: &str) -> String {
+        self.objects.read().unwrap().get(key).unwrap().storage_class.to_owned()
+    }
 }
 
 #[derive(Clone)]
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -99,10 +99,6 @@ impl MockClient {
     pub fn is_upload_in_progress(&self, key: &str) -> bool {
         self.in_progress_uploads.read().unwrap().contains(key)
     }
-
-    pub fn get_object_storage_class(&self, key: &str) -> String {
-        self.objects.read().unwrap().get(key).unwrap().storage_class.to_owned()
-    }
 }
 
 #[derive(Clone)]
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -255,7 +255,7 @@ pub struct PutObjectParams {
     /// Enable Crc32c trailing checksums.
     pub trailing_checksums: bool,
     /// Storage class to be used when creating new S3 object
-    pub storage_class: String,
+    pub storage_class: Option<String>,
 }
 
 impl PutObjectParams {
@@ -271,7 +271,7 @@ impl PutObjectParams {
     }
 
     /// Set the storage class.
-    pub fn storage_class(mut self, value: &str) -> Self {
+    pub fn storage_class(mut self, value: Option<String>) -> Self {
         self.storage_class = value.to_owned();
         self
     }
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -254,6 +254,8 @@ pub enum GetObjectAttributesError {
 pub struct PutObjectParams {
     /// Enable Crc32c trailing checksums.
     pub trailing_checksums: bool,
+    /// Storage class to be used when creating new S3 object
+    pub storage_class: String,
 }
 
 impl PutObjectParams {
@@ -265,6 +267,12 @@ impl PutObjectParams {
     /// Set Crc32c trailing checksums.
     pub fn trailing_checksums(mut self, value: bool) -> Self {
         self.trailing_checksums = value;
+        self
+    }
+
+    /// Set the storage class.
+    pub fn storage_class(mut self, value: &str) -> Self {
+        self.storage_class = value.to_owned();
         self
     }
 }
--- a/mountpoint-s3-client/tests/put_object.rs
+++ b/mountpoint-s3-client/tests/put_object.rs
@@ -18,6 +18,8 @@ use mountpoint_s3_crt::checksums::crc32c;
 use rand::Rng;
 use test_case::test_case;
 
+use test_case::test_case;
+
 // Simple test for PUT object. Puts a single, small object as a single part and checks that the
 // contents are correct with a GET.
 async fn test_put_object(client: &impl ObjectClient, bucket: &str, prefix: &str) {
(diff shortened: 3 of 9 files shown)


mountpoint-s3/src/main.rs
```diff
@@ -149,6 +149,9 @@ struct CliArgs {
     #[clap(long, help = "Set the 'x-amz-request-payer' to 'requester' on S3 requests", help_heading = BUCKET_OPTIONS_HEADER)]
     pub requester_pays: bool,
 
+    #[clap(long, help = "Set the storage class", help_heading = BUCKET_OPTIONS_HEADER)]
```
passaro, line 152: We should display the default here.

passaro (changes requested):
Looks good. It does need integration tests, e.g. default and non-default storage class, invalid values.

mountpoint-s3-client/src/object_client.rs
```diff
@@ -254,6 +254,8 @@ pub enum GetObjectAttributesError {
 pub struct PutObjectParams {
     /// Enable Crc32c trailing checksums.
     pub trailing_checksums: bool,
+    /// Storage class to be used when creating new S3 object
+    pub storage_class: String,
```
jamesbornholt, line 258: Does this work if you use `PutObjectParams::default()`, which will give an empty string? I don't know if the empty string is a valid argument to `x-amz-storage-class`. My guess is we probably want this to be `Option<String>` and only include the header when it's `Some`. For Mountpoint we'll always set it to `Some`, but other users of this `mountpoint-s3-client` crate might have other preferences or no preference.

mountpoint-s3/src/main.rs
jamesbornholt: And "for new objects" in the help

jamesbornholt (commented):
+1 on needing integration tests for other storage classes and for invalid ones. We don't need to test every storage class, just something other than the default `INTELLIGENT_TIERING`, plus `GLACIER` to represent the non-instant-access tiers.

mountpoint-s3/src/main.rs
dannycjones: In case you missed it (since we do set defaults in a different way for some arguments): we can provide the default to Clap directly, and then we document and set the default value in one place.

We do the same for `part_size`: https://github.com/awslabs/mountpoint-s3/blob/ed542e726de9232f8cfbdbeeb3a5a37988ada891/mountpoint-s3/src/main.rs#L243-L250

dannycjones:
> **Does this change impact existing behavior?**
> No

We'll be changing the default storage class for new objects.

We're happy with that, but worth calling out in the PR description.

mountpoint-s3/src/upload.rs
```diff
@@ -20,12 +20,16 @@ pub struct Uploader<Client> {
...
 struct UploaderInner<Client> {
     client: Arc<Client>,
+    storage_class: Option<String>,
 }
 
 impl<Client: ObjectClient> Uploader<Client> {
     /// Create a new [Uploader] that will make requests to the given client.
-    pub fn new(client: Arc<Client>) -> Self {
-        let inner = UploaderInner { client };
+    pub fn new(client: Arc<Client>, storage_class: Option<String>) -> Self {
```
jamesbornholt, line 28: This one should still be `storage_class: String` -- our Uploader will always set the storage class, it's just the `mountpoint-s3-client` crate that shouldn't force a storage class to be specified.

mountpoint-s3/src/main.rs
```diff
@@ -149,6 +149,9 @@ struct CliArgs {
     #[clap(long, help = "Set the 'x-amz-request-payer' to 'requester' on S3 requests", help_heading = BUCKET_OPTIONS_HEADER)]
     pub requester_pays: bool,
 
+    #[clap(long, help = "Set the storage class for new objects", help_heading = BUCKET_OPTIONS_HEADER)]
```
jamesbornholt, line 152: needs a default, and then it will be `pub storage_class: String`

mountpoint-s3/src/fs.rs
```diff
@@ -175,6 +175,8 @@ pub struct S3FilesystemConfig {
     pub file_mode: u16,
     /// Prefetcher configuration
     pub prefetcher_config: PrefetcherConfig,
+    /// Storage class to be used
+    pub storage_class: Option<String>,
```
jamesbornholt, line 179: just `String`

jamesbornholt (commented):
Just to be explicit rather than commenting everywhere: every mention of storage class in `mountpoint-s3` should be a `String`, and every mention of storage class in `mountpoint-s3-client` should be a `Option<String>`, except the setter method on `PutObjectParams` which should take a `String` as input (the assumption is that if you're calling that method you want to set a storage class).

mountpoint-s3-client/tests/put_object.rs
```diff
@@ -277,6 +279,65 @@ async fn test_put_review(pass_review: bool) {
     }
 }
 
+fn write_with_storage_class_test<F>(creator_fn: F, storage_class: &str)
```
passaro, line 282: nit: these are tests on the s3 client, so maybe pick a name like `test_put_with_storage_class`

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -435,6 +437,65 @@ where
...
+    let (mount_point, _session, test_client) = creator_fn("write_with_storage_class_test", config);
+
+    let path = mount_point.path().join(KEY);
+
+    let mut f = open_for_write(&path, false).unwrap();
+
+    let data = [0xaa; 16];
+    f.write_all(&data).unwrap();
+    f.sync_all().unwrap();
+    drop(f);
```
passaro, line 464: Could you use `write_file` here as well? Provided you add the `sync_all` call to it.

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -435,6 +437,65 @@ where
...
+        filesystem_config,
+        ..Default::default()
+    };
+    let (mount_point, _session, _test_client) = creator_fn("write_with_storage_class_test", config);
+
+    let path = mount_point.path().join(KEY);
+    write_file(path).expect_err("write with invalid storage class should fail");
+}
+
+fn write_file(path: PathBuf) -> std::io::Result<()> {
```
passaro, line 492: Prefer `path: impl AsRef<Path>` (look at `open_for_write`, for example). 
With `PathBuf` you are requiring callers to give you ownership of the buffer containing the path, while something like `&Path` would just borrow a reference to the path. This is similar to the difference between `String` and `&str`. Here `impl AsRef<Path>` is slightly more complicated, but it means a type that implements the `AsRef<Path>` trait: roughly something that can be converted into a `&Path`.

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -435,6 +437,65 @@ where
     assert_eq!(err.raw_os_error(), Some(libc::ENOENT));
 }
 
+fn write_with_storage_class_test<F>(creator_fn: F, storage_class: &str)
```
passaro, line 440: Move just before `write_with_storage_class_test_s3`

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -435,6 +437,65 @@ where
...
+    f.write_all(&data).unwrap();
+    f.sync_all().unwrap();
+    drop(f);
+
+    let s3_storage_class = test_client.get_storage_class(KEY).unwrap();
+
+    assert_eq!(s3_storage_class, storage_class);
+}
+
+fn write_with_invalid_storage_class_test<F>(creator_fn: F, storage_class: &str)
```
passaro, line 471: Move just before `write_with_invalid_storage_class_test_s3`

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -447,3 +508,15 @@ fn out_of_order_write_test_s3(offset: i64) {
...
+
+#[cfg(feature = "s3_tests")]
+#[test_case("INTELLIGENT_TIERING")]
+#[test_case("GLACIER")]
+fn write_with_storage_class_test_s3(storage_class: &str) {
+    write_with_storage_class_test(crate::fuse_tests::s3_session::new, storage_class);
+}
+
+#[test_case("INVALID_CLASS")]
+fn write_with_invalid_storage_class_test_s3(storage_class: &str) {
```
passaro, line 520: This also needs `#[cfg(feature = "s3_tests")]`

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -435,6 +437,65 @@ where
     assert_eq!(err.raw_os_error(), Some(libc::ENOENT));
 }
 
+fn write_with_storage_class_test<F>(creator_fn: F, storage_class: &str)
+where
+    F: FnOnce(&str, TestSessionConfig) -> (TempDir, BackgroundSession, TestClientBox),
+{
+    const KEY: &str = "new.txt";
+
+    let filesystem_config: S3FilesystemConfig = S3FilesystemConfig {
```
passaro, line 446: nit: I believe the type annotation (`: S3FilesystemConfig`) is redundant here.

passaro: super nit: this is only setting one field, I would probably inline it in `TestSessionConfig` (line 452)

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -99,6 +99,10 @@ impl MockClient {
     pub fn is_upload_in_progress(&self, key: &str) -> bool {
         self.in_progress_uploads.read().unwrap().contains(key)
     }
+
+    pub fn get_object_storage_class(&self, key: &str) -> String {
+        self.objects.read().unwrap().get(key).unwrap().storage_class.to_owned()
```
passaro, line 104: Don't we need to set the storage class on `MockObject`s?

mountpoint-s3-client/tests/put_object.rs
passaro, line 282: In fact, these don't belong here. We just want to test that `put_object` is correctly setting the storage class (or not). No need for the file system.

passaro (changes requested):

PR #356, opened 2023-07-05
Introduce release workflow.

eslrahc-swa:
We adopt two Github Action to automate release for Mounpoint-S3
* Github Action ID: [taiki-e/create-gh-release-action](https://github.com/taiki-e/create-gh-release-action) for create Github Release. * During the release, operator must have consensus with team on version, draft, prefix and title as workflow inputs. * This workflow can also take changelog as inputs, we can opt in that once we have a change log format ready.
* Github Action ID: [taiki-e/upload-rust-binary-action](https://github.com/taiki-e/upload-rust-binary-action) for uploading Mountpoint-S3 binary and all other artifacts. * This workflow uploads compiled binary on target platforms. When we do the release, artifacts must includes: bin, LICENSE, archive and checksum.

TODO: create cross compilation for EC2 AARCH64
## Description of change

Relevant issues: 

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,64 @@
+name: Release
+
+permissions:
+  contents: write
+
+on:
+  push:
+    tags:
+      - mountpoint-s3-[0-9]+.*
+      - mountpoint-s3-[a-z]+-[0-9]+.*
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: taiki-e/create-gh-release-action@v1
+      with:
+        prefix: mountpoint-s3(-[a-z]+)?
+        draft: true
+        # TODO: set it true after we have changlog template reviewed.
+        changelog: false
+        # TODO: set it false after we have changlog template reviewed.
+        allow-missing-changelog: true
+        branch: main
+        title: "Mountpoint for Amazon S3 v$version"
+        token: ${{ secrets.GITHUB_TOKEN }}
+
+  upload-assets:
+    name: Build and release on target ${{ matrix.runner.target}}
+    runs-on: ${{ matrix.runner.tags }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        runner:
+        - tags: [ubuntu-20.04] # GitHub-hosted
+          target: x86_64-unknown-linux-gnu
+        - tags: [self-hosted, linux, arm64] 
+          target: aarch64-unknown-linux-gnu
+
+    steps:
+    - name: Checkout source code
+      uses: actions/checkout@v3
+      with:
+        submodules: true
+    - name: Set up stable Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        override: true
+    - name: Install operating system dependencies
+      uses: ./.github/actions/install-dependencies
+      with:
+        fuseVersion: 2 
+    - name: Release mount-s3 binary
+      uses: taiki-e/upload-rust-binary-action@v1
+      with:
+        bin: mount-s3
+        token: ${{ secrets.GITHUB_TOKEN }}
+        checksum: sha512
+        asset: LICENSE
+
+
--- a/.github/actions/install-dependencies/install.sh
+++ b/.github/actions/install-dependencies/install.sh
@@ -159,6 +159,10 @@ if [[ $dry_run == false ]]; then
     case $os_release_id in
         amzn)
             sudo yum install -y $package_list
+            type -p yum-config-manager >/dev/null || sudo yum install yum-utils -y
+            sudo yum-config-manager --add-repo https://cli.github.com/packages/rpm/gh-cli.repo -y
+            sudo yum install gh -y
+            sudo yum update gh -y
             ;;
         ubuntu)
             sudo apt-get -q update


.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+
+on:
+  push:
+    tags:
+      - mountpoint-s3-[0-9]+.*
+      - mountpoint-s3-[a-z]+-[0-9]+.*
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
```
eslrahc-swa, line 14: Need to add self host EC2 instance on x86_64 and aarch_64.

eslrahc-swa:
per-release example: https://github.com/eslrahc-swa/mountpoint-s3/releases/tag/mountpoint-s3-0.0.1

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - uses: taiki-e/create-gh-release-action@v1
+      with:
```
jamesbornholt, line 18: think we also want `branch: main` here to force releases to only come from main

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+        fuseVersion: 2 
+    - name: Release mount-s3 binary
+      uses: taiki-e/upload-rust-binary-action@v1
+      with:
+        bin: mount-s3
+        token: ${{ secrets.GITHUB_TOKEN }}
+        asset: LICENSE,README.md
+        target:  ${{ matrix.target }}
+        archive: $bin-$tag-$target
```
jamesbornholt, line 55: probably don't need `$bin` here since the `tag` already starts with `mountpoint-s3-`

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+        token: ${{ secrets.GITHUB_TOKEN }}
+
+  upload-assets:
+    name: Build and release on target (${{ matrix.target }})
+    runs-on: ${{ matrix.os }}
+    
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
```
jamesbornholt, line 31: Let's not build on `ubuntu-latest`, because it changes when there's a new release. I think we want `ubuntu-20.04`, which is the oldest Ubuntu that GitHub Actions still offers.

Can you test that a binary built this way runs successfully on AL2?

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: Install operating system dependencies
+      uses: ./.github/actions/install-dependencies
+      with:
+        fuseVersion: 2 
+    - name: Release mount-s3 binary
+      uses: taiki-e/upload-rust-binary-action@v1
+      with:
+        bin: mount-s3
+        token: ${{ secrets.GITHUB_TOKEN }}
+        asset: LICENSE,README.md
```
jamesbornholt, line 53: I don't think we need these in the archive for a binary release. I also don't think we need the upload to be an archive at all — just the binary on its own should be fine — but I don't know if this action supports that.

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: taiki-e/create-gh-release-action@v1
+      with:
+        prefix: mountpoint-s3(-[a-z]+)?
+        draft: true
+        title: "$prefix $version"
```
jamesbornholt, line 21: Let's use a nicer title: "Mountpoint for Amazon S3 $version"

.github/workflows/release.yml
eslrahc-swa: will add

.github/workflows/release.yml
eslrahc-swa: changed

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
+name: Release
+
+permissions:
+  contents: write
+  id-token: write
```
monthonk, line 5: Do we need `id-token: write`? I don't see where we use OIDC token in the workflow.

.github/workflows/release.yml
eslrahc-swa: this workflow does not support release raw binaries.

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
...
+        fuseVersion: 2 
+    - name: Release mount-s3 binary
+      uses: taiki-e/upload-rust-binary-action@v1
+      with:
+        bin: mount-s3
+        token: ${{ secrets.GITHUB_TOKEN }}
+        checksum: sha512
+        asset: LICENSE
+
+
```
eslrahc-swa, line 64: release example: https://github.com/eslrahc-swa/mountpoint-s3/releases/tag/mountpoint-s3-0.0.1

client test:
```
# zyaoshen @ u40d82267bd7952 in ~/Downloads [13:03:11] 
$ ./mount-s3 jepsen-canary /tmp/mnt

# zyaoshen @ u40d82267bd7952 in ~/Downloads [13:03:22] 
$ cd /tmp/mnt 

# zyaoshen @ u40d82267bd7952 in /tmp/mnt [13:03:29] 
$ ls
0-src####                             16-src####                            20-src####                            27-src####  34-src####                            
```

monthonk: Where does the release message come from? I saw this from the example
```
mountpoint-s3-0.0.1

mountpoint-s3-0.0.1
```
Can we modify this text when doing a release?

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
...
+    - uses: taiki-e/create-gh-release-action@v1
+      with:
+        prefix: mountpoint-s3(-[a-z]+)?
+        draft: true
+        # TODO: set it true after we have changlog template reviewed.
+        changelog: false
+        # TODO: set it false after we have changlog template reviewed.
+        allow-missing-changelog: true
+        branch: main
+        title: "Mountpoint for Amazon S3 $version"
```
monthonk, line 26: Probably add `v` before the version number, so it will be like `Mountpoint for Amazon S3 v0.0.1`

.github/workflows/release.yml
eslrahc-swa: We can modify those before release, given the release is a draft, before we click the release button, team must double check all the artifacts is valid

monthonk (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
...
+    runs-on: ${{ matrix.runner.tags }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        runner:
+        - tags: [ubuntu-20.04] # GitHub-hosted
+          target: x86_64-unknown-linux-gnu
+        - tags: [self-hosted, linux, arm64] 
+          target: aarch64-unknown-linux-gnu
```
jamesbornholt, line 40: these need to build on the same OS

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
...
+    name: Build and release on target ${{ matrix.runner.target}}
+    runs-on: ${{ matrix.runner.tags }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        runner:
+        - tags: [ubuntu-20.04] # GitHub-hosted
+          target: x86_64-unknown-linux-gnu
+        - tags: [self-hosted, linux, arm64]
```
jamesbornholt, line 39: we need a different tag for release build hosts, which we want to be running AL2 even if we update benchmarking et al to AL2023 later.

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
+name: Release
+
+permissions:
+  contents: write
+
+on:
+  push:
+    tags:
+      - mountpoint-s3-[0-9]+.*
+      - mountpoint-s3-[a-z]+-[0-9]+.*
```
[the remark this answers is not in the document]
eslrahc-swa: Nice catch, I think what we need is `mountpoint-s3` only

.github/workflows/release.yml
eslrahc-swa: I think the tag here is confusing, I will modified that to AMI, given terraform config we have today is `amzn2-ami-kernel-5.10-hvm-*-arm64-gp2`

Issue #277, opened 2023-06-05
Implement full lifecycle of inodes (`forget`)


dannycjones:
When we see a new file in S3, we create a new `Inode` and maintain a list of these.

Today, the client does not implement FUSE `forget` and so we never remove inodes when the Kernel no longer cares for them.

We should implement this part of the lifecycle.

dannycjones:
Incomplete changes live here: #221

PR #359, opened 2023-07-06
Implement forget

passaro:
## Description of change

Implement forget and maintain a lookup count on inodes. Ensure that only inodes with non-zero lookup count are registered with the superblock, and thus can be looked up by inode number.

Relevant issues: #277

Internal change to prevent memory leaking.

---

### DIFF
--- a/mountpoint-s3/src/fuse.rs
+++ b/mountpoint-s3/src/fuse.rs
@@ -152,7 +152,7 @@ where
 
         let replier = ReplyDirectory { inner: &mut reply };
 
-        match block_on(self.fs.readdir(parent, fh, offset, false, replier).in_current_span()) {
+        match block_on(self.fs.readdir(parent, fh, offset, replier).in_current_span()) {
             Ok(_) => reply.ok(),
             Err(e) => reply.error(e),
         }
@@ -187,7 +187,7 @@ where
 
         let replier = ReplyDirectoryPlus { inner: &mut reply };
 
-        match block_on(self.fs.readdir(parent, fh, offset, true, replier).in_current_span()) {
+        match block_on(self.fs.readdirplus(parent, fh, offset, replier).in_current_span()) {
             Ok(_) => reply.ok(),
             Err(e) => reply.error(e),
         }
--- a/mountpoint-s3/src/fuse.rs
+++ b/mountpoint-s3/src/fuse.rs
@@ -59,6 +59,11 @@ where
         }
     }
 
+    #[instrument(level="debug", skip_all, fields(req=_req.unique(), ino, nlookup))]
+    fn forget(&self, _req: &Request<'_>, ino: u64, nlookup: u64) {
+        block_on(self.fs.forget(ino, nlookup));
+    }
+
     #[instrument(level="debug", skip_all, fields(req=_req.unique(), ino=ino))]
     fn open(&self, _req: &Request<'_>, ino: InodeNo, flags: i32, reply: ReplyOpen) {
         match block_on(self.fs.open(ino, flags).in_current_span()) {
@@ -147,7 +152,7 @@ where
 
         let replier = ReplyDirectory { inner: &mut reply };
 
-        match block_on(self.fs.readdir(parent, fh, offset, replier).in_current_span()) {
+        match block_on(self.fs.readdir(parent, fh, offset, false, replier).in_current_span()) {
             Ok(_) => reply.ok(),
             Err(e) => reply.error(e),
         }
@@ -182,7 +187,7 @@ where
 
         let replier = ReplyDirectoryPlus { inner: &mut reply };
 
-        match block_on(self.fs.readdir(parent, fh, offset, replier).in_current_span()) {
+        match block_on(self.fs.readdir(parent, fh, offset, true, replier).in_current_span()) {
             Ok(_) => reply.ok(),
             Err(e) => reply.error(e),
         }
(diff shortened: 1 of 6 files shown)


mountpoint-s3/src/inode/readdir.rs
```diff
@@ -65,33 +65,31 @@ impl ReaddirHandle {
...
-        let local_entries = {
+        let cached_entries = {
             let inode = inner.get(dir_ino)?;
             let kind_data = &inode.get_inode_state()?.kind_data;
-            let local_files = match kind_data {
+            let cached_entries = match kind_data {
                 InodeKindData::File { .. } => return Err(InodeError::NotADirectory(dir_ino)),
-                InodeKindData::Directory { writing_children, .. } => writing_children.iter().map(|ino| {
-                    let inode = inner.get(*ino)?;
+                InodeKindData::Directory { children, .. } => children.values().map(|inode| {
```
jamesbornholt, line 73: This seems problematic for large directories. Before we only needed to look at the children being written to, but now we need to look at every child before we can start streaming. So the first `ls` on a directory will be fast, but the second one will be slow.

jamesbornholt (commented):
Let's split this into two changes -- implement `forget`, and then separately handle cleaning up directory entries that don't exist remotely any more. For the second one, we need to think harder about how to make it work without listing/copying every cached child upfront, of which there could be millions/billions.

mountpoint-s3/src/fuse.rs
```diff
@@ -59,6 +59,11 @@ where
         }
     }
 
+    #[instrument(level="debug", skip_all, fields(req=_req.unique(), ino, nlookup))]
+    fn forget(&self, _req: &Request<'_>, ino: u64, nlookup: u64) {
```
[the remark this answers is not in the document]
passaro: This is already the default implementation in fuser:
```
    /// Like forget, but take multiple forget requests at once for performance. The default
    /// implementation will fallback to forget.
    #[cfg(feature = "abi-7-16")]
    fn batch_forget(&self, req: &Request<'_>, nodes: &[fuse_forget_one]) {
        for node in nodes {
            self.forget(req, node.nodeid, node.nlookup);
        }
    }
```

mountpoint-s3/src/fuse.rs
```diff
@@ -182,7 +187,7 @@ where
 
         let replier = ReplyDirectoryPlus { inner: &mut reply };
 
-        match block_on(self.fs.readdir(parent, fh, offset, replier).in_current_span()) {
+        match block_on(self.fs.readdir(parent, fh, offset, true, replier).in_current_span()) {
```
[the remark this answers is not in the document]
passaro: I agree, the flag is not great. But I cannot immediately see how else to avoid too much duplication between `readdir` and `readdirplus`, besides pushing the flag down to a `readdir_common` they both call. Suggestions welcome... or I'll give it another go tomorrow.

mountpoint-s3/src/inode/readdir.rs
passaro: Okay, removed this part from this PR.

mountpoint-s3/src/fuse.rs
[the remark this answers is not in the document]
jamesbornholt: I think pushing the flag down into a common readdir in `fuse.rs` is fine. The fact that we were reusing `readdir` here is really just laziness -- in reality, `fs.rs` needs to know whether it's readdir or readdirplus, so it should just have both.

mountpoint-s3/src/inode/readdir.rs
```diff
@@ -123,6 +127,9 @@ impl ReaddirHandle {
                     warn!("{} has an invalid name and will be unavailable", next.description());
                 } else {
                     let lookup = self.instantiate_remote_inode(next)?;
+                    if inc_lookup_count {
+                        self.inner.remember(&lookup.inode);
+                    }
```
jamesbornholt, line 132: So this means non-readdirplus inodes just get deallocated immediately and never get added to the `inodes` dict, right? Probably worth a comment.

mountpoint-s3/src/inode.rs
```diff
@@ -529,6 +417,168 @@ impl SuperblockInner {
...
+    pub fn remember(&self, inode: &Inode) -> u64 {
+        let lookup_count = inode.inc_lookup_count();
+        if lookup_count == 1 {
+            let previous = self.inodes.write().unwrap().insert(inode.ino(), inode.clone());
+            assert!(previous.is_none(), "inode numbers are never reused");
+        }
+        lookup_count
+    }
+
+    /// Lookup an inode in the parent directory with the given name.
```
jamesbornholt, line 431: Worth a comment that the caller is responsible for calling `remember` if the inode is expected to survive beyond the lookup.

mountpoint-s3/src/inode.rs
```diff
@@ -1211,6 +1284,84 @@ mod tests {
...
+        };
+        assert_eq!(lookup_count, 0, "lookup should have been reduced");
+        assert!(
+            superblock.inner.inodes.read().unwrap().get(&ino).is_none(),
+            "inode should not be present in superblock"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_forget_can_remove_inodes() {
```
jamesbornholt, line 1333: Maybe also a test that `readdir` (the non-plus version) inodes are never added to the superblock?

jamesbornholt (approved):

PR #362, opened 2023-07-06
Add some tests for UTF-8 filenames

jamesbornholt:
## Description of change

Linux filesystems allow UTF-8 in filenames, and S3 allows UTF-8 in keys,
so we expect this to work. It does! But it did expose a bug in our mock
client's ListObjects implementation, which was dealing with bytes
instead of characters (unlike the real S3).

No -- test code only.

---


mountpoint-s3-client/src/mock_client.rs
```diff
@@ -734,6 +733,48 @@ mod tests {
...
+            bucket: "test_bucket".to_string(),
+            part_size: 1024,
+        });
+
+        let mut keys = vec![];
+        for i in 0..5 {
+            keys.push(format!("dirs/こんにちは/file{i}.txt"));
+        }
+        for i in 0..5 {
+            keys.push(format!("dirs/😄🥹/file{i}.txt"));
```
monthonk, line 748: We probably want to test with negative case here? for example, if we change the second prefix to `dirs/😄🥹😮` the test would panick without your code change.

jamesbornholt: Yep, good call, added a couple of negative tests.

monthonk (approved):

PR #363, opened 2023-07-07
Access point test

sauraank:
## Description of change
Added test for Multi-region access point(needs an endpoint-url for now), Object Lambda Access Point, Single Region access point as alias. Also added test for Transfer acceleration.

Just added integration tests.
Relevant issues: 
#4 
No, this does impact existing behaviour. But, in order for these tests to pass we need to add the resources like accesspoints to CI.

No, there is no breaking change.
---

### DIFF
--- a/mountpoint-s3-client/tests/common/mod.rs
+++ b/mountpoint-s3-client/tests/common/mod.rs
@@ -53,6 +53,10 @@ pub fn get_test_domain() -> String {
     std::env::var("S3_DOMAIN").unwrap_or(String::from("amazonaws.com"))
 }
 
+pub fn get_subsession_iam_role() -> String {
+    std::env::var("S3_SUBSESSION_IAM_ROLE").expect("Set S3_SUBSESSION_IAM_ROLE to run integration tests")
+}
+
 pub async fn get_test_sdk_client() -> s3::Client {
     let config = aws_config::from_env()
         .region(Region::new(get_test_region()))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1979,6 +1979,7 @@ dependencies = [
  "aws-config",
  "aws-credential-types",
  "aws-sdk-s3",
+ "aws-sdk-sts",
  "base64ct",
  "built",
  "bytes",
--- a/mountpoint-s3-client/Cargo.toml
+++ b/mountpoint-s3-client/Cargo.toml
@@ -39,6 +39,7 @@ anyhow = { version = "1.0.64", features = ["backtrace"] }
 aws-config = "0.54.1"
 aws-credential-types = "0.54.1"
 aws-sdk-s3 = "0.24.0"
+aws-sdk-sts = "0.24.0"
 bytes = "1.2.1"
 clap = "4.1.9"
 ctor = "0.1.23"
--- a/mountpoint-s3-client/tests/auth.rs
+++ b/mountpoint-s3-client/tests/auth.rs
@@ -12,8 +12,9 @@ use aws_sdk_s3::Region;
 use bytes::Bytes;
 use common::*;
 use futures::StreamExt;
-use mountpoint_s3_client::{EndpointConfig, ObjectClient, S3CrtClient};
-use mountpoint_s3_client::{S3ClientAuthConfig, S3ClientConfig};
+use mountpoint_s3_client::{
+    EndpointConfig, ObjectClient, ObjectClientError, S3ClientAuthConfig, S3ClientConfig, S3CrtClient, S3RequestError,
+};
 use mountpoint_s3_crt::auth::credentials::{CredentialsProvider, CredentialsProviderStaticOptions};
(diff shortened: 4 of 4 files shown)


mountpoint-s3-client/tests/endpoint.rs
```diff
@@ -88,3 +88,88 @@ async fn test_addressing_style_uri_fips_dualstack(addressing_style: AddressingSt
...
+        let uri = "https://s3-accelerate.amazonaws.com".to_string();
+        Endpoint::from_uri(&uri, addressing_style).unwrap()
+    })
+    .await;
+}
+
+async fn run_access_points<F: FnOnce(&str) -> Endpoint>(f: F) {
+    let sdk_client = get_test_sdk_client().await;
+    let (access_point, prefix) =
+        get_test_access_point_alias_and_prefix("test_access_point", AccessPointType::SingleRegion);
```
passaro, line 106: Is this the only difference with `run_test`?  If so, it may be worth to refactor to merge them.

mountpoint-s3-client/tests/endpoint.rs
```diff
@@ -88,3 +88,88 @@ async fn test_addressing_style_uri_fips_dualstack(addressing_style: AddressingSt
...
+
+#[test_case(AddressingStyle::Automatic)]
+#[test_case(AddressingStyle::Virtual)]
+#[test_case(AddressingStyle::Path)]
+#[tokio::test]
+async fn test_single_region_access_point_alias(addressing_style: AddressingStyle) {
+    run_access_points(|region| Endpoint::from_region(region, addressing_style).unwrap()).await;
+}
+
+async fn run_other_access_points<F: FnOnce(&str) -> Endpoint>(f: F, access_point_type: AccessPointType) {
```
passaro, line 140: Why is the actual test different from the one in `run_access_points` (besides `access_point_type`)?

mountpoint-s3-client/tests/common/mod.rs
```diff
@@ -52,6 +58,32 @@ pub async fn get_test_sdk_client() -> s3::Client {
...
+            std::env::var("S3_MRAP_ALIAS").expect("Set S3_MRAP_ALIAS to run integration tests")
+        }
+    };
+
+    // Generate a random nonce to make sure this prefix is truly unique
+    let nonce = OsRng.next_u64();
+
+    // Prefix always has a trailing "/" to keep meaning in sync with the S3 API.
+    let prefix =
+        std::env::var("S3_BUCKET_TEST_PREFIX").expect("Set S3_ACCESS_POINT_TEST_PREFIX to run integration tests");
```
passaro, line 79: `S3_BUCKET_TEST_PREFIX` or `S3_ACCESS_POINT_TEST_PREFIX`?

passaro: Depending on whether we just want `S3_BUCKET_TEST_PREFIX` here, it may be worth to just split out the prefix part from `get_test_bucket_and_prefix` and use it everywhere.

mountpoint-s3-client/tests/endpoint.rs
```diff
@@ -88,3 +88,88 @@ async fn test_addressing_style_uri_fips_dualstack(addressing_style: AddressingSt
...
+    let endpoint = f(&region);
+    let config = S3ClientConfig::new().endpoint(endpoint);
+    let client = S3CrtClient::new(&region, config).expect("could not create test client");
+    client
+        .list_objects(&access_point, None, "/", 10, &prefix)
+        .await
+        .expect("list_object should succeed");
+}
+
+#[test_case(AddressingStyle::Virtual)]
```
passaro, line 153: If not all combinations are supported (here and below), maybe it's worth a comment.

Here, in particular, do we need `test_case`?

passaro (changes requested):

mountpoint-s3-client/tests/endpoint.rs
sauraank: I have added test case here because it can be more clear that it is only working with Virtual addressing style. Also, for multi-region access point and object lambda access point path style addressing of endpoint does not work. I tried looking up the documentation but could not find any particular reason. I can dive deep into it further.

mountpoint-s3-client/tests/endpoint.rs
sauraank: Because, for Object Labda Access Point, Lambda function needs to be configured to add API support for Put Object. For multi region access points, Rust SDK is not supported. I have added comments for it.

passaro: typo: "Object Lambda"

passaro: maybe rename to `run_list_objects_test`, which is what this method does. The comment explains why.

passaro (changes requested):

PR #367, opened 2023-07-12
Compute checksums on write and compare with trailing checksums

passaro:
## Description of change

Maintain a running checksum of the data received from write requests by computing a crc32c checksum on the slice received from fuse and combining it with the previous checksum. Before an upload completes, the checksum is compared with the trailing checksums of all the uploaded parts. If they do not match, the upload is aborted.

Notes for reviewers:
* It may be useful to review the commits individually. Happy to split into separate PRs if preferred - but let's first discuss the overall approach.
* I've added a few methods to the `Crc32c` type in `mountpoint-s3-crt` (and related dependencies), but I'm not sure it is the right place. Maybe we should move it to `mountpoint-s3-client`? Or even ``mountpoint-s3`?

* `mountpoint-s3-client`: only adds new method `PutObjectRequest::review_and_complete`,
* `mountpoint-s3`: performs an additional checksum calculation on `write`s.

---


mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -868,3 +907,93 @@ impl ChecksumConfig {
...
+    }
+}
+
+/// Info for the caller to review before an upload completes.
+#[derive(Debug)]
+pub struct UploadReview {
+    /// Info about each part uploaded.
+    pub parts: Vec<UploadPartReview>,
+    /// The checksum algorithm used.
+    pub checksum_algorithm: ChecksumAlgorithm,
```
jamesbornholt, line 957: Maybe slightly more idiomatic if this is `Option<ChecksumAlgorithm>` rather than literally copying the `None` variant?

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -868,3 +907,93 @@ impl ChecksumConfig {
...
+        }
+    }
+
+    /// Failure error code.
+    pub const FAILURE: i32 = aws_s3_errors::AWS_ERROR_S3_CANCELED as i32;
+}
+
+/// Info about a single part, for the caller to review before the upload completes.
+#[derive(Debug)]
+pub struct UploadPartReview {
```
jamesbornholt, line 982: nit: I feel like the CRT name is backwards because they're trying to get namespacing, but `UploadReviewPart` makes more sense to me as a name

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -868,3 +907,93 @@ impl ChecksumConfig {
...
+    /// Size in bytes of this part.
+    pub size: u64,
+
+    /// Checksum string (usually base64-encoded). This is empty if no checksum is used.
+    pub checksum: String,
+}
+
+impl UploadPartReview {
+    fn new(part: &aws_s3_upload_part_review) -> Self {
+        // SAFETY: `part` is a valid aws_byte_cursor. The returned slice is only used in current scope.
```
jamesbornholt, line 992:
```suggestion
        // SAFETY: `part.checksum` is a valid aws_byte_cursor. The returned slice is only used in current scope.
```

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -868,3 +907,93 @@ impl ChecksumConfig {
...
+        let slice = unsafe { aws_byte_cursor_as_slice(&part.checksum) };
+        let str = core::str::from_utf8(slice).expect("Checksum should be a valid UTF-8 string.");
```
jamesbornholt, line 994:
```suggestion
        let str = std::str::from_utf8(slice).expect("checksum should be a valid UTF-8 string");
```

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -868,3 +907,93 @@ impl ChecksumConfig {
...
+}
+
+/// Info about a single part, for the caller to review before the upload completes.
+#[derive(Debug)]
+pub struct UploadPartReview {
+    /// Size in bytes of this part.
+    pub size: u64,
+
+    /// Checksum string (usually base64-encoded). This is empty if no checksum is used.
+    pub checksum: String,
```
jamesbornholt, line 987: Again should probably be `Option<String>` rather than exposing the "empty byte cursor" logic

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -442,6 +461,26 @@ unsafe extern "C" fn meta_request_shutdown_callback(user_data: *mut libc::c_void
...
+    _request: *mut aws_s3_meta_request,
+    upload_review: *const aws_s3_upload_review,
+    user_data: *mut libc::c_void,
+) -> i32 {
+    // SAFETY: user_data always will be a MetaRequestOptionsInner since that's what we set it to
+    // in MetaRequestOptions::new.
+    let user_data = MetaRequestOptionsInner::from_user_data_ptr(user_data);
+
+    if let Some(callback) = user_data.on_upload_review.take() {
+        return match callback(UploadReview::new(&*upload_review)) {
```
jamesbornholt, line 475: would rather we do the dereference here explicitly, and sanity check it first, so the safety argument is obvious:
```rust
assert!(!upload_review.is_null(), "CRT should provide a valid upload_review");
let upload_review = &*upload_review;
```

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -442,6 +461,26 @@ unsafe extern "C" fn meta_request_shutdown_callback(user_data: *mut libc::c_void
...
+unsafe extern "C" fn meta_request_upload_review_callback(
+    _request: *mut aws_s3_meta_request,
+    upload_review: *const aws_s3_upload_review,
+    user_data: *mut libc::c_void,
+) -> i32 {
+    // SAFETY: user_data always will be a MetaRequestOptionsInner since that's what we set it to
+    // in MetaRequestOptions::new.
+    let user_data = MetaRequestOptionsInner::from_user_data_ptr(user_data);
+
+    if let Some(callback) = user_data.on_upload_review.take() {
```
jamesbornholt, line 474: The `return match` below confused me (I didn't notice we were returning the value of `aws_raise_error(err)`). Maybe write this as
```rust
let Some(callback) = user_data.on_upload_review.take() else {
    return AWS_OP_SUCCESS;
};
```
so the later control flow is clearer

mountpoint-s3/src/fs.rs
```diff
@@ -536,17 +537,18 @@ where
         ino: InodeNo,
         fh: u64,
         offset: i64,
-        data: &[u8],
+        data: ChecksummedSlice<'_>,
```
jamesbornholt, line 540: I know we're doing it this way because we want to "do checksums as early as possible", but I think we might want to push the checksum calculation down into the `Uploader`. It's a very small difference in time (a couple of lock acquisitions?) and therefore risk, and moving it will make it much easier to disable checksums or allow different checksum algorithms in the future. I think moving it also saves you the need for `ChecksummedSlice` at all -- you can just maintain a running `Hasher`.

In the future you could imagine actually doing this in `fuser` instead, as the bytes are being read off the FUSE device, and then the entire interface would have this checksummed type.

mountpoint-s3-client/tests/put_object.rs
```diff
@@ -227,7 +230,48 @@ async fn test_put_checksums() {
...
+            pass_review
+        })
+        .await;
+
+    if pass_review {
+        _ = result.expect("putobject should succeed when review passes");
+    } else {
+        let err = result.expect_err("putobject should abort when review fails");
+        assert!(
+            matches!(err, ObjectClientError::ClientError(S3RequestError::ResponseError(MetaRequestResult { crt_error, .. })) if crt_error.raw_error() == UploadReview::FAILURE)
```
jamesbornholt, line 274: Checking the error code is probably overkill, I think. And then we don't have to expose this weird `FAILURE` constant.

mountpoint-s3-client/tests/put_object.rs
```diff
@@ -227,7 +230,48 @@ async fn test_put_checksums() {
...
-        let encoded = Base64::encode_string(&expected_checksum.value().to_be_bytes());
+        let encoded = expected_checksum.to_base64();
         assert_eq!(checksum, encoded);
     }
 }
+
+#[test_case(true; "pass review")]
+#[test_case(false; "fail review")]
+#[tokio::test]
+async fn test_put_review(pass_review: bool) {
```
jamesbornholt, line 241: This test should also check that `get_object` works if `true` and doesn't if `false`? And perhaps that the MPU was in fact aborted in the false case.

mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -17,117 +16,114 @@ impl S3CrtClient {
...
+            writer,
+            review_callback,
+        })
     }
 }
 
+type ReviewCallback = dyn FnOnce(UploadReview) -> bool + Send;
+
+#[derive(Clone, Default)]
+struct ReviewCallbackBox {
```
jamesbornholt, line 63: Probably deserves a brief comment about why this exists (because the callback is provided at completion time, but its invocation is async).

mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -17,117 +16,114 @@ impl S3CrtClient {
...
+
+impl std::fmt::Debug for ReviewCallbackBox {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ReviewCallbackBox").finish()
+    }
+}
+
+impl ReviewCallbackBox {
+    fn set(&mut self, callback: impl FnOnce(UploadReview) -> bool + Send + 'static) {
+        let _ = self.callback.lock().unwrap().insert(Box::new(callback));
```
jamesbornholt, line 75: Should probably assert the old one was None?

mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -17,117 +16,114 @@ impl S3CrtClient {
...
+    }
+
+    fn invoke(self, review: UploadReview) -> Result<(), i32> {
+        let mut callback = self.callback.lock().unwrap();
+        if let Some(callback) = callback.take() {
```
jamesbornholt, line 80: Somewhere it would be a nice sanity check to make sure this callback is invoked at most once, and maybe return `Err` on a second call. I don't know if the right place to do that check is here or in the CRT binding -- probably in the binding since what we're really checking is that the CRT is holding up its end of the contract?

mountpoint-s3/src/upload.rs
```diff
@@ -119,10 +130,31 @@ impl<Client: ObjectClient> Debug for UploadRequest<Client> {
...
+            Err(error) => {
+                error!(?error, "error decoding part checksum");
+                return false;
+            }
+        };
+
+        combined_checksum = combined_checksum.combine(checksum, part.size as usize);
+    }
+
+    total_size == expected_size && combined_checksum == expected_checksum
```
jamesbornholt, line 155: We should log if this fails and why.

mountpoint-s3/src/upload.rs
```diff
@@ -119,16 +128,66 @@ impl<Client: ObjectClient> Debug for UploadRequest<Client> {
...
+        );
+        return false;
+    }
+
+    true
+}
+
+/// Calculates the combined checksum for `AB` where `prefix_crc` is the checksum for `A`,
+/// `suffix_crc` is the checksum for `B`, and `suffic_len` is the length of `B`.
+pub fn combine_checksums(prefix_crc: Crc32c, suffix_crc: Crc32c, suffix_len: usize) -> Crc32c {
```
jamesbornholt, line 179: doesn't need to be pub?

mountpoint-s3-client/tests/put_object.rs
```diff
@@ -227,7 +217,75 @@ async fn test_put_checksums() {
...
+        ));
+
+        let err = get_result.expect_err("getobject should fail for aborted put");
+        assert!(matches!(
+            err,
+            ObjectClientError::ServiceError(GetObjectError::NoSuchKey)
+        ));
+
+        // Allow for the AbortMultipartUpload to complete.
+        tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
```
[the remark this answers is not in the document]
passaro: Well spotted. It does wait in both cases, so this is not needed here. I had copied it from `test_put_object_abort`, where it's needed after `drop`.

mountpoint-s3-client/tests/put_object.rs
```diff
@@ -227,7 +217,75 @@ async fn test_put_checksums() {
...
         assert_eq!(checksum, encoded);
     }
 }
+
+#[test_case(true; "pass review")]
+#[test_case(false; "fail review")]
+#[tokio::test]
+async fn test_put_review(pass_review: bool) {
+    const PART_SIZE: usize = 5 * 1024 * 1024;
+    let (bucket, prefix) = get_test_bucket_and_prefix("test_put_checksums");
```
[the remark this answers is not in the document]
passaro: Sorry... I keep missing these :(

mountpoint-s3-client/src/checksums/crc32.rs
```diff
@@ -1,4 +1,4 @@
-use mountpoint_s3_crt_sys::aws_checksums_crc32;
```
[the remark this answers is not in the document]
passaro: Another downside is that we go back to use the CRT crate directly from mountpoint, which I thought we would like to avoid (I know, log adapter).

Anyway, my last commit moves them back. We can always re-discuss later.

jamesbornholt (approved):

PR #370, opened 2023-07-13
Support open with O_RDWR flag

monthonk:
## Description of change

Currently, Mountpoint supports either open with `O_WRONLY` or `O_RDONLY` because we don't allow applications to do both read and write at the same time. However, it's possible support `O_RDWR` flag too since we can decide at open time whether to give a read handle or a write handle back, and for any inode it's never possible for both start_reading and start_writing to work.

No

---

### DIFF
--- a/doc/SEMANTICS.md
+++ b/doc/SEMANTICS.md
@@ -58,7 +58,7 @@ Mountpoint for Amazon S3 intentionally does not support all POSIX file system op
 #### Reads
 
 Basic read-only operations are fully supported, including both sequential and random reads:
-* `open`, `openat`, in read-only mode (`O_RDONLY`)
+* `open`, `openat`
 * `read`, `readv`, `pread`, `preadv`
 * `lseek`
 * `close`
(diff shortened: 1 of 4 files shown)


mountpoint-s3/src/fs.rs
```diff
@@ -375,43 +423,16 @@ where
...
-            match self.uploader.put(&self.bucket, key).await {
-                Err(e) => {
-                    error!(key, "put failed to start: {e:?}");
-                    return Err(libc::EIO);
-                }
-                Ok(request) => FileHandleType::Write {
-                    request: UploadState::InProgress(request).into(),
-                    handle,
-                },
+            let remote_file = lookup.inode.is_remote()?;
```
jamesbornholt, line 426: probably log here which option we're choosing for this `O_RDWR` file

doc/SEMANTICS.md
```diff
@@ -58,7 +58,7 @@ Mountpoint for Amazon S3 intentionally does not support all POSIX file system op
 #### Reads
 
 Basic read-only operations are fully supported, including both sequential and random reads:
-* `open`, `openat`, in read-only mode (`O_RDONLY`)
```
jamesbornholt, line 61: we should mention that `O_RDWR` will work but you can't both read and write to the same fd

jamesbornholt (approved):

PR #371, opened 2023-07-13
Endpoint Resolver for each request

sauraank:
## Description of change
Using CRT endpoint resolver to resolve endpoint for each request. With this support for ARN, mount options for FIPS, transfer acceleration and dual stack will be enabled.

Added the mount options. Created endpointConfig which contains all the configuration for endpoint resolution in mountpoint client. This configuration is used to resolve endpoint for each request in mountpoint. 
Still need to add some tests to see if it is right. So, adding it as a draft.

Relevant issues: 
#4 

Yes, new mount options will get added after this change. Also, ARN will get supported as bucket name.

No breaking change for mountpoint. Just we are not resolving endpoint in 2 step process like earlier (first, for region and then for bucket and addressing for each request)
---

### DIFF
--- a/mountpoint-s3-client/src/endpoint.rs
+++ b/mountpoint-s3-client/src/endpoint.rs
@@ -1,109 +1,80 @@
-use lazy_static::lazy_static;
 use mountpoint_s3_crt::common::allocator::Allocator;
 use mountpoint_s3_crt::common::uri::Uri;
 use mountpoint_s3_crt::s3::endpoint_resolver::{RequestContext, ResolverError, RuleEngine};
-use regex::Regex;
 use std::ffi::OsStr;
 use std::os::unix::prelude::OsStrExt;
 use thiserror::Error;
 
-lazy_static! {
-    /// Bucket names that are acceptable as virtual host names for DNS
-    static ref VALID_DNS_REGEX: Regex = Regex::new(r"[a-z0-9][a-z0-9\-]*[a-z0-9]").unwrap();
-}
+use crate::EndpointConfig;
 
 #[derive(Debug, Clone)]
 #[non_exhaustive]
 pub struct Endpoint {
     uri: Uri,
-    addressing_style: AddressingStyle,
 }
 
 impl Endpoint {
-    /// Create a new endpoint for the given S3 region. This method automatically resolves the right
+    /// get uri from Endpoint
+    fn get_uri(&self) -> Uri {
+        self.uri.clone()
+    }
+
+    /// Create a new endpoint URI for the given Endpoint Configuration. This method automatically resolves the right
     /// endpoint URI to target.
-    pub fn from_region(region: &str, addressing_style: AddressingStyle) -> Result<Self, EndpointError> {
+    pub fn set_endpoint(endpoint_config: EndpointConfig) -> Result<Uri, EndpointError> {
         let allocator = Allocator::default();
-        let endpoint_rule_engine = RuleEngine::new(&allocator).unwrap();
         let mut endpoint_request_context = RequestContext::new(&allocator).unwrap();
+        let endpoint_rule_engine = RuleEngine::new(&allocator).unwrap();
         endpoint_request_context
-            .add_string(&allocator, "Region", region)
+            .add_string(&allocator, "Region", endpoint_config.get_region())
             .unwrap();
+        if let Some(endpoint_uri) = endpoint_config.get_endpoint() {
+            endpoint_request_context
+                .add_string(&allocator, "Endpoint", endpoint_uri.get_uri().as_os_str())
+                .unwrap()
+        };
+        if let Some(bucket) = endpoint_config.get_bucket() {
+            // TODO: Handle the case of Invalid bucket name/ Alias/ ARN
+            endpoint_request_context
+                .add_string(&allocator, "Bucket", bucket)
+                .unwrap()
+        };
+        if endpoint_config.is_accelerate() {
+            endpoint_request_context
+                .add_boolean(&allocator, "UseFIPS", true)
+                .unwrap()
+        };
+        if endpoint_config.is_dual_stack() {
+            endpoint_request_context
+                .add_boolean(&allocator, "UseDualStack", true)
+                .unwrap()
+        };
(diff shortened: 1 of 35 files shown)


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -67,12 +67,127 @@ macro_rules! event {
...
+#[derive(Debug, Clone)]
+pub struct EndpointConfig {
+    bucket: Option<String>,
+    region: String,
+    use_fips: bool,
+    use_accelerate: bool,
+    use_dual_stack: bool,
+    endpoint: Option<Endpoint>,
+    addressing_style: AddressingStyle,
+}
```
dannycjones, line 79: I don't think `bucket` nor `region` should live here.

- `bucket` should be passed in per-request.
- `region` should be a member of `S3ClientConfig` I think.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -67,12 +67,127 @@ macro_rules! event {
...
+
+impl Default for EndpointConfig {
+    fn default() -> Self {
+        Self {
+            bucket: None,
+            region: "us-east-1".to_owned(),
```
dannycjones, line 85: I don't think `mountpoint-s3-client` shouldn't make this assumption, even if that's what we end up doing in `mountpoint-s3`.

Didn't fully review yet, so if that sounds off - it might be! :)

dannycjones (commented):
Just put a few initial comments. I reviewed only the `s3_crt_client.rs` very quickly to give early feedback.

This is a very very large PR. I think some of the changes I recommend will make it much smaller. Generally though a large PR makes it very difficult to review, so please be mindful of it for future changes.

mountpoint-s3-client/src/object_client.rs
```diff
@@ -72,6 +74,7 @@ pub trait ObjectClient {
         &self,
         bucket: &str,
         key: &str,
+        endpoint_config: EndpointConfig,
```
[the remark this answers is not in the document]
jamesbornholt: I also think this should live inside the client.

mountpoint-s3-client/src/endpoint.rs
```diff
@@ -1,109 +1,80 @@
...
 
-lazy_static! {
-    /// Bucket names that are acceptable as virtual host names for DNS
-    static ref VALID_DNS_REGEX: Regex = Regex::new(r"[a-z0-9][a-z0-9\-]*[a-z0-9]").unwrap();
-}
+use crate::EndpointConfig;
 
 #[derive(Debug, Clone)]
 #[non_exhaustive]
 pub struct Endpoint {
```
jamesbornholt, line 12: My rough thinking was that this struct doesn't need to exist any more, and instead things work something like this:
- the `S3CrtClient` has an `endpoint_config: EndpointConfig` field
- in `S3CrtClient::new_request_template`, the first line:
    ```rust
    let (uri, path_prefix) = self.endpoint.for_bucket(bucket)?;
    ```
    is now something like
    ```rust
    let (uri, path_prefix) = self.endpoint_config.resolve_for_bucket(bucket)?;
    ````
    and that `resolve_for_bucket` function is what invokes the endpoint resolver.
    - I don't know exactly how `ForcePathStyle` works with the endpoint resolver, but something needs to arrange for `path_prefix` to be set correctly so it can be used later. Or maybe the path prefix is already in the `Uri` and `S3Message::set_request_path_and_query` can take care of setting the request up correctly?

mountpoint-s3-client/src/s3_crt_client.rs
jamesbornholt: `bucket` definitely should be passed in per-request like it is today. For `region`, I think living either here or in `S3ClientConfig` is fine, since regions are pretty related to endpoints.

mountpoint-s3-client/src/s3_crt_client.rs
jamesbornholt: I think this probably just shouldn't implement `Default` if it includes a region field -- instead maybe there's a `EndpointConfig::default_for_region(region: &str)` method to get one.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -67,12 +67,127 @@ macro_rules! event {
...
+
+    /// Set addressing style for endpoint configuration
+    #[must_use = "EndpointConfig follows a builder pattern"]
+    pub fn addressing_style(mut self, addressing_style: AddressingStyle) -> Self {
+        self.addressing_style = addressing_style;
+        self
+    }
+
+    /// Get bucket name if set for endpoint resolution
+    pub fn get_bucket(&self) -> Option<String> {
```
jamesbornholt, line 150: Do any of these getters need to be called anywhere other than the endpoint resolver? I think you end up just adding a `resolve_for_bucket` method to this `EndpointConfig`, and don't need these getters any more.

mountpoint-s3-client/src/s3_crt_client.rs
sauraank: For the large PR, it was atomic change. I think not passing endpoint_config for each method will make this PR much shorter as you mentioned.

mountpoint-s3-client/src/s3_crt_client.rs
sauraank: Sure. But, the endpoint can be configured even without a bucket. And since this is mountpoint-s3-client, shouldn't we keep it more generic ratehr than resolving endpoint only when bucket is provided. I agree that we should add the bucket information for each S3 request.

sauraank: Yes. But I am keeping the setters like `pub fn endpoint(mut self, endpoint: Endpoint) -> Self` to set the config in the `main.rs` .

mountpoint-s3-client/src/endpoint.rs
sauraank: Looking into how path_prefix works with the resolver.

Issue #303, opened 2023-06-20
Remove default logging to file by default


dannycjones:
Logging to a file was configured in #39. By default, it logs to `~/.mountpoint-s3/`.

In this issue, we'd like to consider what the default logging configuration should be to be unsurprising and safe for customers.

dannycjones:
Completed in #373. Also, we added default syslog logging in #387.

PR #373, opened 2023-07-13
Disable logging to file by default

dannycjones:
## Description of change

Logging by default to a location on disk is unexpected. It risks accumulating logs over time since the user may not be aware that they are written there, and there is no log rotation or truncation in place.

This change disables logging to a file by default. Logging can still be configured as usual using `--log-directory <DIRECTORY>`, `-l <LOG_DIRECTORY>`.

This ensures users are making a conscious decision to enable logging and handle log rotation.

Relevant issues: #303

Yes. When updating to a version of Mountpoint containing this change, logs will no longer be emitted to `$HOME/.mountpoint-s3` if no configuration was provided by `-l` or `--log-directory`.

---

### DIFF
--- a/doc/LOGGING.md
+++ b/doc/LOGGING.md
@@ -13,7 +13,7 @@ A new log file will be created for each execution of `mount-s3`.
 Both the directory and log files are created with permissions such that the process owner has read/write access and the group has read access.
 Log files are not rotated or cleaned up.
 
-When running in foreground mode (`-f, --foreground`), `mount-s3` will emit logs to stdout irrespective of the `--log-directory` configuration.
+When running in foreground mode (`-f, --foreground`), `mount-s3` will emit logs to stdout in addition to any configured log directory.
 
 ## Log details
 
--- a/doc/LOGGING.md
+++ b/doc/LOGGING.md
@@ -1,18 +1,23 @@
 # Logging
 
-Mountpoint for Amazon S3 uses the [tracing](https://docs.rs/tracing/latest/tracing/) ecosystem for logging. This makes it easy to configure the verbosity and target of log output. By default we output minimal log information, but for reporting issues or debugging application problems, you may want to customize this logging behavior.
+Mountpoint for Amazon S3 uses the [tracing](https://docs.rs/tracing/latest/tracing/) ecosystem for logging. This makes it easy to configure the verbosity and target of log output.
+The logging level is minimal by default, but for reporting issues or debugging application problems, you may want to customize this logging behavior.
 
 ## Log outputs
 
-By default, Mountpoint for Amazon S3 outputs logs to the `~/.mountpoint-s3/` directory, creating it if it doesn't exist. This destination can be changed using the `-l, --log-directory` command-line argument.
+By default, Mountpoint for Amazon S3 will not emit any logs to disk.
 
-A new log file is created for each execution of `mount-s3`. Log files are never automatically rotated or cleaned up.
+You can enable logging to disk by providing a destination directory using the `-l, --log-directory` command-line argument.
+The directory will be created if it doesn't exist.
+A new log file will be created for each execution of `mount-s3`.
+Both the directory and log files are created with permissions such that the process owner has read/write access and the group has read access.
+Log files are not rotated or cleaned up.
 
-When running in foreground mode (`-f, --foreground`), `mount-s3` also emits the same log information to stdout.
+When running in foreground mode (`-f, --foreground`), `mount-s3` will emit logs to stdout irrespective of the `--log-directory` configuration.
 
 ## Log details
 
-By default, we output minimal (error-level) information to the log file. For reporting issues or debugging application problems, it can be helpful to increase this verbosity. We use the common `RUST_LOG` environment variable for controlling log verbosity and subjects.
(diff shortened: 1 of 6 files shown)


mountpoint-s3/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Unreleased
+
+Breaking changes:
+
+* Logging to disk is now disabled by default.
+  Logs will not longer be written to `$HOME/.mountpoint-s3/` and should be configured using `--log-directory <DIRECTORY>`.
```
jamesbornholt, line 6: I think what we probably actually want here is to start drafting v1.0.0, with a summary of breaking changes we've made during the pre-release period.

doc/LOGGING.md
```diff
@@ -1,18 +1,23 @@
...
 
-When running in foreground mode (`-f, --foreground`), `mount-s3` also emits the same log information to stdout.
+When running in foreground mode (`-f, --foreground`), `mount-s3` will emit logs to stdout irrespective of the `--log-directory` configuration.
 
 ## Log details
 
-By default, we output minimal (error-level) information to the log file. For reporting issues or debugging application problems, it can be helpful to increase this verbosity. We use the common `RUST_LOG` environment variable for controlling log verbosity and subjects.
+By default, the logging level is minimal. For reporting issues or debugging application problems, it can be helpful to increase this verbosity. We use the common `RUST_LOG` environment variable for controlling log verbosity and subjects.
 
 To control the log verbosity, set the `RUST_LOG` environment variable. If unset, it defaults to `error` to enable only error-level log messages. Verbosity can be increased by instead setting `RUST_LOG` to `warn`, `info`, `debug`, or `trace`.
```
jamesbornholt, line 22: We've talked about it before, but we should make this a command-line flag instead of (or in addition to?) the environment variable. Probably in a separate PR.

doc/LOGGING.md
```diff
@@ -1,18 +1,23 @@
...
 
-A new log file is created for each execution of `mount-s3`. Log files are never automatically rotated or cleaned up.
+You can enable logging to disk by providing a destination directory using the `-l, --log-directory` command-line argument.
+The directory will be created if it doesn't exist.
+A new log file will be created for each execution of `mount-s3`.
+Both the directory and log files are created with permissions such that the process owner has read/write access and the group has read access.
+Log files are not rotated or cleaned up.
 
-When running in foreground mode (`-f, --foreground`), `mount-s3` also emits the same log information to stdout.
+When running in foreground mode (`-f, --foreground`), `mount-s3` will emit logs to stdout irrespective of the `--log-directory` configuration.
```
jamesbornholt, line 16: Slightly unclear -- in foreground mode we'll emit logs to stdout _in addition_ to any configured log directory.

mountpoint-s3/src/main.rs
```diff
@@ -103,51 +99,48 @@ mod logging {
...
-        let filename = OffsetDateTime::now_utc()
-            .format(LOG_FILE_NAME_FORMAT)
-            .context("couldn't format log file name")?;
-
-        // log directories and files should not be accessible by other users
-        let mut file_options = OpenOptions::new();
-        file_options.mode(0o640).write(true).create(true);
+        let file_layer = if let Some(path) = log_directory {
+            const LOG_FILE_NAME_FORMAT: &[FormatItem<'static>] =
+                macros::format_description!("mountpoint_s3_[year][month][day][hour][minute][second].log");
```
jamesbornholt, line 104: Let's make this more readable while we're here
```suggestion
                macros::format_description!("mountpoint-s3-[year]-[month]-[day]-[hour]-[minute]-[second].log");
```

mountpoint-s3/CHANGELOG.md
dannycjones: That's effectively what I'm going for here with the "unreleased" heading, which will eventually be v1.0.0 if there's no minor version bump before it.

Was there a specific change you had in mind?

doc/LOGGING.md
dannycjones: Agree, though separate change. I've opened #377 to track it.

doc/LOGGING.md
dannycjones: Thanks, that's the wording I was looking for but couldn't figure out. 😄

mountpoint-s3/src/main.rs
dannycjones: How do you feel about something closer (but sadly not meeting) ISO8601?

```suggestion
                macros::format_description!("mountpoint-s3-[year]-[month]-[day]T[hour]-[minute]-[second]Z.log");
```

Example:

```
mountpoint-s3-2023-07-14T09-26-03Z.log
```

It's not quite meeting the spec, since the dashes between time should be colon. If someone really wants ISO8601, they can drop the dashes from the whole name when processing which will meet spec.

I'm also throwing in `Z` to indicate the timezone, since this timestamp is UTC.

mountpoint-s3/CHANGELOG.md
jamesbornholt: No, just that we should go back and add the ones we've missed before release.

jamesbornholt (approved):

PR #375, opened 2023-07-14
Adds region in tests

ahmarsuhail:
## Description of change

To gets tests to pass in the gov cloud and china partitions, the region needs to be specified in some cases. Without this, the head bucket probe to us-east-1 one fails as us-east-1 is in a different partition and so credentials will not work there.

Tested by running in cn-north-1, all tests pass. 

No

---

### DIFF
--- a/mountpoint-s3-client/tests/common/mod.rs
+++ b/mountpoint-s3-client/tests/common/mod.rs
@@ -44,6 +44,10 @@ pub fn get_test_bucket_without_permissions() -> String {
     std::env::var("S3_FORBIDDEN_BUCKET_NAME").expect("Set S3_FORBIDDEN_BUCKET_NAME to run integration tests")
 }
 
+pub fn get_secondary_test_region() -> String {
+    std::env::var("S3_SECONDARY_REGION").unwrap_or(String::from("ap-southeast-2"))
+ }
+ 
 pub async fn get_test_sdk_client() -> s3::Client {
     let config = aws_config::from_env()
         .region(Region::new(get_test_region()))
--- a/mountpoint-s3-client/tests/common/mod.rs
+++ b/mountpoint-s3-client/tests/common/mod.rs
@@ -48,6 +48,10 @@ pub fn get_secondary_test_region() -> String {
     std::env::var("S3_SECONDARY_REGION").unwrap_or(String::from("ap-southeast-2"))
  }
  
+pub fn get_test_domain() -> String {
+    std::env::var("S3_DOMAIN").unwrap_or(String::from("com"))
+ }
+
 pub async fn get_test_sdk_client() -> s3::Client {
     let config = aws_config::from_env()
         .region(Region::new(get_test_region()))
--- a/mountpoint-s3-client/tests/common/mod.rs
+++ b/mountpoint-s3-client/tests/common/mod.rs
@@ -49,7 +49,7 @@ pub fn get_secondary_test_region() -> String {
  }
  
 pub fn get_test_domain() -> String {
-    std::env::var("S3_DOMAIN").unwrap_or(String::from("com"))
+    std::env::var("S3_DOMAIN").unwrap_or(String::from("amazonaws.com"))
  }
 
 pub async fn get_test_sdk_client() -> s3::Client {
--- a/mountpoint-s3-client/tests/common/mod.rs
+++ b/mountpoint-s3-client/tests/common/mod.rs
@@ -46,11 +46,11 @@ pub fn get_test_bucket_without_permissions() -> String {
 
 pub fn get_secondary_test_region() -> String {
     std::env::var("S3_SECONDARY_REGION").unwrap_or(String::from("ap-southeast-2"))
- }
- 
+}
+
 pub fn get_test_domain() -> String {
     std::env::var("S3_DOMAIN").unwrap_or(String::from("amazonaws.com"))
- }
+}
 
 pub async fn get_test_sdk_client() -> s3::Client {
     let config = aws_config::from_env()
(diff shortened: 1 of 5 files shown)


mountpoint-s3-client/tests/endpoint.rs
```diff
@@ -62,29 +63,41 @@ async fn test_addressing_style_uri(addressing_style: AddressingStyle) {
...
     })
     .await;
 }
+
+fn get_test_domain(region: &str) -> String {
+    match region {
+        "cn-north-1" | "cn-northwest-1" => "com.cn".to_string(),
+        _ => "com".to_string(),
+    }
+}
```
dannycjones, line 103: Can we do this as an environment variable?

Other partitions may have different suffixes again, not just the `aws-cn` partition.

Something like:

```
AWS_REGION_DOMAIN_SUFFIX=amazonaws.com
AWS_REGION_DOMAIN_SUFFIX=amazonaws.com.cn
```

Of course, default is `amazonaws.com` if not set.

mountpoint-s3-client/tests/endpoint.rs
```diff
@@ -62,29 +63,41 @@ async fn test_addressing_style_uri(addressing_style: AddressingStyle) {
...
-        let uri = format!("https://s3.dualstack.{region}.amazonaws.com");
+        let domain = get_test_domain(region);
+        let uri = format!("https://s3.dualstack.{region}.amazonaws.{domain}");
         Endpoint::from_uri(&uri, addressing_style).unwrap()
     })
     .await;
 }
 
 // FIPS endpoints can only be used with virtual-hosted-style addressing
+#[cfg(feature = "fips_tests")]
```
dannycjones, line 74: Please enable this feature in the CI, since it currently points at `us-east-1`.

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -24,6 +24,7 @@ fn run_in_background() -> Result<(), Box<dyn std::error::Error>> {
         .arg(mount_point.path())
         .arg(format!("--prefix={prefix}"))
         .arg("--auto-unmount")
+        .arg(format!("--region={region}"))
```
dannycjones, line 27: I'm guessing this is because the tests fail if region isn't specified outside of `aws` partition?

Not sure how I feel about this, still want to keep a version of this test as is.

dannycjones (commented):
Nice changes, thanks Ahmar.

Few comments. One thing we need to figure out for the fork test.

mountpoint-s3/tests/fuse_tests/fork_test.rs
monthonk: I would prefer to keep the test as is for `aws` partition too. Maybe we need another environment variable to tell whether it's in `aws` partition, then we will pass in `--region` argument if it's not. Something like
```
if !commercial_partition {
        cmd.arg(format!("--region={region}"));
}
```

dannycjones: Discussed a little offline.

It might be just a test that we recommend skipping outside of the commercial partition, or that is ignored by default and we can opt into.

Ultimately, I want to keep this test as "region cannot otherwise be determined, does it work by probing `us-east-1`?".

Thinking to have a test specifically for this case, and add region elsewhere like in this change.

mountpoint-s3-client/tests/endpoint.rs
jamesbornholt: I don't think this function should exist at all -- all these tests need to be rewritten to use the endpoint resolver once it gains support for setting the FIPS/dualstack/etc flags. At that point none of these tests should have any hardcoded URI components at all, they should just be different `EndpointConfig`s or whatever we end up calling the way you configure the endpoint resolver.

ahmarsuhail:
thanks for the reviews, have updated with the following:
- remove region arg from `run_in_background` test, if running in any partition other than aws, test can be skipped using `-- --skip run_in_background`
- Use env variable for domain
- enable fips_tests feature in CI

mountpoint-s3-client/tests/common/mod.rs
```diff
@@ -44,6 +44,14 @@ pub fn get_test_bucket_without_permissions() -> String {
     std::env::var("S3_FORBIDDEN_BUCKET_NAME").expect("Set S3_FORBIDDEN_BUCKET_NAME to run integration tests")
 }
 
+pub fn get_secondary_test_region() -> String {
+    std::env::var("S3_SECONDARY_REGION").unwrap_or(String::from("ap-southeast-2"))
+ }
+ 
+pub fn get_test_domain() -> String {
+    std::env::var("S3_DOMAIN").unwrap_or(String::from("com"))
+ }
```
dannycjones, line 53: Please use the full domain `amazonaws.com`.

Some of the US Government regions do not contain "amazonaws" at all.

mountpoint-s3/tests/fuse_tests/fork_test.rs
dannycjones: @ahmarsuhail can you add a copy of this test that does not include the region, and keep the `run_in_background` test including the region.

Basically want to make it very clear what the test is testing - i.e. that automatic region resolution works as expected when not provided as a CLI flag.

dannycjones (changes requested):
Almost LGTM.

Two things:
- Update the test domain var to take the full domain
- Add separate test for omission of region flag https://github.com/awslabs/mountpoint-s3/pull/375#discussion_r1265444183

mountpoint-s3-client/tests/endpoint.rs
dannycjones: Sorry, just saw this. You're right.

Not sure when that'll be in, but hopefully this is something we can remove very soon.

dannycjones:
@ahmarsuhail You have a few conflicts that need resolving, it would be worth to rebase or merge from `main`.

dannycjones (dismissed):
LGTM, thanks Ahmar.

jamesbornholt (approved):

PR #380, opened 2023-07-14
Remove inodes from their parent in `forget`

jamesbornholt:
## Description of change

The parent directory still holds onto an `Inode` (an `Arc<InodeInner>`), so right now our `forget` is leaking the actual inode. We need to remove it from its parent at `forget` time. Also updated the tests to check that the inode is in fact free'd.

I tested this by listing a directory with 2M objects on an instance with 1GiB of memory, and saw constant memory usage.

No.

---

### DIFF
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -142,7 +142,12 @@ impl Superblock {
                     let InodeKindData::Directory { children, writing_children, .. } = &mut parent_state.kind_data else {
                         unreachable!("parent is always a directory");
                     };
-                    children.remove(inode.name());
+                    if let Some(child) = children.get(inode.name()) {
+                        // Don't accidentally remove a newer inode (e.g. remote shadowing local)
+                        if child.ino() == ino {
+                            children.remove(inode.name());
+                        }
+                    }
                     writing_children.remove(&ino);
                 }
             }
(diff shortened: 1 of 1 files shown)


mountpoint-s3/src/inode.rs
```diff
@@ -130,7 +130,20 @@ impl Superblock {
...
-                    inodes.remove(&ino);
+                    let inode = inodes.remove(&ino).expect("we just retrieved it");
+
+                    // Should be impossible for this to fail (VFS inodes reference their parent, so
+                    // children need to be freed first), but let's not crash in a `forget` function...
+                    let Some(parent) = inodes.get(&inode.parent()) else {
+                        debug_assert!(false, "children should be forgotten before parents");
+                        return;
+                    };
+                    let mut parent_state = parent.inner.sync.write().unwrap();
```
passaro, line 141: We typically use `parent.get_mut_inode_state()`, unless we also want directories marked as deleted?

mountpoint-s3/src/inode.rs
```diff
@@ -130,7 +130,20 @@ impl Superblock {
...
+                    let InodeKindData::Directory { children, writing_children, .. } = &mut parent_state.kind_data else {
+                        unreachable!("parent is always a directory");
+                    };
+                    children.remove(inode.name());
```
passaro, line 145: Should we check that we are in fact removing the same inode? vs a new child with the same name?

passaro: It should be possible to write a test for this scenario, e.g.:
* unlink child
* mknod new child with same name
* forget original child

passaro (changes requested):
Great follow up to forget!

mountpoint-s3/src/inode.rs
jamesbornholt: We do want those directories here, otherwise we'd leak them, since `forget` is the only way to free them.

mountpoint-s3/src/inode.rs
jamesbornholt: Nice catch. I wrote a slightly different test with the same idea -- the child changes from a file to a directory and so gets recreated, forgetting the original child.

passaro (approved):
LGTM

PR #387, opened 2023-07-18
Emit warning-level logs to syslog when log directory is unset

jamesbornholt:
## Description of change

We disabled logging to a file by default in #373. But we'd still like to have a default home for high-severity messages, which is exactly what syslog is for.

This PR is the bare minimum to get `tracing`-style structured logging into syslog. I hardcoded the verbosity to `warn,awscrt=off` and didn't think too hard about configuration: the syslog logger is on whenever the `--log-directory` flag isn't set. Because I didn't think too hard about it, I also didn't update docs/LOGGING.md yet.

This PR is two commits:
1. Move the logging module from `main.rs` to its own file, just because it was getting big. No other code changes here.
2. Actually implement the syslog logger and plumb it into `logging.rs`.

No.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -40,7 +40,8 @@ use crate::s3_crt_client::put_object::S3PutObjectRequest;
 macro_rules! request_span {
     ($self:expr, $method:expr) => {{
         let counter = $self.next_request_counter();
-        tracing::debug_span!($method, id = counter)
+        // Request failures are emitted at warning verbosity, so emit their spans there too
+        tracing::warn_span!($method, id = counter)
     }};
 }
 
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -40,8 +40,7 @@ use crate::s3_crt_client::put_object::S3PutObjectRequest;
 macro_rules! request_span {
     ($self:expr, $method:expr) => {{
         let counter = $self.next_request_counter();
-        // Request failures are emitted at warning verbosity, so emit their spans there too
-        tracing::warn_span!($method, id = counter)
+        tracing::debug_span!($method, id = counter)
     }};
 }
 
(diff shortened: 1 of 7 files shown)


dannycjones:
Plan to finish review here tomorrow morning. Needed to learn a bit about `syslog` and `tracing_subscriber` layers.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -40,7 +40,8 @@ use crate::s3_crt_client::put_object::S3PutObjectRequest;
 macro_rules! request_span {
     ($self:expr, $method:expr) => {{
         let counter = $self.next_request_counter();
-        tracing::debug_span!($method, id = counter)
+        // Request failures are emitted at warning verbosity, so emit their spans there too
+        tracing::warn_span!($method, id = counter)
```
dannycjones, line 44: We are referring to the CRT here?

```suggestion
        // Request failures are emitted at warning verbosity by the CRT, so emit their spans as warning too
        tracing::warn_span!($method, id = counter)
```

dannycjones (commented):
Thanks James, this looks great and I'm glad to have this as the default if logging is not configured. The additional comments in the last commit also helped a lot.

The main concern I have is that it will become a file that most will not want to touch, so I'd like to make it as accessible as possible. The comments are along these lines.

mountpoint-s3-client/src/s3_crt_client.rs
jamesbornholt: No, referring to the `warn!` messages we emit in this file. But this is actually wrong -- I always get `span` levels confused. It's the _lowest_ level at which the span should be emitted, and we want this span emitted in `debug!` messages, so the old code is correct.

mountpoint-s3/src/logging/syslog.rs
```diff
@@ -0,0 +1,217 @@
...
+        Self {
+            logger: Mutex::new(logger),
+        }
+    }
+
+    fn formatter() -> Formatter3164 {
+        Formatter3164 {
+            facility: Facility::LOG_USER,
+            hostname: None,
+            process: "mount-s3".into(),
```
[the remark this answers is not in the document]
dannycjones: Fair enough, something we can chase in another PR if we feel the need to.

mountpoint-s3-client/src/s3_crt_client.rs
dannycjones: Gotcha! I was wondering the same but couldn't pin it down.

PR #388, opened 2023-07-18
Maintain metadata checksums and verify on unlink

passaro:
## Description of change

Adds a checksum to each inode computed on the inode number and the full key. On unlink, verifies the checksum before deleting the object from the bucket.

No impact on the happy path. Unlink could fail if inode metadata gets corrupted.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -804,7 +804,7 @@ impl From<InodeError> for i32 {
             InodeError::CannotRemoveRemoteDirectory(_) => libc::EPERM,
             InodeError::DirectoryNotEmpty(_) => libc::ENOTEMPTY,
             InodeError::UnlinkNotPermittedWhileWriting(_) => libc::EPERM,
-            InodeError::CorruptedMetadata(_, _) => libc::EPERM,
+            InodeError::CorruptedMetadata(_, _) => libc::EIO,
         }
     }
 }
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -804,6 +804,7 @@ impl From<InodeError> for i32 {
             InodeError::CannotRemoveRemoteDirectory(_) => libc::EPERM,
             InodeError::DirectoryNotEmpty(_) => libc::ENOTEMPTY,
             InodeError::UnlinkNotPermittedWhileWriting(_) => libc::EPERM,
+            InodeError::CorruptedMetadata(_, _) => libc::EPERM,
         }
     }
 }
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -28,6 +28,7 @@ use std::time::{Duration, Instant};
 use fuser::FileType;
 use futures::{select_biased, FutureExt};
 use mountpoint_s3_client::{HeadObjectError, HeadObjectResult, ObjectClient, ObjectClientError};
+use mountpoint_s3_crt::checksums::crc32c::{self, Crc32c};
 use thiserror::Error;
 use time::OffsetDateTime;
 use tracing::{debug, error, trace, warn};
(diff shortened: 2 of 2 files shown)


mountpoint-s3/src/fs.rs
```diff
@@ -804,6 +804,7 @@ impl From<InodeError> for i32 {
             InodeError::CannotRemoveRemoteDirectory(_) => libc::EPERM,
             InodeError::DirectoryNotEmpty(_) => libc::ENOTEMPTY,
             InodeError::UnlinkNotPermittedWhileWriting(_) => libc::EPERM,
+            InodeError::CorruptedMetadata(_, _) => libc::EPERM,
```
passaro, line 807: Not sure if there is a more fitting error.

jamesbornholt: I think maybe `EIO` is slightly more appropriate, but it's pretty niche either way.

mountpoint-s3/src/inode.rs
```diff
@@ -1803,6 +1827,39 @@ mod tests {
...
+        client.add_object(file_name.as_ref(), MockObject::constant(0xaa, 30, ETag::for_tests()));
+
+        let parent_ino = FUSE_ROOT_INODE;
+        let LookedUp { inode, .. } = superblock
+            .lookup(&client, parent_ino, file_name.as_ref())
+            .await
+            .expect("file should exist");
+
+        // Inject a key mutation in the Inode.
+        // SAFETY: Inode and String not accessed concurrently. Mutated string is still valid utf8.
```
jamesbornholt, line 1850: This is actually undefined behavior and so not safe (mutating data to which a shared reference exists is [always UB](https://doc.rust-lang.org/reference/behavior-considered-undefined.html)). The risk here is low in practice — the way this would go wrong is if an extremely smart compiler previously cached the `full_key` in a register somewhere, which would be OK because it's a shared immutable field, and so didn't observe this mutation to the key the next time it read it. But in theory the compiler is free to do whatever it wants with the entire program when it sees this.

As an alternative, maybe you could manually construct a new `Inode` struct here as a copy of the existing one, and then overwrite it in the right places (the superblock and the parent of this inode)? It's a lot more coupled to the internals, but maybe unavoidable.

mountpoint-s3/src/inode.rs
```diff
@@ -360,6 +360,10 @@ impl Superblock {
             return Err(InodeError::IsDirectory(inode.ino()));
         }
 
+        if !inode.verify_checksum() {
```
jamesbornholt, line 363: Maybe we should just make `self.inner.get` do this check every time?

jamesbornholt (commented):
Neat! Just a couple of suggestions.

mountpoint-s3/src/inode.rs
passaro: Maybe. And/or in `self.inner.lookup` (or it would miss unlink for example)?

passaro: Done in latest commit.

passaro: Or maybe instead we should verify the checksum when we get the full key?

mountpoint-s3/src/inode.rs
```diff
@@ -425,12 +425,15 @@ impl Superblock {
...
-        self.inodes
+        let inode = self
+            .inodes
             .read()
             .unwrap()
             .get(&ino)
             .cloned()
-            .ok_or(InodeError::InodeDoesNotExist(ino))
+            .ok_or(InodeError::InodeDoesNotExist(ino))?;
+        inode.verify_checksum()?;
```
jamesbornholt, line 435: We should also check that `inode.ino() == ino` here to be safe.

mountpoint-s3/src/inode.rs
```diff
@@ -1013,6 +1011,46 @@ impl Inode {
...
+    fn verify_inode(&self, expected_ino: InodeNo) -> Result<(), InodeError> {
+        let computed = Self::compute_checksum(self.ino(), self.full_key());
+        if computed == self.inner.checksum && self.ino() == expected_ino {
+            Ok(())
+        } else {
+            Err(InodeError::CorruptedMetadata(self.ino(), self.full_key().to_owned()))
+        }
+    }
+
+    fn verify_child(&self, expected_parent: InodeNo, expected_name: &str) -> Result<(), InodeError> {
```
jamesbornholt, line 1039: Nice!

jamesbornholt (approved):

PR #390, opened 2023-07-19
Add detailed CRT metrics and tweak per-request logging

jamesbornholt:
This change is a few related things to make logging more useful:
- Include verbose CRT request metrics at trace level. We emit what we
  think is the "interesting" stuff at higher levels, but for detailed
  investigation we might want to see the raw CRT view.
- Add parameters to request spans. This ensures that we know _which_
  request is going wrong when we see log messages about requests.
- Don't log the bucket redirection at warning level, since that's going
  to syslog by default

## Description of change

Relevant issues: 

---


mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -792,12 +868,104 @@ impl RequestMetrics {
...
+    /// Get the IP address the request connected to
+    pub fn ip_address(&self) -> Option<String> {
+        let mut out: *const aws_string = std::ptr::null();
+        // SAFETY: `inner` is a valid aws_s3_request_metrics
+        unsafe {
+            aws_s3_request_metrics_get_ip_address(self.inner.as_ptr(), &mut out)
+                .ok_or_last_error()
+                .ok()?
+        };
+        assert!(!out.is_null(), "IP address should be available if call succeeded");
```
dannycjones, line 918: Should this be debug assert, and return `None` for release builds?

```suggestion
        debug_assert!(!out.is_null(), "IP address should be available if call succeeded");
        if (out.is_null()) {
            return None;
        }
```

dannycjones (approved):
Great QOL logging improvement!

One possible query, otherwise LGTM.

Issue #358, opened 2023-07-06
`touch` does not work


jchorl:
### Mountpoint for Amazon S3 version

mountpoint-s3 0.3.0

### AWS Region

us-west-2

### Describe the running environment

Running on my local machine. Compiled from source. Latest master. ubuntu 22.04 LTS.

### What happened?

```
root@75bda5985fb6:/work# ./target/release/mount-s3 --prefix $S3_BUCKET_TEST_PREFIX $S3_BUCKET_NAME /mnt
root@75bda5985fb6:/work# ls /mnt
# some files, great
root@75bda5985fb6:/work# touch /mnt/foo
touch: setting times of '/mnt/foo': Function not implemented
```

I was debating between a bug report and feature req. However, `touch` seems like it would be expected for https://github.com/awslabs/mountpoint-s3/milestone/3

Notably, this patch fixes the functionality: https://github.com/awslabs/mountpoint-s3/commit/7a10bcc27e5c25123434110accf96726c30fbbce

It's pretty hacky though, it just no-ops `setattr`. I'm not sure if it is useful or important to have a real implementation.

### Relevant log output

```shell
touch: setting times of '/mnt/foo': Function not implemented
```

passaro:
Hi, thank you for opening this issue. We will look into it.

monthonk:
I'm going to close this issue, `touch` should now work from https://github.com/awslabs/mountpoint-s3/pull/391. Please open a new bug report if it's still a problem.

PR #391, opened 2023-07-19
Implement setattr to support changing time attributes

monthonk:
## Description of change

Some applications like `touch` requires the file system to support changing file last access and modification times. We don't support this operation because the last modification time for objects can't be set via S3 API. However, it's possible to allow this only for the files that are being written because at that time it's still a temporary stat in Mountpoint.

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/358

Yes, `setattr` will now work for the files that are being written.

---

### DIFF
--- a/doc/SEMANTICS.md
+++ b/doc/SEMANTICS.md
@@ -77,6 +77,8 @@ further writes.
 
 Space allocation (`fallocate`, `posix_fallocate`) are not supported.
 
+Changing file last access and modification times (`utime`) are supported on files that are being written only.
+
 #### Deletes
 
 File deletion (`unlink`) is planned but not yet supported (see [#78](https://github.com/awslabs/mountpoint-s3/issues/78)).
--- a/doc/SEMANTICS.md
+++ b/doc/SEMANTICS.md
@@ -77,7 +77,7 @@ further writes.
 
 Space allocation (`fallocate`, `posix_fallocate`) are not supported.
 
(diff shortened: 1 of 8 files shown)


mountpoint-s3/src/inode.rs
```diff
@@ -189,6 +189,39 @@ impl Superblock {
...
+
+        if sync.write_status == WriteStatus::Remote {
+            return Err(InodeError::SetAttrNotPermittedOnRemoteInode(ino));
+        }
+
+        // Should be impossible since local file stat never expire.
+        if !sync.stat.is_valid() {
+            warn!(?ino, "local inode stat already expired");
+            return Err(InodeError::SetAttrOnExpiredStat(ino));
+        }
```
monthonk, line 211: Not really sure how we should handle this case, another option is doing lookup to update the validity, but it should not happen so I think it makes sense to return an error immediately.

doc/SEMANTICS.md
```diff
@@ -77,6 +77,8 @@ further writes.
 
 Space allocation (`fallocate`, `posix_fallocate`) are not supported.
 
+Changing file last access and modification times (`utime`) are supported on files that are being written only.
```
passaro, line 80: Maybe something like:
```suggestion
Changing last access and modification times (`utime`) is supported only on files that are being written.
```

passaro (approved):
LGTM

passaro (approved):

PR #393, opened 2023-07-20
Fix sequential write test with O_RDWR flag

passaro:
## Description of change

Some of the sequential write tests were testing that once a new file is opened with O_RDWR and written to, a read on it will fail with EBADF. The issue was that the read in the test was performed at the EOF, which would result in different fuse calls depending on the kernel version: for example on 5.15, fuse `read` would be called with a "valid" offset and the expected EBADF would be returned (as we were observing in GitHub Actions), but on 6.1, a `getattr` would check that the current position was EOF and succeed the `read` call by returning `0`. 

The fix seeks back `1` byte before trying to read it.

No. Only tests are affected.

---

### DIFF
--- a/mountpoint-s3/tests/fuse_tests/write_test.rs
+++ b/mountpoint-s3/tests/fuse_tests/write_test.rs
@@ -111,6 +111,7 @@ where
     }
 
     // We shouldn't be able to read from a file mid-write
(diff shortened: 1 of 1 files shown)


monthonk (approved):
LGTM

PR #397, opened 2023-07-20
Add always-successful workflow for DCO on merge_group event only

dannycjones:
## Description of change

The DCO app that currently gates pushes to `main` to ensure that all commits have DCO trailers does not appear to run on the `merge_group` event. There is an open ticket at dcoapp/app#199.

This change introduces a workaround by having a workflow always succeed with the same name as the DCO app. We already check for DCO at PR approval time.

If dcoapp/app#199 is resolved, we should be able to remove this workflow.

Relevant issues: #392, dcoapp/app#199

No change.

---

### DIFF
--- a/.github/workflows/merge_queue_dco_mock.yml
+++ b/.github/workflows/merge_queue_dco_mock.yml
@@ -1,4 +1,7 @@
 ---
+# Workflow that always succeeds with name "DCO".
+# This is temporarily needed to ensure we can keep DCO requirement on PRs and allow merge queue to succeed.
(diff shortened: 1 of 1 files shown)


.github/workflows/merge_queue_dco_mock.yml
```diff
@@ -0,0 +1,12 @@
+---
```
jamesbornholt, line 1: Can you add a comment to this file with the same point you make in the PR that this can be removed if https://github.com/dcoapp/app/issues/199 is fixed?

jamesbornholt (approved):

PR #400, opened 2023-07-21
Adds storage class

ahmarsuhail:
## Description of change

Adds in new CLI option to configure storage class at mount time. Users can now do mount --storage-class = "INTELLIGENT_TIERING", default storage class remains as standard.

Most of this code has been previously reviewed in https://github.com/awslabs/mountpoint-s3/pull/355. Had to close that and opening new one after rebasing. 

No.

---


mountpoint-s3/src/fs.rs
```diff
@@ -226,6 +226,8 @@ pub struct S3FilesystemConfig {
     pub prefetcher_config: PrefetcherConfig,
     /// Allow delete
     pub allow_delete: bool,
+    /// Storage class to be used
```
jamesbornholt, line 229:
```suggestion
    /// Storage class to be used for new object uploads
```

mountpoint-s3-client/src/object_client.rs
```diff
@@ -267,6 +269,12 @@ impl PutObjectParams {
         self.trailing_checksums = value;
         self
     }
+
+    /// Set the storage class.
+    pub fn storage_class(mut self, value: Option<String>) -> Self {
```
jamesbornholt, line 274: nit: the idiomatic way to do builders like this is that it should take just `String` as input, and the callers only call it if they want to set a non-None value

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -448,3 +452,76 @@ fn out_of_order_write_test_s3(offset: i64) {
...
+        .await
+        .unwrap();
+
+    assert_eq!(storage_class, attributes.storage_class.unwrap().as_str());
+}
+
+#[cfg(feature = "s3_tests")]
+#[test_case("INTELLIGENT_TIERING")]
+#[test_case("GLACIER")]
+fn write_with_storage_class_test_s3(storage_class: &str) {
```
jamesbornholt, line 492: Do we need to add storage classes to the mock client to be able to test with that here?

mountpoint-s3/src/main.rs
```diff
@@ -94,6 +94,9 @@ struct CliArgs {
     )]
     pub read_only: bool,
 
+    #[clap(long, help = "Set the storage class for new objects", help_heading = BUCKET_OPTIONS_HEADER)]
+    pub storage_class: Option<String>,
```
jamesbornholt, line 98: This should be just `String` and we set the default to `INTELLIGENT_TIERING`.

mountpoint-s3/tests/fuse_tests/mod.rs
```diff
@@ -31,6 +31,8 @@ pub trait TestClient {
     fn contains_key(&self, key: &str) -> Result<bool, Box<dyn std::error::Error>>;
 
     fn is_upload_in_progress(&self, key: &str) -> Result<bool, Box<dyn std::error::Error>>;
+
+    fn get_object_storage_class(&self, key: &str) -> Result<String, Box<dyn std::error::Error>>;
```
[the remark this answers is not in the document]
ahmarsuhail: done

mountpoint-s3/src/upload.rs
```diff
@@ -72,9 +72,12 @@ impl<Client: ObjectClient> UploadRequest<Client> {
...
         key: &str,
     ) -> ObjectClientResult<Self, PutObjectError, Client::ClientError> {
-        let params = PutObjectParams::new()
-            .trailing_checksums(true)
-            .storage_class(inner.storage_class.to_owned().unwrap());
+        let mut params = PutObjectParams::new().trailing_checksums(true);
+
+        if let Some(storage_class) = inner.storage_class.to_owned() {
+            params = params.storage_class(storage_class);
+        }
```
passaro, line 79: nit: my preference would be for something like this, where `if let` just takes a reference and `clone` is called when passing the argument
```suggestion
        let mut params = PutObjectParams::new().trailing_checksums(true);
        if let Some(storage_class) = inner.storage_class {
            params.storage_class(storage_class.clone());
        }
```

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -99,6 +99,11 @@ impl MockClient {
     pub fn is_upload_in_progress(&self, key: &str) -> bool {
         self.in_progress_uploads.read().unwrap().contains(key)
     }
+
+    /// Returns the objects storage class
+    pub fn get_object_storage_class(&self, key: &str) -> Option<String> {
```
passaro, line 104: This should probably return an error if the key is not found, rather than panicking.

passaro (approved):

PR #403, opened 2023-07-24
Bump some dependencies to mostly remove `atty`

jamesbornholt:
## Description of change

It's no longer in our release dependency closure, but still in the test
closure through two sources:
* fuser uses clap v3 in its examples, which still depends on atty
* fuser uses env-logger v0.9 in its examples, which still depends on
  atty

No.

---


dannycjones (changes requested):
Compiles but Clap upgrade has a breaking change for arguments.

This example breaks, we just need to remove the `--` from arg definitions on the list example.

```
❯ cargo run --example list -- --bucket doc-example-bucket
    Finished dev [unoptimized + debuginfo] target(s) in 0.80s
     Running `/Users/djonesoa/devel/mountpoint-s3/target/debug/examples/list --bucket doc-example-bucket`
thread 'main' panicked at 'Argument delimiter: long "--delimiter" must not start with a `-`, that will be handled by the parser', /Users/djonesoa/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.3.9/src/builder/debug_asserts.rs:79:13
```

dannycjones (approved):

PR #408, opened 2023-07-25
fuse: log all operation failures

jamesbornholt:
## Description of change

I was going to do this in a proc macro but this is way simpler. This
just follows up on #404 by recording all the errors, using a small macro
in place of the existing calls to `reply.error(libc::c_int)`.

No.

---


passaro (approved):
Very nice! And much simpler than the proc macro approach :)

PR #409, opened 2023-07-25
Fix region detection with specified region

jamesbornholt:
## Description of change

We weren't setting the region in the endpoint config if it was specified manually (with `--region`). This passed our tests because it was defaulting to us-east-1 and that's where our CI is. Instead, let's start with an obviously wrong placeholder region and fill it in when creating the client.

No.

---

### DIFF
--- a/mountpoint-s3/src/main.rs
+++ b/mountpoint-s3/src/main.rs
@@ -435,7 +435,7 @@ fn create_client_for_bucket(
         if endpoint_url.is_some() {
             tracing::warn!(
                 "endpoint specified but region unspecified. using {} as the signing region.",
-                endpoint_config.get_region()
+                DEFAULT_REGION
             );
         }
         DEFAULT_REGION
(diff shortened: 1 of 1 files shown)


mountpoint-s3/src/main.rs
```diff
@@ -429,24 +429,27 @@ fn create_client_for_bucket(
...
             tracing::warn!(
                 "endpoint specified but region unspecified. using {} as the signing region.",
-                endpoint_config.get_region()
+                DEFAULT_REGION
             );
         }
-        endpoint_config.get_region().to_owned()
+        DEFAULT_REGION
     });
+    endpoint_config = endpoint_config.region(region_to_try);
```
sauraank, line 443: Right. I was only once setting the region that was default and not after getting the `region_to_try` in endpoint config. Which meant it was always `us-east-1` in `EndpointConfig`. Thanks.

sauraank (approved):
Thanks for correcting.

PR #410, opened 2023-07-26
Add new subsession IAM role for auth integration tests

jamesbornholt:
## Description of change

We want to be able to write tests with various permutations of IAM
credentials and policies (read only, prefix only, etc). Rather than
manually building new infrastructure for them, we're creating a single
new IAM role that tests can call AssumeRole on, using a session policy
to scope down the credentials to those they want to test.

I'll be using this in a follow-up commit to switch from HeadBucket to
ListObjects for region detection. I suspect we can also use it to get
rid of our "forbidden" bucket, but don't plan on doing that right now.

No, only new tests.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1979,6 +1979,7 @@ dependencies = [
  "aws-config",
  "aws-credential-types",
  "aws-sdk-s3",
+ "aws-sdk-sts",
  "base64ct",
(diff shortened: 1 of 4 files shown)


dannycjones (approved):
LGTM, pending rebase on #412.

PR #411, opened 2023-07-26
Dynamically scale fuse threads

passaro:
## Description of change

Introduce a pool of fuse worker threads that will scale dynamically up to a `max_workers` limit when receiving kernel requests.

The implementation relies on the following changes in `fuser/fork`:
* `Session::run_with_callbacks()`: version of `Session::run()` with callbacks before/after dispatch.
* `Request::is_forget()`: we do not want to spawn new threads on spikes of forget/batch_forget requests.

Relevant issues: #7 

New `--max-threads` option replaces `--thread-count` and sets the **maximum** number of threads to spawn. 

---

### DIFF
--- a/mountpoint-s3/src/fuse/session.rs
+++ b/mountpoint-s3/src/fuse/session.rs
@@ -1,11 +1,13 @@
-use std::sync::Arc;
+use std::io;
 
 use anyhow::Context;
 use fuser::{Filesystem, Session, SessionUnmounter};
-use tracing::{debug, error, trace};
+use tracing::{debug, error, trace, warn};
 
-use crate::sync::mpsc;
-use crate::sync::thread;
+use crate::sync::atomic::{AtomicUsize, Ordering};
+use crate::sync::mpsc::{self, Sender};
+use crate::sync::thread::{self, JoinHandle};
+use crate::sync::Arc;
 
 /// A multi-threaded FUSE session that can be joined to wait for the FUSE filesystem to unmount or
 /// this process to be interrupted.
@@ -18,32 +20,15 @@ impl FuseSession {
     /// Create worker threads to dispatch requests for a FUSE session.
     pub fn new<FS: Filesystem + Send + Sync + 'static>(
         mut session: Session<FS>,
-        worker_threads: usize,
+        max_worker_threads: usize,
     ) -> anyhow::Result<Self> {
-        assert!(worker_threads > 0);
+        assert!(max_worker_threads > 0);
 
         let unmounter = session.unmount_callable();
 
-        let session = Arc::new(session);
-
         let (tx, rx) = mpsc::channel();
 
-        let workers: Vec<_> = (0..worker_threads)
-            .map(|i| {
-                let session = session.clone();
-                thread::Builder::new().name(format!("fuse-worker-{i}")).spawn(move || {
-                    #[cfg(target_os = "linux")]
-                    {
-                        // SAFETY: this syscall is available since Linux 2.4.11 but glibc didn't
-                        // wrap it until very recently.
-                        let tid = unsafe { libc::syscall(libc::SYS_gettid) };
-                        debug!("fuse worker {i} is thread ID {tid}");
-                    }
-                    session.run()
-                })
-            })
-            .collect::<Result<_, _>>()
-            .context("failed to spawn worker threads")?;
+        let (workers_tx, workers_rx) = mpsc::channel::<JoinHandle<io::Result<()>>>();
 
         // A thread that waits for all workers to exit and then sends a message on the channel
         let _waiter = {
@@ -51,7 +36,7 @@ impl FuseSession {
             thread::Builder::new()
                 .name("fuse-worker-waiter".to_owned())
                 .spawn(move || {
-                    for thd in workers {
+                    while let Ok(thd) = workers_rx.recv() {
                         let thread_name = thd.thread().name().map(ToOwned::to_owned);
                         match thd.join() {
                             Err(panic_param) => {
@@ -80,6 +65,8 @@ impl FuseSession {
         })
         .context("failed to set interrupt handler")?;
 
+        WorkerPool::start(session, workers_tx, max_worker_threads).context("failed to start worker thread pool")?;
+
         Ok(Self {
             unmounter,
             receiver: rx,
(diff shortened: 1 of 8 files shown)


mountpoint-s3/src/main.rs
```diff
@@ -135,7 +135,7 @@ struct CliArgs {
 
     #[clap(
         long,
-        help = "Number of FUSE daemon threads",
+        help = "Maximum number of FUSE daemon threads",
```
jamesbornholt, line 138: should probably rename the actual flag to` --max-threads` too

jamesbornholt (commented):
If you want to think more about testing, you could factor out the `Session::run` part of the worker pool into a shared closure, and write a stress test with a different `run` function. Could even use shuttle for that if you were inclined.

mountpoint-s3/src/fuse/session.rs
```diff
@@ -101,3 +86,137 @@ enum Message {
...
 }
+
+type WorkerHandle = JoinHandle<std::io::Result<()>>;
+
+#[derive(Debug)]
+struct WorkerPool<FS: Filesystem> {
+    state: Arc<WorkerPoolState<FS>>,
+    workers: Sender<WorkerHandle>,
+    max_workers: usize,
+    max_idle_workers: Option<usize>,
```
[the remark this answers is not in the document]
passaro: Scaling down removed

passaro:
Still thinking about testing

mountpoint-s3/src/fuse/session.rs
```diff
@@ -101,3 +88,249 @@ enum Message {
...
+                message.process();
+                after();
+            }
+        }
+    }
+
+    #[test_case(10, 10)]
+    #[test_case(10, 30)]
+    #[test_case(30, 10)]
+    fn test_worker_pool(max_worker_threads: usize, concurrent_messages: usize) {
```
jamesbornholt, line 295: It took me a long time to parse what this test is doing, so probably deserves a comment. Here's my understanding: it tests that the spawning logic never under-spawns threads, by assigning each worker thread a work item that only completes when a flag is flipped, and then arranges for the flag to flip only once `max_worker_threads` have been spawned. Neat!

I thought maybe you could use `std::sync::Barrier` to make this simpler (set the barrier to `max_worker_threads + 1`), but I guess it doesn't have a `wait_timeout` so you'd block forever if the test broke.

mountpoint-s3/src/fuse/session.rs
```diff
@@ -101,3 +88,249 @@ enum Message {
...
+        drop(tx);
+
+        // The pool tries to spawn an extra idle worker.
+        if let Ok(worker) = workers_rx.recv() {
+            workers.push(worker);
+            assert_eq!(workers.len(), min_expected_workers + 1);
+        } else {
+            assert_eq!(workers.len(), min_expected_workers);
+        }
+    }
```
jamesbornholt, line 335: Wasn't too hard to run this to a Shuttle test as well:

```rust
    #[cfg(feature = "shuttle")]
    mod shuttle_tests {
        use shuttle::rand::Rng;
        use shuttle::{check_pct, check_random};

        fn test_worker_pool_helper() {
            let mut rng = shuttle::rand::thread_rng();
            let num_worker_threads = rng.gen_range(1..=8);
            let num_concurrent_messages = rng.gen_range(1..=16);
            super::test_worker_pool(num_worker_threads, num_concurrent_messages);
        }

        #[test]
        fn test_worker_pool() {
            check_random(test_worker_pool_helper, 10000);
            check_pct(test_worker_pool_helper, 10000, 3);
        }
    }
```

mountpoint-s3/src/fuse/session.rs
jamesbornholt: Also, I think we should write one other, simpler test: the work item is just incrementing a shared counter. We check that the counter got incremented exactly as many times as we expected, and that the number of spawned threads was no greater than `max_worker_threads`.

mountpoint-s3/src/fuse/session.rs
```diff
@@ -101,3 +88,249 @@ enum Message {
     WorkersExited,
     Interrupted,
 }
+
+trait Work: Send + Sync + 'static {
+    type Result: Send;
+
+    fn run<FB, FA>(&self, before: FB, after: FA) -> Self::Result
```
jamesbornholt, line 95: Add a brief comment about the semantics of this function.

vendor/fuser/src/session.rs
```diff
@@ -116,6 +116,18 @@ impl<FS: Filesystem> Session<FS> {
...
     /// calls into the filesystem.
     pub fn run(&self) -> io::Result<()> {
+        self.run_and_notify(|_| {}, |_| {})
+    }
+
+    /// Run the session loop that receives kernel requests and dispatches them to method
+    /// calls into the filesystem.
+    /// This version also notifies callers of kernel requests before and after they
+    /// are dispatched to the filesystem.
+    pub fn run_and_notify<FA, FB>(&self, mut before_dispatch: FB, mut after_dispatch: FA) -> io::Result<()>
```
jamesbornholt, line 126: naming nit: I'd probably call it `run_with_callbacks` or something.

jamesbornholt (commented):
Just a couple suggestions. The `fuser` changes look good to me, but we don't commit those directly to mainline because we've been trying to maintain fuser history separately. Instead, pull those into their own commit on the `fuser/fork` branch (you can push that without reviewing), and then run `vendor-fuser.sh` on the PR branch to pull the change in.

mountpoint-s3/src/main.rs
```diff
@@ -135,13 +135,13 @@ struct CliArgs {
...
         value_name = "N",
         default_value = "1",
```
jamesbornholt, line 140: Let's make the default here 16 (just a number I made up).

mountpoint-s3/src/fuse/session.rs
passaro: Thanks! Added together with the new test.

vendor/fuser/src/session.rs
passaro: reverted!

jamesbornholt (approved):

PR #413, opened 2023-07-26
Use ListObjectsV2 instead of HeadBucket for region detection 

jamesbornholt:
## Description of change

HeadBucket requires `s3:ListBucket` permissions for the root of the bucket, but some customers scope their users' access down to only a prefix of the bucket. This makes it impossible for them to use prefix mounts today. Instead, we want to use ListObjects on the prefix as the region detection mechanism.

It turned out that the cleanest way to do this was to refactor how the S3CrtClient does error handling, to centralize the generic cases like permission denied or incorrect region. So this PR is two commits that might be easier to read individually: the first one is that refactoring, and the second does the actual change to `mount-s3`.

Yes, Mountpoint no longer uses HeadBucket at mount time, preferring ListObjects instead.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -1,5 +1,6 @@
 use std::ffi::{OsStr, OsString};
 use std::future::Future;
+use std::ops::Deref;
 use std::ops::Range;
 use std::os::unix::prelude::OsStrExt;
 use std::pin::Pin;
@@ -29,7 +30,7 @@ use futures::channel::oneshot;
 use percent_encoding::{percent_encode, AsciiSet, NON_ALPHANUMERIC};
 use pin_project::pin_project;
 use thiserror::Error;
-use tracing::{error, trace, Span};
+use tracing::{debug, error, trace, Span};
 
 use crate::endpoint_config::EndpointError;
 use crate::object_client::*;
@@ -299,6 +300,7 @@ impl S3CrtClientInner {
     /// object data.
     fn new_request_template(&self, method: &str, bucket: &str) -> Result<S3Message, ConstructionError> {
         let uri = self.endpoint_config.resolve_for_bucket(bucket)?;
+        trace!(?uri, "resolved endpoint");
         let hostname = uri.host_name().to_str().unwrap();
         let path_prefix = uri.path().to_os_string().into_string().unwrap();
         let port = uri.host_port();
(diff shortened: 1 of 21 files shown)


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -343,34 +345,40 @@ impl S3CrtClientInner {
...
         let options = Self::new_meta_request_options(message, meta_request_type);
         self.make_meta_request_from_options(options, request_span, on_headers, on_body, on_finish)
     }
 
     /// Make an HTTP request using this S3 client that invokes the given callbacks as the request
-    /// makes progress. The `on_finish` callback is invoked on both successful and failed requests;
-    /// it should call `.is_err()` on the [MetaRequestResult] to decide whether the request
-    /// succeeded.
+    /// makes progress. See [make_meta_request] for arguments.
     #[allow(clippy::too_many_arguments)] // TODO: review
```
passaro, line 372: unrelated: does this still apply?

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -444,9 +452,27 @@ impl S3CrtClientInner {
...
                 }
 
+                // Let the `on_finish` callback decide whether the request failed or not. If it did,
+                // the callback can decide whether it wants to parse the error, or return
+                // `Err(None)` if it wants to fall back to generic error handling.
+                let parsed_result = on_finish(&request_result);
+                let parsed_result = match parsed_result {
+                    Ok(t) => Ok(t),
+                    Err(Some(e)) => Err(Some(e)),
+                    // If `try_parse_generic_error` fails we'll eventually generate a
```
passaro, line 462: This took me a while to digest. Do we use `parsed_result` only for logging? (I mean until line 488 when we are ready to return it)
I wonder if it would get clearer if we moved the logging into the different branches of this `match`.

mountpoint-s3-client/src/s3_crt_client/head_object.rs
```diff
@@ -72,6 +72,11 @@ impl S3CrtClient {
         bucket: &str,
         key: &str,
     ) -> ObjectClientResult<HeadObjectResult, HeadObjectError, S3RequestError> {
+        // We use this header to stash the response from the head_object during the on_headers
+        // callback, and send it back on the oneshot when we finish.
```
passaro, line 76: Update comment? I assume the "oneshot" was the `on_finish` callback.

mountpoint-s3/src/main.rs
```diff
@@ -499,19 +502,20 @@ fn create_client_for_bucket(
...
+    let list_request = client.list_objects(bucket, None, "", 0, prefix.as_str());
+    match futures::executor::block_on(list_request) {
         Ok(_) => Ok(client),
         // Don't try to automatically correct the region if it was manually specified incorrectly
-        Err(ObjectClientError::ServiceError(HeadBucketError::IncorrectRegion(region))) if supposed_region.is_none() => {
+        Err(ObjectClientError::ClientError(S3RequestError::IncorrectRegion(region))) if supposed_region.is_none() => {
             tracing::warn!("bucket {bucket} is in region {region}, not {region_to_try}. redirecting...");
             let new_client = S3CrtClient::new(client_config.endpoint_config(endpoint_config.region(&region)))?;
-            let head_request = new_client.head_bucket(bucket);
+            let head_request = new_client.list_objects(bucket, None, "", 0, prefix.as_str());
```
passaro, line 512: rename

passaro (commented):
Just a couple of minor comments.

passaro (approved):
LGTM!

PR #416, opened 2023-07-27
Make inode-related errors more user-friendly

jamesbornholt:
## Description of change

Inode numbers aren't very useful for debugging, since you have to go
reconstruct which object they refer to, which is only possible with
fairly detailed (trace-level?) logs. Since many of these errors become
customer-facing in logs (e.g. opening an existing file with O_WRONLY),
we can make them at least a little more parseable by attaching some info
to them.

Here's an example of trying to `rmdir` a nested remote directory:
```
2023-07-27T19:35:25.509970Z  WARN rmdir{req=12 parent=2 name="dir1"}: mountpoint_s3::fuse: rmdir failed: inode error: remote directory cannot be removed at inode 3 (full key "dirs-small/dir1/")
```

No

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -447,7 +447,7 @@ where
         let lookup = self.superblock.getattr(&self.client, ino, true).await?;
 
         match lookup.inode.kind() {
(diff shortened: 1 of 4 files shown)


passaro (approved):
LGTM

PR #419, opened 2023-07-28
Map all HTTP 403 errors to S3RequestError::Forbidden

jamesbornholt:
## Description of change

There's a long list of potential error codes that all map to 403, and we
only picked a few. Mapping them all makes what happened a little clearer
in logs. They're not all "permissions" errors, so I renamed it to the
more generic "forbidden" (which is what HTTP 403 is).

No.

---


passaro (approved):
LGTM

PR #420, opened 2023-07-31
Build infrastructure for release packages

jamesbornholt:
## Description of change

This commit adds a first pass at build infrastructure for assembling Mountpoint releases. It builds RPM and DEB packages and a tar.gz archive for distributions that don't use those package managers. The RPM and DEB are preferred because they check for the necessary dependencies. The new README explains more about how to use it.

No.

---


jamesbornholt:
The artifacts show up at the bottom of the workflow summary page, if you want to poke at them: https://github.com/awslabs/mountpoint-s3/actions/runs/5710344533?pr=420

.github/workflows/tests.yml
```diff
@@ -269,3 +269,41 @@ jobs:
           override: true
       - name: Run cargo deny
         uses: EmbarkStudios/cargo-deny-action@v1
+
+  packages:
```
monthonk, line 273: I think we should move it into another config file, we probably don't want to run it on every pull request?

release/release.py
```diff
@@ -0,0 +1,331 @@
...
+    return artifacts
+
+
+if __name__ == "__main__":
+    p = argparse.ArgumentParser()
+    p.add_argument("--root-dir", help="override the path to the Cargo workspace")
+    p.add_argument("--expected-version", help="expected version number for the Mountpoint binary")
+    p.add_argument("--no-rpm", action="store_true", help="do not build an RPM")
+    p.add_argument("--no-deb", action="store_true", help="do not build a DEB")
+    p.add_argument("--unofficial", action="store_true", help="tag the release as unofficial (for CI use)")
```
monthonk, line 326: I would prefer this to be `unofficial` by default, people could run this script on their own and miss adding the flag, result in unintentional "official" release.

monthonk (commented):
Looks good and the script works well on my machine, only have a few comments.

release/release.py
monthonk, line 1: Maybe we should call it `package.py`?

release/attribution.hbs
```diff
@@ -0,0 +1,13 @@
...
+
+{{#each licenses}}
+## {{name}}
+
+Used by:
+{{#each used_by}}
+* {{crate.name}} {{crate.version}} - {{#if crate.repository}}{{crate.repository}}{{else}}https://crates.io/crates/{{crate.name}}{{/if}}
+{{/each}}
+
+{{text}}
```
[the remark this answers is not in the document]
jamesbornholt: Fixed this.

release/attribution.toml
```diff
@@ -0,0 +1,34 @@
...
+license = "BSD-2-Clause"
+[[zerocopy.clarify.files]]
+path = "LICENSE"
+checksum = "83c1763356e822adde0a2cae748d938a73fdc263849ccff6b27776dff213bd32"
+
+[zerocopy-derive.clarify]
+license = "BSD-2-Clause"
+[[zerocopy-derive.clarify.files]]
+path = "LICENSE"
+checksum = "83c1763356e822adde0a2cae748d938a73fdc263849ccff6b27776dff213bd32"
```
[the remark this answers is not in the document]
jamesbornholt: Not sure! Without this, cargo-about complains that it can't find the license. cargo-deny seems a little better at finding the license file:
```
help[accepted]: license requirements satisfied
  ┌─ zerocopy 0.6.1 (registry+https://github.com/rust-lang/crates.io-index):5:15
  │
4 │ license = ""
  │            - license expression was not specified
5 │ files-expr = "BSD-2-Clause"
  │               ^^^^^^^^^^^^
  │               │
  │               license expression retrieved via LICENSE
  │               accepted: license is explicitly allowed
  │
  = zerocopy v0.6.1
    └── fuser v0.12.0
        └── mountpoint-s3 v0.3.0

help[accepted]: license requirements satisfied
  ┌─ zerocopy-derive 0.3.2 (registry+https://github.com/rust-lang/crates.io-index):5:15
  │
4 │ license = ""
  │            - license expression was not specified
5 │ files-expr = "BSD-2-Clause"
  │               ^^^^^^^^^^^^
  │               │
  │               license expression retrieved via LICENSE
  │               accepted: license is explicitly allowed
  │
  = zerocopy-derive v0.3.2
    └── zerocopy v0.6.1
        └── fuser v0.12.0
            └── mountpoint-s3 v0.3.0
```

release/mount-s3.debian-control
```diff
@@ -0,0 +1,15 @@
+Package: mount-s3
+Maintainer: Amazon Web Services <mountpoint-s3@amazon.com>
```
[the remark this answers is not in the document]
jamesbornholt: Yep!

release/mount-s3.spec
```diff
@@ -0,0 +1,39 @@
...
+Version:        %{MOUNTPOINT_VERSION}
+Release:        1
+Summary:        Mountpoint for Amazon S3
+
+License:        Apache-2.0
+URL:            https://github.com/awslabs/mountpoint-s3
+Source0:        mount-s3.tar.gz
+
+Requires:       fuse
+Requires:       fuse-libs
```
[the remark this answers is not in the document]
jamesbornholt: `rpmbuild` infers the required libc dependency automatically. This should include the ca-certificates dependency, though.

release/release.py
```diff
@@ -0,0 +1,331 @@
...
+        if shutil.which(dep) is None:
+            raise Exception(f"`{dep}` must be installed")
+
+    output = run(["whereis", "libfuse"])
+    if b"libfuse.so" not in output:
+        raise Exception(f"libfuse not found (whereis output: {output})")
+
+    output = run(["whereis", "libfuse3"])
+    if b"libfuse3.so" in output:
+        raise Exception(f"libfuse3 should not be installed (whereis output: {output})")
```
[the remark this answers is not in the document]
jamesbornholt: This is checking that libfuse3 _isn't_ present, so we don't accidentally link against it.

release/release.py
jamesbornholt: Yeah, done.

release/release.py
jamesbornholt: Done (and the directory itself too)

.github/workflows/tests.yml
jamesbornholt: Actually, I switched it back to running on PRs too — I think we want to find out early if we broke something in our packaging process, like taking a dependency on a too-new crate or something that doesn't build in our packaging environment.

dannycjones: Good point, makes sense.

release/attribution.toml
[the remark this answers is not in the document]
dannycjones: Weird. The two projects are both owned by the same org.

Ah well, not an issue.

monthonk (approved):

package/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+        wget && \
+    yum clean all
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH="/root/.cargo/bin:$PATH"
+
+RUN wget -q "https://github.com/EmbarkStudios/cargo-about/releases/download/0.5.6/cargo-about-0.5.6-$(uname -p)-unknown-linux-musl.tar.gz" && \
+    wget -q "https://github.com/EmbarkStudios/cargo-about/releases/download/0.5.6/cargo-about-0.5.6-$(uname -p)-unknown-linux-musl.tar.gz.sha256" && \
+    echo -n "$(sha256sum cargo-about-*.tar.gz | cut -d' ' -f1)" > checksum.sha256 && \
+    diff cargo-about-*.tar.gz.sha256 checksum.sha256 && \
```
dannycjones, line 26: I was thinking that we compared w/ a hardcoded SHA256 here.

jamesbornholt: It's actually pretty annoying to do this because of the two architectures.

Issue #274, opened 2023-06-02
Support configuring AWS region via `AWS_REGION`


dannycjones:
### Tell us more about this new feature.

I should be able to configure the region using environment variables.

Specifically, this is the case we should support: `AWS_REGION`.

dannycjones:
Open question: if this is set but the bucket is in a different region, should we correct or fail when we discover this?

I suspect fail.

jamesbornholt:
We'd just use the `AWS_REGION` environment variable as a fallback for the `--region` command-line argument, I think. It should work the same way as `AWS_PROFILE`/`--profile`: prefer the command line argument if provided, otherwise the environment variable if provided, otherwise None.

jamesbornholt:
#422

PR #422, opened 2023-08-01
Use AWS_REGION and IMDS to determine region

passaro:
## Description of change

Determine the region using the following sources (in order):
  * `--region` flag,
  * `AWS_REGION` environment variable,
  * EC2 instance region (using the IMDS client),
  * default region (us-east-1).
 

Relevant issues: #274 

The change extends existing behavior by adding support for the `AWS_REGION` environment variable and by using IMDS to retrieve the EC2 instance region (if the user did not specify a region).

---

### DIFF
--- a/mountpoint-s3-client/src/imds_crt_client.rs
+++ b/mountpoint-s3-client/src/imds_crt_client.rs
@@ -1,16 +1,11 @@
-use std::pin::Pin;
-use std::task::Context;
-use std::task::Poll;
-
 use futures::channel::oneshot;
-use futures::Future;
 use mountpoint_s3_crt::auth::imds_client::{ImdsClient, ImdsClientConfig};
 use mountpoint_s3_crt::common::allocator::Allocator;
 use mountpoint_s3_crt::common::error;
 use mountpoint_s3_crt::io::channel_bootstrap::{ClientBootstrap, ClientBootstrapOptions};
 use mountpoint_s3_crt::io::event_loop::EventLoopGroup;
 use mountpoint_s3_crt::io::host_resolver::{HostResolver, HostResolverDefaultOptions};
-use pin_project::pin_project;
+use serde_json::Value;
 use thiserror::Error;
 
 #[derive(Debug)]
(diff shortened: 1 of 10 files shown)


mountpoint-s3-crt/src/auth/imds_client.rs
```diff
@@ -83,6 +83,29 @@ impl ImdsClient {
                 .ok_or_last_error()
         }
     }
+
+    /// Get the value for the given resource path.
+    pub fn get_resource<F>(&self, resource_path: &str, callback: F) -> Result<(), Error>
```
jamesbornholt, line 88: Should we just use this for `get_instance_type` too? It seems semi-random to me which IMDS resources have methods in `aws_imds_client` and which don't, so maybe we should just have this one generic method instead?

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -56,32 +50,30 @@ impl ImdsCrtClient {
     }
 
     /// Query the type of the EC2 instance this code is executed on.
-    pub fn make_instance_type_query(&self) -> Result<ImdsQueryRequest, ImdsQueryRequestError> {
+    pub async fn get_instance_type(&self) -> Result<String, ImdsQueryRequestError> {
```
jamesbornholt, line 53: Not for this PR, but maybe we should switch to just grabbing the [instance identity document](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html) since we can do it in one query instead of two.

mountpoint-s3-client/src/imds_crt_client.rs
```diff
@@ -56,32 +50,30 @@ impl ImdsCrtClient {
...
-/// ImdsQueryRequest is a data encapsulation wrapper for asnychronous instance metadata query.
-pub struct ImdsQueryRequest {
-    #[pin]
-    receiver: oneshot::Receiver<Result<String, ImdsQueryRequestError>>,
-}
-
-impl Future for ImdsQueryRequest {
-    type Output = Result<String, ImdsQueryRequestError>;
+    /// Query the region of the EC2 instance this code is executed on.
+    pub async fn get_region(&self) -> Result<String, ImdsQueryRequestError> {
```
jamesbornholt, line 66: So the test failures you're getting are because the tests are hitting Azure's IMDS, which listens on the same IP and I guess is where GitHub Actions runs. That's an unusual thing but we probably shouldn't totally choke. Maybe some cheap regex check on the string, like `^[A-Za-z0-9-]+$`, is enough to get only valid-ish regions without overfitting.

jamesbornholt (commented):
Mostly LGTM, just need to decide what to do about the test failures when running on GitHub Actions.

mountpoint-s3-client/src/imds_crt_client.rs
passaro: Yes, maybe we should do that. But I'm not sure I understand why that string is returned:
* locally I get an empty string, but the status code is 403
* github actions, I *guess* the status code is 404
It seems that the crt does not check the status code before calling our callback.

mountpoint-s3-client/src/imds_crt_client.rs
passaro: Tempted about this: it would also "solve" the issue with spurious replies below.

passaro: I've switched to getting the instance identity document and parsing it. This saves us a query and mitigates the issue with spurious responses, since they will fail to parse correctly. The drawback is some added complexity in `main.rs` where we create and pass around a lazy `InstanceInfo` to query the Imds once and only if required. We may eventually want to refactor the logic around that, but I would say not in this PR.

mountpoint-s3/src/main.rs
```diff
@@ -371,16 +372,25 @@ fn mount(args: CliArgs) -> anyhow::Result<FuseSession> {
...
-                    DEFAULT_TARGET_THROUGHPUT
-                }
-            });
+    let region = args.region.or_else(|| {
+        let region = env_region();
+        if let Some(region) = &region {
+            tracing::debug!("using AWS_REGION: {region}");
+        };
+        region
+    });
```
jamesbornholt, line 381: push this logic into `get_region`? check the env var first if `provided_region` is None

passaro: Slight complication that the retry logic needs to know whether the user provided a region. But I'll think how to do that.

jamesbornholt (approved):

PR #423, opened 2023-08-01
Parse endpoint property of Auth Scheme from endpoint resolver

sauraank:
## Description of change
Got the properties in json string format from CRT. Parsed it to struct `AuthScheme` whose fields can later be mapped to signing config to have the values of signing names and signing region for each request separately.

Got the properties of endpoint in the Rust Wrapper of CRT and added tests for it. Parsed the properties to get the `AuthScheme` struct.

Relevant issues: 
#4 

Not yet. But this might change the way signing config is handled later.

No

---

### DIFF
--- a/mountpoint-s3-client/src/endpoint_config.rs
+++ b/mountpoint-s3-client/src/endpoint_config.rs
@@ -243,7 +243,7 @@ pub enum EndpointError {
     InvalidUri(#[from] InvalidUriError),
     #[error("endpoint could not be resolved")]
     UnresolvedEndpoint(#[from] ResolverError),
-    #[error("Endpoint properties not resolved")]
+    #[error("Endpoint properties could not be parsed")]
     ParseError(#[from] serde_json::Error),
     #[error("AuthScheme field missing: {0}")]
     InvalidAuthScheme(String),
--- a/mountpoint-s3-client/src/endpoint_config.rs
+++ b/mountpoint-s3-client/src/endpoint_config.rs
@@ -1,6 +1,8 @@
+use std::os::unix::prelude::OsStrExt;
+
 use mountpoint_s3_crt::{
     common::{allocator::Allocator, uri::Uri},
-    s3::endpoint_resolver::{RequestContext, ResolverError, RuleEngine},
+    s3::endpoint_resolver::{RequestContext, ResolvedEndpoint, ResolverError, RuleEngine},
 };
 use thiserror::Error;
 
@@ -13,6 +15,37 @@ pub enum AddressingStyle {
     Path,
 }
 
+#[derive(Debug, Clone)]
+pub struct AuthScheme {
+    disable_double_encoding: bool,
+    scheme_name: String,
+    signing_name: String,
+    signing_region: String,
+}
+
+impl AuthScheme {
+    /// Get the siging name from [AuthScheme]
+    pub fn signing_name(&self) -> &str {
+        &self.signing_name
+    }
+
+    /// Get the signing region from [AuthScheme]
+    pub fn signing_region(&self) -> &str {
+        &self.signing_region
+    }
+
+    /// Get Disable double encoding value for [AuthScheme]
+    pub fn disable_double_encoding(&self) -> bool {
+        self.disable_double_encoding
+    }
+
+    /// Get the name of [AuthScheme]
+    pub fn scheme_name(&self) -> &str {
+        &self.scheme_name
+    }
+}
+
+/// Configuration for resolution of Endpoint
 #[derive(Debug, Clone)]
 pub struct EndpointConfig {
     region: String,
@@ -92,6 +125,7 @@ impl EndpointConfig {
     pub fn get_accelerate(&self) -> bool {
         self.use_accelerate
     }
+
     /// get the dual stack config from the [EndpointConfig]
     pub fn get_dual_stack(&self) -> bool {
         self.use_dual_stack
@@ -108,7 +142,7 @@ impl EndpointConfig {
     }
 
     /// resolve the endpoint from the [EndpointConfig] and the bucket name
-    pub fn resolve_for_bucket(&self, bucket: &str) -> Result<Uri, EndpointError> {
+    pub fn resolve_for_bucket(&self, bucket: &str) -> Result<ResolvedEndpointInfo, EndpointError> {
         let allocator = Allocator::default();
         let mut endpoint_request_context: RequestContext = RequestContext::new(&allocator).unwrap();
         let endpoint_rule_engine = RuleEngine::new(&allocator).unwrap();
@@ -148,9 +182,58 @@ impl EndpointConfig {
         let resolved_endpoint = endpoint_rule_engine
             .resolve(endpoint_request_context)
             .map_err(EndpointError::UnresolvedEndpoint)?;
-        let endpoint_uri = resolved_endpoint.get_url();
-        Uri::new_from_str(&allocator, endpoint_uri)
+
+        Ok(ResolvedEndpointInfo(resolved_endpoint))
+    }
+}
+
+/// Wrapper for [ResolvedEndpoint] from CRT to get [Uri] and [AuthScheme]
+#[derive(Debug)]
+pub struct ResolvedEndpointInfo(ResolvedEndpoint);
+
+impl ResolvedEndpointInfo {
+    /// Get the [Uri] from [ResolvedEndpointInfo]
+    pub fn uri(&self) -> Result<Uri, EndpointError> {
+        let allocator = Allocator::default();
+        let endpoint_url = self.0.get_url();
+        Uri::new_from_str(&allocator, endpoint_url)
             .map_err(|e| EndpointError::InvalidUri(InvalidUriError::CouldNotParse(e)))
+    }
+
+    /// Get the [AuthScheme] from [ResolvedEndpointInfo] for the signing config
+    pub fn auth_scheme(&self) -> Result<AuthScheme, EndpointError> {
+        // ResolvedEndpoint is wrapper for aws_endpoints_resolved_endpoint which has url, properties and header for the endpoint.
+        // Property if in json format containing the AuthScheme. Egs. -
+        // {\"authSchemes\":[{\"disableDoubleEncoding\":true,\"name\":\"sigv4\",\"signingName\":\"s3\",\"signingRegion\":\"us-east-2\"}]}
+        let endpoint_properties = self.0.get_properties();
+        let auth_scheme_data: serde_json::Value = serde_json::from_slice(endpoint_properties.as_bytes())?;
+        let auth_scheme_value = auth_scheme_data
+            .get("authSchemes")
+            .ok_or_else(|| EndpointError::InvalidAuthScheme("AuthScheme".to_owned()))?;
+        let auth_scheme_value = &auth_scheme_value[0];
+        let disable_double_encoding = auth_scheme_value
+            .get("disableDoubleEncoding")
+            .and_then(|t| t.as_bool())
+            .ok_or_else(|| EndpointError::InvalidAuthScheme("disableDoubleEncoding".to_owned()))?;
+        let scheme_name = auth_scheme_value
+            .get("name")
+            .and_then(|t| t.as_str())
+            .ok_or_else(|| EndpointError::InvalidAuthScheme("name".to_owned()))?;
+        let signing_name = auth_scheme_value
+            .get("signingName")
+            .and_then(|t| t.as_str())
+            .ok_or_else(|| EndpointError::InvalidAuthScheme("signingName".to_owned()))?;
+        let signing_region = auth_scheme_value
+            .get("signingRegion")
+            .and_then(|t| t.as_str())
+            .unwrap_or("*");
+
+        Ok(AuthScheme {
+            disable_double_encoding,
+            scheme_name: scheme_name.to_owned(),
+            signing_name: signing_name.to_owned(),
+            signing_region: signing_region.to_owned(),
+        })
     }
 }
 
@@ -160,6 +243,10 @@ pub enum EndpointError {
     InvalidUri(#[from] InvalidUriError),
     #[error("endpoint could not be resolved")]
     UnresolvedEndpoint(#[from] ResolverError),
+    #[error("Endpoint properties not resolved")]
+    ParseError(#[from] serde_json::Error),
+    #[error("AuthScheme field missing: {0}")]
+    InvalidAuthScheme(String),
 }
 
 #[derive(Debug, Error)]
@@ -175,7 +262,11 @@ mod test {
     #[test]
     fn test_virtual_addr() {
         let endpoint_config = EndpointConfig::new("eu-west-1").addressing_style(AddressingStyle::Automatic);
-        let endpoint_uri = endpoint_config.resolve_for_bucket("doc-example-bucket").unwrap();
+        let endpoint_uri = endpoint_config
+            .resolve_for_bucket("doc-example-bucket")
+            .unwrap()
+            .uri()
+            .unwrap();
         assert_eq!(
             "https://doc-example-bucket.s3.eu-west-1.amazonaws.com",
             endpoint_uri.as_os_str()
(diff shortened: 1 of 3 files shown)


mountpoint-s3-client/Cargo.toml
```diff
@@ -33,9 +33,10 @@ thiserror = "1.0.34"
 time = { version = "0.3.17", features = ["formatting", "parsing"] }
 tracing = { version = "0.1.35", default-features = false, features = ["std", "log"] }
 xmltree = "0.10.3"
+serde_json = "1.0.95"
+anyhow = { version = "1.0.64", features = ["backtrace"] }
```
passaro, line 37: Revert. We do not want to use anyhow in the client.

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -108,7 +113,7 @@ impl EndpointConfig {
     }
 
     /// resolve the endpoint from the [EndpointConfig] and the bucket name
-    pub fn resolve_for_bucket(&self, bucket: &str) -> Result<Uri, EndpointError> {
+    pub fn resolve_for_bucket(&self, bucket: &str) -> Result<ResolvedEndpoint, EndpointError> {
```
passaro, line 116: We probably need to wrap `ResolvedEndpoint` (which is from `mountpoint-s3-crt`) in a new type defined here

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -145,15 +150,50 @@ impl EndpointConfig {
...
             .resolve(endpoint_request_context)
-            .map_err(EndpointError::UnresolvedEndpoint)?;
-        let endpoint_uri = resolved_endpoint.get_url();
-        Uri::new_from_str(&allocator, endpoint_uri)
-            .map_err(|e| EndpointError::InvalidUri(InvalidUriError::CouldNotParse(e)))
+            .map_err(EndpointError::UnresolvedEndpoint)
     }
 }
 
+pub fn get_uri_from_endpoint(resolved_endpoint: ResolvedEndpoint) -> Result<Uri, EndpointError> {
```
passaro, line 159: both this and `get_auth_scheme_from_endpoint` should really be methods of a new `ResolvedEndpoint` wrapper to be defined here

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -145,15 +150,50 @@ impl EndpointConfig {
...
+    let allocator = Allocator::default();
+    let endpoint_url = resolved_endpoint.get_url();
+    Uri::new_from_str(&allocator, endpoint_url)
+        .map_err(|e| EndpointError::InvalidUri(InvalidUriError::CouldNotParse(e)))
+}
+
+pub fn get_auth_scheme_from_endpoint(resolved_endpoint: ResolvedEndpoint) -> Result<AuthScheme, anyhow::Error> {
```
passaro, line 166: The error here should be a new case in `EndpointError` (or even a new error type - but okay to keep it simple)

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -74,6 +74,55 @@ macro_rules! event {
...
 
+#[derive(Debug, Clone)]
+pub struct AuthScheme {
+    disable_double_encoding: bool,
+    scheme_name: String,
+    signing_name: String,
+    signing_region: String,
+}
+
+impl Default for AuthScheme {
```
passaro, line 85: Is the builder pattern (and `Default`) needed here? It seems to me that a simple `new()` would work fine.

passaro (changes requested):

mountpoint-s3-client/src/endpoint_config.rs
sauraank: But, Uri is also from `mountpoint-s3-crt`?

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -148,10 +152,57 @@ impl EndpointConfig {
...
             .map_err(EndpointError::UnresolvedEndpoint)?;
-        let endpoint_uri = resolved_endpoint.get_url();
-        Uri::new_from_str(&allocator, endpoint_uri)
+
+        Ok(ResolvedEndpointInfo(resolved_endpoint))
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct ResolvedEndpointInfo(ResolvedEndpoint);
```
jamesbornholt, line 161: nit: needs a doc comment

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -148,10 +152,57 @@ impl EndpointConfig {
...
+
+impl ResolvedEndpointInfo {
+    pub fn get_uri_from_endpoint(&self) -> Result<Uri, EndpointError> {
```
jamesbornholt, line 164: nit: `uri`, and needs a doc comment

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -148,10 +152,57 @@ impl EndpointConfig {
...
+        let allocator = Allocator::default();
+        let endpoint_url = self.0.get_url();
+        Uri::new_from_str(&allocator, endpoint_url)
             .map_err(|e| EndpointError::InvalidUri(InvalidUriError::CouldNotParse(e)))
     }
+
+    pub fn get_auth_scheme_from_endpoint(&self) -> Result<AuthScheme, EndpointError> {
```
jamesbornholt, line 171: nit: `auth_scheme` and needs a doc comment

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -148,10 +152,57 @@ impl EndpointConfig {
...
+        let endpoint_properties = self.0.get_properties();
```
jamesbornholt, line 172: Might be useful to add a comment inside this method with an example of an output from the resolver, so people reading the code know what the shape of it is.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -74,6 +74,48 @@ macro_rules! event {
     }
 }
 
+#[derive(Debug, Clone, Default)]
```
jamesbornholt, line 77: I don't think empty string is the right default for most of these, so should manually implement `Default`. But actually, I don't think we want this to implement `Default`, because there's no sensible default value for `signing_region`?

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -74,6 +74,48 @@ macro_rules! event {
...
+pub struct AuthScheme {
+    disable_double_encoding: bool,
+    scheme_name: String,
+    signing_name: String,
+    signing_region: String,
+}
+
+impl AuthScheme {
+    pub fn new() -> Self {
```
jamesbornholt, line 86: `signing_region` should probably be a required argument here

mountpoint-s3-crt/src/s3/endpoint_resolver.rs
```diff
@@ -147,7 +147,7 @@ impl Drop for RequestContext {
 }
 
 /// [ResolvedEndpoint] for the given context using [RuleEngine] and rule set
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
jamesbornholt, line 150: `aws_endpoints_resolved_endpoint` is reference counted, so this can't derive `Clone`. If you need to clone, you need to manually implement `Clone` to include a call to `aws_endpoints_resolved_endpoint_acquire`.

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -160,6 +211,16 @@ pub enum EndpointError {
     InvalidUri(#[from] InvalidUriError),
     #[error("endpoint could not be resolved")]
     UnresolvedEndpoint(#[from] ResolverError),
+    #[error("Endpoint properties not resolved")]
+    SerdeError(serde_json::Error),
```
jamesbornholt, line 215: nit: `Serde` shouldn't be in the name, and we want the `from` part too
```suggestion
    ParseError(#{from] serde_json::Error),
```

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -148,10 +152,57 @@ impl EndpointConfig {
...
+            .and_then(|t| t.as_str())
+            .ok_or_else(|| {
+                EndpointError::InvalidAuthScheme("signingName is missing from authScheme json.".to_owned())
+            })?;
+        let signing_region = auth_scheme_value
+            .get("signingRegionSet")
+            .and_then(|t| t.as_str())
+            .ok_or_else(|| {
+                EndpointError::InvalidAuthScheme("signingRegionSet is missing from authScheme json.".to_owned())
+            })?;
```
jamesbornholt, line 198: Are all five of these fields always present -- is failing the right thing to do if any of them are missing?

I think you were saying earlier that not every resolution result includes this structure at all, so maybe this should return `Result<Option<AuthScheme>, ...>`?

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -148,10 +152,57 @@ impl EndpointConfig {
...
+        Uri::new_from_str(&allocator, endpoint_url)
             .map_err(|e| EndpointError::InvalidUri(InvalidUriError::CouldNotParse(e)))
     }
+
+    pub fn get_auth_scheme_from_endpoint(&self) -> Result<AuthScheme, EndpointError> {
+        let endpoint_properties = self.0.get_properties();
+        let auth_scheme_data: serde_json::Value = serde_json::from_slice(endpoint_properties.as_bytes())?;
+        let auth_scheme_value = auth_scheme_data
+            .get("authSchemes")
+            .ok_or_else(|| EndpointError::InvalidAuthScheme("AuthScheme is missing from properties".to_owned()))?;
```
jamesbornholt, line 176: These error strings can all just be e.g. `EndpointError::InvalidAuthScheme("authSchemes")`, don't need the extra words (the `#[error("AuthScheme field missing: {0}")]` already provides them).

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -1,9 +1,13 @@
+use std::os::unix::prelude::OsStrExt;
+
 use mountpoint_s3_crt::{
     common::{allocator::Allocator, uri::Uri},
-    s3::endpoint_resolver::{RequestContext, ResolverError, RuleEngine},
+    s3::endpoint_resolver::{RequestContext, ResolvedEndpoint, ResolverError, RuleEngine},
 };
 use thiserror::Error;
 
+use crate::s3_crt_client::AuthScheme;
```
jamesbornholt, line 9: I think `AuthScheme` probably belongs in this file. You can make the fields `pub(crate)` if you need to read them from `s3_crt_client`.

mountpoint-s3-crt/src/s3/endpoint_resolver.rs
```diff
@@ -307,4 +322,17 @@ mod test {
...
+    #[test_case("arn:aws:s3:us-east-2:AccountID:accesspoint/access-point-test", "{\"authSchemes\":[{\"disableDoubleEncoding\":true,\"name\":\"sigv4\",\"signingName\":\"s3\",\"signingRegion\":\"us-east-2\"}]}"; "Access Point ARN")]
+    #[test_case("arn:aws:s3::accountID:accesspoint/s3-bucket-test.mrap", "{\"authSchemes\":[{\"disableDoubleEncoding\":true,\"name\":\"sigv4a\",\"signingName\":\"s3\",\"signingRegionSet\":[\"*\"]}]}"; "MRAP")]
+    fn test_endpoint_properties_for_arn(arn: &str, endpoint_property: &str) {
+        let new_allocator = Allocator::default();
+        let endpoint_rule_engine = RuleEngine::new(&new_allocator).unwrap();
+        let mut endpoint_request_context = RequestContext::new(&new_allocator).unwrap();
+        test_endpoint_resolver_init(arn, "us-east-1", &new_allocator, &mut endpoint_request_context);
+        let resolved_endpoint = endpoint_rule_engine.resolve(endpoint_request_context).unwrap();
+        let property = resolved_endpoint.get_properties();
+        assert_eq!(property, endpoint_property);
```
jamesbornholt, line 336: This test should also call `get_auth_scheme_for_endpoint` (or whatever you rename it to) and check the results. Might be easier without the `test_case` macro, not sure though.

mountpoint-s3-client/src/s3_crt_client.rs
sauraank: I think we can have empty string here. If we find empty string or '*' (as in the case of Multi Region Access Points), we can set it to `EndpointConfig` region.

mountpoint-s3-crt/src/s3/endpoint_resolver.rs
sauraank: I can implement `Clone`. But, clone was required because I wanted to have a wrapper struct on `ResolvedEndpoint`. But I wanted to know, why do we actually want it?

mountpoint-s3-crt/src/s3/endpoint_resolver.rs
sauraank: This is inside CRT right? I can add tests in `mountpoint-client` for the `AuthScheme`.

mountpoint-s3-client/src/endpoint_config.rs
sauraank: Like the region is set as '*' when it is not part of ARN. Not found any case where they are missing in testcases. Need to test normal bucket and aliases.

mountpoint-s3-crt/src/s3/endpoint_resolver.rs
passaro: Do we need `Clone` on the wrapper in client?

The reason for the wrapper is that `ResolveEndpoint` is a lower-level type that exposes `get_uri` and `get_properties` which return `OsString`s. In the client we want to return a similar "Resolved Endpoint" type, but which provides methods returning already parsed `Uri` and `AuthScheme` types.

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -160,6 +243,10 @@ pub enum EndpointError {
     InvalidUri(#[from] InvalidUriError),
     #[error("endpoint could not be resolved")]
     UnresolvedEndpoint(#[from] ResolverError),
+    #[error("Endpoint properties not resolved")]
```
jamesbornholt, line 246: nit:
```suggestion
    #[error("Endpoint properties could not be parsed")]
```

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -148,10 +182,59 @@ impl EndpointConfig {
...
             .map_err(EndpointError::UnresolvedEndpoint)?;
-        let endpoint_uri = resolved_endpoint.get_url();
-        Uri::new_from_str(&allocator, endpoint_uri)
+
+        Ok(ResolvedEndpointInfo(resolved_endpoint))
+    }
+}
+
+/// Wrapper for [ResolvedEndpoint] from CRT to get [Uri] and [AuthScheme]
+#[derive(Debug, Clone)]
```
jamesbornholt, line 191: Maybe I'm not looking hard enough, but I don't see anywhere that needs this to be `Clone`, which also means you don't need `ResolvedEndpoint` to be `Clone`.

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -13,6 +15,37 @@ pub enum AddressingStyle {
...
+
+    /// Get Disable double encoding value for [AuthScheme]
+    pub fn get_disable_double_encoding(&self) -> bool {
+        self.disable_double_encoding
+    }
+
+    /// Get the name of [AuthScheme]
+    pub fn get_scheme_name(&self) -> String {
+        self.scheme_name.clone()
+    }
```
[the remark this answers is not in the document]
sauraank: Why do we not need get on the name? We would need those values for signing config, right? (At least signing name)

jamesbornholt (approved):

passaro (approved):

PR #427, opened 2023-08-03
Print a message when mount succeeds

jamesbornholt:
## Description of change

It's not obvious that mounting succeeded in the default background mode -- the process just silently exits. Let's make it a bit more obvious that things went well.

No.

---

### DIFF
--- a/mountpoint-s3/src/main.rs
+++ b/mountpoint-s3/src/main.rs
@@ -249,6 +249,15 @@ impl CliArgs {
             default_filter,
         }
     }
+
+    /// Human-readable description of the bucket being mounted
+    fn bucket_description(&self) -> String {
+        if let Some(prefix) = self.prefix.as_ref() {
+            format!("prefix {} of bucket {}", prefix, self.bucket_name)
+        } else {
+            format!("bucket {}", self.bucket_name)
+        }
+    }
 }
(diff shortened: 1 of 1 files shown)


mountpoint-s3/src/main.rs
```diff
@@ -309,6 +318,9 @@ fn main() -> anyhow::Result<()> {
                 }
             }
             ForkResult::Parent { child } => {
+                // Re-parse CLI args in parent process because the child might move them
+                let args = CliArgs::parse();
```
dannycjones, line 322: I was reviewing this and what we already have. The child arm looks a little odd. I can see the child parses again too, but it _also_ uses the original args (in a PR I approved 😄 ).

Should we drop the original args so its clear? At least, I think we should shadow them in the `child` arm.

When reading into fork again, its mapping Copy-on-Write memory for both processes around the original memory of the process before the fork. So re-creating the args is just an ownership thing?

jamesbornholt: Yeah. It should be fine to be COW because it's just plain data. But just in case we ever added something fancier to that struct for some reason, better to just re-parse on both sides.

dannycjones (approved):
LGTM!

PR #428, opened 2023-08-03
Build SigningConfig from AuthScheme for each request

passaro:
## Description of change

Use the `AuthScheme` obtained from the `EndpointResolver` to build the `SigningConfig`
for each request. Also extend the initializer for `SigningConfig` to accept the
additional parameters: service name, signing algorithm, and the `use_double_uri_encode`
flag.

The `AuthScheme` will now validate the `scheme_name` field (i.e. signing algorithm)
on parsing and store it as a `SigningAlgorithm`.

## Does this change impact existing behavior?

No

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

PR #430, opened 2023-08-04
Added default signing config for each request

sauraank:
## Description of change
Now the signing config is set for each request in request template (although still with default values) rather than once in the client.

Added credential provider in S3CrtClientInner to have signing config set for each request template of S3CrtClient. Added the signing config in S3Message which later get set in options.

Relevant issues: 
#4 

No

No, there’s no breaking change.

---


passaro:
This change is now included in #428

PR #431, opened 2023-08-04
Change artifact names for aarch64 package

monthonk:
## Description of change

We will be using `arm64` in the package names for `aarch64` architecture.

No

---

### DIFF
--- a/package/package.py
+++ b/package/package.py
@@ -37,9 +37,10 @@ class BuildMetadata:
     version_string: str
     buildroot: str
     arch: str
+    arch_name: str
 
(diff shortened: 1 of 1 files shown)


package/package.py
```diff
@@ -37,9 +37,10 @@ class BuildMetadata:
     version_string: str
     buildroot: str
     arch: str
```
passaro, line 39: nit: do we still need it? may be confusing

passaro (approved):

package/package.py
monthonk: we still use it in some other places

PR #437, opened 2023-08-07
Remove --fips command-line flag

passaro:
## Description of change

Remove --fips command-line flag.

FIPS endpoints need to be manually specified.

---

### DIFF
--- a/mountpoint-s3/src/main.rs
+++ b/mountpoint-s3/src/main.rs
@@ -72,9 +72,6 @@ struct CliArgs {
     #[clap(long, help = "Use dual-stack endpoints when accessing S3", help_heading = BUCKET_OPTIONS_HEADER)]
     pub dual_stack: bool,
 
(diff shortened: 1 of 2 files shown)


mountpoint-s3/src/main.rs
```diff
@@ -391,7 +388,6 @@ fn mount(args: CliArgs) -> anyhow::Result<FuseSession> {
     // Placeholder region will be filled in by [create_client_for_bucket]
     let endpoint_config = EndpointConfig::new("PLACEHOLDER")
         .addressing_style(args.addressing_style())
-        .use_fips(args.fips)
```
sauraank, line 394: I think we should also remove it from EndpointConfig struct?

dannycjones: We should leave that it. In the future, we want to make FIPS endpoints easy to use through a flag when ready.

jamesbornholt (approved):

PR #440, opened 2023-08-09
Add an example using Mountpoint with PyTorch

jamesbornholt:
I'd like to start collecting a few examples of how to use Mountpoint for
stuff. This is the first one: using Mountpoint as a PyTorch data loader.
The goal is really just to show how to do it, and maybe say a little
about how well it works.

For now, this doesn't run in CI (need a GPU instance), will work on that
later.

I can imagine other examples we'd like to build and stick in this new `examples` directory:
* ffmpeg (like the blog post)
* genomics stuff

---

### DIFF
--- a/examples/pytorch/README.md
+++ b/examples/pytorch/README.md
@@ -1,0 +1,42 @@
+# PyTorch data loading with Mountpoint
+
+This directory shows a simple example of how to use Mountpoint for Amazon S3 as a PyTorch data loader via the torchdata library.
+It trains a ResNet-50 model for a few epochs using [PyTorch Lightning](https://www.pytorchlightning.ai/index.html), with synthetic ImageNet-style training data, stored in S3 as shards in WebDataset format.
+The idea is that because Mountpoint exposes your S3 bucket as a local file system, you can just use standard file-oriented loaders rather than writing S3-specific code, and get great performance.
+
+To get started, first install the dependencies (we recommend using the [AWS Deep Learning AMI for PyTorch](https://aws.amazon.com/machine-learning/amis/)):
+
+    pip install -r requirements.txt
+
+Now install Mountpoint if you don't already have it:
+
+    wget https://s3.amazonaws.com/mountpoint-s3-release/latest/x86_64/mount-s3.rpm
+    sudo yum install ./mount-s3.rpm
+
+To generate and upload the training shards to an S3 bucket, run:
+
+    python resnet.py make s3://DOC-EXAMPLE-BUCKET/shard-data/ --num-images 50000
+
+This will upload about 5GB worth of shards to your bucket.
+
+Now to run the training loop:
+
+    python resnet.py train s3://DOC-EXAMPLE-BUCKET/shard-data/ --source-kind mountpoint --batch-size 256 --max-epochs 3
+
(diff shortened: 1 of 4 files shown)


examples/pytorch/README.md
```diff
@@ -0,0 +1,42 @@
...
+    pip install -r requirements.txt
+
+Now install Mountpoint if you don't already have it:
+
+    wget https://s3.amazonaws.com/mountpoint-s3-release/latest/x86_64/mount-s3.rpm
+    sudo yum install ./mount-s3.rpm
+
+To generate and upload the training shards to an S3 bucket, run:
+
+    python resnet.py make s3://DOC-EXAMPLE-BUCKET/shard-data/ --num-images 50000
```
monthonk, line 18: I got a `ModuleNotFoundError` complaining about `certifi` module when running this command. Adding this module to the requirements file solved the problem for me. I think we need to update yours too.

examples/pytorch/README.md
```diff
@@ -0,0 +1,42 @@
+# PyTorch data loading with Mountpoint
+
+This directory shows a simple example of how to use Mountpoint for Amazon S3 as a PyTorch data loader via the torchdata library.
+It trains a ResNet-50 model for a few epochs using [PyTorch Lightning](https://www.pytorchlightning.ai/index.html), with synthetic ImageNet-style training data, stored in S3 as shards in WebDataset format.
```
monthonk, line 4: Would be great to have more info about the system you tested this workload, like os name or python version. Just by looking at the commands I would assume it's amazonlinux2023 and probably python2?

monthonk (changes requested):
Thanks for adding this. It's really nice to see how we can use Mountpoint in real world applications.

examples/pytorch/README.md
dannycjones: I followed it using AL2023 and python3, and had to use `python3` instead of `python`. It would be good to be explicit in terms of "The following instructions were run using Amazon Linux 2023 and Python 3."

Should we consider making sure these instructions work out of the box for Python 3 (regardless of OS)?

examples/pytorch/resnet.py
```diff
@@ -0,0 +1,187 @@
...
+
+def make_sharded_dataset(bucket: str, prefix: str, num_images: int, max_shard_size: int):
+    """Make a fake dataset in WebDataset format and upload it to S3. In reality you'd already have
+    this dataset in S3, so this is just for benchmarking purposes. We choose the image sizes to be
+    roughly ImageNet-esque."""
+
+    s3 = boto3.client("s3")
+    ds = torchvision.datasets.FakeData(size=num_images, image_size=(3, 224, 224), num_classes=100)
+
+    with tempfile.TemporaryDirectory() as tempdir:
```
tchaton, line 43: Would be nice to have the with / without shared benchmark. This requires code changes and this isn't always an assumption to be made about user dataset.

monthonk (approved):

dannycjones (approved):
Thanks James, super nice to have this example accessible with zero ML experience required.

examples/pytorch/resnet.py
jamesbornholt: Agreed! This is just a first example, but we're looking at non-sharded versions too.

PR #443, opened 2023-08-10
decrease closed-channel log severity

jchorl:
## Description of change

This is more of a proposal/asking for thoughts.

I've noticed this log is extremely noisy. There's a few bad omens here:
1. The log message doesn't contain much data itself:
    ```
    WARN mountpoint_s3::prefetch::part_queue: closed channel
    ```
2. It's noisy
3. It doesn't feel actionable - what am I supposed to do with this

Given the above, it feels `WARN` is too high. Feels more like info or debug. However I'm not sure if the log is useful in other contexts. Open to guidance!

### DIFF
--- a/mountpoint-s3/src/prefetch/part_queue.rs
+++ b/mountpoint-s3/src/prefetch/part_queue.rs
@@ -1,6 +1,6 @@
 use std::time::Instant;
 
-use tracing::info;
+use tracing::trace;
 
 use crate::prefetch::part::Part;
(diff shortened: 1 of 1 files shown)


jamesbornholt:
Thanks! This is definitely too spammy for what's actually expected behavior if you're doing random reads or otherwise not reading an entire object (is that where you're seeing this?). It should probably just be `trace!`.

jamesbornholt (approved):

PR #445, opened 2023-08-11
Refactor metrics to have sharded shared state

jamesbornholt:
## Description of change

I wanted to add some new metrics but they are gauges, and our current
thread-local approach makes gauges hard -- each thread gets its own copy
of the gauge. So instead, this change refactors the metrics
infrastructure to have just one copy of each metric, stored in a sharded
map to hopefully reduce contention.

I didn't actually add any new metrics, so this should be a pure
refactoring change. I beefed up the tests a little bit, too.

No, refactoring only.

---


monthonk (commented):
LGTM, just one typo.

monthonk (approved):

Issue #446, opened 2023-08-11
Improve error for no signing credential


sauraank:
### Tell us more about this new feature.

It is an enhancement of error message. Earlier we used to get -
```
Error: Failed to create S3 client

Caused by:
    0: HeadBucket failed for bucket bornholt-test-bucket in region us-east-1
    1: Client error
    2: Unknown response error: MetaRequestResult { response_status: 0, crt_error: Error(6146, "aws-c-auth: AWS_AUTH_SIGNING_NO_CREDENTIALS, Attempt to sign an http request without credentials"), error_response_headers: None, error_response_body: None }
Error: Failed to create mount process
```
which suggested that it is failed attempt rather than there are no aws signing credential provided.

jamesbornholt:
#447

PR #447, opened 2023-08-11
Improved Error message for Invalid Credential

sauraank:
## Description of change
Changed the error message for no signing credential.

Added another S3RequestError

Relevant issues: 
#446 

No

There is no breaking change.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -878,7 +878,7 @@ fn try_parse_generic_error(request_result: &MetaRequestResult) -> Option<S3Reque
         if crt_error_code == 6146 {
             Some(S3RequestError::NoSigningCredentials)
         } else {
-            None
+            Some(S3RequestError::CrtError(crt_error_code.into()))
         }
     }
 
@@ -1190,10 +1190,12 @@ mod tests {
     fn parse_test_other_crt_error() {
         // 6144 is crt error code for AWS_AUTH_SIGNING_UNSUPPORTED_ALGORITHM, which is another signing error,
         // but not no signing credential error
-        let result = make_crt_error_result(0, 6144.into());
-        let error_result = try_parse_generic_error(&result);
-        let None = error_result else {
-            panic!("wrong result, got: {:?}", error_result);
+        let error_code = 6144;
+        let result = make_crt_error_result(0, error_code.into());
+        let result = try_parse_generic_error(&result);
+        let Some(S3RequestError::CrtError(error)) = result else {
+            panic!("wrong result, got: {:?}", result);
         };
+        assert_eq!(error, error_code.into());
     }
 }
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -766,6 +766,9 @@ pub enum S3RequestError {
     /// Forbidden
     #[error("Forbidden: {0}")]
     Forbidden(String),
+
+    #[error("No signing credentials provided")]
+    NoSigningCredentials,
 }
 
 impl S3RequestError {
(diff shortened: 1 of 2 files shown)


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -868,12 +871,23 @@ fn try_parse_generic_error(request_result: &MetaRequestResult) -> Option<S3Reque
         }
     }
 
+    /// Try to look for error related to no signing credentials
+    fn try_parse_no_credentials(request_result: &MetaRequestResult) -> Option<S3RequestError> {
+        let crt_error_code = request_result.crt_error.raw_error();
+        if crt_error_code == 6146 {
+            Some(S3RequestError::NoSigningCredentials)
+        } else {
+            None
```
monthonk, line 880: I think these cases are all CRT errors because normal http status shouldn't be zero? if that's true we can return `CrtError` here.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -868,12 +871,23 @@ fn try_parse_generic_error(request_result: &MetaRequestResult) -> Option<S3Reque
...
+        }
+    }
+
     match request_result.response_status {
         301 => try_parse_redirect(request_result),
         // 400 is overloaded, it can be an access error (invalid token) or (for MRAP) a bucket
         // redirect
         400 => try_parse_forbidden(request_result).or_else(|| try_parse_redirect(request_result)),
         403 => try_parse_forbidden(request_result),
+        0 => try_parse_no_credentials(request_result),
```
monthonk, line 890: Can we add some unit tests for this arm?

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -766,6 +766,9 @@ pub enum S3RequestError {
     /// Forbidden
     #[error("Forbidden: {0}")]
     Forbidden(String),
+
+    #[error("No signing credentials provided")]
+    NoSigningCredentials,
```
monthonk, line 771: Add a comment to explain what is this error.

mountpoint-s3-client/src/s3_crt_client.rs
sauraank: But, if they are `None`, it will still return `ObjectClientError::ClientError(S3RequestError::ResponseError(request_result)))` in the `make_meta_request_from_options` method. So, I am not sure whether we should return `CrtError` here.

monthonk: I think that will turn into `Unknown response error` when being printed out to the console, it's more meaningful to return `CrtError` when we already know that.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -868,12 +872,23 @@ fn try_parse_generic_error(request_result: &MetaRequestResult) -> Option<S3Reque
         }
     }
 
+    /// Try to look for error related to no signing credentials
+    fn try_parse_no_credentials(request_result: &MetaRequestResult) -> Option<S3RequestError> {
+        let crt_error_code = request_result.crt_error.raw_error();
+        if crt_error_code == 6146 {
```
monthonk, line 878: Add a comment for this number that it's `AWS_AUTH_SIGNING_NO_CREDENTIALS` from the CRT. Would be great if we can use enum instead of raw number. I think `bindgen` already generated these enum for us, what we have to do is creating a wrapper for them in `mountpoint-s3-crt`.

Anyway, not blocking on that since we only want this specific error code. Adding a comment is fine for me.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -766,6 +766,10 @@ pub enum S3RequestError {
     /// Forbidden
     #[error("Forbidden: {0}")]
     Forbidden(String),
+
+    /// No signing credential is set for requests
+    #[error("No signing credentials provided")]
```
jamesbornholt, line 771: nit: "No signing credentials found", I think

mountpoint-s3-client/src/s3_crt_client.rs
jamesbornholt: Agree we can do this in a separate change. I think the right thing would be for the `mountpoint_s3_crt::common::error` to export all the raw error codes from the bindings, maybe in a `raw` submodule. But it gets a bit weird because of how bindgen converts enums (what integer type to use), so will need some fiddling around.

sauraank: yes, there is `pub enum aws_auth_errors`. For now, I am only adding the comments here.

monthonk (approved):
LGTM

PR #458, opened 2023-08-15
Improve benchmark script

monthonk:
## Description of change

Increase number of iteration we run the throughput benchmarks from 1 to 10. This should make the benchmark result more stable. Also change the tool we use  for write benchmark from `dd` to `fio` for consistency with other benchmarks.

No

---

### DIFF
--- a/mountpoint-s3/scripts/fs_bench.sh
+++ b/mountpoint-s3/scripts/fs_bench.sh
@@ -33,17 +33,48 @@ results_dir=results
 runtime_seconds=30
 startdelay_seconds=30
 max_threads=4
+iteration=10
 
 rm -rf ${results_dir}
 mkdir -p ${results_dir}
+
+run_fio_job() {
+  job_file=$1
+  bench_file=$2
+  mount_dir=$3
+
+  job_name=$(basename "${job_file}")
+  job_name="${job_name%.*}"
+
+  for i in $(seq 1 $iteration);
+  do
+    fio --thread \
+      --output=${results_dir}/${job_name}_${i}.json \
+      --output-format=json \
+      --directory=${mount_dir} \
+      --filename=${bench_file} \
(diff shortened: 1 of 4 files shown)


.github/workflows/bench.yml
```diff
@@ -33,6 +33,7 @@ jobs:
       with:
         role-to-assume: ${{ vars.ACTIONS_IAM_ROLE }}
         aws-region: ${{ vars.S3_REGION }}
+        role-duration-seconds: 21600
```
monthonk, line 36: The default assume role duration is 1 hour which is apparently not enough when the benchmark now takes 10x times.

mountpoint-s3/scripts/fs_bench.sh
```diff
@@ -65,122 +96,45 @@ read_bechmark () {
...
+        --prefix=${S3_BUCKET_TEST_PREFIX} \
+        --max-threads=${max_threads}
+    mount_status=$?
+    if [ $mount_status -ne 0 ]; then
+        echo "Failed to mount file system"
+        exit 1
+    fi
+
+    # set bench file
+    bench_file=${mount_dir}/${job_name}_${RANDOM}.dat
```
jamesbornholt, line 127: Is this going to be annoying if we want to do a four thread version (since they need to write to four different files)? Maybe that's just a problem for future us.

jamesbornholt (commented):
Just one question.

mountpoint-s3/scripts/fs_bench.sh
monthonk: It will be a bit annoying. The reason we use random file names is to be able to run multiple benchmarks at the same time. For four thread version, I think instead of using random file names we can create a random directory instead and make some changes in the `run_fio_job` function.

Issue #434, opened 2023-08-05
Allow reading already restored Glacier Flexible Retrieval/Glacier Deep Archive objects


jamesbornholt:
### Tell us more about this new feature.

We disabled this in #414 to avoid weird EIO errors because we couldn't distinguish restored objects from unrestored ones without a HeadObject request. But there's a [new S3 feature](https://aws.amazon.com/about-aws/whats-new/2023/06/amazon-s3-restore-status-s3-glacier-objects-list-api/) that includes restore status in ListObjectsV2, which we should be able to use to make only restored objects readable (non-000 permissions).

odellso:
Regarding this issue: I am quite keen to have this enhancement released. Apologies if this is not the proper git etiquette (not a professional developer myself), but is there anything I can do to push this along? Would be happy to contribute at-scale user testing if that would help.

jamesbornholt:
@odellso we actually have an open PR for this right now (#467) that should be merged shortly. You'll still have to manually issue restore requests for the objects, but once they're restored they'll be readable with Mountpoint. Does that fit your use case?

odellso:
> @odellso we actually have an open PR for this right now (#467) that should be merged shortly. You'll still have to manually issue restore requests for the objects, but once they're restored they'll be readable with Mountpoint. Does that fit your use case?

@jamesbornholt: thank you for the reply, that's great news. Yes, our use case is exactly as you have described.

jamesbornholt:
#467 is merged and released as v1.0.1: https://github.com/awslabs/mountpoint-s3/releases/tag/mountpoint-s3-1.0.1

PR #467, opened 2023-08-21
Allow reading restored GFR/GDA objects (#434)

vladem:
## Description of change

- Adjusting `mountpoint-s3-client` to propagate information about object's restoration status from `head` and `list` methods, this is accounted in `ObjectInfo::restored` boolean field;
- This information is accounted in `InodeStat::restored` boolean field in `mountpoint-s3`;
- Based on `InodeStat::restored` I adjust  `crate::fs::is_archived` function (`is_flexible_retrieval_storage_class ` previously) which is used to restrict permissions for GLACIER/DEEP_ARCHIVE objects. 
- Two long-running (~2 mins) tests are added to `mountpoint-s3-client/tests/head_object.rs` and `mountpoint-s3/tests/fuse_tests/read_test.rs` -- these are communicating with s3 and wait for it to restore objects.
- Manually tested that for objects that previously have been restored but expired permissions are 000 (corresponding `RestoreStatus`/`x-amz-restore` field not returned by list/head API)

API reference (ctrl+f: "restore"):
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html

Relevant issues: #434 

This is not a breaking change. This change will change permissions for restored GLACIER/DEEP_ARCHIVE objects, which previously were reported as 000. This also will allow reading those objects. 

---

### DIFF
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -4,6 +4,7 @@ use std::ops::Range;
 use std::pin::Pin;
 use std::sync::{Arc, RwLock};
 use std::task::{Context, Poll};
+use std::time::{Duration, SystemTime};
 
 use async_trait::async_trait;
 use futures::{Stream, StreamExt};
@@ -20,7 +21,7 @@ use crate::object_client::{
     ObjectClient, ObjectClientError, ObjectClientResult, ObjectInfo, PutObjectError, PutObjectParams, PutObjectResult,
     UploadReview, UploadReviewPart,
 };
-use crate::{Checksum, ETag, ObjectAttribute, PutObjectRequest};
+use crate::{Checksum, ETag, ObjectAttribute, PutObjectRequest, RestoreStatus};
 
 pub const RAMP_MODULUS: usize = 251; // Largest prime under 256
 static_assertions::const_assert!((RAMP_MODULUS > 0) && (RAMP_MODULUS <= 256));
@@ -56,12 +57,12 @@ pub struct MockClientConfig {
 #[derive(Debug)]
 pub struct MockClient {
     config: MockClientConfig,
-    objects: Arc<RwLock<BTreeMap<String, Arc<MockObject>>>>,
+    objects: Arc<RwLock<BTreeMap<String, MockObject>>>,
     in_progress_uploads: Arc<RwLock<BTreeSet<String>>>,
 }
 
-fn add_object(objects: &Arc<RwLock<BTreeMap<String, Arc<MockObject>>>>, key: &str, value: MockObject) {
-    objects.write().unwrap().insert(key.to_owned(), Arc::new(value));
+fn add_object(objects: &Arc<RwLock<BTreeMap<String, MockObject>>>, key: &str, value: MockObject) {
+    objects.write().unwrap().insert(key.to_owned(), value);
 }
 
 impl MockClient {
@@ -108,6 +109,30 @@ impl MockClient {
             Err(MockClientError("object not found".into()))
         }
     }
+
+    /// Returns error if object does not exist
+    pub fn restore_object(&self, key: &str) -> Result<(), MockClientError> {
+        match self.objects.write().unwrap().get_mut(key) {
+            Some(mock_object) => {
+                mock_object.restore_status = Some(RestoreStatus::Restored {
+                    expiry: SystemTime::now() + Duration::from_secs(3600),
+                });
+                Ok(())
+            }
+            None => Err(MockClientError("object not found".into())),
+        }
+    }
+
+    pub fn is_object_restored(&self, key: &str) -> Result<bool, MockClientError> {
+        if let Some(mock_object) = self.objects.read().unwrap().get(key) {
+            Ok(matches!(
+                mock_object.restore_status,
+                Some(RestoreStatus::Restored { expiry: _ })
+            ))
+        } else {
+            Err(MockClientError("object not found".into()))
+        }
+    }
 }
 
 #[derive(Clone)]
@@ -115,6 +140,7 @@ pub struct MockObject {
     generator: Arc<dyn Fn(u64, usize) -> Box<[u8]> + Send + Sync>,
     size: usize,
     storage_class: Option<String>,
+    restore_status: Option<RestoreStatus>,
     last_modified: OffsetDateTime,
     etag: ETag,
 }
@@ -131,6 +157,7 @@ impl MockObject {
             size: bytes.len(),
             generator: Arc::new(move |offset, size| bytes[offset as usize..offset as usize + size].into()),
             storage_class: None,
+            restore_status: None,
             last_modified: OffsetDateTime::now_utc(),
             etag,
         }
@@ -141,6 +168,7 @@ impl MockObject {
             generator: Arc::new(move |_offset, size| vec![v; size].into_boxed_slice()),
             size,
             storage_class: None,
+            restore_status: None,
             last_modified: OffsetDateTime::now_utc(),
             etag,
         }
@@ -161,6 +189,7 @@ impl MockObject {
             }),
             size,
             storage_class: None,
+            restore_status: None,
             last_modified: OffsetDateTime::now_utc(),
             etag,
         }
@@ -172,6 +201,10 @@ impl MockObject {
 
     pub fn set_storage_class(&mut self, storage_class: Option<String>) {
         self.storage_class = storage_class;
+    }
+
+    pub fn set_restored(&mut self, restore_status: Option<RestoreStatus>) {
+        self.restore_status = restore_status;
     }
 
     pub fn len(&self) -> usize {
@@ -200,13 +233,14 @@ impl std::fmt::Debug for MockObject {
             .field("storage_class", &self.storage_class)
             .field("last_modified", &self.last_modified)
             .field("etag", &self.etag)
+            .field("restored", &self.restore_status)
             .finish()
     }
 }
 
 #[derive(Debug)]
 pub struct GetObjectResult {
-    object: Arc<MockObject>,
+    object: MockObject,
     next_offset: u64,
     length: usize,
     part_size: usize,
@@ -316,7 +350,7 @@ impl ObjectClient for MockClient {
             };
 
             Ok(GetObjectResult {
-                object: Arc::clone(object),
+                object: object.clone(),
                 next_offset,
                 length,
                 part_size: self.config.part_size,
@@ -347,6 +381,7 @@ impl ObjectClient for MockClient {
                     last_modified: object.last_modified,
                     etag: object.etag.as_str().to_string(),
                     storage_class: object.storage_class.clone(),
+                    restore_status: object.restore_status,
                 },
             })
         } else {
@@ -442,6 +477,7 @@ impl ObjectClient for MockClient {
                     last_modified: object.last_modified,
                     etag: object.etag.as_str().to_string(),
                     storage_class: object.storage_class.clone(),
+                    restore_status: object.restore_status,
                 });
             }
         }
@@ -523,7 +559,7 @@ pub struct MockPutObjectRequest {
     buffer: Vec<u8>,
     part_size: usize,
     params: PutObjectParams,
-    objects: Arc<RwLock<BTreeMap<String, Arc<MockObject>>>>,
+    objects: Arc<RwLock<BTreeMap<String, MockObject>>>,
     in_progress_uploads: Arc<RwLock<BTreeSet<String>>>,
 }
 
@@ -532,7 +568,7 @@ impl MockPutObjectRequest {
         key: &str,
         part_size: usize,
         params: &PutObjectParams,
-        objects: &Arc<RwLock<BTreeMap<String, Arc<MockObject>>>>,
+        objects: &Arc<RwLock<BTreeMap<String, MockObject>>>,
         in_progress_uploads: &Arc<RwLock<BTreeSet<String>>>,
     ) -> Self {
         in_progress_uploads.write().unwrap().insert(key.to_owned());
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -2,6 +2,7 @@ use async_trait::async_trait;
 use auto_impl::auto_impl;
 use futures::Stream;
 use std::str::FromStr;
+use std::time::SystemTime;
 use std::{
     fmt::{self, Debug},
     ops::Range,
@@ -315,6 +316,19 @@ pub enum PutObjectError {
     NoSuchBucket,
 }
 
+/// Restoration status for S3 objects in GLACIER/DEEP_ARCHIVE storage class
+/// See https://docs.aws.amazon.com/AmazonS3/latest/userguide/restoring-objects.html#restore-archived-objects-status for more details.
+#[derive(Debug, Clone, Copy)]
+pub enum RestoreStatus {
+    /// S3 returns this status after it accepted a restoration request, but not have completed it yet.
+    /// Objects with this status are not readable.
+    InProgress,
+
+    /// This status means that restoration is fully completed. Note that restored objects are stored only
+    /// for the number of days that was specified in the request.
+    Restored { expiry: SystemTime },
+}
+
 /// Metadata about a single S3 object.
 /// See https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html for more details.
 #[derive(Debug)]
@@ -332,6 +346,9 @@ pub struct ObjectInfo {
     /// the storage class in its response for Standard objects. See examples here:
     /// https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_Examples
     pub storage_class: Option<String>,
+
+    /// Objects with GLACIER or DEEP_ARCHIVE storage classes are only acessable after restoration
+    pub restore_status: Option<RestoreStatus>,
 
     /// Entity tag of this object.
     pub etag: String,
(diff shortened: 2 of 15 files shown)


vladem:
Hey folks, noticed that 2 new tests are failing in CI (while passing locally). It fails with [a permission error](https://github.com/awslabs/mountpoint-s3/actions/runs/5927941339/job/16072505813?pr=467#step:8:817) when trying to restore the object:
`RestoreObjectError { kind: Unhandled(Unhandled { source: Error { code: Some("AccessDenied"), message: Some("Access Denied"), request_id: Some("8SH7Q3RSY58CE1M8")`

From the restore API [docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_RestoreObject.html):
> To use this operation, you must have permissions to perform the **s3:RestoreObject action**. The bucket owner has this permission by default and can grant this permission to others.

Can it be the case that tests are run by a IAM user with a restricted access to the bucket? Can we give this IAM user permission for **s3:RestoreObject** action?

Here is [an example](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-with-s3-actions.html#using-with-s3-actions-related-to-bucket-subresources) of policy specification just in case.

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -108,13 +108,28 @@ impl MockClient {
...
+        let mut cloned: MockObject;
+        if let Some(mock_object) = self.objects.write().unwrap().get(key) {
+            cloned = (**mock_object).clone();
+            cloned.restored = true;
+        } else {
+            return Err(MockClientError("object not found".into()));
+        }
+        self.remove_object(key);
+        self.add_object(key, cloned);
+        Ok(())
```
jamesbornholt, line 123: Do we need to remove and re-add the object? Maybe I'm missing why it isn't enough to just set `restored` to `true`.

mountpoint-s3-client/src/object_client.rs
```diff
@@ -336,6 +336,11 @@ pub struct ObjectInfo {
     /// https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_Examples
     pub storage_class: Option<String>,
 
+    /// Objects with GLACIER or DEEP_ARCHIVE storage classes are only acessable after restoration
+    /// This field is is set to true when \1 object has either GLACIER or DEEP_ARCHIVE storage class \2 restoration is fully
+    /// completed (i.e. head_object response contains "x-amz-restore" header with ongoing-request="false").
+    pub restored: bool,
+
```
jamesbornholt, line 343: While we're here, maybe we should do the slightly richer thing, define an enum like this:
```rust
pub enum RestoreStatus {
    InProgress,
    Restored { expiry: SystemTime },
}
```
and then make this field `pub restored: Option<RestoreStatus>`.

I don't think we actually need to _use_ the expiry time for anything, though it's nice to have.

mountpoint-s3-client/src/s3_crt_client/list_objects.rs
```diff
@@ -143,7 +153,9 @@ impl S3CrtClient {
                 .inner
                 .new_request_template("GET", bucket)
                 .map_err(S3RequestError::construction_failure)?;
-
+            message
+                .add_header(&Header::new("x-amz-optional-object-attributes", "RestoreStatus"))
+                .map_err(S3RequestError::construction_failure)?;
```
jamesbornholt, line 158: This should _probably_ be optional, because I'm not sure it's supported everywhere, but let's tackle that separately.

mountpoint-s3-client/tests/head_object.rs
```diff
@@ -105,3 +106,66 @@ async fn test_head_object_no_perm() {
...
+    let mut i = 0;
+    while i < timeout_s
+        && !client
+            .head_object(&bucket, &key)
+            .await
+            .expect("head_object failed")
+            .object
+            .restored
+    {
+        println!("restoration is in progress, timeout in: {}s", timeout_s - i);
```
jamesbornholt, line 166: delete

mountpoint-s3-client/tests/head_object.rs
```diff
@@ -105,3 +106,66 @@ async fn test_head_object_no_perm() {
...
+        std::thread::sleep(std::time::Duration::from_secs(1));
```
jamesbornholt, line 167: `std::time::sleep` is fine, but we should `use std::time::Duration;`

mountpoint-s3/src/inode.rs
```diff
@@ -1255,6 +1255,8 @@ pub struct InodeStat {
     pub etag: Option<String>,
     /// Storage class for the file (object), if known
     storage_class: Option<StorageClass>,
+    /// For objects with GLACIER or DEEP_ARCHIVE storage class shows if the object is restored (=> readable)
+    pub restored: bool,
```
jamesbornholt, line 1259: Now we're putting `restored` here, I think we can probably just delete `storage_class` and instead make this a single `is_readable` flag or something.

mountpoint-s3/tests/fuse_tests/read_test.rs
```diff
@@ -75,22 +75,45 @@ fn basic_read_test_mock_prefix() {
     basic_read_test(crate::fuse_tests::mock_session::new, "basic_read_test");
 }
 
-fn read_flexible_retrieval_test<F>(creator_fn: F, prefix: &str)
+#[derive(PartialEq)]
+enum RestorationOptions {
+    None,
+    RestoreAndWait,
+    RestoreInProgress,
```
jamesbornholt, line 82: Cool tests!

jamesbornholt (commented):
Thanks for taking this one on!

jamesbornholt:
> Hey folks, noticed that 2 new tests are failing in CI (while passing locally). It fails with [a permission error](https://github.com/awslabs/mountpoint-s3/actions/runs/5927941339/job/16072505813?pr=467#step:8:817) when trying to restore the object: `RestoreObjectError { kind: Unhandled(Unhandled { source: Error { code: Some("AccessDenied"), message: Some("Access Denied"), request_id: Some("8SH7Q3RSY58CE1M8")`
> 
> From the restore API [docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_RestoreObject.html):
> 
> > To use this operation, you must have permissions to perform the **s3:RestoreObject action**. The bucket owner has this permission by default and can grant this permission to others.
> 
> Can it be the case that tests are run by a IAM user with a restricted access to the bucket? Can we give this IAM user permission for **s3:RestoreObject** action?
> 
> Here is [an example](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-with-s3-actions.html#using-with-s3-actions-related-to-bucket-subresources) of policy specification just in case.

Yeah, the role we use for CI doesn't have s3:RestoreObject. We'll fix that.

monthonk:
I just updated the role for CI to allow `s3:RestoreObject`, so the test should work now.

mountpoint-s3-client/tests/head_object.rs
```diff
@@ -105,3 +106,66 @@ async fn test_head_object_no_perm() {
...
+                ))
+                .build(),
+        )
+        .send()
+        .await
+        .unwrap();
+
+    let timeout_s = 300;
+    let mut i = 0;
+    while i < timeout_s
```
monthonk, line 158: I would prefer this
```suggestion
    let timeout_s = 300;
    let start = Instant::now();
    while start.elapsed().as_secs() < timeout_s
```

mountpoint-s3/tests/fuse_tests/mod.rs
```diff
@@ -166,6 +170,17 @@ mod mock_session {
...
+
+        fn restore_object(&mut self, key: &str, expedited: bool) -> Result<(), Box<dyn std::error::Error>> {
+            let full_key = format!("{}{}", self.prefix, key);
+            self.client
+                .restore_object(&full_key)
+                .map_err(|e| Box::new(e) as Box<dyn std::error::Error>)
+        }
+
+        fn is_object_restored(&mut self, key: &str) -> Result<bool, Box<dyn std::error::Error>> {
+            Ok(true)
```
monthonk, line 182: Can we get value from the `MockClient`? should be just a simple function like `get_object_storage_class`.

mountpoint-s3/tests/fuse_tests/read_test.rs
```diff
@@ -75,22 +75,45 @@ fn basic_read_test_mock_prefix() {
...
-            .unwrap();
+        let key = format!("{file}.txt");
+        test_client.put_object_params(&key, b"hello world", put_params).unwrap();
+        match restore {
+            RestorationOptions::None => (),
+            RestorationOptions::RestoreAndWait => {
+                test_client.restore_object(&key, true).unwrap();
+                let mut i = 0;
+                let timeout_s = 300;
+                while i < timeout_s
```
monthonk, line 104: same as above comment, we can check by duration instead.

mountpoint-s3-client/src/object_client.rs
monthonk: +1 that we should make it `Option` and using enum would be great.

monthonk (commented):
Great work! I think we should update the documentation too, there is a few places we mention these storage classes e.g. [SEMANTIC.md](https://github.com/awslabs/mountpoint-s3/blob/main/doc/SEMANTICS.md#reading-and-writing-files) and [CONFIGURATION.md](https://github.com/awslabs/mountpoint-s3/blob/main/doc/SEMANTICS.md#reading-and-writing-files).

mountpoint-s3-client/src/mock_client.rs
vladem: So, object, which Arc points to, is immutable. From the [documentation](https://doc.rust-lang.org/std/sync/struct.Arc.html):

> Shared references in Rust disallow mutation by default, and Arc is no exception: you cannot generally obtain a mutable reference to something inside an Arc. If you need to mutate through an Arc, use [Mutex](https://doc.rust-lang.org/std/sync/struct.Mutex.html), [RwLock](https://doc.rust-lang.org/std/sync/struct.RwLock.html), or one of the [Atomic](https://doc.rust-lang.org/std/sync/atomic/index.html) types.

A program like this:
```
use std::{sync::Arc};

struct MockObject {
    restored: bool,
}

fn main() {
    let mut obj1 = Arc::new(MockObject{restored: false});
    obj1.restored = true;
}
```
Won't compile: 
```
error[E0594]: cannot assign to data in an `Arc`
 --> src/main.rs:9:5
  |
9 |     obj1.restored = true;
  |     ^^^^^^^^^^^^^^^^^^^^ cannot assign
  |
  = help: trait `DerefMut` is required to modify through a dereference, but it is not implemented for `Arc<MockObject>
```

While we can do something like [this](https://doc.rust-lang.org/book/ch16-03-shared-state.html#atomic-reference-counting-with-arct):
```
use std::{sync::{Arc, RwLock}};

struct MockObject {
    restored: bool,
}

fn main() {
    let mut obj1 = Arc::new(RwLock::new(MockObject{restored: false}));
    obj1.write().unwrap().restored = true;
}
```

I am wondering why are we having reference counted objects here in the first place. Probably something simpler will work:
```
use std::{sync::{Arc, RwLock}, collections::BTreeMap};

#[derive(Debug)]
struct MockObject {
    restored: bool,
}

#[derive(Debug)]
struct MockClient {
    objects: RwLock<BTreeMap<String, MockObject>>,  // <-------------------- no Arc's here
}

impl MockClient {
    fn new() -> Self {
        Self {
            objects: RwLock::new(BTreeMap::new()),
        }
    }

    fn add_object(&self, key: &str, value: MockObject) {
        self.objects.write().unwrap().insert(key.to_owned(), value);
    }

    fn restore(&mut self, key: &str) {
        if let Some(mock_object) = self.objects.write().unwrap().get_mut(key) {
            mock_object.restored = true;
        } else {
            panic!("not impl");
        }
    }
}

fn main() {
    let mut client = MockClient::new();
    client.add_object("key", MockObject { restored: false });
    println!("{:?}", client);
    client.restore("key");
    println!("{:?}", client);
}
```

So if you don't mind refactoring this a little, I'll try to remove Arcs tomorrow.

jamesbornholt: Oh right, I missed that it's in an `Arc` (embarrassing since I was the one who did that). It's there because it makes cloning the object cheap, and we make a clone of it when you start a get:

https://github.com/awslabs/mountpoint-s3/blob/b9f7e913fb2e0ddf5beff1f1f3121b4ca1a20601/mountpoint-s3-client/src/mock_client.rs#L318-L323

Without that `Arc`, we'd have to clone the entire object body (the `generator` field in `MockObject`) every time we start a get. BUT! Since we wrote that code, we've moved the `generator` inside `MockObject` into an `Arc` too:

https://github.com/awslabs/mountpoint-s3/blob/b9f7e913fb2e0ddf5beff1f1f3121b4ca1a20601/mountpoint-s3-client/src/mock_client.rs#L114-L120

So `MockObject` should now be fairly cheap to clone even if we get rid of the `Arc<MockObject>` you're tripping over here. Not free to clone — still have to make a copy of the `String`s for `storage_class` and `etag` — but cheap enough. Seems like a good refactor!

mountpoint-s3-client/src/object_client.rs
vladem: its a good thing to do, it lets us analyse expiry for 2 purposes:
1. object is not readable if `expiry` is less than `now()`;
2. inode ttl is better to be less than `expiry`;

While \1 is simple, wasn't able to tackle \2 as inode expiry is measured in `Instant` which is not comparable with "human" dates. I gave up since I assume inode ttl's precision is in fractions of a second, which means that its unlikely that we'll cache "valid" restoration status after the object's been put back to archive by S3. Let me know, if you have any other thoughts on that, pls.

mountpoint-s3-client/src/s3_crt_client/list_objects.rs
vladem: I suspect you're referring to `Outposts`, what I've noticed in this regard is that [restoration](https://docs.aws.amazon.com/AmazonS3/latest/API/API_RestoreObject.html) itself is not supported there:
> This action is not supported by Amazon S3 on Outposts.

So at least new tests under `s3_tests` feature will fail if they will ever run against Outputs S3 installation.

mountpoint-s3/src/inode.rs
vladem: This inclined me to [print](https://github.com/awslabs/mountpoint-s3/pull/467/files#diff-72de80783ffd916a6b6d2b704c855eadcb03d30e1d6315a0e33a30196cb7ca1aR1286) a warning on `InodeStat` instantiation. Previously it was printed from `new_read_handle ` or from `make_attr ` when listing a directory).

Generic `is_readable` implies that we may add new cases when the object has restricted permissions.

mountpoint-s3-client/tests/head_object.rs
vladem: A little more code, since I now need to set `timeout` flag, but more precise in terms of running time. Seems reasonable, fixed.

mountpoint-s3/tests/fuse_tests/mod.rs
vladem: Yep, thanks for spotting this loose end

vladem:
Addressed comments from the first round, ready to be reviewed again.

mountpoint-s3-client/src/s3_crt_client/head_object.rs
```diff
@@ -131,10 +167,14 @@ fn parse_head_object_error(result: &MetaRequestResult) -> Option<HeadObjectError
 
 #[cfg(test)]
 mod tests {
-    use std::ffi::OsString;
+    use mountpoint_s3_crt::common::allocator::Allocator;
+
```
jamesbornholt, line 171: nit: no blank line here

mountpoint-s3/tests/fuse_tests/read_test.rs
```diff
@@ -75,22 +76,47 @@ fn basic_read_test_mock_prefix() {
...
-        test_client
-            .put_object_params(&format!("{file}.txt"), b"hello world", put_params)
-            .unwrap();
+        let key = format!("{file}.txt");
+        test_client.put_object_params(&key, b"hello world", put_params).unwrap();
+        match restore {
+            RestorationOptions::None => (),
+            RestorationOptions::RestoreAndWait => {
+                test_client.restore_object(&key, true).unwrap();
+                let timeout_s = 300;
```
jamesbornholt, line 103: nit: same thing here about `Duration`.

mountpoint-s3/tests/fuse_tests/read_test.rs
```diff
@@ -115,11 +143,60 @@ where
...
+    const FILES: &[&str] = &["GLACIER", "DEEP_ARCHIVE"];
+    read_flexible_retrieval_test(
+        crate::fuse_tests::mock_session::new,
+        prefix,
+        FILES,
+        RestorationOptions::RestoreAndWait,
+    );
+}
+
+// This test relies on s3's expedited object restoration, it takes 1-5 minutes to complete
```
jamesbornholt, line 179: nit: just say explicitly that we don't test `DEEP_ARCHIVE` because it has no expedited restore available

vladem: going to test with **DEEP_ARCHIVE** object manually, need to wait it to be restored in my bucket though :) 

also it'll be good to test with **S3 Intelligent-Tiering**, because I have a suspicion that expiry date [may not be returned](https://docs.aws.amazon.com/AmazonS3/latest/userguide/restoring-objects.html#restore-archived-objects-status) for them:
> Objects restored from S3 Intelligent-Tiering Archive Access and Deep Archive Access storage tiers do not have expiration dates and instead move back to the Frequent Access tier.

but it seems not feasible, since object is put into Archive Access [90 days](https://docs.aws.amazon.com/AmazonS3/latest/userguide/intelligent-tiering-overview.html) after last access

jamesbornholt (approved):

PR #470, opened 2023-08-22
Small fixes for S3 on Outposts

jamesbornholt:
## Description of change

This fixes two issues that were preventing Mountpoint from working against Outposts buckets:
1. Outposts doesn't include the bucket name in ListObjectsV2 responses. We weren't actually using that output anyway, so I just removed it.
2. For GetObject requests, we were sending a HTTP header like `Accept: application/xml,*/*`. While technically valid HTTP, it's weird to accept `*/*` as well as something else, and it was confusing Outposts' request signing. So I switched to overwriting the existing header, which is what the comment suggested the code was intended to do anyway.

I also took this chance to make a little cleanup to parsing ListObjectsV2 responses: the `parse` functions shouldn't be defined on the generic `ListObjectsResult` structs, which are shared by all clients.

This change doesn't make Mountpoint completely work on Outposts. The outstanding issue is that Outposts don't support S3 additional checksums, but we turn those on by default. We'll need a way to turn those off for Outposts buckets. But read-only usage should work.

Yes, removing the bucket name is a breaking change to the client crate, and we now send a different `Accept` header on GetObject requests. There shouldn't be any customer-visible effects.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -634,12 +634,13 @@ struct S3Message {
 }
 
 impl S3Message {
-    /// Add a header to this message.
-    fn add_header(
+    /// Add a header to this message. The header is added if necessary and any existing values for
+    /// this header are removed.
+    fn set_header(
         &mut self,
         header: &Header<impl AsRef<OsStr>, impl AsRef<OsStr>>,
     ) -> Result<(), mountpoint_s3_crt::common::error::Error> {
-        self.inner.add_header(header)
+        self.inner.set_header(header)
     }
 
     /// Set the request path and query for this message. The components should not be URL-encoded;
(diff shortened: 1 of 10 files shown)


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -634,12 +634,13 @@ struct S3Message {
 }
 
 impl S3Message {
-    /// Add a header to this message.
-    fn add_header(
+    /// Add a header to this message. The header is added if necessary and any existing values for
+    /// this header are removed.
+    fn set_header(
```
sauraank, line 639: Was there any use of adding a new header with different value rather than overwriting it, earlier?

jamesbornholt: Nope, that's why I changed it. Multiple headers with the same name is a pretty niche HTTP thing that we don't need.

ale-rinaldi:
I can confirm that this also fixes the GetObject operation on the Seeweb provider; it previously failed with an invalid signature error. Thank you

monthonk (approved):
LGTM

PR #473, opened 2023-08-23
erase "accept" header before add new

huangnauh:
Authentication for certain object storage systems, such as Ceph RGW, does not support multi-valued headers.

## Description of change

GetObject 

Origin

```
GET /xxx/ HTTP/1.1
accept: application/xml
accept: */*
```

Now

```
GET /xxx/ HTTP/1.1
accept: */*
```

No

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -642,6 +642,14 @@ impl S3Message {
         self.inner.add_header(header)
     }
 
+    /// Erase a header with the given name from this message.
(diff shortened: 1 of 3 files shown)


monthonk:
Hi @huangnauh, thanks for contributing to Mountpoint! we'are experiencing the same problem with outpost buckets. In https://github.com/awslabs/mountpoint-s3/pull/470 we proposed to overwrite existing HTTP header for GetObject, and I think that should fix your problem as well.

jamesbornholt:
#470 is merged and should fix this; please let us know if it doesn’t!

PR #466, opened 2023-08-21
Update --allow-root and --allow-other documentation

Alexj12:
## Description of change

Using `--allow-root` and `--allow-other` together resulted in unexpected results. This makes a documentation change to improve understanding.

Relevant issues: N/A

## Does this change impact existing behavior?

No functionality changes.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

PR #475, opened 2023-08-23
Make allow_other and allow_root mutually exclusive

monthonk:
## Description of change

Reported in https://github.com/awslabs/mountpoint-s3/pull/466 that `--allow-other` infers `--allow-root`. If both `--allow-root` and `--allow-other` are defined, only `--allow-root` will function. So, we should make the two options mutually exclusive.

Yes, cli arguments `--allow-root` and `--allow-other` are now mutually exclusive.

---

### DIFF
--- a/mountpoint-s3/src/main.rs
+++ b/mountpoint-s3/src/main.rs
@@ -110,8 +110,9 @@ struct CliArgs {
 
     #[clap(
         long,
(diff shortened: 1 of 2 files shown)


sauraank (approved):
LGTM. conflicts_with option is added. And there is test for using allow-others and allow-root together.

PR #481, opened 2023-08-25
Add build instructions for Ubuntu

ale-rinaldi:
## Description of change
I added some basic build instructions for Ubuntu, that lead me to a successful build. I understand there should be instructions for more platforms, but I think some instructions are better than no instructions at all; I feel it's useful at least to have an immediate understanding of the required dependencies and save from the fail-and-retry approach I had to use :)

No

---


jamesbornholt:
Thanks for this! We actually have build instructions in the [install docs](https://github.com/awslabs/mountpoint-s3/blob/main/doc/INSTALL.md#building-mountpoint-for-amazon-s3-from-source), but that's not super obvious from the README right now.

If those instructions look right to you, then probably what we should do in this PR is add a pointer to that section from your new `## Building` section in the README?

ale-rinaldi:
Wow, I can't believe I missed it! Actually it's super clear if you keep on reading after the commands 😅
Sorry for that, closing.

PR #487, opened 2023-08-29
Allow file overwrites

monthonk:
## Description of change

Allow file overwrites on mountpoint. This is quite straightforward for files opened with `O_WRONLY`, we can just return a new write handle instead of `InodeNotWritable` error when the write status is remote. However, it's a bit tricky for files opened with `O_RDWR` because we can either read or write to remote files and it means we can't decide that at `open()` time. So, I changed the concept of file handle to be able to do both write and read based on the operation that comes after `open()`.

Same as "delete", I disable this feature by default and provide a new option `--allow-overwrite` to enable it.

Yes, mountpoint now supports overwrites when mounting with `--allow-overwrite` option.

---

### DIFF
--- a/doc/SEMANTICS.md
+++ b/doc/SEMANTICS.md
@@ -13,7 +13,7 @@ While the rest of this document gives details on specific file system behaviors,
 
 Mountpoint supports opening and reading existing objects from your S3 bucket. It is optimized for reading large files sequentially, and will automatically make multiple concurrent requests to S3 to improve throughput when reads are sequential. Mountpoint also supports random reads from an existing object, including seeking in an open file.
 
-Mountpoint supports writing only to new files, and writes to new files must be made sequentially. If you try to open an existing file with write access, the open operation will fail with a permissions error. Mountpoint uploads new files to S3 asynchronously, and optimizes for high write throughput using multiple concurrent upload requests. If your application needs to guarantee that a new file has been uploaded to S3, it should call `fsync` on the file before closing it. You cannot continue writing to the file after calling `fsync`.
+Mountpoint supports writing only to new files by default. Writes to existing files are allowed if `--allow-overwrite` flag is set at startup time, but only when the `O_TRUNC` flag is used at open time to truncate the existing file. All writes must start from the beginning of the file and must be made sequentially. Mountpoint uploads new files to S3 asynchronously, and optimizes for high write throughput using multiple concurrent upload requests. If your application needs to guarantee that a new file has been uploaded to S3, it should call `fsync` on the file before closing it. You cannot continue writing to the file after calling `fsync`.
 
 By default, Mountpoint does not allow deleting existing objects with commands like `rm`. To enable deletion, pass the `--allow-delete` flag to Mountpoint at startup time. Delete operations immediately delete the object from S3, even if the file is being read from. We recommend that you enable [Bucket Versioning](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Versioning.html) to help protect against unintentionally deleting objects. You cannot delete a file while it is being written.
 
@@ -153,19 +153,23 @@ Windows-style path delimiters (`\`) are not supported.
 #### Reads
 
 Basic read-only operations are fully supported, including both sequential and random reads:
-* `open`, `openat`. A file can be opened in read-write mode (`O_RDWR`), but you cannot both read and write to the same file descriptor even in this mode.
+* `open`, `openat`
 * `read`, `readv`, `pread`, `preadv`
 * `lseek`
 * `close`
+
+`open` creates a file handle and returns it back to the kernel. A file handle can only be used for one type of operation, either read or write, for its lifetime. You can open a file in read-write mode (`O_RDWR`), but you cannot both read and write to the same file descriptor even in this mode. The first `read` or `write` will determine the type of operation you can do with the file descriptor.
 
 #### Writes
 
 Mountpoint supports sequential write operations (through `write`, `writev`, `pwrite`, `pwritev`),
 but with some limitations:
 
-* Writes are only supported to new files, and must be done sequentially.
-* Modifying existing files is not supported.
-* Truncation is not supported.
+* Writes must start at the beginning of the file and be done sequentially.
+* Modifying an existing file is only allowed with the `--allow-overwrite` flag and only when the file is opened in truncate mode (`O_TRUNC`).
+    * You cannot overwrite files that are currently being read.
+    * The upload to S3 starts as soon as Mountpoint receives the first `write` request and cannot be cancelled.
+* Modifying an existing file without using truncate mode is not supported.
 
 Synchronization operations (`fsync`, `fdatasync`) complete the upload of the object to S3 and disallow
 further writes.
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+### New features
+* Allow file overwrites when mounting with `--allow-overwrite` option. The upload will start as soon as Mountpoint receives `write` request and cannot be aborted. Once it is started the file is guaranteed to be overwritten.
+
 ## v1.3.2 (January 11, 2024)
 
 ### Breaking changes
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -58,8 +58,39 @@ where
 {
     inode: Inode,
     full_key: String,
-    object_size: u64,
-    typ: FileHandleType<Client, Prefetcher>,
+    state: AsyncMutex<FileHandleState<Client, Prefetcher>>,
+}
+
+enum FileHandleState<Client, Prefetcher>
+where
+    Client: ObjectClient + Send + Sync + 'static,
+    Prefetcher: Prefetch,
+{
+    /// A state where the file handle is created but the type is not yet determined
+    Created { lookup: LookedUp, flags: i32, pid: u32 },
+    /// A state where type is already assigned to the file handle
+    Assigned(FileHandleType<Client, Prefetcher>),
+    /// The file handle is already closed, currently only used to tell that the read is finished
+    Closed,
+}
+
+impl<Client, Prefetcher> std::fmt::Debug for FileHandleState<Client, Prefetcher>
+where
+    Client: ObjectClient + Send + Sync + 'static + std::fmt::Debug,
+    Prefetcher: Prefetch,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            FileHandleState::Assigned(typ) => f.debug_tuple("Assigned").field(typ).finish(),
+            FileHandleState::Created { lookup, flags, pid } => f
+                .debug_struct("Created")
+                .field("lookup", lookup)
+                .field("flags", flags)
+                .field("pid", pid)
+                .finish(),
+            FileHandleState::Closed => f.debug_struct("Closed").finish(),
+        }
+    }
 }
 
 enum FileHandleType<Client, Prefetcher>
@@ -68,8 +99,8 @@ where
     Prefetcher: Prefetch,
 {
     Read {
-        request: AsyncMutex<Option<Prefetcher::PrefetchResult<Client>>>,
-        etag: ETag,
+        request: AsyncMutex<Prefetcher::PrefetchResult<Client>>,
+        pid: u32,
     },
     Write(AsyncMutex<UploadState<Client>>),
 }
@@ -81,7 +112,7 @@ where
 {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Self::Read { request: _, etag } => f.debug_struct("Read").field("etag", etag).finish(),
+            Self::Read { request: _, pid } => f.debug_struct("Read").field("pid", pid).finish(),
             Self::Write(arg0) => f.debug_tuple("Write").field(arg0).finish(),
         }
     }
@@ -99,18 +130,23 @@ where
         pid: u32,
         fs: &S3Filesystem<Client, Prefetcher>,
     ) -> Result<FileHandleType<Client, Prefetcher>, Error> {
-        // We can't support O_SYNC writes because they require the data to go to stable storage
-        // at `write` time, but we only commit a PUT at `close` time.
-        if flags & (libc::O_SYNC | libc::O_DSYNC) != 0 {
-            return Err(err!(libc::EINVAL, "O_SYNC and O_DSYNC are not supported"));
+        if flags & libc::O_ACCMODE == libc::O_RDONLY {
+            return Err(err!(libc::EBADF, "file handle is not open for writes"));
         }
 
-        let handle = match fs.superblock.write(&fs.client, ino, lookup.inode.parent(), pid).await {
-            Ok(handle) => handle,
-            Err(e) => {
-                return Err(e.into());
-            }
-        };
+        let is_truncate = flags & libc::O_TRUNC != 0;
+        let handle = fs
+            .superblock
+            .write(
+                &fs.client,
+                ino,
+                lookup.inode.parent(),
+                pid,
+                fs.config.allow_overwrite,
+                is_truncate,
+            )
+            .await
+            .start_writing()?;
         let key = lookup.inode.full_key();
         let handle = match fs.uploader.put(&fs.bucket, key).await {
             Err(e) => {
@@ -122,7 +158,15 @@ where
         Ok(handle)
     }
 
-    async fn new_read_handle(lookup: &LookedUp) -> Result<FileHandleType<Client, Prefetcher>, Error> {
+    async fn new_read_handle(
+        lookup: &LookedUp,
+        flags: i32,
+        pid: u32,
+        fs: &S3Filesystem<Client, Prefetcher>,
+    ) -> Result<FileHandleType<Client, Prefetcher>, Error> {
+        if flags & libc::O_WRONLY != 0 {
+            return Err(err!(libc::EBADF, "file handle is not open for reads",));
+        }
         if !lookup.stat.is_readable {
             return Err(err!(
                 libc::EACCES,
@@ -130,13 +174,17 @@ where
             ));
         }
         lookup.inode.start_reading()?;
-        let handle = FileHandleType::Read {
-            request: Default::default(),
-            etag: match &lookup.stat.etag {
-                None => return Err(err!(libc::EBADF, "no E-Tag for inode {}", lookup.inode.ino())),
-                Some(etag) => ETag::from_str(etag).expect("E-Tag should be set"),
-            },
+        let full_key = lookup.inode.full_key().to_owned();
+        let object_size = lookup.stat.size as u64;
+        let etag = match &lookup.stat.etag {
+            None => return Err(err!(libc::EBADF, "no E-Tag for inode {}", lookup.inode.ino())),
+            Some(etag) => ETag::from_str(etag).expect("E-Tag should be set"),
         };
+        let request = fs
+            .prefetcher
+            .prefetch(fs.client.clone(), &fs.bucket, &full_key, object_size, etag.clone())
+            .into();
+        let handle = FileHandleType::Read { request, pid };
         metrics::increment_gauge!("fs.current_handles", 1.0, "type" => "read");
         Ok(handle)
     }
@@ -180,28 +228,27 @@ impl<Client: ObjectClient> UploadState<Client> {
     }
 
     async fn complete(&mut self, key: &str, ignore_if_empty: bool, pid: Option<u32>) -> Result<(), Error> {
-        match self {
-            Self::InProgress { request, handle } => {
-                if ignore_if_empty && request.size() == 0 {
-                    trace!(key, "not completing upload because file is empty");
-                    return Ok(());
-                }
-                if let Some(pid) = pid {
-                    let open_pid = handle.pid();
-                    if !are_from_same_process(open_pid, pid) {
-                        trace!(
-                            key,
-                            pid,
-                            open_pid,
-                            "not completing upload because current pid differs from pid at open"
-                        );
-                        return Ok(());
-                    }
-                }
-            }
+        let (request_size, open_pid) = match self {
+            Self::InProgress { request, handle } => (request.size(), handle.pid()),
             Self::Completed => return Ok(()),
             Self::Failed(e) => return Err(err!(*e, "upload already aborted for key {:?}", key)),
         };
+
+        if ignore_if_empty && request_size == 0 {
+            trace!(key, "not completing upload because file is empty");
+            return Ok(());
+        }
+        if let Some(pid) = pid {
+            if !are_from_same_process(open_pid, pid) {
+                trace!(
+                    key,
+                    pid,
+                    open_pid,
+                    "not completing upload because current pid differs from pid at open"
+                );
+                return Ok(());
+            }
+        }
 
         let (upload, handle) = match std::mem::replace(self, Self::Completed) {
             Self::InProgress { request, handle } => (request, handle),
@@ -330,6 +377,8 @@ pub struct S3FilesystemConfig {
     pub file_mode: u16,
     /// Allow delete
     pub allow_delete: bool,
+    /// Allow overwrite
+    pub allow_overwrite: bool,
     /// Storage class to be used for new object uploads
     pub storage_class: Option<String>,
     /// S3 personality (for different S3 semantics)
@@ -349,6 +398,7 @@ impl Default for S3FilesystemConfig {
             dir_mode: 0o755,
             file_mode: 0o644,
             allow_delete: false,
+            allow_overwrite: false,
             storage_class: None,
             s3_personality: S3Personality::Standard,
         }
@@ -485,6 +535,16 @@ where
 {
     pub async fn init(&self, config: &mut KernelConfig) -> Result<(), libc::c_int> {
         let _ = config.add_capabilities(fuser::consts::FUSE_DO_READDIRPLUS);
+        if self.config.allow_overwrite {
+            // Overwrites require FUSE_ATOMIC_O_TRUNC capability on the host, so we will panic if the
+            // host doesn't support it.
+            //
+            // This should makes it clear to users that they cannot enable overwrite on their host
+            // rather than silently disable it and let users find out later when their writes fail.
+            config
+                .add_capabilities(fuser::consts::FUSE_ATOMIC_O_TRUNC)
+                .expect("The host must support FUSE_ATOMIC_O_TRUNC capability in order to allow overwrites");
+        }
         Ok(())
     }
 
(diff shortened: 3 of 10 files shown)


mountpoint-s3/src/fs.rs
```diff
@@ -485,28 +499,39 @@ where
...
 
+        let inode = lookup.inode.clone();
+        let full_key = lookup.inode.full_key().to_owned();
+        let object_size = lookup.stat.size as u64;
+        let remote_file = lookup.inode.is_remote()?;
+
+        // Open with O_APPEND is ok for new files because it's same as creating a new one.
+        // but we can't support it on existing files and we should explicitly say we don't allow that.
+        if remote_file && (flags & libc::O_APPEND != 0) {
+            return Err(err!(libc::EPERM, "O_APPEND is not supported on existing files"));
```
jamesbornholt, line 510: EINVAL, I think

mountpoint-s3/src/fs.rs
```diff
@@ -547,6 +572,28 @@ where
                 file_etag = etag.clone();
                 request.lock().await
             }
+            FileHandleType::Uninitialized { lookup, .. } => {
```
jamesbornholt, line 575: Just thinking out loud: is there a way to refactor this so that this logic is shared with the `O_RDONLY` path in `open`, and similar for `O_WRONLY` and the logic in `write`? Would be nice to have those be shared so there's only one place each type of handle can be created.

Might also be nice to find a way to represent `Uninitialized` so that not every code path has to deal with that case. Maybe `typ` can be another enum with `Open(FileHandleType)` and `NotYetOpen(Lookup)` variants or something?

mountpoint-s3/src/fs.rs
```diff
@@ -833,6 +906,10 @@ where
                 metrics::decrement_gauge!("fs.current_handles", 1.0, "type" => "read");
                 Ok(())
             }
+            FileHandleType::Uninitialized { .. } => {
+                metrics::decrement_gauge!("fs.current_handles", 1.0, "type" => "uninitialized");
+                Ok(())
+            }
```
jamesbornholt, line 912: I guess we're effectively pretending it was a read if it was never read/written?

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -153,38 +156,33 @@ where
...
     assert_eq!(err.raw_os_error(), Some(libc::EBADF));
-
-    // Existing files can be opened O_RDWR but only reading should work on them
-    let mut file = File::options().read(true).write(true).create(true).open(&path).unwrap();
-    let err = file
-        .write(b"hello world")
-        .expect_err("write to an existing file should fail");
-    assert_eq!(err.raw_os_error(), Some(libc::EBADF));
-    // However, read should work
-    assert!(file.read(&mut [0u8; 1]).is_ok());
```
jamesbornholt, line 187: Are we missing a test that checks you can't read an `O_RDWR` file once it's open in write mode, and vice versa?

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -79,70 +93,59 @@ where
...
 #[cfg(feature = "s3_tests")]
-#[test_case(true; "append")]
-#[test_case(false; "no append")]
-fn sequential_write_test_s3(append: bool) {
-    sequential_write_test(crate::fuse_tests::s3_session::new, "sequential_write_test", append);
+#[test_case(true, false; "append write only")]
+#[test_case(true, true; "append readwrite")]
+#[test_case(false, false; "no append write only")]
+#[test_case(false, true; "no append readwrite")]
+fn sequential_write_test_s3(append: bool, readwrite: bool) {
```
jamesbornholt, line 130: Do we have a test somewhere that tries to open a file for writing while it's currently open for reading?

mountpoint-s3/src/inode.rs
```diff
@@ -937,12 +937,11 @@ impl WriteHandle {
         let inode = inner.get(ino)?;
         let mut state = inode.get_mut_inode_state()?;
         match state.write_status {
-            WriteStatus::LocalUnopened => {
+            WriteStatus::LocalUnopened | WriteStatus::Remote => {
```
jamesbornholt, line 940: Think we need to do something here with files that are already open for reading. If an application has a file open `O_RDONLY`, and then opens it again `O_WRONLY`, seems like here the second open would succeed, which isn't what we want.

jamesbornholt (commented):
Nice, looks like this was a fairly simple change. A couple of high level questions for now.

mountpoint-s3/src/fs.rs
monthonk: Yeah, we can also do it that way. I'll update the handle structure.

mountpoint-s3/src/fs.rs
monthonk: Not quite sure why it's read?

mountpoint-s3/src/inode.rs
monthonk: I was planning to do that as a separate change but might be a good idea to include it in this PR.

mountpoint-s3/src/fs.rs
jamesbornholt: If we were pretending it was a write, we'd have to upload a 0-byte object here (which seems like the wrong thing to do!).

monthonk: Right, if we have to pick one it should be a read.

mountpoint-s3/src/fs.rs
```diff
@@ -547,6 +572,28 @@ where
...
+                };
+                let new_handle = Arc::new(new_handle);
+
+                self.file_handles.write().await.insert(fh, new_handle.clone());
+                metrics::decrement_gauge!("fs.current_handles", 1.0, "type" => "uninitialized");
+                handle = new_handle;
+                if let FileHandleType::Read { request, etag } = &handle.typ {
+                    file_etag = etag.clone();
+                    request.lock().await
+                } else {
```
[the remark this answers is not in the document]
monthonk: I didn't mean it should, just for this case, in `read()` operation, it always be initialised as read handle.

mountpoint-s3/src/inode.rs
jamesbornholt: I'm OK with doing it in a separate PR if you think it'll make reviewing easier, but let's put a TODO or open an issue or something.

mountpoint-s3/src/fs.rs
monthonk: What do you think if we delay the file handle initialization until the first read or write? That way there should be one place each type can be created, but it will also change current file system behavior.

mountpoint-s3/tests/fuse_tests/read_test.rs
```diff
@@ -54,26 +71,78 @@ where
...
+
+    test_client.put_object("hello.txt", b"hello world").unwrap();
+
+    let file_path = mount_point.path().join("hello.txt");
+
+    let read_dir_iter = read_dir(mount_point.path()).unwrap();
+    let dir_entry_names = read_dir_to_entry_names(read_dir_iter);
+    assert_eq!(dir_entry_names, vec!["hello.txt"]);
+
+    // We can't write to a file opened in O_WRONLY
```
jamesbornholt, line 111:
```suggestion
    // We can't read from a file opened in O_WRONLY
```

mountpoint-s3/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+* Allow file overwrites when mounting with `--allow-overwrite` option.
```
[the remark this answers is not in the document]
monthonk: The previous commits also change **when** we start the upload. It will no longer start at `open` but happen at first `write` instead. That means the file won't be overwritten unless the user call `write` or `fsync`.

Anyway, the message should be clearer. I will update them and add some more tests.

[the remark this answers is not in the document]
jamesbornholt: Doesn't that make it impossible to write a zero-sized file if you don't call fsync? That seems like the wrong behavior for a file opened `O_WRONLY`.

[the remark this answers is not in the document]
monthonk: Not possible for existing files, but it still possible for new files because we can check the initial file status. If it the previous status is not Remote, then we can start the upload and complete it on `release`.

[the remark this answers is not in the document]
monthonk: I tried this with POSIX file system by opening a file with only `File::options().write(true)` which I think should use `O_WRONLY` flag. The file is not overwritten after I close it, and it's still not overwritten even after `fsync`. So, this might be the right behavior?

dannycjones:
I think we need a `TODO` or update to harness in this area too.

https://github.com/awslabs/mountpoint-s3/blob/16efe55c27781f8a8728491b10d5040a492bd014/mountpoint-s3/tests/reftests/harness.rs#L299-L306

jeremyruffell:
Hey, was there any update on this issue? We are looking at migrating to S3 Mountpoint (Using the [mountpoint-s3-csi-driver](https://github.com/awslabs/mountpoint-s3-csi-driver). The only thing holding us back is the ability to update/overwrite files.

mountpoint-s3/CHANGELOG.md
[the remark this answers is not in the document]
monthonk: Checked that fuser doesn't support `O_TRUNC` (https://docs.rs/fuser/latest/fuser/trait.Filesystem.html#method.open) so we can't make it required.

mountpoint-s3/src/fs.rs
```diff
@@ -58,91 +58,84 @@ where
...
     full_key: String,
     object_size: u64,
-    typ: FileHandleType<Client, Prefetcher>,
+    state: FileHandleState<Client, Prefetcher>,
+    read_only: bool,
+    write_only: bool,
 }
 
-enum FileHandleType<Client, Prefetcher>
+struct FileHandleState<Client, Prefetcher>
```
passaro, line 66: A few issues with this state:
* Shouldn't the mutex be outside the state struct?
* Is the ETag part of the state? Does it ever change? Shouldn't it change (to None) when an O_RDWR open on an existing file "becomes" a write?
* It would seem natural for this to be an enum (maybe something like: Read/Write/StillUnknown). Is there a reason not to do that? If so, it would help to document it.

mountpoint-s3/src/fs.rs
```diff
@@ -171,40 +185,69 @@ impl<Client: ObjectClient> UploadState<Client> {
...
-                if let Some(pid) = pid {
+    async fn complete(
+        &mut self,
+        key: &str,
+        ignore_if_empty: bool,
+        pid: Option<u32>,
+        uploader: &Uploader<Client>,
+    ) -> Result<(), Error> {
+        let (request_size, open_pid) = match self {
+            Self::NotStarted {
```
passaro, line 203: The NotStarted case adds a lot of redundant code in all methods. It would be nice to refactor at least part of it. Or even better to avoid this case altogether, if possible.

passaro (commented):
I still have to look at some of the changes (e.g. tests), but I've left my thoughts so far.

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -682,3 +666,129 @@ fn multi_thread_test() {
...
+    let err = write_fh
+        .write(b"hello world")
+        .expect_err("writing to a file is being read should fail");
+    assert_eq!(err.raw_os_error(), Some(libc::EPERM));
+
+    drop(fh);
+    drop(write_fh);
+
+    // Add some delay because release is asynchronous
+    std::thread::sleep(Duration::from_secs(1));
```
passaro, line 732: We mostly removed this problem when we implemented flush. Why are we relying on release?

mountpoint-s3/src/fs.rs
monthonk: Agree with the first two, the mutex should cover the state struct and the Etag should be updated.

One reason for implementing it this way is complexity, I have tried the Read/Write/StillUnknown enum approach and it requires us to handle the StillUnknown case in both read and write path. Another reason is that I think we probably want to allow read after write on the same file handle if we open it with O_RDWR flag.

mountpoint-s3/src/fs.rs
```diff
@@ -668,16 +741,24 @@ where
...
-            FileHandleType::Read { request, etag } => {
-                file_etag = etag.clone();
-                request.lock().await
+        if handle.write_only {
+            return reply.error(err!(libc::EBADF, "file handle is not open for reads"));
+        }
+        let state = &handle.state;
+        let file_etag = match &state.etag {
+            None => {
+                return reply.error(err!(libc::EBADF, "no E-Tag for inode {}", ino));
```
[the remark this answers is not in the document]
monthonk: yes, it could happen if you open with O_RDWR to create a new file and try to read from it.

mountpoint-s3/src/fs.rs
passaro: I'm still not entirely convinced. The ambiguity of maintaining both fields may only push the complexity down the stack - see for example UploadState::NotStarted - and still make top-level methods more brittle (e.g. what if we "forget" to check for read-only in `FileSystem::write`?).

An enum here would instead always force a check in the top-level read/write methods, but I'd argue that is the right place and where we actually want to transition from StillUnknown to Read or Write. Happy to discuss further though!

passaro: About your last point, I don't see an enum as a blocker for implementing read after write at some point in the future: it would be a transition from Write to Read, right?

mountpoint-s3/tests/fuse_tests/write_test.rs
monthonk: I still think it makes sense that users must close the file handle before they can start the overwrite. Maybe what we should do is `finish_reading` at flush? I updated the semantics document for now.

monthonk:
Rebased on top of https://github.com/awslabs/mountpoint-s3/commit/6e7252dd2e54932e277e5b5ee7000f9bc816a682

mountpoint-s3/src/fs.rs
```diff
@@ -58,8 +58,39 @@ where
 {
     inode: Inode,
     full_key: String,
-    object_size: u64,
-    typ: FileHandleType<Client, Prefetcher>,
+    state: FileHandleState<Client, Prefetcher>,
+}
+
+enum FileHandleState<Client, Prefetcher>
```
passaro, line 64: Not a blocker, but why the 2 layers of enums? It seems that in most (all?) cases where we need to match on one enum we also immediately have to match on the other as well. We could "inline" the `FileHandleType`and replace `Assigned` with its 2 cases.

mountpoint-s3/src/fs.rs
```diff
@@ -58,8 +58,39 @@ where
 {
     inode: Inode,
     full_key: String,
-    object_size: u64,
-    typ: FileHandleType<Client, Prefetcher>,
+    state: FileHandleState<Client, Prefetcher>,
```
passaro, line 61: Doesn't this have to be inside a mutex? Or we'll get a race condition when re-inserting a new value?

passaro (changes requested):
I've only focused on the new state, but I think it needs to be protected by a mutex.

mountpoint-s3/src/fs.rs
monthonk: It's just easier to read in my opinion. Previously, the read handle can be an option which represents created but not yet started, now we also want to do the same with the write handle. It's true that we can add new variants of `FileHandleType` to represent `None` cases or `Closed` handles, but I think it's clearer if we represent them in a new enum and leave the `FileHandleType` to be only read or write.

passaro: Sorry, maybe I wasn't too clear. I'm thinking something like this (with better names, possibly!):

```rust
enum FileHandleState<Client, Prefetcher>
where
    Client: ObjectClient + Send + Sync + 'static,
    Prefetcher: Prefetch,
{
    Created { lookup: LookedUp, flags: i32, pid: u32 },
    Read {
        request: Prefetcher::PrefetchResult<Client>,
        pid: u32,
    },
    Write(UploadState<Client>),
    Closed,
}
```

Definitely not as important as the mutex issue, though.

monthonk: Got it, I see what you mean now.

jamesbornholt (commented):
I think we have the wrong semantics here and need to require `O_TRUNC` for overwrites. 

Consider this little Rust program:
```rust
use std::env;
use std::io::prelude::*;
use std::fs::OpenOptions;

fn main() {
    let args: Vec<String> = env::args().collect();
    let file_path = &args[1];

    // Open for writing but without O_TRUNC or O_APPEND
    let mut file = OpenOptions::new()
        .write(true)
        .open(file_path).expect("Failed to open second file");
    file.write_all(b"222222").expect("Failed to write to second file");
}
```

In POSIX, this should overwrite the first six bytes of the file, but leave the rest intact:
```
[21:02:25] ~/s/rust-scratch (master|✔) $ echo "hello world" > /tmp/hello.txt
[21:02:35] ~/s/rust-scratch (master|✔) $ cat /tmp/hello.txt
hello world
[21:02:38] ~/s/rust-scratch (master|✔) $ cargo run --bin overwrite -- /tmp/hello.txt
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/overwrite /tmp/hello.txt`
[21:02:47] ~/s/rust-scratch (master|✔) $ cat /tmp/hello.txt
222222world
```

But our semantics truncates the file so we get this:
```
[21:02:54] ~/s/rust-scratch (master|✔) $ echo "hello world" > ~/mnt/hello.txt
[21:03:34] ~/s/rust-scratch (master|✔) $ cat ~/mnt/hello.txt
hello world
[21:03:38] ~/s/rust-scratch (master|✔) $ cargo run --bin overwrite -- ~/mnt/hello.txt
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/overwrite /home/bornholt/mnt/hello.txt`
[21:03:50] ~/s/rust-scratch (master|✔) $ cat ~/mnt/hello.txt
222222
```

That's wrong, and while possibly explainable, loses data in a way that is pretty surprising for applications that just want to e.g. overwrite a header. So I think we need to only accept overwrites when `O_TRUNC` is specified.

I believe the fuser docs are incomplete about `O_TRUNC`. There's a [FUSE init option called `FUSE_ATOMIC_O_TRUNC`](https://github.com/torvalds/linux/commit/6ff958edbf39c014eb06b65ad25b736be08c4e63) that makes FUSE include `O_TRUNC` in the open flags. It tells FUSE that our file system handles `O_TRUNC` as part of the `open`, rather than first doing `open` and then sending a `setattr` to set the file size to 0, which is the default way of truncating a file (and which we don't support). I tested this very quickly and turning it on does indeed include `O_TRUNC` in the `open` flags.

monthonk:
Thanks James, didn't know we could make it works with `FUSE_ATOMIC_O_TRUNC` option. I updated the PR with your suggestion.

mountpoint-s3/src/fs.rs
```diff
@@ -130,14 +130,14 @@ where
...
-        if flags == libc::O_RDONLY {
+        if flags & libc::O_ACCMODE == libc::O_RDONLY {
             return Err(err!(libc::EBADF, "file handle is not open for writes"));
         }
 
+        let overwritable = fs.config.allow_overwrite && (flags & libc::O_TRUNC != 0);
         let handle = fs
             .superblock
-            .write(&fs.client, ino, lookup.inode.parent(), pid, fs.config.allow_overwrite)
+            .write(&fs.client, ino, lookup.inode.parent(), pid, overwritable)
```
jamesbornholt, line 140: We should probably have different log messages for the "overwrite not enabled" and "overwrite enabled but not `O_TRUNC`" cases.

mountpoint-s3/src/fs.rs
```diff
@@ -528,6 +528,7 @@ where
 {
     pub async fn init(&self, config: &mut KernelConfig) -> Result<(), libc::c_int> {
         let _ = config.add_capabilities(fuser::consts::FUSE_DO_READDIRPLUS);
+        let _ = config.add_capabilities(fuser::consts::FUSE_ATOMIC_O_TRUNC);
```
jamesbornholt, line 531: This one we shouldn't ignore the return value — if the host doesn't support `FUSE_ATOMIC_O_TRUNC` we should fail, or disable overwrites? I don't know off the top of my head if this works on macOS, for example.

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -731,7 +731,7 @@ where
 
     // Open with O_RDWR and write something to the file
     let mut options = File::options();
-    let mut write_fh = options.read(true).write(true).open(&path).unwrap();
+    let mut write_fh = options.read(true).write(true).truncate(true).open(&path).unwrap();
```
jamesbornholt, line 734: Do we have tests for the non-truncate case (which should fail)? And a test for opened with O_TRUNC but closed without writing anything?

doc/SEMANTICS.md
```diff
@@ -13,7 +13,7 @@ While the rest of this document gives details on specific file system behaviors,
...
 
-Mountpoint supports writing only to new files, and writes to new files must be made sequentially. If you try to open an existing file with write access, the open operation will fail with a permissions error. Mountpoint uploads new files to S3 asynchronously, and optimizes for high write throughput using multiple concurrent upload requests. If your application needs to guarantee that a new file has been uploaded to S3, it should call `fsync` on the file before closing it. You cannot continue writing to the file after calling `fsync`.
+Mountpoint supports writing only to new files by default. Writes to existing files are allowed only if `--allow-overwrite` flag is set at startup time. All the writes must start from the beginning of the files and must be made sequentially. Mountpoint uploads new files to S3 asynchronously, and optimizes for high write throughput using multiple concurrent upload requests. If your application needs to guarantee that a new file has been uploaded to S3, it should call `fsync` on the file before closing it. You cannot continue writing to the file after calling `fsync`.
```
jamesbornholt, line 16:
```suggestion
Mountpoint supports writing only to new files by default. Writes to existing files are allowed if `--allow-overwrite` flag is set at startup time, but only when the `O_TRUNC` flag is used at open time to truncate the existing file. All writes must start from the beginning of the file and must be made sequentially. Mountpoint uploads new files to S3 asynchronously, and optimizes for high write throughput using multiple concurrent upload requests. If your application needs to guarantee that a new file has been uploaded to S3, it should call `fsync` on the file before closing it. You cannot continue writing to the file after calling `fsync`.
```

doc/SEMANTICS.md
```diff
@@ -153,19 +153,23 @@ Windows-style path delimiters (`\`) are not supported.
...
 
 Mountpoint supports sequential write operations (through `write`, `writev`, `pwrite`, `pwritev`),
 but with some limitations:
 
 * Writes are only supported to new files, and must be done sequentially.
 * Modifying existing files is not supported.
-* Truncation is not supported.
+* Append is not supported, the writes must start from the beginning of the files.
+* Truncation is supported on overwrites.
+* Overwrites can be enabled with `--allow-overwrite` flag and only work when the file is opened in truncate mode (`O_TRUNC`). You cannot overwrite files that are being read, your file descriptor must have exclusive access to the file to start writing. The upload will start as soon as Mountpoint receives the first `write` request and cannot be aborted. Once it is started the file is guaranteed to be overwritten.
```
jamesbornholt, line 172: The first bullet is wrong now, and we need to shuffle this around a bit.

Maybe something like:
```
* Writes must start at the beginning of the file and be done sequentially.
* Modifying an existing file is only allowed with the `--allow-overwrite` flag and only when the file is opened in truncate mode (`O_TRUNC`).
    * You cannot overwrite files that are currently being read.
    * The upload to S3 starts as soon as Mountpoint receives the first `write` request and cannot be cancelled.
* Modifying an existing file without using truncate mode is not supported.
```

mountpoint-s3/tests/fuse_tests/write_test.rs
monthonk: Added them in the new commit

passaro (changes requested):
A few more suggestions.

mountpoint-s3/src/inode.rs
```diff
@@ -1045,31 +1051,69 @@ impl LookedUp {
...
-                    parent_ino,
-                    pid,
-                })
+                state.stat.size = 0;
+                Ok(self)
             }
             WriteStatus::LocalOpen => Err(InodeError::InodeAlreadyWriting(inode.err())),
-            WriteStatus::Remote => Err(InodeError::InodeNotWritable(inode.err())),
+            WriteStatus::Remote => {
+                if !self.allow_overwrite {
```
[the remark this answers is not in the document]
monthonk: I think mentioning truncate mode in the first message should be enough?

mountpoint-s3/src/fs.rs
```diff
@@ -961,7 +1037,28 @@ where
             }
         };
         logging::record_name(file_handle.inode.name());
-        let mut request = match &file_handle.typ {
+        let mut state = file_handle.state.lock().await;
+        let typ: &FileHandleType<Client, Prefetcher> = match &*state {
+            FileHandleState::Assigned(typ) => typ,
+            FileHandleState::Created { lookup, flags, pid } => {
+                let is_truncate = flags & libc::O_TRUNC != 0;
+                if !is_truncate && lookup.inode.is_remote()? {
```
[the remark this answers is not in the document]
monthonk: No, we will do nothing only when the file is remote and it’s not open in truncate mode. But, probably swapping the two branches would make it clearer.

monthonk:
Rebased and addressed @passaro's comments.

mountpoint-s3/src/fs.rs
```diff
@@ -991,7 +1073,61 @@ where
...
+        };
+        logging::record_name(file_handle.inode.name());
+        let mut state = file_handle.state.lock().await;
+        match &mut *state {
+            FileHandleState::Created { lookup, flags, pid } => {
+                // This happens when users close a file without calling any read() or write(),
+                // since we don't know what type of handle it would be we need to consider what
+                // to do next for both cases.
+                // * if the file is new or opened in truncate mode, we know it must be a write
+                //   handle so we can start an upload from here.
```
passaro, line 1095: Shouldn't we also complete the upload, like we do in `fsync`?

monthonk: Actually no, because the file is still empty when when start the upload and we want to ignore empty files on flush from that special case with multiple file descriptors.

passaro: I see. Do we even need to start the upload in that case? Is there any difference with just letting `release` do it?

passaro: Just worried that we have similar (but subtly different) logic in a few places.

monthonk: The upload will be started a bit earlier but there should be no other difference. I can remove the logic here and let `release` handle it.

passaro (approved):

PR #504, opened 2023-09-06
Use default thread config for benchmark

monthonk:
## Description of change

Our benchmark scripts are currently running mountpoint with a specific number of thread. However, I think we should use use default configuration instead since that is what most real users would do and it should be a baseline for benchmark.

No

---


passaro (approved):
LGTM

Issue #488, opened 2023-08-30
Readahead reordering causes prefetcher resets


jamesbornholt:
### Mountpoint for Amazon S3 version

mount-s3 1.0.0-unofficial+09f556e

### Describe the running environment

EC2

### What happened?

I ran this command against my mount:
```
cat ~/mnt/test20MiB > /dev/null & cat ~/mnt/test20MiB2 > /dev/null & cat ~/mnt/test20MiB3 > /dev/null &
```
i.e., reading three separate files concurrently.

In the logs, I see:
```
2023-08-30T03:04:41.030144Z  INFO mountpoint_s3::metrics: prefetch.out_of_order: 6 (n=6)
```
so the prefetcher is resetting even though each `cat` is reading sequentially.

It looks like what we're seeing is readahead and multithreading conflicting. For example:
```
2023-08-30T03:04:39.541005Z DEBUG fuser::request: FUSE( 42) ino 0x0000000000000003 READ fh FileHandle(2), offset 131072, size 262144
2023-08-30T03:04:39.541049Z DEBUG fuser::request: FUSE( 44) ino 0x0000000000000003 READ fh FileHandle(2), offset 393216, size 262144
2023-08-30T03:04:39.541118Z TRACE read{req=44 ino=3 fh=2 offset=393216 size=262144}: mountpoint_s3::fs: fs:read with ino 3 fh 2 offset 393216 size 262144
2023-08-30T03:04:39.541144Z TRACE read{req=44 ino=3 fh=2 offset=393216 size=262144}: mountpoint_s3::prefetch: read offset=393216 length=262144 next_seq_offset=131072
2023-08-30T03:04:39.541153Z TRACE read{req=42 ino=3 fh=2 offset=131072 size=262144}: mountpoint_s3::fs: fs:read with ino 3 fh 2 offset 131072 size 262144
2023-08-30T03:04:39.541168Z TRACE read{req=44 ino=3 fh=2 offset=393216 size=262144}: mountpoint_s3::prefetch: out-of-order read, resetting prefetch expected=131072 actual=393216
2023-08-30T03:04:39.541213Z TRACE read{req=44 ino=3 fh=2 offset=393216 size=262144}: mountpoint_s3::prefetch: spawning request range=393216..1572864 size=1179648
[... 80 ms pass ...]
[...]

PR #505, opened 2023-09-06
Update prefetcher to cancel discarded tasks

dannycjones:
## Description of change

Previously, mountpoint-s3 would not cancel prefetch tasks that it was going to ignore. Instead, they would continue to be polled by the executor despite the results never being checked. This could mean up to two (current and at most one "next" GET) would continue to run after the prefetcher had forgotten about them. This change ensures that the task handles are dropped which cancels the task/future.

In the future, we may want to retain some of these tasks where the prefetcher may still be able to make use of them. This refers to the todo comment that is retained from before.

This change was recommended in #488 as a distinct but loosely related issue.

No breaking changes. This should cancel unused GET requests, freeing up bandwidth for GETs that will be used to fulfill mount-s3 reads.

---

### DIFF
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -16,6 +16,7 @@ use std::fmt::Debug;
 use std::time::Duration;
 
 use bytes::Bytes;
+use futures::future::RemoteHandle;
 use futures::pin_mut;
 use futures::stream::StreamExt;
 use futures::task::{Spawn, SpawnExt};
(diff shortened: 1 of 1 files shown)


passaro (approved):
LGTM

Edit: on second look, dropping the task does not cancel the request, so we'll need to implement that on the client if we want this to work.

jamesbornholt (approved):

jamesbornholt:
Won't fix #488 since this won't cancel the inflight GetObject request, but should help free up some resources spent on checksumming at least.

PR #507, opened 2023-09-07
Improve logging and error handling in benchmark script

dannycjones:
## Description of change

I've been running some benchmarks on my own EC2 instance and found the output messy (since fio is rewriting stdout) and also it failed to exit when jq wasn't working. This change ensures the script exits (unless we want to check for exit code), and cleans up the script output a bit.

It also retains intermediary files since we don't need to delete them. They are ignored by the benchmark job in the end.

No.

---

### DIFF
--- a/mountpoint-s3/scripts/fs_bench.sh
+++ b/mountpoint-s3/scripts/fs_bench.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
 
 if ! command -v fio &> /dev/null; then
   echo "fio must be installed to run this benchmark"
@@ -32,7 +33,7 @@ cd ${project_dir}
 results_dir=results
 runtime_seconds=30
 startdelay_seconds=30
-iteration=10
+iterations=10
 
 rm -rf ${results_dir}
 mkdir -p ${results_dir}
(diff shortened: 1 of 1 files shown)


mountpoint-s3/scripts/fs_bench.sh
```diff
@@ -45,43 +46,43 @@ run_fio_job() {
...
   jobs_dir=mountpoint-s3/scripts/fio/read
 
   for job_file in "${jobs_dir}"/*.fio; do
     mount_dir=$(mktemp -d /tmp/fio-XXXXXXXXXXXX)
 
-    echo "Running ${job_name}"
-
     # mount file system
-    cargo run --release ${S3_BUCKET_NAME} ${mount_dir} \
+    set +e
```
sauraank, line 79: I am not very sure about the usage of set -e. Having set +e turns exit on non-zero return off right? Should not that be default?

dannycjones: Exit on non-zero is not on by default (which always surprises me!).

With these changes, I enable exit on non-zero for the whole script. However for this section, I want to tolerate a non-zero exit code and capture it, so that I can provide a more informative error message. I allow non-zero exits using `set +e` and then turn it off again once the mount has succeeded using `set -e`.

jamesbornholt (approved):

Issue #454, opened 2023-08-14
mountpoint-S3 not compatible on centos-7.9


aupadh12:
### Tell us more about this new feature.

Hi Team,

While the mountpoint-s3 is getting installed perfectly fine with RHEL 8 onwards, we are facing dependency issues when trying to install it on centos 7.9
The GLIBC with version  2.17 is default on centos7.9 while the mountpoint-s3 requires GLIBC 2.25 and GLIBC 2.18.
Unfortuantely due to other dependencies, we cannot upgrade to the newer version of GLIBC since we are also running the EC2 instance as submit node for HPC related activities.
The use case for using mountpoint-s3 is to have a shared storage which has  high throughput and can be mounted on multiple instances.
Is there a workaround for this or something which can fixed to get this support?

sauraank:
Hey! To use Mountpoint for amazon S3, you need glibc version >= 2.26. There does not appear to be a workaround as CRT depends on newer version of glibc. ANd Mountpoint tajes direct dependency on CRT.

PR #517, opened 2023-09-15
Build releases on CentOS 7

jamesbornholt:
## Description of change

This gets us compatibility back to glibc 2.17. The tricky part is that CentOS 7 by default packages a GCC that's too old to build the CRT and a Clang that's too old to run bindgen. But they also distribute optional packages (devtoolsets) that update these toolchains and stick them in a separate directory. So this change adopts those, and tweaks the environment variables on the builder to point at the newer tools.

Relevant issues: #454

No.

---

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,12 +1,11 @@
-FROM public.ecr.aws/amazonlinux/amazonlinux:2
+FROM public.ecr.aws/docker/library/centos:7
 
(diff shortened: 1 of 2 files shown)


jamesbornholt:
I've tested the built packages on CentOS 7 and AL2023 and confirmed they work.

passaro (approved):

PR #526, opened 2023-09-20
Complete the upload on flush

passaro:
## Description of change

Currently, Mountpoint will complete an upload in two cases:
* on `release`, that is when the last file descriptor pointing to an open file handle is closed. This is transparent for the caller, but does not allow for reporting the outcome of the upload, nor for blocking until it is completed. This means that a read-after-close may not succeed because the upload is still in progress.
* on `fsync`, which is blocking and can return an error to the caller, but needs to be explicitly invoked before closing a file.

This change implements the `flush` operation, which is invoked when a file descriptor is closed. On `flush`, like on `fsync`, Mountpoint will complete the upload, block, and return on success or failure. In order to support common usage patterns where it is invoked multiple times, `flush`, unlike `fsync`, will be a no-op when invoked before any data has been written or by a different process than the one that originally opened the file.

Yes. Objects will be uploaded before `close` returns.

---

### DIFF
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -1,10 +1,10 @@
 ## Unreleased
 
 ### Breaking changes
-* ...
+* Mountpoint will now complete file uploads at `close` time, and `close` will return an error if the upload was not successful. Before this change, `close` did not wait for the upload to complete, which could cause confusing results for applications that try to read a file they just wrote. ([#526](https://github.com/awslabs/mountpoint-s3/pull/526))
 
 ### Other changes
-* Fixed a bug that cause poor performance for sequential reads in some cases ([#488](https://github.com/awslabs/mountpoint-s3/pull/488)). A workaround we have previously shared for this issue (setting the `--max-threads` argument to `1`) is no longer necessary with this fix. ([#556](https://github.com/awslabs/mountpoint-s3/pull/556))
+* Fixed a bug that caused poor performance for sequential reads in some cases ([#488](https://github.com/awslabs/mountpoint-s3/pull/488)). A workaround we have previously shared for this issue (setting the `--max-threads` argument to `1`) is no longer necessary with this fix. ([#556](https://github.com/awslabs/mountpoint-s3/pull/556))
 
 ## v1.0.2 (September 22, 2023)
 
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -69,6 +69,7 @@ impl<Client: ObjectClient, Runtime> FileHandleType<Client, Runtime> {
         lookup: &LookedUp,
         ino: InodeNo,
         flags: i32,
+        pid: u32,
         fs: &S3Filesystem<Client, Runtime>,
     ) -> Result<FileHandleType<Client, Runtime>, Error> {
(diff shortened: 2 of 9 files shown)


mountpoint-s3/src/fs.rs
```diff
@@ -139,20 +144,29 @@ impl<Client: ObjectClient> UploadState<Client> {
...
     }
 
-    async fn complete(&mut self, key: &str) -> Result<(), Error> {
-        // Check that the upload is still in progress.
-        _ = self.get_upload_in_progress(key)?;
+    async fn complete(&mut self, key: &str, ignore_if_empty: bool) -> Result<(), Error> {
+        match self {
+            Self::InProgress { request, .. } => {
+                if ignore_if_empty && request.size() == 0 {
+                    return Ok(());
```
jamesbornholt, line 158: Maybe log here to help our future selves debugging `flush`-related stuff:
```suggestion
                    trace!(key, "not completing upload because file is empty");
                    return Ok(());
```

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -494,3 +494,53 @@ fn write_file(path: impl AsRef<Path>) -> std::io::Result<()> {
 fn write_with_invalid_storage_class_test_s3(storage_class: &str) {
     write_with_invalid_storage_class_test(crate::fuse_tests::s3_session::new, storage_class);
 }
+
+fn flush_test<F>(creator_fn: F, prefix: &str, append: bool)
```
jamesbornholt, line 498: Another useful test (that maybe we should have written the first time we dealt with `flush`-related weirdness) would be one that literally just shells out to `touch /tmp/mount/file-that-doesnt-exist` and makes sure it doesn't fail. Don't do it if it's too hard, but might be helpful.

jamesbornholt (commented):
Neat!

mountpoint-s3/src/fs.rs
```diff
@@ -764,14 +770,22 @@ where
...
             ret => ret,
         }
     }
 
+    pub async fn fsync(&self, _ino: InodeNo, fh: u64, _datasync: bool) -> Result<(), Error> {
+        self.complete_upload(fh, false).await
+    }
+
+    pub async fn flush(&self, _ino: InodeNo, fh: u64, _lock_owner: u64) -> Result<(), Error> {
+        self.complete_upload(fh, true).await
```
jamesbornholt, line 786: probably worth a comment here summarizing why ignoring on zero length is the behavior we've chosen

jamesbornholt (approved):

mountpoint-s3/CHANGELOG.md
```diff
@@ -1,10 +1,10 @@
 ## Unreleased
 
 ### Breaking changes
-* ...
+* Mountpoint will complete file uploads on close in most cases, and return an error if the upload was not successful. ([#526](https://github.com/awslabs/mountpoint-s3/pull/526))
```
jamesbornholt, line 4:
```suggestion
* Mountpoint will now complete file uploads at `close` time, and `close` will return an error if the upload was not successful. Before this change, `close` did not wait for the upload to complete, which could cause confusing results for applications that try to read a file they just wrote. ([#526](https://github.com/awslabs/mountpoint-s3/pull/526))
```

jamesbornholt (approved):

PR #528, opened 2023-09-21
Set timeout for benchmark jobs and expose mountpoint logs when failing.

monthonk:
## Description of change

We have seen multiple timeouts from the benchmark recently and the default 6 hours timeout is too long. We should be able to fail faster because we know how long each benchmark should be running from the job definitions.

We also want to get mountpoint logs from the failed job so that we can investigate into the problem.

No.

---

### DIFF
--- a/mountpoint-s3/scripts/fs_bench.sh
+++ b/mountpoint-s3/scripts/fs_bench.sh
@@ -42,6 +42,7 @@ run_fio_job() {
   job_file=$1
   bench_file=$2
   mount_dir=$3
+  log_dir=$4
 
   job_name=$(basename "${job_file}")
   job_name="${job_name%.*}"
@@ -51,13 +52,23 @@ run_fio_job() {
   for i in $(seq 1 $iterations);
   do
     echo -n "${i};"
-    fio --thread \
+    # we know each fio job will be running for exactly 1 minute from the job definitions.
+    # there must be something wrong if it takes longer than that.
+    set +e
+    timeout 300s fio --thread \
       --output=${results_dir}/${job_name}_iter${i}.json \
       --output-format=json \
       --directory=${mount_dir} \
       --filename=${bench_file} \
       --eta=never \
       ${job_file}
+    job_status=$?
+    set -e
+    if [ $job_status -ne 0 ]; then
+      tail -1000 ${log_dir}/mountpoint-s3-*
+      echo "Job ${job_name} failed with exit code ${job_status}"
+      exit 1
+    fi
   done
   echo "done"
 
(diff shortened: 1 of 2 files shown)


mountpoint-s3/scripts/fs_bench.sh
```diff
@@ -75,11 +86,19 @@ read_benchmark () {
   for job_file in "${jobs_dir}"/*.fio; do
     mount_dir=$(mktemp -d /tmp/fio-XXXXXXXXXXXX)
 
+    job_name=$(basename "${job_file}")
+    job_name="${job_name%.*}"
+    log_dir=logs/${job_name}
+    rm -rf ${log_dir}
```
sauraank, line 92: Why are we removing the log files for every read job? Same question for write.

mountpoint-s3/scripts/fs_latency_bench.sh
```diff
@@ -67,7 +72,14 @@ do
     iteration=10
     for i in $(seq 1 $iteration);
     do
-        /usr/bin/time -o ${results_dir}/time_output.txt -v ls -f "${target_dir}" >/dev/null 2>&1
+        # we don't know for sure how long does it take, but 5 minutes should be enough
```
sauraank, line 75: Was checking the previous benchmarks. max was around 10-11 seconds. So, 300 s should be fine.

monthonk:
An example of the output when a job has timed out https://github.com/awslabs/mountpoint-s3/actions/runs/6260702935/job/16999287198.

mountpoint-s3/scripts/fs_bench.sh
```diff
@@ -51,13 +52,23 @@ run_fio_job() {
...
       --output=${results_dir}/${job_name}_iter${i}.json \
       --output-format=json \
       --directory=${mount_dir} \
       --filename=${bench_file} \
       --eta=never \
       ${job_file}
+    job_status=$?
+    set -e
+    if [ $job_status -ne 0 ]; then
+      tail -1000 ${log_dir}/mountpoint-s3-*
```
dannycjones, line 68: I'm not familiar with that syntax for tail, but presumably it works.

If we do another revision of this thread, I'd recommend using `-n <NUM>`.

mountpoint-s3/scripts/fs_bench.sh
monthonk: For this PR, I also enable debug logs, that means the log files will be pretty big. I thought we probably run into "no space left" kind of error and want to avoid that.

dannycjones (approved):
LGTM.

I think we can do some more things to cleanup (ensure latency scripts exits on non-zero w/ `set -e`, etc.) but lets merge as is to move forward with the timeout issue.

mountpoint-s3/scripts/fs_bench.sh
monthonk: ah, that might be better. I always use it this way, so I'm quite sure it works.

PR #535, opened 2023-09-26
Correct region warning when endpoint url provided

sauraank:
## Description of change
Corrected region warn message when endpoint url is provided. 
Also added unit test for endpoint resolution when endpoint_url is provided. It is to document the behaviour of resolver. endpoint_url overrides the region argument when endpoint resolution is done. But, signing region is still the region provided. In case no region is provided, Mountpoint detects the right region, so we don't need to make any further changes.

Relevant issues: 
NA

No.

---


dannycjones (commented):
LGTM, just looking for an extra assert on the test

dannycjones (approved):
LGTM, thanks Ankit!

Issue #150, opened 2023-03-15
CSI (Container Storage Interface) driver for Kubernetes and other container orchestrators


cskinfill:
### Tell us more about this new feature.

Any plans to implement a CSI driver(container storage interface) that would allow a kubernetes cluster to mount an s3 bucket as a Persistent Volume?

dannycjones:
Thank you for the feedback. We've heard this ask from a few customers and we're looking into it, but nothing to share right now.

That being said, I encourage you to explore if existing CSI drivers such as the [CSI for Amazon FSx for Lustre](https://github.com/kubernetes-sigs/aws-fsx-csi-driver) may work for your use case if you haven't already. An FSx for Lustre filesystem can be linked to an Amazon S3 bucket, as documented here: https://docs.aws.amazon.com/fsx/latest/LustreGuide/create-dra-linked-data-repo.html

MrMarkW:
@dannycjones Can you provide any updates now that Mountpoint for S3 has been released for GA?  https://aws.amazon.com/blogs/aws/mountpoint-for-amazon-s3-generally-available-and-ready-for-production-workloads/

dannycjones:
I have no update to share right now.

Our goal has been to get Mountpoint into customers' hands as a production product, which we've shared this week with the release of v1.0.0. We're still looking into how we can best support customers using mountpoint-s3 with Kubernetes, but there is no information for me to share at this time.

everpeace:
Hi, I'm devloping/operating an in-house ML platform on the top of EKS clusters and now providing "CSI for Amazon FSx for Lustre" as a high-throughput cache for users to read their S3 data.

And now, we can see mountpoint-s3 as a more cost effective alternative. So, we would be very happy if we will hear that mountpoint-s3 have a plan to support kubernetes CSI in the near future!!

Moreover, normally FUSE(filesystem in user namespace) require privilege operation. But we, as a platform provider, we don't want to give any privilege permission to user pods even to side-car containers. 

So, we would expect mountpoint-s3
- supports kubernetes CSI
- with an appropriate privilege isolation
  - similar to https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver
  - In gcs-fuse-csi-driver,
    - fuse process runs in a sidecar container in a user pod and user pod does NOT require any privilege
    - but, only csi driver daemonset pods require privilege
[...]

PR #537, opened 2023-09-27
Allow mounting to inherited fuse fd (part of #150)

vladem:
## Description of change

- This change allows passing `/dev/fd/N` as a `mount_dir`, which will make mountpoint to bypass mounting and serve fuse requests at the provided file descriptor;
- This allows mountpoint being executed without `CAP_SYS_ADMIN` and with [no_new_privs](https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt) flag, which [corresponds](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) to k8s's `allowPrivilegeEscalation=false` and is enabled by default;
- Similar feature is [provided](https://github.com/libfuse/libfuse/pull/291/files) by `mount.fuse3` command and by [gcs-fuse](https://github.com/GoogleCloudPlatform/gcsfuse/blob/8ab11cd07016a247f64023697383c6e88bc022b0/vendor/github.com/jacobsa/fuse/mount_linux.go#L128-L134);
- Changes are mostly located in `vendor/fuser`, making `fuser::mnt::Mount` struct public is required for testing and may be avoided.

Relevant issues: #150 

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1267,6 +1267,7 @@ dependencies = [
  "memchr",
  "page_size",
  "pkg-config",
+ "regex",
  "serde",
  "smallvec",
  "tempfile",
--- a/mountpoint-s3/src/main.rs
+++ b/mountpoint-s3/src/main.rs
@@ -621,6 +621,15 @@ fn get_region(args_region: Option<String>, instance_info: &InstanceInfo) -> (Str
 fn validate_mount_point(path: impl AsRef<Path>) -> anyhow::Result<()> {
     let mount_point = path.as_ref();
 
+    #[cfg(target_os = "linux")]
+    {
+        // Example: /dev/fd/3
+        let file_descriptor_pattern = Regex::new(r#"^/dev/fd/\d+$"#).unwrap();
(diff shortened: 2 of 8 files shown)


antoniomdk:
Awesome work! this is a feature I'm really interested in using. Any chance this can get reviewed and merged?

jamesbornholt:
@antoniomdk we're still evaluating whether we actually need this or not. Could you tell us a little more about your use case?

antoniomdk:
@jamesbornholt thanks for the heads up. We are using mountpoint-s3 to load ML training data efficiently in K8s, but it requires a sidecar container with CAP_SYS_ADMIN capabilities. Probably this PR would resolve that. The other alternative we are considering is using a device plugin to register the /dev/fuse device and make it available without elevated permissions.

jamesbornholt:
Gotcha. We're actively working on a CSI driver that wraps Mountpoint, which you can track the progress of here: https://github.com/awslabs/mountpoint-s3/issues/150. The sidecar container permissions issue is definitely on our radar there.

PR #543, opened 2023-10-06
Improve the instructions to add benchmark files manually

sauraank:
Improved the instructions to add benchmark files manually in doc/BENCHMARKING.md.

## Description of change
As write is enabled on mountpoint, easiest way to manually add benchmark files is using fio on mounted directory for the bucket. Also, added instruction for adding directories for `readdir` workload using fio.

Relevant issues: 
NA

No

---

### DIFF
--- a/doc/BENCHMARKING.md
+++ b/doc/BENCHMARKING.md
@@ -41,9 +41,9 @@ You can use the following steps.
 
 3. Create the bench files manually in your S3 bucket. You can do that by mounting the bucket on your machine using Mountpoint. Then, running fio jobs against your mount directory to let fio create the files for you.
 
-        mount-s3 DOC-EXAMPLE-BUCKET path/to/mount/ --prefix benchmark/ --part-size=16777216
-        cd mountpoint-s3/scripts/fio/
         MOUNT_DIR = path/to/mount
+        mount-s3 DOC-EXAMPLE-BUCKET $MOUNT_DIR/ --prefix benchmark/ --part-size=16777216
+        cd mountpoint-s3/scripts/fio/
         fio --directory=$MOUNT_DIR/ --filename=bench5MB.bin --create_only=1 read/seq_read_small.fio
         fio --directory=$MOUNT_DIR/ --filename=bench100GB.bin --create_only=1 read/seq_read.fio
         mkdir $MOUNT_DIR/bench_dir_100/ $MOUNT_DIR/bench_dir_1000/ $MOUNT_DIR/bench_dir_10000/ $MOUNT_DIR/bench_dir_100000/
--- a/doc/BENCHMARKING.md
+++ b/doc/BENCHMARKING.md
@@ -26,7 +26,7 @@ We keep the records of benchmarking results in `gh-pages` branch and the perform
 While our benchmark script is written for CI testing only, it is possible to run manually.
 You can use the following steps.
 
-1. Install dependencies and configure FUSE by running the following script in the repository:
+1. Install dependencies and configure FUSE by running the following script in the `mountpoint-s3` repository:
 
         bash .github/actions/install-dependencies/install.sh \
                 --fuse-version 2 \
@@ -39,10 +39,15 @@ You can use the following steps.
         export S3_BUCKET_BENCH_FILE=bench_file_name
         export S3_BUCKET_SMALL_BENCH_FILE=small_bench_file_name
 
-3. Create the bench files manually in your bucket. The size of the files must be exactly the same as the size defined in fio configuration files. The easiest way to do this is running fio against your local file system first to let fio create the files for you, and then upload them to your S3 bucket using the AWS CLI. For example:
+3. Create the bench files manually in your bucket. The size of the files must be exactly the same as the size defined in fio configuration files. The easiest way to do this is mounting your bucket on local file system using mountpoint-s3.Then, running fio against your mount directory to let fio create the files for you. For example:
+* To create small benchmark file for read workload use:
 
-        fio --directory=your_local_dir --filename=your_file_name mountpoint-s3/scripts/fio/read/seq_read_small.fio
-        aws s3 cp your_local_dir/your_file_name s3://${S3_BUCKET_NAME}/${S3_BUCKET_TEST_PREFIX}
+        fio --directory=your_mount_dir --filename=your_file_name mountpoint-s3/scripts/fio/read/seq_read_small.fio
+
+* To create directory with 10000 files for readdir workload use:
+
+        mkdir bench_dir_10000
+        fio --directory=your_mount_dir/bench_dir_10000 --filename=your_file_name mountpoint-s3/scripts/fio/create/create_files_10000.fio
 
 4. Run the benchmark script for [throughput](../mountpoint-s3/scripts/fs_bench.sh) or [latency](../mountpoint-s3/scripts/fs_latency_bench.sh).
 
--- a/doc/BENCHMARKING.md
+++ b/doc/BENCHMARKING.md
@@ -43,13 +43,14 @@ You can use the following steps.
 
         mount-s3 DOC-EXAMPLE-BUCKET path/to/mount/ --prefix benchmark/ --part-size=16777216
         cd mountpoint-s3/scripts/fio/
-        fio --directory=path/to/mount/ --filename=bench5MB.bin read/seq_read_small.fio --create-only=1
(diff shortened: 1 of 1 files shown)


doc/BENCHMARKING.md
```diff
@@ -39,10 +39,17 @@ You can use the following steps.
         export S3_BUCKET_BENCH_FILE=bench_file_name
         export S3_BUCKET_SMALL_BENCH_FILE=small_bench_file_name
 
-3. Create the bench files manually in your bucket. The size of the files must be exactly the same as the size defined in fio configuration files. The easiest way to do this is running fio against your local file system first to let fio create the files for you, and then upload them to your S3 bucket using the AWS CLI. For example:
-
-        fio --directory=your_local_dir --filename=your_file_name mountpoint-s3/scripts/fio/read/seq_read_small.fio
-        aws s3 cp your_local_dir/your_file_name s3://${S3_BUCKET_NAME}/${S3_BUCKET_TEST_PREFIX}
+3. Create the bench files manually in your bucket(create a `benchmark` directory in it if not present). The size of the files must be exactly the same as the size defined in fio configuration files. The easiest way to do this is mounting your bucket on local file system using mountpoint-s3. Then, running fio against your mount directory to let fio create the files for you.
```
monthonk, line 42: The command below already mount with `--prefix benchmark/`, so I think creating a `benchmark` directory is not necessary. More like
```suggestion
3. Create the bench files manually in your S3 bucket. You can do that by mounting the bucket on your machine using Mountpoint. Then, running fio jobs against your mount directory to let fio create the files for you.
```

doc/BENCHMARKING.md
```diff
@@ -39,10 +39,17 @@ You can use the following steps.
...
+
+        mount-s3 your-bucket-name your_mount_dir/ --allow-delete --prefix benchmark/
```
monthonk, line 44: We don't need `--allow-delete`. Also, we use `DOC-EXAMPLE-BUCKET` and `/path/to/mount` in other documents. Probably use the same for consistency.
```suggestion
        mount-s3 DOC-EXAMPLE-BUCKET /path/to/mount --prefix benchmark/
```

doc/BENCHMARKING.md
```diff
@@ -39,10 +39,17 @@ You can use the following steps.
...
+        cd your_mount_dir/
```
monthonk, line 45: cd into the mount directory will make the next commands fail, right? we will get file not found when referencing `mountpoint-s3/scripts/fio/read/seq_read_small.fio`.

Instead, I would prefer to cd into `mountpoint-s3/scripts/fio` all the fio commands will be shorter.

doc/BENCHMARKING.md
```diff
@@ -39,10 +39,17 @@ You can use the following steps.
...
+        fio --directory=your_mount_dir --filename=bench5MB.bin mountpoint-s3/scripts/fio/read/seq_read_small.fio
```
monthonk, line 46: Can we use `--create_only=1` option so we don't really have to run the benchmark? We only want to create the files.

monthonk (changes requested):

doc/BENCHMARKING.md
monthonk: Oh, I forget one thing that we will need `--part-size` because the default part size won't be enough to create a 100GB file.

doc/BENCHMARKING.md
```diff
@@ -39,10 +39,17 @@ You can use the following steps.
...
 
-3. Create the bench files manually in your bucket. The size of the files must be exactly the same as the size defined in fio configuration files. The easiest way to do this is running fio against your local file system first to let fio create the files for you, and then upload them to your S3 bucket using the AWS CLI. For example:
-
-        fio --directory=your_local_dir --filename=your_file_name mountpoint-s3/scripts/fio/read/seq_read_small.fio
-        aws s3 cp your_local_dir/your_file_name s3://${S3_BUCKET_NAME}/${S3_BUCKET_TEST_PREFIX}
+3. Create the bench files manually in your S3 bucket. You can do that by mounting the bucket on your machine using Mountpoint. Then, running fio jobs against your mount directory to let fio create the files for you.
+
+        mount-s3 DOC-EXAMPLE-BUCKET path/to/mount/ --prefix benchmark/ --part-size=16777216
+        cd mountpoint-s3/scripts/fio/
+        fio --directory=path/to/mount/ --filename=bench5MB.bin read/seq_read_small.fio --create-only=1
```
monthonk, line 46: Does this really work? I thought job options are supposed to come before job file. Same for other commands.
```suggestion
        fio --directory=path/to/mount/ --filename=bench5MB.bin --create-only=1 read/seq_read_small.fio
```

doc/BENCHMARKING.md
```diff
@@ -39,10 +39,17 @@ You can use the following steps.
         export S3_BUCKET_BENCH_FILE=bench_file_name
         export S3_BUCKET_SMALL_BENCH_FILE=small_bench_file_name
 
-3. Create the bench files manually in your bucket. The size of the files must be exactly the same as the size defined in fio configuration files. The easiest way to do this is running fio against your local file system first to let fio create the files for you, and then upload them to your S3 bucket using the AWS CLI. For example:
-
-        fio --directory=your_local_dir --filename=your_file_name mountpoint-s3/scripts/fio/read/seq_read_small.fio
-        aws s3 cp your_local_dir/your_file_name s3://${S3_BUCKET_NAME}/${S3_BUCKET_TEST_PREFIX}
+3. Create the bench files manually in your S3 bucket. You can do that by mounting the bucket on your machine using Mountpoint. Then, running fio jobs against your mount directory to let fio create the files for you.
+
+        mount-s3 DOC-EXAMPLE-BUCKET path/to/mount/ --prefix benchmark/ --part-size=16777216
```
monthonk, line 44: Looks like we will have to replace every path/to/mount/ with the real path when following these steps. Maybe it's easier to put it in a variable and use that variable instead.

doc/BENCHMARKING.md
```diff
@@ -39,10 +39,17 @@ You can use the following steps.
...
+        cd mountpoint-s3/scripts/fio/
+        fio --directory=path/to/mount/ --filename=bench5MB.bin read/seq_read_small.fio --create-only=1
+        fio --directory=path/to/mount/ --filename=bench100GB.bin read/seq_read.fio --create-only=1
+        mkdir path/to/mount/bench_dir_100/ path/to/mount/bench_dir_1000/ path/to/mount/bench_dir_10000/ path/to/mount/bench_dir_100000/
+        fio --directory=path/to/mount/bench_dir_100 create/create_files_100.fio --create-only=1
```
monthonk, line 49: We don't need `--create-only=1` for these `create_files_*` jobs. They're doing just that anyway.

monthonk (changes requested):

doc/BENCHMARKING.md
sauraank: `create_only=1` works with it being after the job file. I just tried it for small files it does not run other jobs and just create the file.

monthonk: nit: still other options are the the front. we should group them together, I think.

doc/BENCHMARKING.md
```diff
@@ -39,10 +39,18 @@ You can use the following steps.
         export S3_BUCKET_BENCH_FILE=bench_file_name
         export S3_BUCKET_SMALL_BENCH_FILE=small_bench_file_name
 
-3. Create the bench files manually in your bucket. The size of the files must be exactly the same as the size defined in fio configuration files. The easiest way to do this is running fio against your local file system first to let fio create the files for you, and then upload them to your S3 bucket using the AWS CLI. For example:
-
-        fio --directory=your_local_dir --filename=your_file_name mountpoint-s3/scripts/fio/read/seq_read_small.fio
-        aws s3 cp your_local_dir/your_file_name s3://${S3_BUCKET_NAME}/${S3_BUCKET_TEST_PREFIX}
+3. Create the bench files manually in your S3 bucket. You can do that by mounting the bucket on your machine using Mountpoint. Then, running fio jobs against your mount directory to let fio create the files for you.
+
+        MOUNT_DIR = path/to/mount
```
monthonk, line 44: nit: remove the space.
```suggestion
        MOUNT_DIR=path/to/mount
```

monthonk (approved):
LGTM.

PR #545, opened 2023-10-06
Refactor signing config and set a default signing config for s3 client

monthonk:
## Description of change

This PR refactors the signing config and also set a default signing config for s3 client to prepare for upcoming changes.

No.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -22,8 +22,8 @@ use mountpoint_s3_crt::io::event_loop::EventLoopGroup;
 use mountpoint_s3_crt::io::host_resolver::{HostResolver, HostResolverDefaultOptions};
 use mountpoint_s3_crt::io::retry_strategy::{ExponentialBackoffJitterMode, RetryStrategy, StandardRetryOptions};
 use mountpoint_s3_crt::s3::client::{
-    init_default_signing_config, ChecksumConfig, Client, ClientConfig, MetaRequestOptions, MetaRequestResult,
-    MetaRequestType, RequestType,
+    init_signing_config, ChecksumConfig, Client, ClientConfig, MetaRequestOptions, MetaRequestResult, MetaRequestType,
+    RequestType,
 };
 
 use async_trait::async_trait;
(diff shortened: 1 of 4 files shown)


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -325,12 +331,15 @@ impl S3CrtClientInner {
                 }
             };
             trace!(?auth_scheme, "resolved auth scheme");
-            Some(init_default_signing_config(
+            let algorithm = Some(auth_scheme.scheme_name());
+            let service = Some(auth_scheme.signing_name());
+            let use_double_uri_encode = Some(!auth_scheme.disable_double_encoding());
+            Some(init_signing_config(
                 auth_scheme.signing_region(),
-                auth_scheme.scheme_name(),
```
sauraank, line 330: Reason we were not keeping it optional is because when we were resolving endpoint, we were explicitly failing when we were not getting any value for any of the fields of AuthScheme. So, all these options should have some value set.

mountpoint-s3-crt/src/auth/signing_config.rs
```diff
@@ -32,6 +33,49 @@ impl Debug for SigningConfigInner {
     }
 }
 
+impl SigningConfigInner {
+    /// Create a new [SigningConfig] with default options.
+    pub fn new(region: &str, credentials_provider: CredentialsProvider) -> Self {
```
sauraank, line 38: Why did we create this constructor for SigningConfigInner when we were doing the same with `default_init_siginig_config()`?

mountpoint-s3-client/src/s3_crt_client.rs
monthonk: That's fine if we want to fail, we can do that in `resolve_for_bucket(..)`. But can't we create a signing config without resolving endpoint? Looking at [the CRT function](https://github.com/awslabs/aws-c-s3/blob/76c6d9f13b6eedb9dd82ff96b5d4fad122b909ff/source/s3_util.c#L248-L264), I think what we really need are region and credential_providers, the others are optional.

mountpoint-s3-crt/src/auth/signing_config.rs
monthonk: Well, some of the configs are now optional. I think this might be easier to read.

mountpoint-s3-client/src/s3_crt_client.rs
sauraank: Yes, we dont need other arguments in the CRT function. But, the signing config is only created after resolving the endpoint ( in [s3_crt_client](https://github.com/awslabs/mountpoint-s3/blob/16efe55c27781f8a8728491b10d5040a492bd014/mountpoint-s3-client/src/s3_crt_client.rs#L319) ). I am not sure if we need to set signing config other than making request where endpoint should be resolved.

monthonk: Yeah, that is a signing config at request level. We can also set it at client level which I added in this change (see `s3_crt_client.rs` line [272-279](https://github.com/awslabs/mountpoint-s3/pull/545/files#diff-f6cffb8f99a71f6683ff83dde299952b18332df781964a35320647da246dfc80R272-R279)), and I don't think we want to resolve the endpoint there.

sauraank (approved):

Issue #255, opened 2023-05-23
Reduce cost for repeated accesses


plgounod:
### Tell us more about this new feature.

After reading a file (or a portion of it) using Mountpoint for Amazon S3, customers want to keep the data on their compute instance for a configurable amount of time. With this enhancement, Mountpoint will make fewer GET requests to Amazon S3 when customers repeatedly access the same file data.

stevew3344:
Local file caching would make a huge performance gain in our workflows, where we access the same file dozens or hundreds of times in some cases. Currently I switch to s3fs when I need to run these workflows.

dannycjones:
> Additionally, I wonder if it would be possible to provide a parameter to avoid re-listing the bucket once done once. This is a costly operation and can lead to slower training.

Hey Thomas, would you be able to give a little more detail here on the access pattern? Maybe a short example code?

I'm wondering if the cost you're referring to is caused by some explicit call like `os.listdir(path)` or to the ListObjectsV2 and HeadObject performed when opening individual files.

Either way, I see why listing the bucket/prefix up front and storing the results would help both of these cases.

tchaton:
Hey @dannycjones,

Yes, let's assume the dataset doesn't change through time and it is used in read-only mode. Right now, everytime I am listing ImageNet from python (~1.2M files), it takes 300 seconds. 
   
Here are 2 additional things I wish to have on top of the possibility to cache the files locally:
* 1) Keep the list of files in RAM or within a file when listed once e.g the second time I list from python, it is as fast as listing locally
* 2) Some python utilities to fast list a bucket folder by recursively listing sub-folders in parallel.

dannycjones:
On-going work on this issue is being integrated behind the build-time feature flag `caching`.

https://github.com/awslabs/mountpoint-s3/blob/fa0d516ccb71c09881ef4e87fd87b396853b1a00/mountpoint-s3/Cargo.toml#L64-L66

tchaton:
@dannycjones What are the steps to compile it with the feature flag ?

goldstar611:
> @dannycjones What are the steps to compile it with the feature flag ?

https://github.com/awslabs/mountpoint-s3/blob/main/doc/INSTALL.md#building-mountpoint-for-amazon-s3-from-source

and I think step 4 should be changed from
`cargo build --release`
to 
`cargo build --release --features "caching"`
[...]

PR #547, opened 2023-10-09
Implement CacheConfig field permitting some operations to return non-expired cache data

dannycjones:
## Description of change

This plumbs in checks for if the filesystem should check with S3 to ensure we're consistent with S3 at open, create, etc.

There is no way to configure mountpoint-s3 itself to relax the consistency model - this change only impacts internals.

Relevant issues: #255

If the new field is set to false, it does impact the consistency model of the `mountpoint-s3::fs::S3Filesystem`. This is not configurable during mounting at this time, but may be in the future.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -194,6 +194,13 @@ impl<Client: ObjectClient> UploadState<Client> {
 
 #[derive(Debug, Clone)]
 pub struct CacheConfig {
+    /// Should the file system check S3 even when a valid cached entry may be available?
+    ///
+    /// When enabled, some operations such as `getattr` are allowed to be served from cache
+    /// with a short TTL since Linux filesystems behave badly when the TTL is zero.
+    /// For example, results from `readdir` will expire immediately, and so the kernel will
+    /// immediately `getattr` every entry returned from `readdir`.
+    pub prefer_s3: bool,
     /// How long the kernel will cache metadata for files
     pub file_ttl: Duration,
     /// How long the kernel will cache metadata for directories
@@ -202,8 +209,9 @@ pub struct CacheConfig {
 
 impl Default for CacheConfig {
     fn default() -> Self {
-        // We want to do as little caching as possible, but Linux filesystems behave badly when the
-        // TTL is exactly zero. For example, results from `readdir` will expire immediately, and so
+        // We want to do as little caching as possible by default,
+        // but Linux filesystems behave badly when the TTL is exactly zero.
+        // For example, results from `readdir` will expire immediately, and so
         // the kernel will immediately re-lookup every entry returned from `readdir`. So we apply
         // small non-zero TTLs. The goal is to be small enough that the impact on consistency is
         // minimal, but large enough that a single cache miss doesn't cause a cascading effect where
@@ -213,7 +221,11 @@ impl Default for CacheConfig {
         let file_ttl = Duration::from_millis(100);
         let dir_ttl = Duration::from_millis(1000);
 
-        Self { file_ttl, dir_ttl }
+        Self {
+            prefer_s3: true,
+            file_ttl,
+            dir_ttl,
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


mountpoint-s3/src/inode.rs
```diff
@@ -1197,11 +1262,10 @@ impl From<InodeKind> for FileType {
...
     File {},
     Directory {
-        /// Mapping from child names to [Inode]s.
+        /// Mapping from child names to previously seen [Inode]s.
         ///
-        /// How should this field be used?:
-        /// - **Many operations should maintain** this list.
-        /// - **Only `mknod` and `mkdir` should read** this list, for checking if a file already exists.
+        /// The existence of a child or lack thereof does not imply the object does not exist,
+        /// nor that it current exists in S3 in that state.
```
passaro, line 1268:
```suggestion
        /// nor that it currently exists in S3 in that state.
```

passaro (commented):
Looks good. Mostly minor comments here and there.

Only slight concern is that we need to check the new cache config flag in quite a few different places. It seems to me that `lookup_by_name` could almost be the central place where we check it, if not for one case (when called by `getattr` on `open`).  
Not critical, but it could interesting to explore alternatives where the flag is only used in a single layer: e.g.  move the `getattr` code to a new `SuperblockInner::lookup_by_ino` method, which would also check the flag (to ignore `force_revalidate`) and then try the remote lookup.

mountpoint-s3/src/fs.rs
```diff
@@ -194,6 +194,13 @@ impl<Client: ObjectClient> UploadState<Client> {
 
 #[derive(Debug, Clone)]
 pub struct CacheConfig {
+    /// Should the file system check S3 even when a valid cached entry may be available?
+    ///
+    /// When enabled, some operations such as `lookup` are allowed to be served from cache
```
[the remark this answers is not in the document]
dannycjones: Ah, yeah!

mountpoint-s3/src/inode.rs
```diff
@@ -518,14 +544,50 @@ impl SuperblockInner {
...
 
-        // TODO use caches. if we already know about this name, we just need to revalidate the stat
-        // cache and then read it.
-        let remote = self.remote_lookup(client, parent_ino, name).await?;
-        let lookup = self.update_from_remote(parent_ino, name, remote)?;
+        let lookup = if skip_cache {
+            None
+        } else {
+            let lookup = self.cache_lookup(parent_ino, name);
+            trace!("lookup returned from cache: {:?}", lookup);
```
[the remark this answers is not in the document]
dannycjones: Sounds good. I can provide different trace message for each case.

mountpoint-s3/src/inode.rs
```diff
@@ -518,14 +545,60 @@ impl SuperblockInner {
...
+        };
+
         lookup.inode.verify_child(parent_ino, name.as_ref())?;
         Ok(lookup)
     }
 
+    /// Lookup an [Inode] against known directory entries in the parent,
+    /// verifying any returned entry has not expired.
+    fn cache_lookup(&self, parent_ino: InodeNo, name: &str) -> Option<LookedUp> {
+        fn do_cache_lookup(parent: Inode, name: &str) -> Option<LookedUp> {
```
passaro, line 569: nit: I'm not sure the local fn is appropriate here. Is there no way to make this logic more linear?

dannycjones: I'm wanting to make the core logic clear while still letting us wrap it in traces and metrics.

The difficulty is that there's a lot of `Option` and `Result` unpacking, and so it gets pretty hard to follow if trying to eliminate the use of the `?` operator.

I gave it a go and played around for a bit, but I'm not happy with it:

```rust


    /// Lookup an [Inode] against known directory entries in the parent,
    /// verifying any returned entry has not expired.
    fn cache_lookup(&self, parent_ino: InodeNo, name: &str) -> Option<LookedUp> {
        let parent = self.get(parent_ino).ok()?;
        let lookup = match &parent.get_inode_state().ok()?.kind_data {
            InodeKindData::File { .. } => unreachable!("parent should be a directory!"),
            InodeKindData::Directory { children, .. } => {
                if let Some(inode) = children.get(name) {
                    let valid_stat = inode.get_inode_state().ok()
                        .map(|state| state.stat)
                        .filter(|stat| stat.is_valid());
                    if let Some(valid_stat) = valid_stat {
                        let lookup = LookedUp {
                            inode: inode.clone(),
                            stat: valid_stat.clone(),
                        };
                        Some(lookup)
                    } else {
                        None
                    }
                } else {
                    None
                }
            }
        };

        match &lookup {
            Some(lookup) => trace!("lookup returned from cache: {:?}", lookup),
            None => trace!("no lookup available from cache"),
        }
        metrics::counter!("metadata_cache.cache_hit", lookup.is_some().into());

        lookup
    }
```

I'll leave it as is unless you have any specific suggestions.

dannycjones: Ah, I want `0` on cache miss for that metric. This way, you have from this counter both the number of hits (value of the counter) as well as the number of times we checked the cache (number of times the metric was emitted).

That's the tricky one to plumb in, otherwise your approach is the better option.

Issue #546, opened 2023-10-06
Allow configurable user agent prefix for mountpoint-s3


dannycjones:
### Tell us more about this new feature.

When distributing mountpoint-s3 as part of a broader project, I'd like to include the project name and version. mountpoint-s3 should provide a mechanism to provide this without needing to compile forked source code.

dannycjones:
I suspect what we want here is just a CLI flag like `--user-agent-prefix <STRING>`, whose value will then be prepended to the user agent we configure our mountpoint-s3-client with during creation.

PR #548, opened 2023-10-09
Add configurable user agent prefix flag for mount-s3

dannycjones:
## Description of change

If mountpoint-s3 is run as part of a larger application, user/customer may want to configure a prefix for the user agent to indicate S3 requests came from that particular application.

This change adds a new `--user-agent-prefix <PREFIX>` flag to achieve this.

Relevant issues: #546

No change to existing behaviour. Users can now add a prefix to the HTTP 'User-Agent' header.

---

### DIFF
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Unreleased
 
+### Breaking changes
+* ...
+
+### Other changes
+* Introduced `--user-agent-prefix <PREFIX>` CLI argument ([#548](https://github.com/awslabs/mountpoint-s3/pull/548)) to optionally allow invoker of `mount-s3` to specify an additional prefix for the HTTP User-Agent header.
+
 ## v1.0.2 (September 22, 2023)
 
 ### Breaking changes
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -4,8 +4,8 @@
 * ...
 
 ### Other changes
-* Introduced `--user-agent-prefix <PREFIX>` CLI argument ([#548](https://github.com/awslabs/mountpoint-s3/pull/548)) to optionally allow invoker of `mount-s3` to specify an additional prefix for the HTTP User-Agent header.
 * Fixed a bug that cause poor performance for sequential reads in some cases ([#488](https://github.com/awslabs/mountpoint-s3/pull/488)). A workaround we have previously shared for this issue (setting the `--max-threads` argument to `1`) is no longer necessary with this fix. ([#556](https://github.com/awslabs/mountpoint-s3/pull/556))
+* Introduced `--user-agent-prefix <PREFIX>` CLI argument ([#548](https://github.com/awslabs/mountpoint-s3/pull/548)) to optionally allow invoker of `mount-s3` to specify an additional prefix for the HTTP User-Agent header.
 
 ## v1.0.2 (September 22, 2023)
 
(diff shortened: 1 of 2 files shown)


dannycjones:
Tested this locally by pointing mount-s3 at a local HTTP server. Looks good.

sauraank (dismissed):
LGTM. 
We just need to see why the tests are failing. The ASAN failure did not repeat when I repeat the Address sanitizer job on my PR. But, both the tests are failing in other PRs as well.

dannycjones:
This change should update the changelog: https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/CHANGELOG.md

I'd like to update it before merging this.

dannycjones:
This will conflict with #556. Prefer to merge in #556, rebase this, and merge.

mountpoint-s3/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * ...
 
 ### Other changes
+* Introduced `--user-agent-prefix <PREFIX>` CLI argument ([#548](https://github.com/awslabs/mountpoint-s3/pull/548)) to optionally allow invoker of `mount-s3` to specify an additional prefix for the HTTP User-Agent header.
```
jamesbornholt, line 7: Move this below the bug fix.

Also:
```suggestion
* Introduced the `--user-agent-prefix <PREFIX>` CLI argument ([#548](https://github.com/awslabs/mountpoint-s3/pull/548)) to allow specifying an additional prefix for the HTTP User-Agent header sent in all requests to S3.
```

mountpoint-s3/src/main.rs
```diff
@@ -119,6 +119,14 @@ struct CliArgs {
     )]
     pub allow_other: bool,
 
+    #[clap(
+        long,
+        help = "Configure a string to be prepended to the 'User-Agent' HTTP request header for all S3 requests",
+        value_name = "PREFIX",
+        help_heading = CLIENT_OPTIONS_HEADER
+    )]
+    pub user_agent_prefix: Option<String>,
```
jamesbornholt, line 128: I kind of think this should be an undocumented option. It's really only useful for us right now, right? Not sure it helps customers for it to show up in `--help`. But it would be our first undocumented option. Maybe an alternative is a new help section for "advanced options" or something, to make clear this isn't something most customers who aren't AWS would ever need to use.

dannycjones: I'm actually less comfortable with moving it into a new "Advanced options", since someone looking for this kind of option would expect it to be with "Client options". Our intention is also for things like `--max-threads <>` and `--part-size <>` to be things most customers don't need to configure either.

At the moment, it shows up at the top of the list (since its based on the order it is defined). How do you feel about moving it to the bottom of the "Client options" list like below?

```
Client options:
      --maximum-throughput-gbps <N>  Maximum throughput in Gbps [default: auto-detected on EC2 instances, 10 Gbps elsewhere]
      --max-threads <N>              Maximum number of FUSE daemon threads [default: 16]
      --part-size <PART_SIZE>        Part size for multi-part GET and PUT [default: 8388608]
      --user-agent-prefix <PREFIX>   Configure a string to be prepended to the 'User-Agent' HTTP request header for all S3 requests
```

jamesbornholt: Is anyone really going to be looking for it? Only AWS really gets to see the user agent. Unlike the other options, this is something a customer would basically never configure — it's really only useful for bundling Mountpoint into something else (and that basically means only useful for us).

dannycjones: OK, let me move it to an "Advanced options" section at the bottom.

jamesbornholt (approved):

PR #552, opened 2023-10-11
Introduce new abstraction between the prefetcher and GetObject calls

passaro:
## Description of change

Preliminary work that will allow in the future to plug in a caching layer between the prefetching logic and the GetObject calls on the client.

This change introduce a `ObjectPartFeed` trait that will be used by the prefetcher to obtain the chunks of object data and to adjust the size of requests so that they are "optimally" aligned. The currently only implementation delegates retrieving object data to a `GetObject` call on the client and aligns to requests to `part_size` boundaries, replicating the logic previously in the prefetcher.

Relevant issues: #255 

No functional change.

---

### DIFF
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -8,6 +8,7 @@
 //! non-sequential read, we abandon the prefetching and start again with the minimum request size.
 
 pub mod checksummed_bytes;
+mod feed;
 mod part;
 mod part_queue;
 
@@ -15,20 +16,17 @@ use std::collections::VecDeque;
 use std::fmt::Debug;
 use std::time::Duration;
 
-use bytes::Bytes;
 use futures::future::RemoteHandle;
-use futures::pin_mut;
-use futures::stream::StreamExt;
 use futures::task::{Spawn, SpawnExt};
 use metrics::counter;
 use mountpoint_s3_client::error::{GetObjectError, ObjectClientError};
 use mountpoint_s3_client::types::ETag;
 use mountpoint_s3_client::ObjectClient;
-use mountpoint_s3_crt::checksums::crc32c;
 use thiserror::Error;
 use tracing::{debug_span, error, trace, Instrument};
 
 use crate::prefetch::checksummed_bytes::{ChecksummedBytes, IntegrityError};
+use crate::prefetch::feed::{ClientPartFeed, ObjectPartFeed};
 use crate::prefetch::part::Part;
 use crate::prefetch::part_queue::{unbounded_part_queue, PartQueue};
 use crate::sync::{Arc, RwLock};
@@ -45,8 +43,6 @@ pub struct PrefetcherConfig {
     pub sequential_prefetch_multiplier: usize,
     /// Timeout to wait for a part to become available
     pub read_timeout: Duration,
-    /// The size of the parts that the prefetcher is trying to align with
-    pub part_alignment: usize,
 }
 
 impl Default for PrefetcherConfig {
@@ -65,7 +61,6 @@ impl Default for PrefetcherConfig {
             max_request_size: 2 * 1024 * 1024 * 1024,
             sequential_prefetch_multiplier: 8,
             read_timeout: Duration::from_secs(60),
-            part_alignment: 8 * 1024 * 1024,
         }
     }
 }
@@ -76,11 +71,16 @@ pub struct Prefetcher<Client, Runtime> {
     inner: Arc<PrefetcherInner<Client, Runtime>>,
 }
 
-#[derive(Debug)]
 struct PrefetcherInner<Client, Runtime> {
-    client: Arc<Client>,
+    part_feed: Arc<dyn ObjectPartFeed<Client> + Send + Sync>,
     config: PrefetcherConfig,
     runtime: Runtime,
+}
+
+impl<Client, Runtime> Debug for PrefetcherInner<Client, Runtime> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("PrefetcherInner").field("config", &self.config).finish()
+    }
 }
 
 impl<Client, Runtime> Prefetcher<Client, Runtime>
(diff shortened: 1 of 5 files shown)


mountpoint-s3/src/prefetch.rs
```diff
@@ -76,22 +71,28 @@ pub struct Prefetcher<Client, Runtime> {
...
 
-#[derive(Debug)]
 struct PrefetcherInner<Client, Runtime> {
-    client: Arc<Client>,
+    part_feed: Arc<dyn ObjectPartFeed<Client> + Send + Sync>,
     config: PrefetcherConfig,
     runtime: Runtime,
 }
 
+impl<Client, Runtime> Debug for PrefetcherInner<Client, Runtime> {
```
passaro, line 80: derive(Debug) did not seem to work with `part_feed: Arc<dyn ObjectPartFeed<Client>...>`. Not completely sure why (wrt `client: Arc<Client>`) or whether there is a better solution.

mountpoint-s3/src/prefetch.rs
```diff
@@ -45,8 +43,6 @@ pub struct PrefetcherConfig {
     pub sequential_prefetch_multiplier: usize,
     /// Timeout to wait for a part to become available
     pub read_timeout: Duration,
-    /// The size of the parts that the prefetcher is trying to align with
-    pub part_alignment: usize,
```
passaro, line 49: Now we retrieve this directly from the client

mountpoint-s3/src/prefetch/feed.rs
```diff
@@ -0,0 +1,134 @@
...
+    ) {
+        // The lower bound is 128k as it is the Linux's readahead size
+        const MIN_PART_SIZE: usize = 128 * 1024;
+        // The upper bound is 1MiB since it should be a common IO size.
+        const MAX_PART_SIZE: usize = 1024 * 1024;
+        let part_size = preferred_part_size
+            .unwrap_or_default()
+            .clamp(MIN_PART_SIZE, MAX_PART_SIZE);
+
+        let get_object_result = match self.client.get_object(bucket, key, Some(range), Some(if_match)).await {
```
passaro, line 70: This is moved from `PrefetchGetObject::spawn_next_request`

mountpoint-s3/src/prefetch.rs
```diff
@@ -115,8 +116,8 @@ pub struct PrefetchGetObject<Client: ObjectClient, Runtime> {
     future_tasks: Arc<RwLock<VecDeque<RequestTask<TaskError<Client>>>>>,
     bucket: String,
     key: String,
-    // preferred part size in the prefetcher's part queue, not the object part
-    preferred_part_size: usize,
+    // Tracks the size of read requests
+    max_read_request_size: usize,
```
dannycjones, line 120:
```suggestion
    /// Tracks the maximum size of read request submitted so far
    max_read_request_size: usize,
```

dannycjones (commented):
looking good

dannycjones (approved):

mountpoint-s3/src/prefetch.rs
```diff
@@ -76,22 +71,28 @@ pub struct Prefetcher<Client, Runtime> {
     inner: Arc<PrefetcherInner<Client, Runtime>>,
 }
 
-#[derive(Debug)]
 struct PrefetcherInner<Client, Runtime> {
-    client: Arc<Client>,
+    part_feed: Arc<dyn ObjectPartFeed<Client> + Send + Sync>,
```
jamesbornholt, line 75: Any thoughts on whether this should be a generic type rather than dynamic dispatch? I guess we can get away with it here (unlike `Client`) because `ObjectPartFeed` is object-safe.

mountpoint-s3/src/prefetch.rs
jamesbornholt: It's because the Debug impl for `part_feed: Arc<dyn ObjectPartFeed<Client>...>` has to work for _any_ `ObjectPartFeed` (because the concrete type isn't known), which can only be possible if the `ObjectPartFeed` trait itself guarantees Debug. On the other hand, the `Arc<Client>` version only has to work for the concrete `Client` type(s) that we instantiate it with, which all happen to implement Debug.

mountpoint-s3/src/prefetch.rs
```diff
@@ -161,15 +162,7 @@ where
...
-        // Currently, we set preferred part size to the current read size.
-        // Our assumption is that the read size will be the same for most sequential
-        // read and it can be aligned to the size of prefetched chunks.
-        //
-        // We initialize this value to 128k as it is the Linux's readahead size
-        // and it can also be used as a lower bound in case the read size is too small.
-        // The upper bound is 1MiB since it should be a common IO size.
-        let max_preferred_part_size = 1024 * 1024;
-        self.preferred_part_size = self.preferred_part_size.max(length).min(max_preferred_part_size);
+        self.max_read_request_size = self.max_read_request_size.max(length);
```
jamesbornholt, line 165: It's a bit weird to me that the clamping happens in the part feed rather than here, which is where we already know about readahead and the like. Is there some reason the part feed shouldn't just take and trust whatever part size this code asks for (with the new catch that it might choose to ignore it)?

mountpoint-s3/src/prefetch.rs
passaro: It is a pragmatic choice which we can revert later on. The downside of a new generic type here is that it would propagate up to the root `S3FuseFilesystem` and end up touching most of the code in this crate. 
In fact, I had started working on an alternative option where I replaced `ObjectClient` with a new `ObjectStore` trait, with `get_object_parts` instead of `get_object` (and allowing for further divergence in the future), but that was a much larger change and would have slowed down work on the caching layer.
I still think something like that would be a preferable solution in the long term, but we can review the approach while we work on the cache or soon after.

mountpoint-s3/src/prefetch.rs
passaro: Fair point about the readahead. In theory, clamping early looses information but I don't actually see how it matters in practice. I'll rework.

PR #557, opened 2023-10-16
Add new DataCache trait and InMemoryDataCache implementation

dannycjones:
## Description of change

Part of #255, this contains some early thinking on what our trait should look like for the data cache itself. It also contains an in-memory implementation that can be used for test cases, without setting up a full data cache (which doesn't exist yet).

Relevant issues: #255

No.

---

### DIFF
--- a/mountpoint-s3/src/data_cache.rs
+++ b/mountpoint-s3/src/data_cache.rs
@@ -35,10 +35,10 @@ pub trait DataCache<Key> {
     /// Get block of data from the cache for the given [Key] and [BlockIndex], if available.
     ///
     /// Operation may fail due to errors, or return [None] if the block was not available in the cache.
-    fn get_block(&self, cache_key: &Key, block_idx: BlockIndex) -> DataCacheResult<Option<&ChecksummedBytes>>;
+    fn get_block(&self, cache_key: &Key, block_idx: BlockIndex) -> DataCacheResult<Option<ChecksummedBytes>>;
 
     /// Put block of data to the cache for the given [Key] and [BlockIndex].
-    fn put_block(&mut self, cache_key: Key, block_idx: BlockIndex, bytes: ChecksummedBytes) -> DataCacheResult<()>;
+    fn put_block(&self, cache_key: Key, block_idx: BlockIndex, bytes: ChecksummedBytes) -> DataCacheResult<()>;
 
     /// Returns the block size for the data cache.
     fn block_size(&self) -> u64;
--- a/mountpoint-s3/src/data_cache.rs
+++ b/mountpoint-s3/src/data_cache.rs
@@ -18,9 +18,10 @@ pub type BlockIndex = u64;
 /// Errors returned by operations on a [DataCache]
 #[derive(Debug, Error)]
 pub enum DataCacheError {
-    /// It was not possible to read from the cache
-    #[error("Failed reading or writing from cache: {0}")]
-    IoFailure(#[source] anyhow::Error),
+    #[error("IO error when reading or writing from cache: {0}")]
+    IoFailure(#[from] std::io::Error),
+    #[error("Block content was not valid/readable")]
+    InvalidBlockContent,
 }
 
 pub type DataCacheResult<Value> = Result<Value, DataCacheError>;
--- a/mountpoint-s3/src/data_cache.rs
+++ b/mountpoint-s3/src/data_cache.rs
@@ -4,7 +4,7 @@
 //! reducing both the number of requests as well as the latency for the reads.
 //! Ultimately, this means reduced cost in terms of S3 billing as well as compute time.
 
-mod in_memory_data_cache;
+pub mod in_memory_data_cache;
 
 use std::ops::Range;
 
--- a/mountpoint-s3/src/data_cache.rs
+++ b/mountpoint-s3/src/data_cache.rs
@@ -40,12 +40,11 @@ pub trait DataCache<Key> {
     /// Put block of data to the cache for the given [Key] and [BlockIndex].
     fn put_block(&mut self, cache_key: Key, block_idx: BlockIndex, bytes: ChecksummedBytes) -> DataCacheResult<()>;
 
-    /// For a given byte range, which blocks should contain the cached data if available?
-    ///
-    /// This does not perform any check to see if the relevant data is cached.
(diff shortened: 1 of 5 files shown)


dannycjones:
There will be a small conflict with #556. I'll rebase this once that's merged.

mountpoint-s3/src/data_cache/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+///
+/// The underlying cache is divided into blocks of equal size.
+///
+/// TODO: Deletion and eviction of cache entries.
+/// TODO: Some version information (ETag) independent from [Key] to allow smarter eviction?
+pub trait DataCache<Key> {
+    /// Get block of data from the cache for the given [Key] and [BlockIndex], if available.
+    ///
+    /// Operation may fail due to errors, or return [None] if the block was not available in the cache.
+    fn get_block(&self, cache_key: &Key, block_idx: BlockIndex) -> DataCacheResult<Option<&ChecksummedBytes>>;
```
jamesbornholt, line 38: Two thoughts here:
- Does the client decide how block indexing works (and the implementation treats them as opaque), or do the implementation and client need to agree on how that works somehow?
- If it's the former, can they just be part of the generic `Key` type?

mountpoint-s3/src/data_cache/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    /// Put block of data to the cache for the given [Key] and [BlockIndex].
+    fn put_block(&mut self, cache_key: Key, block_idx: BlockIndex, bytes: ChecksummedBytes) -> DataCacheResult<()>;
```
passaro, line 41: Should we require interior mutability for that data cache? I don't think we want the caller to lock the whole cache, even more for a future disk cache.

passaro (commented):
Good first stab at the trait. Left a few comments.

mountpoint-s3/src/prefetch/checksummed_bytes.rs
```diff
@@ -123,6 +124,23 @@ pub enum IntegrityError {
     #[error("Checksum mismatch. expected: {0:?}, actual: {1:?}")]
     ChecksumMismatch(Crc32c, Crc32c),
 }
+
+/// Compare two [ChecksummedBytes].
```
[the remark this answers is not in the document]
dannycjones: I actually did implement it! However, the meaning felt off.

There were a few options:
- Compare everything. However, it means two ChecksummedBytes that have the same content but different underlying bytes would not be considered equal.
- Compare only the byte slices relevant, but then we don't compare checksum and one could be invalid while the other is valid.

It felt strange and not like something to consider now since its only for testing, however I am happy to reconsider.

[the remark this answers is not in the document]
passaro: I'd say the first option sounds right: an invalid ChecksummedBytes is not equal to a valid one with the same content.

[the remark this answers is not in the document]
dannycjones: Sorry, I wasn't clear. In that first case, I am referring to two `ChecksummedBytes` that represent the same end slice, both with non-corrupt contents but based on different underlying bytes in memory. The `orig_bytes` and checksums would be two different pairings, but the checksum would be valid for its respective `orig_bytes`.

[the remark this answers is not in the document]
passaro: Oh, sorry. I had completely misunderstood.

I see your doubts now, but all considered my preference would still be to implement `PartialEq` and compare the slices and the "integrity states".

mountpoint-s3/src/data_cache/mod.rs
dannycjones: >Does the client decide how block indexing works (and the implementation treats them as opaque), or do the implementation and client need to agree on how that works somehow?

I hadn't fully thought this through yet. I think for now we will move towards the implementation knowing about "blocks" and surfacing things like the cache's block size in the trait. Client will need to use that information to calculate things such as:

- If I want to fetch some data, what ranges should I use for full blocks?
- If I read from blocks, what do I need to trim at the start and end?

I'm hoping we can move that logic into the DataCache trait later, but for now I'm not sure exactly how it'll look like.

mountpoint-s3/src/prefetch/checksummed_bytes.rs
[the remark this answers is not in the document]
dannycjones: Discussed offline, let's implement `PartialEq` but only for tests.

There's weird behaviour where if we compare the validity, what should we do with it? For testing-only implementation, let's go ahead and panic.

mountpoint-s3/src/data_cache/mod.rs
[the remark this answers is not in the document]
dannycjones: For now, we'll build the block size into the trait with some `block_size(&self) -> u64` and leave this responsibility to the caller, which we can move into cache later if it makes sense.

mountpoint-s3/src/data_cache/mod.rs
dannycjones: Yes! Big oversight, let me fix :) I'll update the in-mem test impl to just have a `RwLock<>` over the whole cache, but we can do much better with the on disk implementation.

inevity: We can provide range/extent lock for parallel write。o

dannycjones: Thanks! For now since it is for testing only, I don't plan to add more complexity to the in-memory implementation of `DataCache` trait but we will certainly be exploring how we can lock only a subset of the on-disk cache, such as individual blocks.

mountpoint-s3/src/data_cache.rs
```diff
@@ -0,0 +1,53 @@
+//! Traits and types for data caching.
+//!
+//! The data cache aims to reduce repeated fetches of S3 object content,
+//! reducing both the number of requests as well as the latency for the reads.
+//! Ultimately, this means reduced cost in terms of S3 billing as well as compute time.
+
+mod in_memory_data_cache;
```
passaro, line 7: Should this be public? We need a way to create an in-memory cache.

passaro (approved):

PR #567, opened 2023-10-20
Add request count tests for FS operations with metadata caching enabled

dannycjones:
## Description of change

When adding metadata caching, we not only want to ensure that cached data is served but also verify we aren't still making S3 requests and the repeated S3 requests are truly eliminated.

This adds counter functionality to the mock client in `mountpoint-s3-client`, and adds a few tests covering the cached cases. We may also want to add more assertions to existing tests or test the cases with caching turned off.

Relevant issues: #255 

No change, tests only.

---

### DIFF
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -138,7 +138,7 @@ impl MockClient {
     }
 
     /// Create a new counter for the given operation, starting at 0.
-    pub fn new_counter<'a>(&'a self, operation: Operation) -> OperationCounter<'a> {
+    pub fn new_counter(&self, operation: Operation) -> OperationCounter<'_> {
         let op_counts = self.operation_counts.read().unwrap();
         let initial_count = op_counts.get(&operation).copied().unwrap_or_default();
 
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -1,7 +1,7 @@
 //! A mock implementation of an object client for use in tests.
 
 use std::borrow::Cow;
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::ops::Range;
 use std::pin::Pin;
 use std::sync::{Arc, RwLock};
@@ -60,6 +60,7 @@ pub struct MockClient {
     config: MockClientConfig,
     objects: Arc<RwLock<BTreeMap<String, MockObject>>>,
     in_progress_uploads: Arc<RwLock<BTreeSet<String>>>,
+    operation_counts: Arc<RwLock<HashMap<Operation, u64>>>,
 }
 
 fn add_object(objects: &Arc<RwLock<BTreeMap<String, MockObject>>>, key: &str, value: MockObject) {
@@ -73,6 +74,7 @@ impl MockClient {
             config,
             objects: Default::default(),
             in_progress_uploads: Default::default(),
+            operation_counts: Default::default(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


mountpoint-s3-client/src/mock_client.rs
```diff
@@ -134,6 +136,55 @@ impl MockClient {
...
+/// Obtainable via `new_counter(&Operation)` method on [MockClient]
+/// Its lifetime is bounded by the client which created it.
+pub struct OperationCounter<'a> {
+    client: &'a MockClient,
+    initial_count: u64,
+    operation: Operation,
+}
+
+impl<'a> OperationCounter<'a> {
+    /// Return number of requests for the given operation counter.
```
jamesbornholt, line 181: since the counter was created (rather than the client), right?

mountpoint-s3/tests/fs.rs
```diff
@@ -164,6 +166,225 @@ async fn test_read_dir_nested(prefix: &str) {
...
+        .expect("should succeed as object is cached and exists");
+    assert_eq!(head_counter.count(), 3);
+    assert_eq!(list_counter.count(), 3);
+}
+
+#[tokio::test]
+async fn test_lookup_then_open_cached() {
+    let fs_config = S3FilesystemConfig {
+        cache_config: CacheConfig {
+            prefer_s3: false,
```
jamesbornholt, line 220: nit: this field was weirdly named — it's not really a preference? it's whether metadata can be served from cache or must be refreshed?

mountpoint-s3/tests/fs.rs
```diff
@@ -164,6 +166,225 @@ async fn test_read_dir_nested(prefix: &str) {
...
+
+    let entry = fs.lookup(FUSE_ROOT_INODE, "file1.txt".as_ref()).await.unwrap();
+    let ino = entry.attr.ino;
+    assert_eq!(head_counter.count(), 1);
+    assert_eq!(list_counter.count(), 1);
+
+    let fh = fs.open(ino, S_IFREG as i32, 0).await.unwrap().fh;
+    fs.release(ino, fh, 0, None, true).await.unwrap();
+    assert_eq!(head_counter.count(), 1);
+    assert_eq!(list_counter.count(), 1);
```
jamesbornholt, line 241: to the previous comment, this surprised me until I understood what `prefer_s3` meant

jamesbornholt: probably also worth an uncached version of this test that confirms we do the requests again on `open`

jamesbornholt (commented):
neat!

mountpoint-s3/tests/fs.rs
dannycjones: agree! I'm wondering whether it is enough to find a way to run existing tests with and without caching: if they pass, then it must have gone to S3.

that being said, maybe it's nice to codify the number of requests we expect.

still thinking here.

jamesbornholt: We don't want to go too overboard with hard-coding the exact request counts for all these scenarios in the uncached case, but one uncached test would be helpful.

mountpoint-s3/tests/fs.rs
dannycjones: I've gone through a lot of names with it that I'm not super happy with, and this is no exception. :)

OK, I think the way you worded it is quite nice. `getattr` unfortunately is a bit of a weird case.

Maybe `serve_from_cache`? All operations will be served from cache, except `getattr`. Not perfect but maybe close enough.

jamesbornholt: `serve_lookup_from_cache` seems good enough, obviously it affects `open` too but maybe not worth trying to shove into the name

dannycjones: Addressing in #573.

mountpoint-s3/tests/fs.rs
dannycjones: Added a case here for lookup only

jamesbornholt (approved):

Issue #534, opened 2023-09-25
Support providing SSE-KMS key ID for all new files


meetreks:
### Tell us more about this new feature.

mountpoint-s3 should support passing of the KMS key ID for SSE-KMS, similar how it can be provided for other S3 clients like the SDKs. This will help with customers who insist that any S3 _PutObject_ is required to explicitly specify a known key ARN as part of their S3 Bucket Policy. The policy is a standard policy and recommended best practices.

dannycjones:
Hey @meetreks, thanks for opening this feature request.

I think this is an ask specifically related to SSE-KMS (rather than another encryption mode), so I've updated the title accordingly. I also clarified that it would be for all new files, rather than a per-file configuration/option. Please correct me if I've assumed wrong!

manokaran3529:
Would be a great and must have feature!
Late to the party but surprised to see this isn’t supported yet. It will be a lot helpful if this gets prioritised please.

justinas-b:
Hey! We also use below S3 bucket policies to enforce encryption, which prevents from having write access to volume

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "DenyIncorrectEncryptionHeader",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::mybucket/*",
            "Condition": {
                "StringNotEquals": {
                    "s3:x-amz-server-side-encryption": "AES256"
                }
            }
        },
        {
            "Sid": "DenyUnEncryptedObjectUploads",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::mybucket/*",
            "Condition": {
                "Null": { 
                    "s3:x-amz-server-side-encryption": "true"
                }
            }
        }
    ]
}
```

I see linked pull requests that are merged, @vladem any ideas when we could expect this feature to be available?

vladem:
Hey @justinas-b, thanks for sharing your use-case! We are working on this issue, but unfortunately, I am not able to share any dates just yet. The policy, that you've attached requires slight adjustments to the implementation, that we're working on. We've created a separate issue to track this enhancement: https://github.com/awslabs/mountpoint-s3/issues/807

dannycjones:
We launched this feature in v1.6.0! https://github.com/awslabs/mountpoint-s3/releases/tag/mountpoint-s3-1.6.0

PR #568, opened 2023-10-20
Add support for SSE-KMS and DSSE-KMS for Mountpoint

sauraank:
## Description of change
Added KMS key type and their IDs as bucket option. Added these headers for Put Object. Still need to get these keys as PutObjectParams to be used by PutObject.
Also need to add tests and changelog.

Relevant issues: 
#534 
Yes, there will be added mount options when this change will get merged.

Yes. It will be a breaking change.

---

### DIFF
--- a/mountpoint-s3-client/src/lib.rs
+++ b/mountpoint-s3-client/src/lib.rs
@@ -51,7 +51,7 @@ pub mod failure_client;
 pub mod imds_crt_client;
 #[doc(hidden)]
 pub mod mock_client;
-mod object_client;
+pub mod object_client;
 mod s3_crt_client;
 mod util;
 
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -282,9 +282,9 @@ pub struct PutObjectParams {
     /// Storage class to be used when creating new S3 object
     pub storage_class: Option<String>,
     /// Server side encryption type for creating new S3 object
-    kms_key: Option<KmsKeys>,
+    pub kms_key: Option<KmsKeys>,
     /// Corresponding key id if SSE-KMS key type is opted
-    key_id: Option<String>
+    pub key_id: Option<String>
 }
 
 impl PutObjectParams {
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -266,6 +266,13 @@ pub enum GetObjectAttributesError {
     NoSuchKey,
 }
 
+/// Type of Server side encryption(SSE) opted for Put Object
+#[derive(Debug, Clone)]
+pub enum KmsKeys {
+    SseKms,
+    DsseKms,
+}
+
 /// Parameters to a [`put_object`](ObjectClient::put_object) request
 #[derive(Debug, Default, Clone)]
 #[non_exhaustive]
@@ -274,6 +281,10 @@ pub struct PutObjectParams {
     pub trailing_checksums: bool,
     /// Storage class to be used when creating new S3 object
     pub storage_class: Option<String>,
+    /// Server side encryption type for creating new S3 object
+    kms_key: Option<KmsKeys>,
+    /// Corresponding key id if SSE-KMS key type is opted
+    key_id: Option<String>
 }
 
 impl PutObjectParams {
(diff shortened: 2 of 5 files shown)


mountpoint-s3-client/src/object_client.rs
```diff
@@ -266,6 +266,13 @@ pub enum GetObjectAttributesError {
     NoSuchKey,
 }
 
+/// Type of Server side encryption(SSE) opted for Put Object
+#[derive(Debug, Clone)]
+pub enum KmsKeys {
+    SseKms,
+    DsseKms,
+}
```
dannycjones, line 274: This isn't KMS type but rather Server-side encryption (SSE) type.

We might include SSE-S3, SSE-C, or something new in the future.

mountpoint-s3/src/fs.rs
```diff
@@ -307,6 +308,10 @@ pub struct S3FilesystemConfig {
     pub allow_delete: bool,
     /// Storage class to be used for new object uploads
     pub storage_class: Option<String>,
+    /// KMS key type for Server side encryption
+    pub kms_key: Option<KmsKeys>,
+    /// KMS key ID for Server side encryption
+    pub key_id: Option<String>
```
dannycjones, line 314: I think we probably want an optional enum "Encryption"/"EncryptionConfig"/something better named which can have variants. Both variants will have an optional key ID.

```suggestion
    /// Server-side encryption for new S3 objects, if specified
    pub encryption: Option<Encryption>,
```

I see we also have something similar in `mountpoint-s3-client`... I'm less sure what we want to do there, maybe we want to specifically take each as a field like a normal SDK would do. **Let's discuss this as a team.**

mountpoint-s3/src/main.rs
```diff
@@ -119,6 +120,31 @@ struct CliArgs {
     )]
     pub allow_other: bool,
 
+    #[clap(
+        long,
+        help = "Add the SSE KMS encryption for PUT requests",
+        help_heading = BUCKET_OPTIONS_HEADER,
+        group = "sse_options"
+    )]
+    pub sse_kms: bool,
```
dannycjones, line 129:
```suggestion
    #[clap(
        long,
        help = "Specify SSE-KMS encryption for new objects",
        help_heading = BUCKET_OPTIONS_HEADER,
        group = "sse_options"
    )]
    pub sse_kms: bool,
```

mountpoint-s3/src/main.rs
```diff
@@ -119,6 +120,31 @@ struct CliArgs {
...
+
+    // Keeping [sse_kms] and [dsse_kms] in same arg group makes them mutually exclusive arguments
+    #[clap(
+        long,
+        help = "Add the DSSE KMS encryption for PUT requests",
+        help_heading = BUCKET_OPTIONS_HEADER,
+        group = "sse_options"
+    )]
+    pub dsse_kms: bool,
```
dannycjones, line 138: Drop the comment, or you can instead define the group above the struct like this:

```rust
#[clap(group(
    clap::ArgGroup::new("sse_options")
        .multiple(false)
        .args(&["dsse_kms", "sse_kms"]),
))]
```

Plus update description alike sse_kms.

dannycjones (commented):
Nice start, provided some early feedback.

I think we want to review the CLI arguments themselves as well as the approach for the interface in `mountpoint-s3-client` as a team - let's discuss.

mountpoint-s3-client/src/lib.rs
```diff
@@ -51,7 +51,7 @@ pub mod failure_client;
 pub mod imds_crt_client;
 #[doc(hidden)]
 pub mod mock_client;
-mod object_client;
+pub mod object_client;
```
passaro, line 54: Revert and instead re-export the required type(s) explicitly under the `types` module (see from line 68).

mountpoint-s3-client/src/object_client.rs
jamesbornholt: Each of these variants will potentially have a different set of config associated with it. So maybe make this a richer enum type:
```rust
pub enum ServerSideEncryption {
    /// Use the default configuration for the bucket
    #[default]
    Default,
    Kms {
        key_id: Option<String>,
    },
    DualLayerKms {
        key_id: Option<String>,
    },
    // not now, but in the future
    CustomerProvided {
        key: Key,
    }
}
```

sauraank:
As @vladem is already has a PR #693  up for it, closing this PR.

PR #572, opened 2023-10-23
Use new ChecksummedBlock in DataCache

passaro:
## Description of change

Introduce a new `ChecksummedBlock` type which represents a bytes buffer and its matching checksum. It is a simpler version of `ChecksummedBytes` in that the checksum is always matching the exposed bytes buffer, rather than potentially a containing larger buffer. The new type is better suited to be used in the `DataCache` because it can be more efficiently serialized/deserialized preserving its checksum.

This change also introduces a new `checksums` module, containing both `ChecksummedBytes` and `ChecksummedBlock`, in addition to other checksum functions and types.

Relevant issues: #255 

No changes.

---

### DIFF
--- a/mountpoint-s3/src/checksums/block.rs
+++ b/mountpoint-s3/src/checksums/block.rs
@@ -94,6 +94,20 @@ impl From<ChecksummedBlock> for ChecksummedBytes {
     }
 }
 
+impl From<Bytes> for ChecksummedBlock {
+    fn from(value: Bytes) -> Self {
+        Self::from_bytes(value)
+    }
+}
+
+impl TryFrom<ChecksummedBlock> for Bytes {
+    type Error = IntegrityError;
+
+    fn try_from(value: ChecksummedBlock) -> Result<Self, Self::Error> {
+        value.into_bytes()
+    }
+}
+
 // Implement equality for tests only. We implement equality, and will panic if the data is corrupted.
 #[cfg(test)]
 impl PartialEq for ChecksummedBlock {
@@ -162,8 +176,7 @@ mod tests {
         let expected = Bytes::from_static(b"some bytes extended");
 
         let mut checksummed_block = ChecksummedBlock::from_bytes(bytes);
-        let extend_block = ChecksummedBlock::from_bytes(extend);
-        checksummed_block.extend(extend_block);
+        checksummed_block.extend(extend.into());
         let actual = checksummed_block.bytes;
         assert_eq!(expected, actual);
     }
@@ -175,8 +188,7 @@ mod tests {
         let mut checksummed_block = ChecksummedBlock::new(currupted_bytes, checksum);
 
         let extend = Bytes::from_static(b" extended");
-        let extend_block = ChecksummedBlock::from_bytes(extend);
-        checksummed_block.extend(extend_block);
+        checksummed_block.extend(extend.into());
         assert!(matches!(
             checksummed_block.validate(),
             Err(IntegrityError::ChecksumMismatch(_, _))
--- a/mountpoint-s3/src/checksums/block.rs
+++ b/mountpoint-s3/src/checksums/block.rs
@@ -1,0 +1,211 @@
+use bytes::{Bytes, BytesMut};
+use mountpoint_s3_crt::checksums::crc32c::{self, Crc32c};
+
+use crate::checksums::{bytes::ChecksummedBytes, combine_checksums, IntegrityError};
+
+/// A `ChecksummedBlock` is a bytes buffer that carries its checksum.
+/// The implementation guarantees that its integrity will be validated when data is accessed.
+#[derive(Debug, Clone)]
+pub struct ChecksummedBlock {
+    bytes: Bytes,
+    checksum: Crc32c,
+}
+
+impl ChecksummedBlock {
+    pub fn new(bytes: Bytes, checksum: Crc32c) -> Self {
+        Self { bytes, checksum }
+    }
+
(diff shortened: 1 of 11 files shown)


mountpoint-s3/src/checksums/block.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    #[test]
+    fn test_extend_self_corrupted() {
+        let bytes = Bytes::from_static(b"some bytes");
+        let checksum = crc32c::checksum(&bytes);
+        let mut checksummed_block = ChecksummedBlock::new(bytes, checksum);
+
+        let currupted_bytes = Bytes::from_static(b"corrupted data");
+        checksummed_block.bytes = currupted_bytes.clone();
```
dannycjones, line 180: nit: We can just put the corrupted bytes straight into the `new` call? Might be a bit clearer

alternatively, we also have `from_bytes` so we can make these tests a bit simpler

dannycjones (commented):
LGTM, let's get a second (third?) opinion on the extend fn.

Plus fix typo.

mountpoint-s3/src/checksums/block.rs
```diff
@@ -0,0 +1,213 @@
+use bytes::{Bytes, BytesMut};
+use mountpoint_s3_crt::checksums::crc32c::{self, Crc32c};
+
+use crate::checksums::{bytes::ChecksummedBytes, combine_checksums, IntegrityError};
+
+/// A `ChecksummedBlock` is a bytes buffer that carries its checksum.
+/// The implementation guarantees that its integrity will be validated when data is accessed.
+#[derive(Debug, Clone)]
+pub struct ChecksummedBlock {
```
jamesbornholt, line 9: I was wondering if it might be nicer to have just one implementation of this stuff, and give `ChecksummedBytes` a [`shrink_to_fit`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.shrink_to_fit)-style method to get the guarantee you're looking for. But then I guess that makes `extend` et al more complicated because you have to handle all the different combinations to decide when you can skip validating the checksums, so probably not worth it?

mountpoint-s3/src/checksums/block.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    /// Append the given bytes to current `ChecksummedBlock`.
+    pub fn extend(&mut self, extend: ChecksummedBlock) {
+        if self.is_empty() {
+            *self = extend;
+            return;
+        }
+        if extend.is_empty() {
+            return;
+        }
```
jamesbornholt, line 57: For these cases you probably need to validate the checksum of the empty side (which will be trivial to compute because they're zero-length slices), because the length might have been corrupted.

mountpoint-s3/src/checksums/block.rs
```diff
@@ -0,0 +1,211 @@
...
+        let mut bytes_mut = BytesMut::with_capacity(total_len);
+        bytes_mut.extend_from_slice(&self.bytes);
+        bytes_mut.extend_from_slice(&extend.bytes);
+        let new_bytes = bytes_mut.freeze();
+        let new_checksum = combine_checksums(self.checksum, extend.checksum, extend.len());
+        *self = ChecksummedBlock {
+            bytes: new_bytes,
+            checksum: new_checksum,
+        };
+    }
```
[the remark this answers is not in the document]
jamesbornholt: Yeah, that sounds right. We know the expected checksum of each side (unlike in the `ChecksummedBytes` case where we only know the checksum of some larger slice of each side), and can compute the new expected checksum from those without actually looking at the bytes.

Can you add a comment here capturing that reasoning?

mountpoint-s3/src/checksums/block.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    /// Validate data integrity in this `ChecksummedBlock`.
+    ///
+    /// Return `IntegrityError` on data corruption.
+    pub fn validate(&self) -> Result<(), IntegrityError> {
```
jamesbornholt, line 74: Do we ever use this as public API? If not, might be better to make it private, since it kinda invites time-of-check/time-of-use problems.

mountpoint-s3/src/checksums/block.rs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        if self.checksum == other.checksum {
+            return true;
+        }
+
+        self.validate().expect("should be valid");
+        other.validate().expect("should be valid");
+
+        true
```
jamesbornholt, line 126: this should be unreachable? here we know the bytes are equal but the checksums aren't, but they both passed validation?

mountpoint-s3/src/checksums/block.rs
```diff
@@ -0,0 +1,213 @@
...
+        if self.bytes != other.bytes {
+            return false;
+        }
+
+        if self.checksum == other.checksum {
+            return true;
+        }
+
+        self.validate().expect("should be valid");
+        other.validate().expect("should be valid");
```
jamesbornholt, line 124: Doesn't really matter since it's just test code, but I think you want to do it this way to be correctly bracketed:
```rust
let result = self.bytes == other.bytes;
self.validate().expect("should be valid");
other.validate().expect("should be valid");
result
```

mountpoint-s3/src/checksums/block.rs
passaro: After all, I think `shrink_to_fit` would be a better approach and can also be used to improve `extend`. I will close this PR and open a new one with that change.

PR #575, opened 2023-10-25
Improve ChecksummedBytes::extend and clarify data integrity guarantee

passaro:
## Description of change

`ChecksummedBytes` maintains a data buffer and a checksum and guarantees that only validated data can be accessed. Transformations such as `split_off`, `extend`, or `slice` (introduced in this change), may trigger a validation (and return an `IntegrityError` on failure), or propagate existing checksum(s) if possible, allowing for later validation.

This change clarifies the data integrity guarantee in the docs and optimizes the `extend` method to avoid re-validation when the checksums for both slices can be combined. It also avoid a redundant buffer allocation.

Also moves `ChecksummedBytes` and related types into a new root module.

No.

---

### DIFF
--- a/mountpoint-s3/src/checksums.rs
+++ b/mountpoint-s3/src/checksums.rs
@@ -12,6 +12,7 @@ use thiserror::Error;
 #[derive(Clone, Debug)]
 pub struct ChecksummedBytes {
     orig_bytes: Bytes,
+    /// Always a subslice of `orig_bytes`
     curr_slice: Bytes,
     /// Checksum for `orig_bytes`
     checksum: Crc32c,
@@ -88,7 +89,12 @@ impl ChecksummedBytes {
             return Ok(self.clone());
         }
 
-        let result = Self::from_bytes(self.curr_slice.clone());
+        // Note that no data is copied: `bytes` still points to a subslice of `orig_bytes`.
+        let bytes = self.curr_slice.clone();
+        let checksum = crc32c::checksum(&bytes);
+        let result = Self::new(bytes, checksum);
+
+        // Check the integrity of the whole buffer.
         self.validate()?;
         Ok(result)
     }
--- a/mountpoint-s3/src/checksums.rs
+++ b/mountpoint-s3/src/checksums.rs
@@ -1,9 +1,10 @@
+use std::ops::RangeBounds;
+
 use bytes::{Bytes, BytesMut};
 use mountpoint_s3_crt::checksums::crc32c::{self, Crc32c};
 use thiserror::Error;
 
 /// A `ChecksummedBytes` is a bytes buffer that carries its checksum.
-/// The implementation guarantees that its integrity will be validated when data transformation occurs.
 #[derive(Clone, Debug)]
 pub struct ChecksummedBytes {
     orig_bytes: Bytes,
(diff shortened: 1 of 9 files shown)


passaro:
For a clearer diff of the changes in `ChecksummedBytes` see just [commit 711d37d](https://github.com/awslabs/mountpoint-s3/pull/575/commits/711d37ddda701ecd181b3a9586ff0c80dfd6e7f6).

dannycjones (dismissed):
LGTM, one small comment on `split_off` tests.

Let's get a second review on `extend` and `shrink_to_fit`.

mountpoint-s3/src/checksums.rs
```diff
@@ -63,47 +67,77 @@ impl ChecksummedBytes {
...
+    /// on the slice, rather than on a larger containing buffer.
+    ///
+    /// Return [IntegrityError] if data corruption is detected.
+    pub fn shrink_to_fit(&self) -> Result<Self, IntegrityError> {
+        if self.curr_slice.len() == self.orig_bytes.len() {
+            return Ok(self.clone());
+        }
+
+        let result = Self::from_bytes(self.curr_slice.clone());
+        self.validate()?;
```
jamesbornholt, line 92: This is actually a little subtle — it's only correct because `Self::from_bytes(self.curr_slice.clone())` doesn't copy the underlying bytes, and so `self.validate()` is checking the same physical bytes that the new checksum was computed over in `from_bytes`. If there was a copy involved, we'd have to do some extra work to make sure that copy had its checksum computed at the right time.

I think I'd rather not call `from_bytes` here and instead just do the construction manually, so we can be explicit that this transformation is safe because it's not copying. (It's very unlikely but not impossible that some future developer comes along and changes `from_bytes` to copy, without realizing some callers are making a bracketing assumption). We should probably also add a comment on the `curr_bytes` field to capture an invariant that it's always a subslice of `orig_bytes`.

dannycjones: Thanks for highlighting this, missed this as a rather big assumption.

Agree, I would prefer to do the construction manually.

dannycjones: >We should probably also add a comment on the curr_bytes field to capture an invariant that it's always a subslice of orig_bytes.

Is there some way we can codify that invariant? For example, not by using `split_off` but instead taking a (real) slice of `orig_bytes` and tying the lifetimes together?

jamesbornholt: Self-referential lifetimes are pretty annoying (usually require `unsafe`), and `Bytes` is pretty heavily designed around reference counting, so probably not something we can codify that way.

If we wanted an executable check, we could do something janky like:
```rust
let curr_slice = &self.curr_slice[..];
// will panic if curr_slice isn't a subslice of orig_slice
let _ = self.orig_slice.slice_ref(curr_slice);
```
maybe as a debug assert? But possibly not worth the trouble.

dannycjones: Got it. Plus even if the lifetimes worked, I don't think it would even be `Send` anymore and unusable in async code.

Random idea: maybe we consider storing the bounds/range in place of `curr_slice`, and create the smaller `Bytes` at the last minute when performing `into_bytes`. Looks like an involved refactor though at first glance.

No opinion here, I'm OK keeping it as a documented invariant.

passaro: I've added the comment to `curr_slice` for now

dannycjones (approved):

jamesbornholt (approved):

PR #579, opened 2023-10-26
Replace use of Range with RangeBounds in DataCache trait

dannycjones:
## Description of change

[RangeBounds](https://doc.rust-lang.org/std/ops/trait.RangeBounds.html) is better than [Range](https://doc.rust-lang.org/std/ops/struct.Range.html), which represents only a range where both start and end is defined. IDE wasn't working last time I tried to use it, now I see the light.

Relevant issues: N/A

No.

---

### DIFF
--- a/mountpoint-s3/src/data_cache/in_memory_data_cache.rs
+++ b/mountpoint-s3/src/data_cache/in_memory_data_cache.rs
@@ -3,7 +3,7 @@
 use std::collections::HashMap;
 use std::default::Default;
 use std::hash::Hash;
-use std::ops::Range;
+use std::ops::RangeBounds;
 
 use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheResult};
 use crate::sync::RwLock;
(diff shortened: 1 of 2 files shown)


mountpoint-s3/src/data_cache/in_memory_data_cache.rs
```diff
@@ -42,12 +42,17 @@ impl<Key: Eq + Hash> DataCache<Key> for InMemoryDataCache<Key> {
...
+        &self,
+        cache_key: &Key,
+        range: R,
+    ) -> DataCacheResult<Vec<BlockIndex>> {
         let data = self.data.read().unwrap();
-        let result = match data.get(cache_key) {
+        let mut result = match data.get(cache_key) {
             None => Vec::new(),
-            Some(blocks) => range.into_iter().filter(|idx| blocks.contains_key(idx)).collect(),
+            Some(blocks) => blocks.keys().filter(|idx| range.contains(idx)).copied().collect(),
```
passaro, line 53: Doesn't this break the guarantee about the order of the blocks?
```
/// Indices in the vector are already sorted.
```

dannycjones: Vector is sorted on line 55. It does mean we need to explicitly sort, but this implementation is only for testing.

passaro (approved):

PR #581, opened 2023-10-27
Allow repeated readdir offsets

### LINKED ISSUE #477: PHP DirectoryIterator returns no files and triggers a "readdirplus failed" error
### Mountpoint for Amazon S3 version

mount-s3 1.0.0

### AWS Region

eu-west-1

### Describe the running environment

Running in EC2 on Ubuntu with static access and secret key configured via awcli

### What happened?

- I created a test mount folder: `mkdir /home/ubuntu/mount`
- I mounted a bucket with some files in it: `mount-s3 my-test-bucket /home/ubuntu/mount`
- I ensured my files show on ls: `ls /home/ubuntu/mount` (list of files appears)
- I installed PHP cli: `sudo apt install php-cli`
- I created this PHP script:
```php
<?php
$count = 0;
foreach (new \DirectoryIterator("/home/ubuntu/mount") as $fileInfo) {
    echo "Found file: " . $fileInfo->getFilename() . "\n";
    $count++;
}
echo "$count files found\n";
```
- I launched the script: `php test.php`

Of course, I expected the list and count of files, but I get a "0 files found" instead.

When I run the script, a warning gets logged on syslog (pasted below). I tried multiple times with the same result, only "req=" and "fh=" values change.

### Relevant log output

```shell
mount-s3[8770]: [WARN] readdirplus{req=30 ino=1 fh=3 offset=0}: mountpoint_s3::fuse: readdirplus failed: offset mismatch, expected=8, actual=0
```

### LINKED ISSUE #520: Cannot see mounted folders in Midnight Commander
### Mountpoint for Amazon S3 version

mount-s3 1.0.1

### AWS Region

us-west-2

### Describe the running environment

Running Amazon Linux 2023 in EC2 with midnight commander (install mc package)

### What happened?

I mount a bucket and can see folders

```
ec2-user@froster:/mnt/s3/froster$ ls -l
total 0
drwxr-xr-x. 2 root root 0 Sep 16 22:27 archive
drwxr-xr-x. 2 root root 0 Sep 16 22:27 deep_archive
```

but when running standard mc there is nothing there:

image

Btw: I noticed that df is reporting all zeros, while this may be correct many other fuse programmers have stumbled over this you might give it 1EB of size and also 1EB of free space just for compatibility reasons ?

```
ec2-user@froster:/mnt/s3/froster$ df -h /mnt/s3/froster
Filesystem      Size  Used Avail Use% Mounted on
mountpoint-s3      0     0     0    - /mnt/s3/froster
```

### Relevant log output

_No response_

jamesbornholt:
## Description of change

POSIX allows seeking an open directory handle, which in FUSE means the
`offset` can be any offset we've previously returned. This is pretty
annoying for us to implement since we're streaming directory entries
from S3 with ListObjects, which can't resume from an arbitrary index,
and can't fit its continuation tokens into a 64-bit offset anyway. So
we're probably never going to truly support seeking a directory handle.

But there's a special case we've seen come up a couple of times (#477, #520):
some applications read one page of directory entries and then seek back
to 0 and do it again. I don't fully understand _why_ they do this, but
it's common enough that it's worth special casing.

This change makes open directory handles remember their most recent
response so that they can repeat it if asked for the same offset again.
It's not too complicated other than needing to make sure we do
readdirplus correctly (managing the lookup counts for entries that are
being returned a second time).

I've tested this by running the PHP example from #477, which now works.

Related issues: fixes #477, fixes #520.

No, it allows a case to succeed that previously failed.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -775,7 +775,7 @@ where
             // request offset 0 twice in a row. So we remember the last response and, if repeated,
             // we return it again.
             let last_response = dir_handle.last_response.lock().await;
-            if let Some((last_offset, entries)) = &*last_response {
+            if let Some((last_offset, entries)) = last_response.as_ref() {
                 if offset == *last_offset {
                     trace!(offset, "repeating readdir response");
                     for entry in entries {
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -3,7 +3,7 @@
 use futures::task::Spawn;
 use nix::unistd::{getgid, getuid};
 use std::collections::HashMap;
-use std::ffi::OsStr;
+use std::ffi::{OsStr, OsString};
 use std::str::FromStr;
 use std::time::{Duration, UNIX_EPOCH};
 use time::OffsetDateTime;
@@ -35,6 +35,7 @@ struct DirHandle {
     ino: InodeNo,
     handle: ReaddirHandle,
     offset: AtomicI64,
+    last_response: AsyncMutex<Option<(i64, Vec<DirectoryEntry>)>>,
 }
 
 impl DirHandle {
@@ -402,15 +403,18 @@ pub struct Opened {
 pub trait DirectoryReplier {
     /// Add a new dentry to the reply. Returns true if the buffer was full and so the entry was not
     /// added.
-    fn add<T: AsRef<OsStr>>(
-        &mut self,
-        ino: u64,
-        offset: i64,
-        name: T,
-        attr: FileAttr,
-        generation: u64,
-        ttl: Duration,
-    ) -> bool;
+    fn add(&mut self, entry: DirectoryEntry) -> bool;
+}
+
+#[derive(Debug, Clone)]
+pub struct DirectoryEntry {
+    pub ino: u64,
+    pub offset: i64,
+    pub name: OsString,
+    pub attr: FileAttr,
+    pub generation: u64,
+    pub ttl: Duration,
+    lookup: LookedUp,
 }
 
 /// Reply to a `read` call. This is funky because we want the reply to happen with only a borrow of
@@ -717,6 +721,7 @@ where
             ino: parent,
             handle: inode_handle,
             offset: AtomicI64::new(0),
+            last_response: AsyncMutex::new(None),
         };
 
         let mut dir_handles = self.dir_handles.write().await;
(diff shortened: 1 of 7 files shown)


mountpoint-s3/src/fs.rs
```diff
@@ -764,19 +769,77 @@ where
         };
 
         if offset != dir_handle.offset() {
+            // POSIX allows seeking an open directory. That's a pain for us since we are streaming
+            // the directory entries and don't want to keep them all in memory. But one common case
+            // we've seen (https://github.com/awslabs/mountpoint-s3/issues/477) is applications that
+            // request offset 0 twice in a row. So we remember the last response and, if repeated,
+            // we return it again.
+            let last_response = dir_handle.last_response.lock().await;
+            if let Some((last_offset, entries)) = &*last_response {
```
dannycjones, line 778: I had to do a bit of juggling in my head to figure out what `&*last_response` is doing: de-referencing the guard to get the value, then taking only the reference to that value.

Is it clearer to use Option's `as_ref()`? De-referencing of the guard behaves like normal (coerced), then we explicitly take a ref to the underlying values of the Option. https://doc.rust-lang.org/std/option/enum.Option.html#method.as_ref

```suggestion
            if let Some((last_offset, entries)) = last_response.as_ref() {
```

mountpoint-s3/src/fs.rs
```diff
@@ -764,19 +769,77 @@ where
...
+                if offset == *last_offset {
+                    trace!(offset, "repeating readdir response");
+                    for entry in entries {
+                        if reply.add(entry.clone()) {
+                            break;
+                        }
```
dannycjones, line 784: I'm a bit uncomfortable with this right now.

If the buffer is full, we should be keeping the entries we've not returned and return them in the next call? I'm guessing this typically works anyway since the entries are the same and the buffer presumably will be the same size, but I'd prefer we handled this case with an error and log that this case occurred.

mountpoint-s3/tests/fs.rs
```diff
@@ -1232,3 +1232,65 @@ async fn test_flexible_retrieval_objects() {
...
+    let mut new_reply = DirectoryReply::new(5);
+    let _ = fs
+        .readdirplus(FUSE_ROOT_INODE, dir_handle, 0, &mut new_reply)
+        .await
+        .unwrap();
+    let new_entries = new_reply
+        .entries
+        .iter()
+        .map(|e| (e.ino, e.name.clone()))
+        .collect::<Vec<_>>();
```
dannycjones, line 1271: Related to the other comment, I'm wondering what the behaviour should be if this was `DirectoryReply::new(3)` for the repeated `readdirplus`. I'm imagining out of order, and then we can still repeat with the same offset and reply size and it would be OK?

dannycjones (commented):
Nice, I was hoping we might add support for this scenario soon.

I have just one concern on the repeated reply sizes matching, see comments.

dannycjones (commented):
I think a `CHANGELOG.md` entry would be worth adding since it'll be a 'bug fix' / 'new feature' to be able to support languages which use readdir in this way!

mountpoint-s3/src/fs.rs
jamesbornholt: I thought about doing this, but it seemed a bit too much complexity -- we'd have to pop the used entries off the vector and update the offset or something. The contract for readdir is just "return as many entries as you can fit into the buffer, starting from this offset", which this satisfies. If the buffer got smaller (which won't happen in practice), it seems fine to just fulfil the contract the first time around and give up on the second attempt.

mountpoint-s3/src/fs.rs
jamesbornholt: Yeah, `&*` is just an old habit.

mountpoint-s3/tests/fs.rs
jamesbornholt: I added more test cases to make this clear -- the size of the buffer never matters, only the offset.

mountpoint-s3/src/fs.rs
dannycjones: Ah nice! I hadn't noticed that we can still continue so long as the next offset is what we expect from the first request. Yep, agree to just error out if its one of those weird edge cases.

Thanks for the new tests, that's clear now.

dannycjones (approved):
LGTM!

Issue #577, opened 2023-10-26
Future creates fail for files that are deleted remotely


sauraank:
### Mountpoint for Amazon S3 version

mountpoint-s3 v1.1.0

### AWS Region

eu-west-1

### Describe the running environment

Running in EC2 on OS Amazon Linux2 with S3 bucket in the same account as EC2 instance.

### What happened?

When we are creating a new object and then deleting the object using CLI/Console it gets deleted successfully. However when we try re-creating the same object name it gives error `No such file or directory`.

### Relevant log output

```shell
Sequence of Fuse requests made: 

2023-10-25T14:31:28.613010Z DEBUG fuser::request: FUSE(  2) ino 0x0000000000000000 INIT kernel ABI 7.37, capabilities 0x73fffffb, max readahead 131072    
2023-10-25T14:31:28.613358Z DEBUG fuser::request: INIT response: ABI 7.28, flags 0x402021, max readahead 131072, max write 16777216    
2023-10-25T14:31:32.648763Z DEBUG fuser::request: FUSE(  4) ino 0x0000000000000001 GETATTR    
2023-10-25T14:31:57.381988Z DEBUG fuser::request: FUSE(  6) ino 0x0000000000000001 LOOKUP name "ini.txt"    
2023-10-25T14:31:57.425159Z DEBUG fuser::request: FUSE(  8) ino 0x0000000000000001 CREATE name "ini.txt", mode 0o100664, flags 0x8241    
2023-10-25T14:31:57.425492Z DEBUG fuser::request: FUSE( 10) ino 0x0000000000000001 MKNOD name "ini.txt", mode 0o100664, rdev 0    
2023-10-25T14:31:57.471106Z DEBUG fuser::request: FUSE( 12) ino 0x0000000000000002 OPEN flags 0x8001    
2023-10-25T14:31:57.511767Z DEBUG fuser::request: FUSE( 14) ino 0x0000000000000002 FLUSH fh FileHandle(1), lock owner LockOwner(10151073290680498340)    
2023-10-25T14:31:57.512014Z DEBUG fuser::request: FUSE( 16) ino 0x0000000000000002 GETXATTR name "security.capability", size GetSize(GetXAttrSize)    
2023-10-25T14:31:57.512157Z DEBUG fuser::request: FUSE( 18) ino 0x0000000000000002 WRITE fh FileHandle(1), offset 0, size 6, write flags 0x0    
[...]

PR #584, opened 2023-10-30
Change local file/directory expiry TTL from NEVER_EXPIRE to 100ms

sauraank:
## Description of change
This change allows mountpoint to rewrite to the same file which was deleted remotely using other methods like console/aws cli. Earlier, Due to large validity of files and directories, the inodes corresponding were remembered by kernel forever. So, even deleting remotely did not make kernel call `lookup` again for those files.

Relevant issues: 
#577 
No.

There is no breaking change.

---

### DIFF
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -834,9 +834,18 @@ impl SuperblockInner {
                     unreachable!("we know parent is a directory");
                 };
                 if writing_children.contains(&existing_inode.ino()) {
-                    let stat = existing_inode.get_inode_state()?.stat.clone();
+                    let mut sync = existing_inode.get_mut_inode_state()?;
+
+                    let validity = match existing_inode.kind() {
+                        InodeKind::File => self.cache_config.file_ttl,
+                        InodeKind::Directory => self.cache_config.dir_ttl,
+                    };
+                    sync.stat.update_validity(validity);
+                    let stat = sync.stat.clone();
+                    drop(sync);
+
                     Ok(LookedUp {
-                        inode: existing_inode,
+                        inode: existing_inode.clone(),
                         stat,
                     })
                 } else {
@@ -2537,15 +2546,7 @@ mod tests {
         let stat = lookup.stat;
         assert_eq!(stat.atime, atime);
         assert_eq!(stat.mtime, mtime);
-
-        // getattr() validates that InodeStat is not expired as the Inode is still Local and it would have not been able to lookup stat remotely.
-        let lookup = superblock
-            .getattr(&client, ino, false)
-            .await
-            .expect("getattr should be successful");
-        let stat = lookup.stat;
-        assert_eq!(stat.atime, atime);
-        assert_eq!(stat.mtime, mtime);
+        assert!(stat.is_valid());
     }
 
     #[test]
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -343,9 +343,9 @@ impl Superblock {
                 None,
                 None,
                 None,
-                Duration::from_millis(100),
+                self.inner.cache_config.file_ttl,
             ),
-            InodeKind::Directory => InodeStat::for_directory(self.inner.mount_time, Duration::from_millis(100)),
+            InodeKind::Directory => InodeStat::for_directory(self.inner.mount_time, self.inner.cache_config.dir_ttl),
         };
 
         let state = InodeState {
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -242,6 +242,15 @@ impl Superblock {
             return Err(InodeError::SetAttrNotPermittedOnRemoteInode(inode.err()));
         }
 
+        let validiy = match inode.kind() {
+            InodeKind::File => self.inner.cache_config.file_ttl,
+            InodeKind::Directory => self.inner.cache_config.dir_ttl,
+        };
+
+        sync.stat.expiry = Instant::now()
+            .checked_add(validiy)
+            .expect("64-bit time shouldn't overflow");
+
         if let Some(t) = atime {
             sync.stat.atime = t;
         }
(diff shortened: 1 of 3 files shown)


mountpoint-s3/src/inode.rs
```diff
@@ -337,8 +337,15 @@ impl Superblock {
...
-            InodeKind::Directory => InodeStat::for_directory(self.inner.mount_time, NEVER_EXPIRE_TTL),
+            InodeKind::File => InodeStat::for_file(
+                0,
+                OffsetDateTime::now_utc(),
+                None,
+                None,
+                None,
+                Duration::from_millis(100),
+            ),
+            InodeKind::Directory => InodeStat::for_directory(self.inner.mount_time, Duration::from_millis(100)),
```
dannycjones, line 348: Let's use the duration specified in `self.inner.cache_config.file_ttl`, etc..

Does this work?

```suggestion
            InodeKind::File => InodeStat::for_file(
                0,
                OffsetDateTime::now_utc(),
                None,
                None,
                None,
                self.inner.cache_config.file_ttl,
            ),
            InodeKind::Directory => InodeStat::for_directory(self.inner.mount_time, self.inner.cache_config.file_ttl),
```

dannycjones (changes requested):

mountpoint-s3/src/inode.rs
sauraank: One should be dir_ttl, right?

dannycjones: oops, yes!

mountpoint-s3/src/inode.rs
```diff
@@ -337,8 +337,15 @@ impl Superblock {
         // Local inode stats never expire, because they can't be looked up remotely
```
dannycjones, line 337: Also need to update this comment

dannycjones (commented):
One comment to fix, otherwise LGTM.

dannycjones:
The integration tests are failing and I suspect it is related to this change, can you investigate @sauraank? For example, `setattr` is failing (although not every time!).

mountpoint-s3/src/inode.rs
```diff
@@ -334,11 +328,17 @@ impl Superblock {
...
             // Objects don't have an ETag until they are uploaded to S3
-            InodeKind::File => InodeStat::for_file(0, OffsetDateTime::now_utc(), None, None, None, NEVER_EXPIRE_TTL),
-            InodeKind::Directory => InodeStat::for_directory(self.inner.mount_time, NEVER_EXPIRE_TTL),
+            InodeKind::File => InodeStat::for_file(
+                0,
+                OffsetDateTime::now_utc(),
+                None,
+                None,
+                None,
+                self.inner.cache_config.file_ttl,
```
dannycjones, line 339: Maybe this will never happen since I'm guessing applications almost always open the file immediately after performing a `mknod`, but an `open` 100ms after the `mknod` will now go to S3 even though we know we'll always choose the local node over a remote one.

Not worried right now, but just noting this observation.

mountpoint-s3/src/inode.rs
```diff
@@ -242,11 +242,14 @@ impl Superblock {
...
-        if !sync.stat.is_valid() {
-            error!(?ino, "local inode stat already expired");
-            return Err(InodeError::SetAttrOnExpiredStat(inode.err()));
-        }
+        let validiy = match inode.kind() {
+            InodeKind::File => self.inner.cache_config.file_ttl,
+            InodeKind::Directory => self.inner.cache_config.dir_ttl,
+        };
+
+        sync.stat.expiry = Instant::now()
```
passaro, line 250: Why are we refreshing the validity here? If required, shouldn't we do it in `update_slow_path`?

passaro: Also, why not call `update_validity`?

mountpoint-s3/src/inode.rs
```diff
@@ -242,11 +242,14 @@ impl Superblock {
             return Err(InodeError::SetAttrNotPermittedOnRemoteInode(inode.err()));
         }
 
-        // Should be impossible since local file stat never expire.
-        if !sync.stat.is_valid() {
-            error!(?ino, "local inode stat already expired");
-            return Err(InodeError::SetAttrOnExpiredStat(inode.err()));
```
passaro, line 248: Is `SetAttrOnExpiredStat` still used anywhere else?

passaro (changes requested):

mountpoint-s3/src/inode.rs
sauraank: Yes, I should use `update_validity`. We did not do this `update_slow_path` because it does not update the Inode Stat for local inode. This case is specifically for Local Inode as we are setting new InodeStat, we want it to have renewed expiry.

mountpoint-s3/src/inode.rs
```diff
@@ -2522,11 +2527,25 @@ mod tests {
         let stat = inode.get_inode_state().unwrap().stat.clone();
         assert!(!stat.is_valid());
 
-        // Should get an error back when calling setattr
+        // Should be able to reset expiry back and make stat valid when calling setattr
         let atime = OffsetDateTime::UNIX_EPOCH + Duration::days(90);
         let mtime = OffsetDateTime::UNIX_EPOCH + Duration::days(60);
-        let result = superblock.setattr(&client, ino, Some(atime), Some(mtime)).await;
-        assert!(matches!(result, Err(InodeError::SetAttrOnExpiredStat(_))));
+        let lookup = superblock
```
sauraank, line 2533: I still need to see if there is any other possible scenario is where InodeStat may be invalid and does resetting the InodeStat is correct in that case.

mountpoint-s3/src/inode.rs
passaro: That's my point: now that we are not using NEVER_EXPIRES_TTL anymore for local inodes, maybe we should update their stats on lookup, likely in `update_slow_path`. Otherwise, local inodes will never have their initial expiry updated, except on calls to `settattr` (why?), and will always immediately trigger an S3 lookup to confirm they are still local.

mountpoint-s3/src/inode.rs
```diff
@@ -2522,11 +2527,25 @@ mod tests {
...
+            .setattr(&client, ino, Some(atime), Some(mtime))
+            .await
+            .expect("setattr should be successful");
+        let stat = lookup.stat;
+        assert_eq!(stat.atime, atime);
+        assert_eq!(stat.mtime, mtime);
+
+        // getattr() validates that InodeStat is not expired as the Inode is still Local and it would have not been able to lookup stat remotely.
```
passaro, line 2541: This is a bit convoluted: why not just test `stat.is_valid()`?

dannycjones (dismissed):

mountpoint-s3/src/inode.rs
```diff
@@ -2522,11 +2527,25 @@ mod tests {
         let stat = inode.get_inode_state().unwrap().stat.clone();
         assert!(!stat.is_valid());
 
-        // Should get an error back when calling setattr
+        // Should be able to reset expiry back and make stat valid when calling setattr
```
passaro, line 2530: Let's review this test, when/if we change the way the stats are updated in `update_slow_path`.

passaro (changes requested):

dannycjones (commented):
Let's address concerns raised by Alex too

mountpoint-s3/src/inode.rs
sauraank: Should I also add a test for lookup on local inode with Invalid InodeStat?

dannycjones (commented):
We should update the release notes in this PR too as we're fixing #577.

mountpoint-s3/src/inode.rs
```diff
@@ -827,9 +835,18 @@ impl SuperblockInner {
...
+                        InodeKind::File => self.cache_config.file_ttl,
+                        InodeKind::Directory => self.cache_config.dir_ttl,
+                    };
+                    sync.stat.update_validity(validity);
+                    let stat = sync.stat.clone();
+                    drop(sync);
+
                     Ok(LookedUp {
-                        inode: existing_inode,
+                        inode: existing_inode.clone(),
```
passaro, line 849: We should not need to clone here:
```suggestion
                        inode: existing_inode,
```

passaro (changes requested):

passaro (approved):
LGTM

PR #587, opened 2023-10-30
Add documentation for object metadata and data caching

dannycjones:
## Description of change

Draft of GitHub documentation changes for the introduction of metadata and data caching, tracked in #255.

Relevant issues: #255 

Live documentation changes, but no change to behavior. This should only be merged once caching features are part of main releases.

---


doc/CONFIGURATION.md
```diff
@@ -6,6 +6,7 @@ In most scenarios, you can use Mountpoint by running the following command, wher
 
 We've tried hard to make this simple command adopt good defaults for most scenarios. However, some scenarios may need additional configuration. This document shows how to configure these elements of Mountpoint:
 * [AWS credentials](#aws-credentials)
+* [Caching configuration](#caching-configuration)
```
jamesbornholt, line 9: Should be below bucket configuration, and probably below file system configuration too. Also needs a sentence description (caching what?).

doc/CONFIGURATION.md
```diff
@@ -80,6 +81,23 @@ Here is an example least-privilege policy document to add to an IAM user or role
 
 Mountpoint also respects access control lists (ACLs) applied to objects in your S3 bucket, but does not allow you to automatically attach ACLs to objects created with Mountpoint. A majority of modern use cases in Amazon S3 no longer require the use of ACLs. We recommend that you keep ACLs disabled for your S3 bucket, and instead use bucket policies to control access to your objects.
 
+## Caching configuration
+
+Mountpoint now offers caching of metadata and object content allowing for reduced requests when reading files.
```
jamesbornholt, line 86: Documentation is independent of time, so no "now".

```suggestion
Mountpoint can optionally cache object metadata and content to reduce cost and improve performance for repeated reads to the same file.
```

doc/CONFIGURATION.md
```diff
@@ -80,6 +81,23 @@ Here is an example least-privilege policy document to add to an IAM user or role
...
+This is particularly useful when reading the same files many times for the same Mountpoint filesystem.
+
+To enable caching, use the `--caching` command-line flag.
+This alone will enable caching of metadata using a default time-to-live (TTL) of 60 minutes.
+Caching of object/file content on disk can also be enabled
+by providing a caching location using the `--data-cache-directory <DIR>` command-line flag.
+
+Enabling caching relaxes the strong read-after-write consistency offered by Mountpoint with default configuration.
+Please read more in the [consistency and concurrency section of the semantics documentaton](./SEMANTICS.md#consistency-and-concurrency).
```
jamesbornholt, line 95:
```suggestion
Caching relaxes the strong read-after-write consistency offered by Amazon S3 and Mountpoint in its default configuration. See the [consistency and concurrency section of the semantics documentaton](./SEMANTICS.md#consistency-and-concurrency) for more details.
```

doc/CONFIGURATION.md
```diff
@@ -80,6 +81,23 @@ Here is an example least-privilege policy document to add to an IAM user or role
...
 
+## Caching configuration
+
+Mountpoint now offers caching of metadata and object content allowing for reduced requests when reading files.
+This is particularly useful when reading the same files many times for the same Mountpoint filesystem.
+
+To enable caching, use the `--caching` command-line flag.
+This alone will enable caching of metadata using a default time-to-live (TTL) of 60 minutes.
+Caching of object/file content on disk can also be enabled
+by providing a caching location using the `--data-cache-directory <DIR>` command-line flag.
```
jamesbornholt, line 92: Naming bikeshedding: `--cache-metadata`/`--cache-data <DIR>`?

And then mention that `--cache-data` implies `--cache-metadata`.

doc/CONFIGURATION.md
```diff
@@ -80,6 +81,23 @@ Here is an example least-privilege policy document to add to an IAM user or role
...
+
+Enabling caching relaxes the strong read-after-write consistency offered by Mountpoint with default configuration.
+Please read more in the [consistency and concurrency section of the semantics documentaton](./SEMANTICS.md#consistency-and-concurrency).
+
+Mountpoint caching can be further configured,
+such as adjusting the metadata time-to-live (TTL) or the maximum space allowed to be used by the data cache.
+Review the caching options available using `mount-s3 --help`.
```
jamesbornholt, line 99: Move this up to the paragraph about the command-line flags.

doc/CONFIGURATION.md
jamesbornholt: Also, maybe this should use the fancy `> [!WARNING]` thing.

doc/SEMANTICS.md
```diff
@@ -63,9 +63,14 @@ Mountpoint has limited support for other file and directory metadata, including
...
 
-However, Mountpoint may return stale metadata for an existing object within 1 second of the object being modified or deleted in your S3 bucket by another client. This occurs only if the object was accessed through Mountpoint immediately before being modified or deleted in your S3 bucket. The stale metadata will only be visible through metadata operations such as `stat` on individual files. Directory listings will never be stale and always reflect the current metadata. These cases do not apply to newly created objects, which are always immediately visible through Mountpoint. Stale metadata can be refreshed by either opening the file or listing its parent directory.
+Mountpoint also offers metadata and object content caching which can be enabled using CLI flags: see the [caching section of the configuration document](./CONFIGURATION.md#caching) for more information.
```
jamesbornholt, line 69: * Put this after the stale metadata paragraph.
* Call it optional
* "configuration documentation"

doc/SEMANTICS.md
```diff
@@ -63,9 +63,14 @@ Mountpoint has limited support for other file and directory metadata, including
...
+When opting into caching, the consistency model is relaxed and you may see stale entries until they have expired.
+Stale entries may live as long as the cache's metadata time-to-live (TTL).
```
jamesbornholt, line 71: Is there an escape hatch to guarantee consistency? (`O_DIRECT` for data, anything for metadata?)

doc/CONFIGURATION.md
dannycjones: makes sense.

I was planning to put "Since v1.2.0, Mountpoint can optionally cache...", but replaced it with "now" as a placeholder. Exclude that?

doc/CONFIGURATION.md
passaro: If we are starting with a single flag, would it make sense to have something like `--cache <DIR>`, which enables both metadata and data cache using `<DIR>` for the latter?

We can always add more specific flags like `--cache-metadata` and `--cache-data <DIR>` if we want to give users more detailed options, without breaking changes or ambiguities.

dannycjones: I updated to use `--cache <DIR>`. We should in a separate PR make an update to the final CLI args which also adopts this (or makes whatever change we decide to make).

doc/SEMANTICS.md
jamesbornholt: `O_DIRECT` should definitely bypass the cached data. I think it's an open question whether it should _also_ bypass cached metadata and force an ETag lookup. Strictly speaking, that's not what `O_DIRECT` is documented to do, but that's because in a real file system `lookup` is supposed to do all the work of validating metadata. But here, if we're bypassing the cached data we also need a way to find out what the latest ETag is, so forcing a re-lookup is probably the right thing to do?

jamesbornholt:
LGTM, but obviously shouldn't be merged until we actually turn this on.

doc/CONFIGURATION.md
```diff
@@ -227,6 +228,27 @@ ExecStop=/usr/bin/fusermount -u /home/ec2-user/s3-bucket-mount
...
 ```
 
+## Caching configuration
+
+Mountpoint can optionally cache object metadata and content to reduce cost and improve performance for repeated reads to the same file.
+
+To enable caching, use the `--cache <CACHE_DIR>` command-line flag.
+This flag will enable caching of metadata using a default time-to-live (TTL) of 60 minutes.
+Object/file content will also be cached within the cache directory specified.
+Mountpoint will create a directory within this path, removing any files or directories within it at mount time and exit.
```
jamesbornholt, line 238: Don't think we should be deleting at exit time — that's something we can't change our mind about if we want to default to reusable cache later.

jamesbornholt (commented):
This PR seems to have picked up some CRT module updates that I don't think you intended :-)

dannycjones:
Fixed the CRT issues (my bad for not checking diff) on GH. I still need to think about options on deleting cache and `O_DIRECT` flag too.

doc/CONFIGURATION.md
dannycjones: I'm not sure what to pick here. Two paths forward.

- We don't do cleanup at mount or exit.
  - There is a risk that we fill up the disk with stale cache content that we don't track, since we don't attempt to ingest old cache data. We eventually get to a point where we have very little space remaining for the real cache data, unless the customer takes manual action.
  - We also provide a `--delete-cache` option which will delete the cache at mount and exit.
  - Note, we haven't defined our recommendations to the user. We haven't tested disk reuse, and I assume we should discourage it here in the documentation until we are happy that cache reuse is both well tested and supported.
- We do cleanup at mount and exit.
  - We mitigate the risk from the disk filling up.
  - We don't need to guide the customer on cache reuse, since its not possible to use it.
  - Allowing cache reuse later would potentially be a breaking change since we'd change the disk usage characteristics, and would likely need to be opt-in through some `--allow-cache-reuse`/`--no-delete-cache` flag.

I do think I actually prefer the former option, but we need to guide users in the right direction to ensure they aren't reusing the cache.

dannycjones: My other thinking is around how we should cleanup the cache.

At the moment, we are planning to delete the `mp-cache-V1` directory. I think this is fine, but in this case I think we also need to guide users to using a unique value for `--cache <CACHE_DIR>` for different buckets so that the caches don't trample over each other. The other option is that we include a hash of the bucket name into the path (which I think we should do anyway) and _delete only that sub-directory, leaving any other buckets intact_. I don't see much issue with this either actually.

doc/SEMANTICS.md
dannycjones: I'd challenge if we should bypass cached data. `O_DIRECT` is a hint to the Kernel to avoid the page cache and read directly from storage. As I understand it, it can be ignored and even if we ignore it, users would still benefit from bypassing the page cache I assume.

I would assume that we want to refresh the ETag, so it should trigger a lookup that will refresh that. If it hasn't changed, then we know that the data on our disk cache is already up to date with S3 at the time of opening. I don't see benefit to the user to bypass that cache?

jamesbornholt: `O_DIRECT` isn't specific to the page cache, the hint is "Try to minimize cache effects of the I/O to and from this file". In practice people use it to mean "hit the storage device directly". It's true that this is a case where we can detect whether the cache is valid or not, but we should probably just do what people expect even if it's slower.

doc/CONFIGURATION.md
dannycjones: Discussed on a call with @jamesbornholt and @passaro, documenting here:

- We'll implement the cache directory argument as a directory we are creating a sub-directory within, and deleting the contents at mount and at exit. This avoids issues where requiring a non-empty directory causes availability issues on unexpected process restarts, or with not deleting and disk space accumulating and not being cleaned.
- Configuration documentation will recommend to use a unique value for this argument when running multiple processes, to avoid any clobbering of each process' cache. If Mountpoint is used with the same path, it may cause performance issues when the cache is clobbered but we also expect mounting and un-mounting on a process to be relatively infrequent and the impact here should be limited.
- If we were to consider cache re-use in the future, its unlikely we can silently change behavior to allow re-use and we'd probably need a new argument for that anyway.

doc/SEMANTICS.md
dannycjones: I've updated this (although implementation is yet to be completed!).

doc/CONFIGURATION.md
```diff
@@ -227,6 +228,32 @@ ExecStop=/usr/bin/fusermount -u /home/ec2-user/s3-bucket-mount
...
+To enable caching, use the `--cache <CACHE_DIR>` command-line flag.
+This flag will enable caching of metadata using a default time-to-live (TTL) of 1 second.
+Object content will also be cached within the cache directory specified.
+Mountpoint will create a directory within this path, removing any files or directories within it at mount time and exit.
+When running multiple Mountpoint processes on the same host,
+you should use unique cache directories to avoid different processes interfering with the others' cache content.
+
+Mountpoint caching can be further configured,
+such as adjusting the metadata time-to-live (TTL) or the maximum space allowed to be used by the data cache.
+Review the caching options available using `mount-s3 --help`.
```
dannycjones, line 244: It may be worth calling out the metadata TTL specifically here also, since we expect this to be something customers may want to tune.

dannycjones:
I've updated the documentation here to cover new proposed default (1 second), and also describe an example of the implications of stale metadata when opening a file for reading.

doc/CONFIGURATION.md
```diff
@@ -227,6 +228,31 @@ ExecStop=/usr/bin/fusermount -u /home/ec2-user/s3-bucket-mount
 WantedBy=remote-fs.target
 ```
 
+## Caching configuration
+
+Mountpoint can optionally cache object metadata and content to reduce cost and improve performance for repeated reads to the same file.
+
+To enable caching, use the `--cache <CACHE_DIR>` command-line flag.
```
jamesbornholt, line 235:
```suggestion
To enable caching, use the `--cache <CACHE_DIR>` command-line flag, specifying the directory in which to store cached object content.
```

doc/CONFIGURATION.md
```diff
@@ -227,6 +228,31 @@ ExecStop=/usr/bin/fusermount -u /home/ec2-user/s3-bucket-mount
...
+This flag will enable caching of metadata using a default time-to-live (TTL) of 1 second.
+Object content will also be cached within the cache directory specified.
```
jamesbornholt, line 237:
```suggestion
This flag will also enable caching of metadata using a default time-to-live (TTL) of 1 second.
```

doc/CONFIGURATION.md
```diff
@@ -227,6 +228,31 @@ ExecStop=/usr/bin/fusermount -u /home/ec2-user/s3-bucket-mount
...
+Mountpoint will create a directory within this path, removing any files or directories within it at mount time and exit.
```
jamesbornholt, line 238:
```suggestion
Mountpoint will create a new subdirectory within the path that you specify, and will remove any existing files or directories within that subdirectory at mount time and at exit.
```

doc/CONFIGURATION.md
```diff
@@ -227,6 +228,31 @@ ExecStop=/usr/bin/fusermount -u /home/ec2-user/s3-bucket-mount
...
+When available space for the cache on disk is limited,
+content of files least recently accessed will be evicted to allow new data to be cached.
```
jamesbornholt, line 240:
```suggestion
By default, Mountpoint will limit the maximum size of the cache such that the free space on the file system does not fall below 5%, and will automatically evict the least recently used content from the cache when caching new content. You can instead manually configure the maximum size of the cache with the `--max-cache-size <MiB>` command-line argument.
```

doc/CONFIGURATION.md
```diff
@@ -227,6 +228,31 @@ ExecStop=/usr/bin/fusermount -u /home/ec2-user/s3-bucket-mount
...
+When running multiple Mountpoint processes on the same host,
+you should use unique cache directories to avoid different processes interfering with the others' cache content.
+
+Mountpoint caching can be further configured,
+such as adjusting the metadata time-to-live (TTL) or the maximum space allowed to be used by the data cache.
+Review the caching options available using `mount-s3 --help`.
```
jamesbornholt, line 246: Don't point to `--help` in documentation, this _is_ the documentation :-) Just describe the options you can configure.

doc/CONFIGURATION.md
```diff
@@ -227,6 +228,55 @@ ExecStop=/usr/bin/fusermount -u /home/ec2-user/s3-bucket-mount
...
+To create a RAM disk on Linux, you can use [tmpfs](https://www.kernel.org/doc/html/latest/filesystems/tmpfs.html) as shown below.
+Replace the mount directory for tmpfs if required.
+
+```
+sudo mkdir /mnt/mp-cache-tmpfs
+sudo mount -o uid=$(id --user),mode=700 -t tmpfs none /mnt/mp-cache-tmpfs
+```
+
+This will create a RAM disk mounted at `/mnt/mp-cache-tmpfs` with access restricted to the current user.
+By default, Linux limits it's size to 50% of the physical memory available on the system.
```
passaro, line 271:
```suggestion
By default, Linux limits its size to 50% of the physical memory available on the system.
```

doc/CONFIGURATION.md
```diff
@@ -227,6 +228,55 @@ ExecStop=/usr/bin/fusermount -u /home/ec2-user/s3-bucket-mount
...
+The size is configurable using the `size` option.
+
+You can then start Mountpoint using the directory where the RAM disk was mounted.
+
+```
+mount-s3 DOC-EXAMPLE-BUCKET /path/to/mount --cache /mnt/tmpfs
```
passaro, line 277:
```suggestion
mount-s3 DOC-EXAMPLE-BUCKET /path/to/mount --cache /mnt/mp-cache-tmpfs
```

passaro (changes requested):
Just a couple of minor comments

passaro (approved):

PR #590, opened 2023-10-31
Add caching ObjectStore implementation

passaro:
## Description of change

EDIT: This is now the follow-up to #592 which introduces a caching store implementation.

Still a **draft** PR to preview the use of the cache:
* uses the `InMemoryCache`, which will be eventually replaced by a disk implementation,
* CLI args are only temporary for initial testing.

Relevant issues: #255

No. All changes hidden under `caching` flag.

---


mountpoint-s3/src/fs.rs
```diff
@@ -49,37 +47,45 @@ impl DirHandle {
 }
 
 #[derive(Debug)]
-struct FileHandle<Client: ObjectClient, Runtime> {
```
jamesbornholt, line 52: Is making this a type parameter going to be annoying when it comes to switching caching on/off? I was imagining that maybe we're choose a different `ObjectStore` implementation based on whether the cache is enabled or not, but maybe that's not what you had in mind.

edit: oh, I see now, you statically switch on it with `create_filesystem`. That's probably fine, although note that we'll end up with two copies of a bunch of code in the final binary (everything with a `Store` generic).

mountpoint-s3/src/main.rs
```diff
@@ -238,6 +240,25 @@ struct CliArgs {
...
+    #[cfg(feature = "caching")]
+    #[clap(
+        long,
+        help = "Block size for the data cache",
+        default_value = "1048576",
+        value_parser = value_parser!(u64).range(1..),
+        help_heading = CACHING_OPTIONS_HEADER,
+        requires = "enable_data_caching",
+    )]
+    pub data_cache_block_size: u64,
```
jamesbornholt, line 260: Is this temporary? It doesn't seem like a knob we'd want to publicly expose.

mountpoint-s3/src/prefetch.rs
```diff
@@ -289,80 +274,53 @@ where
...
         self.next_sequential_read_offset = offset;
+        self.next_request_size = self.inner.config.first_request_size;
         self.next_request_offset = offset;
     }
 
     /// Try to seek within the current inflight requests without restarting them. Returns true if
     /// the seek succeeded, in which case self.next_sequential_read_offset will be updated to the
     /// new offset. If this returns false, the prefetcher is in an unknown state and must be reset.
-    async fn try_seek(&mut self, offset: u64) -> Result<bool, PrefetchReadError<TaskError<Client>>> {
+    async fn try_seek(&mut self, offset: u64) -> Result<bool, ObjectClientError<PrefetchReadError, Feed::ClientError>> {
```
jamesbornholt, line 323: This error type seems inverted — seeking should return a prefetcher-level error, one of the causes of which (the only cause of which?) is a client error.

mountpoint-s3/src/prefetch/feed.rs
```diff
@@ -1,128 +1,266 @@
...
+use crate::prefetch::part_queue::unbounded_part_queue;
+use crate::prefetch::task::RequestTask;
+use crate::store::PrefetchReadError;
 
 /// A generic interface to retrieve data from objects in a S3-like store.
-#[async_trait]
-pub trait ObjectPartFeed<Client: ObjectClient> {
-    /// Get the content of an object in fixed size parts. The parts are pushed to the provided `part_sink`
-    /// and are guaranteed to be contiguous and in the correct order. Callers need to specify a preferred
+pub trait ObjectPartFeed {
```
jamesbornholt, line 17: bikeshedding on naming: `ObjectPartStream`?

mountpoint-s3/src/store.rs
```diff
@@ -0,0 +1,262 @@
...
+use crate::{
+    prefetch::{self, feed::ObjectPartFeed, Prefetcher},
+    sync::Arc,
+};
+
+/// A generic interface to S3-like object store.
+/// Similar to [ObjectClient], but provides a [ObjectStore::prefetch] method instead
+/// of [ObjectClient::get_object].
+#[async_trait]
+pub trait ObjectStore: Clone {
```
jamesbornholt, line 36: Would it make more sense to do `pub trait ObjectStore: ObjectClient` rather than duplicating all the methods?

jamesbornholt (commented):
Just a quick high-level pass, will try to get more detailed tomorrow. In particular, I haven't read `cached_feed.rs` yet.

mountpoint-s3/src/prefetch.rs
```diff
@@ -535,6 +459,77 @@ mod tests {
...
+                get_fn: Box::new(move |bucket: &str, key: &str, size: u64, etag: ETag| {
+                    Box::new(prefetcher.get(bucket, key, size, etag))
+                }),
+            }
+        }
+
+        fn get(&self, bucket: &str, key: &str, size: u64, etag: ETag) -> Box<dyn PrefetchGetObject<ClientError = E>> {
+            (self.get_fn)(bucket, key, size, etag)
+        }
+    }
```
dannycjones, line 498: not sure what the purpose of this is, it could use some rustdoc

mountpoint-s3/src/prefetch/feed.rs
dannycjones: now that we moved the part queue channels to be internal, i think stream makes sense.

mountpoint-s3/src/prefetch.rs
passaro: If I understand correctly, you mean `try_seek` (and the `read` method) should return an error about like this:

```rust
enum PrefetchReadError<ClientError> {
  GetRequestFailed(ObjectClientError<GetObjectError,ClientError>),
  ..
  Integrity(IntegrityError),
}
```
which is not too far from what we had before, where `type TaskError<Client> = ObjectClientError<GetObjectError, <Client as ObjectClient>::ClientError>`.

My thinking, though, is that using `ObjectClientError` here with `PrefetchReadError` as the "`ServiceError`" fits nicely with the other `ObjectStore`/`ObjectClient` methods. Compare for example:

```rust
ObjectStore::head_object(..) -> ObjectClientResult<HeadObjectResult, HeadObjectError, Self::ClientError>
PrefetchGetObject::read(..) -> ObjectClientResult<ChecksummedBytes, PrefetchReadError, Self::ClientError>
```
It also means we don't need `PrefetchReadError` to be generic at all, but just wrap a `GetObjectError`.

mountpoint-s3/src/store.rs
passaro: The downside I see is that if `ObjectStore` extends `ObjectClient`, it will expose `get_object` as well: fine for the default implementation, but what about the "caching" implementation? Will it have to implement caching for it? Or just accept that `get_object` and `prefetch` may behave differently?

mountpoint-s3/src/prefetch.rs
passaro: You're right. This is a wrapper I quickly built to dynamically switch between prefetchers with and without cache in the tests. I'll either rethink the tests or document it.

passaro:
Replaced by #598

PR #592, opened 2023-11-01
Introduce ObjectStore trait to replace ObjectClient in mountpoint-s3

passaro:
## Description of change

Introduce a new `ObjectStore` trait that replaces `ObjectClient` in the mountpoint-s3 crate. In addition to most of `ObjectClient` methods, `ObjectStore` also declares a new `prefetch` method returning a `PrefetchGetObject` which allows callers to read the object content. `PrefetchGetObject` is where `ObjectStore` implementations can add object data caching.

This change also reworks the `Prefetcher` so that `ObjectStore` implementations can delegate `prefetch` to it. The main changes to `Prefetcher` are:
* it is now generic on the `ObjectPartStream` (previously `ObjectPartFeed`), rather than using dynamic dispatch.
* the logic to spawn a new task for each `GetObject` request and handle the object body parts returned was moved into `ObjectPartStream`.

Relevant issues: #255 

No changes in behavior.

---


passaro:
Note that this PR supersedes the first part of #590, which I'll close shortly.

mountpoint-s3/src/prefetch.rs
```diff
@@ -143,37 +121,21 @@ pub struct PrefetchGetObject<Client: ObjectClient, Runtime> {
...
     /// Read some bytes from the object. This function will always return exactly `size` bytes,
     /// except at the end of the object where it will return however many bytes are left (including
     /// possibly 0 bytes).
-    pub async fn read(
+    async fn read(
         &mut self,
         offset: u64,
         length: usize,
-    ) -> Result<ChecksummedBytes, PrefetchReadError<TaskError<Client>>> {
+    ) -> ObjectClientResult<ChecksummedBytes, PrefetchReadError, Self::ClientError> {
```
passaro, line 138: See discussion about error type here: https://github.com/awslabs/mountpoint-s3/pull/590#discussion_r1378802481

mountpoint-s3/src/store.rs
```diff
@@ -0,0 +1,235 @@
...
+
+use crate::checksums::{ChecksummedBytes, IntegrityError};
+use crate::prefetch::{self, ClientPartStream, ObjectPartStream, Prefetcher, PrefetcherConfig};
+use crate::sync::Arc;
+
+/// A generic interface to S3-like object store.
+/// Similar to [ObjectClient], but provides a [ObjectStore::prefetch] method instead
+/// of [ObjectClient::get_object].
+#[async_trait]
+pub trait ObjectStore: Clone {
```
passaro, line 24: See discussion on extending `ObjectClient`: https://github.com/awslabs/mountpoint-s3/pull/590#discussion_r1378813014

jamesbornholt: I think extending `ObjectClient`, or otherwise exposing the `ObjectClient` trait directly, is still the right thing to do here. 

Actually, thinking more broadly: right now this is a pretty heavyweight abstraction for what's essentially one method (`prefetch`) with two implementations (cached or not). Seems like we could have a narrower interface that just focuses on "where object bytes come from". For example, I don't see how we'd use this abstraction to handle cached writes in the future.

Basically: can we just start with a smaller abstraction and expand it once we need it?

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -42,7 +44,7 @@ impl<E: std::error::Error + Send + Sync> PartQueue<E> {
     /// empty.
     ///
     /// If this method returns an Err, the PartQueue must never be accessed again.
-    pub async fn read(&self, length: usize) -> Result<Part, PrefetchReadError<E>> {
+    pub async fn read(&self, length: usize) -> ObjectClientResult<Part, PrefetchReadError, E> {
```
jamesbornholt, line 47: Still think this error type is inverted; it expands to
```
Result<Part, ObjectClientError<PrefetchReadError, E>>
```
which seems wrong: `ObjectClientError` is about errors in the `ObjectClient`, but the prefetcher lives higher in the stack than that.

mountpoint-s3/src/prefetch/part_stream.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    pub fn start(&self) -> u64 {
+        self.offset
+    }
+
+    pub fn end(&self) -> u64 {
+        self.offset + self.size as u64
+    }
+
+    pub fn trim_start(&self, start_offset: u64) -> Self {
```
jamesbornholt, line 72: Is it intentional that `start_offset > self.end()` is valid and just silently becomes a zero-length range? (Similar thing for `trim_end`).

mountpoint-s3/src/store.rs
passaro: An alternative approach could be to introduce a `Prefetcher` trait, roughly something like:

```rust
trait Prefetcher {
  type PrefetchGetObject: PrefetchGetObject<ClientError = Self::ClientError>;
  fn prefetch(&self, bucket: &str, key: &str, size: u64, etag: ETag) -> Self::PrefetchGetObject;
}
```

Comparing with `ObjectStore`:

Pros:
* focus on "where object bytes come from"
* no impact on write path, readdir, etc.

Cons:
* the prefetcher type would still need to be propagated together with the client in many places, e.g.:
  `struct S3Filesystem<Client, Prefetcher>`.
* If we wanted to introduce cached writes in the future, we would have to carry even more types, e.g.: `struct S3Filesystem<Client, Prefetcher, Uploader>`.
* building a filesystem may become a bit complex and error prone, e.g.:
  ```rust
  let client = Arc::new(client); // Annoyingly, ObjectClient is not Clone
  let prefetcher = SomePrefetcher::new(client.clone());
  let filesystem = S3Filesystem::new(client, ...);
  ```
  Although maybe solvable it we had `Prefetcher::prefetch` take `Client` as an argument (or rather `Arc<Client>`). 

On balance, I would still lean towards `ObjectStore`.

passaro: Another idea could be something like:

```rust
trait ObjectStore {
  ...
  type Client: ObjectClient + ...
  
  fn client(&self) -> &ObjectClient;

  fn prefetch(...) -> ...

  // in the future 
  fn write(...)
}
```

Not too sure if it would qualify as a smaller abstraction, but it would make it clear what is the relationship with `ObjectClient`.

mountpoint-s3/src/prefetch/part_stream.rs
passaro: Yes. Rustdoc added to clarify.

mountpoint-s3/src/prefetch/part_queue.rs
passaro: You're right. Inverted now.

mountpoint-s3/src/store.rs
jamesbornholt: I don't think "too many type parameters" is much of a concern — it's annoying to get right one time, and then shouldn't really matter. I like the idea of having small, separate abstractions for this until we know for sure that we need a shared one (for e.g. reads and writes).

passaro: I'll start reworking with something like the `Prefetcher` above. When I'm done, shall I update this PR, or open a new one?

passaro: Opened #595

passaro:
Closing and replacing with #595

PR #593, opened 2023-11-01
Implement disk-based DataCache with no eviction

dannycjones:
## Description of change

Initial version of the data cache on disk.

We now write data to disk including some metadata to check we haven't mixed up blocks.

Related: #255

No. This code is not yet adopted in main releases.

---


mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,312 @@
...
+
+        // An S3 key may be up to 1024 UTF-8 bytes long, which exceeds the maximum UNIX file name length.
+        // Instead, we encode the key and split into 255 character long directory names.
+        let encoded_s3_key = Base64Url::encode_string(cache_key.s3_key.as_bytes());
+        let mut slice = encoded_s3_key.as_str();
+        while !slice.is_empty() {
+            let (chunk, remaining) = slice.split_at(255.min(slice.len()));
+            path.push(chunk);
+            slice = remaining;
+        }
```
dannycjones, line 48: I am reconsidering if we actually need the base 64 encoding. It does obfuscate the path a bit, but I'm not convinced we actually have any other reason at this point.

Instead, could we just put the blocks in directories based on the actually file path they represent? (Of course, that makes this a cache suited towards files with etag rather than objects with an etag.)

mountpoint-s3/src/checksums.rs
```diff
@@ -151,6 +151,16 @@ impl ChecksummedBytes {
         }
         Ok(())
     }
+
+    /// Provide the underlying bytes and the associated checksum,
+    /// which may be recalulated if the checksum covers a larger slice than the current slice.
```
passaro, line 156: typo:
```suggestion
    /// which may be recalculated if the checksum covers a larger slice than the current slice.
```

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,312 @@
...
+    pub fn new(cache_directory: PathBuf, block_size: u64) -> Self {
+        DiskDataCache {
+            block_size,
+            cache_directory,
+        }
+    }
+
+    /// Get the relative path for the given block.
+    fn get_path_for_key(&self, cache_key: &CacheKey) -> PathBuf {
+        let mut path = self.cache_directory.join("v1");
```
passaro, line 38: Besides/instead of the version (which is not really needed for a temporary cache), should we include the process id, or something equivalent, to ensure the path is unique?

Maybe for this type it could just be a token to be passed at construction.

mountpoint-s3/src/data_cache/disk_data_cache.rs
passaro: I like the idea, but it may be a bit cumbersome to pass the file path here (or the mounted prefix?). Happy to leave the base64 encoding for now.

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,312 @@
...
+
+        // An S3 key may be up to 1024 UTF-8 bytes long, which exceeds the maximum UNIX file name length.
+        // Instead, we encode the key and split into 255 character long directory names.
+        let encoded_s3_key = Base64Url::encode_string(cache_key.s3_key.as_bytes());
```
passaro, line 42: Or [`Base64UrlUnpadded`](https://docs.rs/base64ct/latest/base64ct/struct.Base64UrlUnpadded.html)?

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,312 @@
...
+                trace!(
+                    ?path_for_cache_key,
+                    "no directory for cache key, assuming no block cached yet"
+                );
+                return Ok(Vec::new());
+            }
+            Err(e) => return Err(e.into()),
+        };
+
+        let mut indicies = Vec::new();
```
passaro, line 116:
```suggestion
        let mut indices = Vec::new();
```

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,312 @@
...
+                };
+            } else {
+                warn!(
+                    path=?path_for_cache_key.join(file_name.as_ref()),
+                    "unexpected file found in cache without \".block\" suffix"
+                );
+            }
+        }
+
+        Ok(indicies)
```
passaro, line 146: don't we need to sort?

passaro (commented):
Few comments from a quick pass.

mountpoint-s3/src/data_cache/disk_data_cache.rs
dannycjones: I'll push an update to this PR now, but not address this today. Let me think about it.

I left in the versioning while I'm not super happy about a) not purging the cache on mount and b) not having things like etag, s3key in the blocks themselves on disk.

mountpoint-s3/src/data_cache/disk_data_cache.rs
jamesbornholt: Obfuscation is important here — we don't want customers looking inside this directory instead of going through Mountpoint. But it seems like we could just take a hash instead. It's also often a good idea to split the directory into subdirectories (based on the first few characters of the hash) to avoid hitting any filesystem-specific maximum number of directory entries.

dannycjones:
Just pushed a few changes based on some feedback, but still need to address tomorrow:

- What to do about any versioning in path (`v1` or PID or something else)
- Look into comments on if we need base64 encoding, and to consider splitting on the first part of the directory
- Identify path forward with macOS since it has a full path limit of 1024 (or maybe smaller), which is why the macOS unit tests are failing

mountpoint-s3/src/data_cache/disk_data_cache.rs
dannycjones: updated to use unpadded, we don't care about the lost length info since we are not concatenating multiple base64 values together

mountpoint-s3/src/data_cache/disk_data_cache.rs
jamesbornholt: Versioning here seems good. We definitely need some safety around (a) not using a non-empty cache directory where we might overwrite existing files, and (b) coordinating enough to avoid multiple Mountpoints trying to use the same directory. But happy for that to come in a followup.

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,315 @@
...
+        let encoded_s3_key = Base64UrlUnpadded::encode_string(cache_key.s3_key.as_bytes());
+        let mut slice = encoded_s3_key.as_str();
+        while !slice.is_empty() {
+            let (chunk, remaining) = slice.split_at(255.min(slice.len()));
+            path.push(chunk);
+            slice = remaining;
+        }
+
+        path.push(cache_key.etag.as_str());
+        path
```
jamesbornholt, line 51: As mentioned elsewhere: think this should just be a SHA256 hash of the object key, etag, and block index. It gets us out of all the concerns about filename size and valid/invalid characters, and obfuscates the cache so customers don't take a dependency on its layout (without a lot of effort). There's always a collision risk with hashes but these are big enough that I wouldn't worry about it, and since you're eventually going to include the key and etag in the file contents we can always disambiguate.

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    fn put_block(&self, cache_key: CacheKey, block_idx: BlockIndex, bytes: ChecksummedBytes) -> DataCacheResult<()> {
+        let path = self.get_path_for_block(&cache_key, block_idx);
+        trace!(?cache_key, ?path, "new block will be created in data cache");
+        fs::create_dir_all(path.parent().expect("path should include cache key in directory name"))?;
+        let mut file = fs::File::create(path)?;
+        let (bytes, _checksum) = bytes.into_inner().map_err(|_| DataCacheError::InvalidBlockContent)?;
+        // TODO: Store checksum
+        file.write_all(&bytes)?;
+        file.sync_data()?;
```
jamesbornholt, line 91: not necessary (once checksums are in, anyway)

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    fn put_block(&self, cache_key: CacheKey, block_idx: BlockIndex, bytes: ChecksummedBytes) -> DataCacheResult<()> {
+        let path = self.get_path_for_block(&cache_key, block_idx);
+        trace!(?cache_key, ?path, "new block will be created in data cache");
+        fs::create_dir_all(path.parent().expect("path should include cache key in directory name"))?;
+        let mut file = fs::File::create(path)?;
+        let (bytes, _checksum) = bytes.into_inner().map_err(|_| DataCacheError::InvalidBlockContent)?;
+        // TODO: Store checksum
+        file.write_all(&bytes)?;
```
jamesbornholt, line 90: Let's make sure to build a correctly checksum-bracketed on-disk layout here. We need to include the key, etag, block index, and data checksum, and then compute a checksum of all that metadata. We should also encode the serialization version somewhere (e.g. as a magic number at the start of the file).

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,315 @@
...
+        file.sync_data()?;
+        Ok(())
+    }
+
+    fn block_size(&self) -> u64 {
+        self.block_size
+    }
+
+    fn cached_block_indices<R: RangeBounds<BlockIndex>>(
```
jamesbornholt, line 99: It's surprising to me that you need this, versus just lazily querying the cache for each block you need.

dannycjones: I'm not super fond of this. The reasoning behind it though is to try and gauge what we can read from disk, and what we need to be creating S3 GetObject for. We'd prefer to avoid doing 1MiB GetObject so determining what blocks are missing up front is useful so we can coalesce missing blocks into up to 8MiB GetObject requests.

@passaro lets consider what other options we have here, not blocking this PR though

mountpoint-s3/src/data_cache/disk_data_cache.rs
dannycjones: Sounds good! I've been putting something together for the on-disk layout, adding a checksum does sound like something we need. I'll add it to the TODO for that next PR.

mountpoint-s3/src/data_cache/disk_data_cache.rs
dannycjones: I had previously preferred to keep the ETag and Block Index exposed.

My reasoning here was that:
- We can wipe all the blocks for a whole file version relatively easily.
- When we know that an object has been replaced, we could purge all cache versions by wiping all blocks that don't match the new ETag.

However, we still plan to use LRU cache eviction and those versions would eventually be purged. The bit that makes me uncomfortable is that we would be allowing the disk to be consumed unnecessarily in those use cases by waiting for them to be evicted when we run out of capacity. Possible alternatives might be that we have a TTL on the data cache entries also.

dannycjones: I am coming around to the idea of using the SHA256 over Base64. Agree that it should be very rare, and if we do get a collision then we will check the S3 key encoded in the block and reject blocks that don't match. Small concern over thrashing with blocks being written over each other, but again: we think this is rare.

If we do move to SHA256 (or similar) then I'll add some metrics for collisions.

jamesbornholt: What scenarios would we need to evict more than one old version of an object from the cache? My thinking is that at any point in time there's at most one cached version of an object — when we discover the remote etag has changed, we just evict whatever the current version is (whose etag we know since it's already in the inode).

Even if it's more complicated than that, I don't think it's worth the trouble to try to accelerate eviction right now; seems fine to just fall back on those objects eventually aging out of the LRU.

dannycjones:
Planning to update this PR as follows to have it in a state we're happy to merge and iterate later on:

- Pull in the block data struct that we serialize to disk rather than just the bytes (w/ TODO for checksumming this struct). dannycjones/mountpoint-s3#15
- Replace Base64 encoding of S3 key _only_ with SHA256. Consider what we want to do about ETag and block numbers later.
- Remove body of `cached_block_indices` and replace with `todo!("implement or deprecate")`. Less to review, since we are now considering removing this outright.

mountpoint-s3/src/data_cache/disk_data_cache.rs
dannycjones: Updating to hash now. I'll take splitting directories into subdirs as a TODO.

mountpoint-s3/src/data_cache/disk_data_cache.rs
dannycjones: OK yeah, sounds good. Agree, I think we should have at most one version in the cache. Maybe the odd block if doing something weird where two file handles have two objects open, but should be cleaned up by LRU.

dannycjones: I'll keep the ETag and Block ID in the path for now just in the interest of getting a new version of this PR out, and we can simplify in another PR.

I will move to SHA256 in this PR though for the S3 key.

mountpoint-s3/src/data_cache/disk_data_cache.rs
dannycjones: Dropping this. I'll implement as a `todo!()` in this PR, and strip out the other implementations in a later change.

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,333 @@
...
+/// Represents a fixed-size chunk of data that can be serialized.
+///
+/// TODO: Add checksum over struct (excl. `data`) to verify block metadata later.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct DataBlock {
+    block_idx: BlockIndex,
+    checksum: u32,
+    data: Bytes,
+    etag: String,
+    s3_key: String,
```
jamesbornholt, line 36: It might be nicer to have a `BlockHeader` struct that derives serde and can be checksummed/validated on its own, and then combine that with `data: ChecksummedBytes` here.

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,333 @@
...
+        path.push(cache_key.etag.as_str());
+        path
+    }
+
+    /// Get path for the given block.
+    fn get_path_for_block(&self, cache_key: &CacheKey, block_idx: BlockIndex) -> PathBuf {
+        let mut path = self.get_path_for_key(cache_key);
+        path.push(format!("{}.block", block_idx));
+        path
+    }
```
jamesbornholt, line 108: Did you decide what you want to do here w.r.t. including the etag and block index in the hash or not?

mountpoint-s3/src/data_cache/disk_data_cache.rs
dannycjones: Oh, I like that! Sounds good.

mountpoint-s3/src/data_cache/disk_data_cache.rs
dannycjones: I'm going to leave it as is for this PR. I think its worth leaving the block index out of the hash, so that we can wipe out all the blocks for a given S3 & ETag pair.

We can have more discussion in the PR that moves in that direction.

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,431 @@
...
+    /// Checksum over the bytes in `data`
+    data_checksum: u32,
+}
+
+impl DataBlock {
+    /// Create a new [DataBlock].
+    ///
+    /// This may return an integrity error if the checksummed byte buffer is found to be corrupt.
+    /// However, this check is not guaranteed and it shouldn't be assumed that the data within the block is not corrupt.
+    fn new(cache_key: CacheKey, block_idx: BlockIndex, bytes: ChecksummedBytes) -> DataCacheResult<Self> {
```
passaro, line 119: nit: both here and in `data()`, it seems early to map errors in `DataCacheError`s, even more without propagating the root cause. But can be reviewed/addressed later.

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -0,0 +1,431 @@
...
+    block_size: u64,
+    cache_directory: PathBuf,
+}
+
+/// Describes additional information about the data stored in the block.
+///
+/// It should be written alongside the block's data
+/// and used to verify it contains the correct contents to avoid blocks being mixed up.
+///
+/// TODO: Add checksum over struct to verify block metadata later.
```
passaro, line 32: Not sure I understand: isn't this already in `header_checksum`? And/or shouldn't we include the block checksum as well?

dannycjones: Sorry, yes! Missed removing that todo.

mountpoint-s3/src/data_cache/disk_data_cache.rs
dannycjones: I'll update at least to help the error types make more sense. I think they need to be split in two. I'll defer mapping the errors too.

passaro (approved):

dannycjones:
I'll merge this now. If there's any more feedback, please still add and I can follow up in another PR.

mountpoint-s3/src/data_cache/disk_data_cache.rs
dannycjones: Opened: #593

mountpoint-s3/src/data_cache/disk_data_cache.rs
dannycjones: Opened #594.

PR #594, opened 2023-11-03
Add ETag into DiskDataCache hashed block path

dannycjones:
## Description of change

We discussed in #593 to hash the S3 key, ETag, and possibly block ID together when defining a path for the disk-based data cache. This is the PR that implements that change.

I'm encoding only the ETag as I still see value in keeping the block ID visible in the path. This allows us to do things like invalidate all of the blocks for a given S3 object version (version defined by its ETag).

Previous discussions:
- https://github.com/awslabs/mountpoint-s3/pull/593#discussion_r1379502875
- https://github.com/awslabs/mountpoint-s3/pull/593#discussion_r1381392422

Relevant issues:
- #255 
- #593

No change to released behavior. This changes file paths for the blocks on disk for the data cache, which is behind a build-time feature flag.

---

### DIFF
--- a/mountpoint-s3/src/data_cache/disk_data_cache.rs
+++ b/mountpoint-s3/src/data_cache/disk_data_cache.rs
@@ -176,11 +176,13 @@ impl DiskDataCache {
     }
 }
 
-/// Hash the cache key using its fields, returning the hash encoded as hexadecimal in a new [String].
+/// Hash the cache key using its fields as well as the [CACHE_VERSION],
+/// returning the hash encoded as hexadecimal in a new [String].
 fn hash_cache_key(cache_key: &CacheKey) -> String {
     let CacheKey { s3_key, etag } = cache_key;
 
     let mut hasher = Sha256::new();
+    hasher.update(CACHE_VERSION.as_bytes());
     hasher.update(s3_key.as_bytes());
     hasher.update(etag.as_str().as_bytes());
     hex::encode(hasher.finalize())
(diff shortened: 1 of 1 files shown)


mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -276,30 +285,30 @@ mod tests {
...
 
-    #[test_case("hello"; "simple string")]
-    #[test_case("foo/bar/baz"; "with forward slashes")]
-    #[test_case("hello+world"; "with plus char")]
-    #[test_case("hello\\ world"; "backslash")]
-    #[test_case("hello=world"; "equals")]
-    #[test_case("look🎡emoji"; "emoji")]
-    fn test_get_path_for_key(s3_key: &str) {
+    #[test]
+    fn test_get_path_for_key() {
```
passaro, line 289: minor point: maybe we could use `hash_cache_key` here and in `test_get_path_for_block` to build the expected path, and write a separate test specifically for `hash_cache_key`.

dannycjones: sounds good! the thing I wanted to do was ensure we know when we change the encoding.

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -176,6 +176,16 @@ impl DiskDataCache {
     }
 }
 
+/// Hash the cache key using its fields, returning the hash encoded as hexadecimal in a new [String].
+fn hash_cache_key(cache_key: &CacheKey) -> String {
```
jamesbornholt, line 180: Might make sense to also include the `CACHE_VERSION` here, just as a defense-in-depth thing.

passaro (approved):
LGTM

PR #595, opened 2023-11-03
Introduce Prefetch trait

passaro:
## Description of change

Introduce a new `Prefetch` trait to abstract how `S3Filesystem` fetches object data from an `ObjectClient`. While this change does not introduce any functional change, this abstraction will be used to implement optional object data caching.

The existing `Prefetcher` struct has been adapted to implement the new `Prefetch` trait. The main changes are:
* it is generic on the `ObjectPartStream` (previously `ObjectPartFeed`), rather than using dynamic dispatch,
* it does not own an `ObjectClient` instance, instead one is required when initiating a `prefetch` request,
* the logic to spawn a new task for each `GetObject` request and handle the object body parts returned was moved into `ObjectPartStream`.

Relevant issues: #255 

No changes in behavior.

---


mountpoint-s3/src/prefetch.rs
```diff
@@ -7,32 +7,82 @@
...
+use crate::prefetch::task::RequestTask;
 use crate::sync::Arc;
 
-type TaskError<Client> = ObjectClientError<GetObjectError, <Client as ObjectClient>::ClientError>;
+/// Generic interface to handle reading data from an object.
+pub trait Prefetch {
+    type PrefetchResult<Client: ObjectClient + Send + Sync + 'static>: PrefetchResult<Client>;
+
+    /// Start a new prefetch request to the specified object.
+    fn prefetch<Client>(
```
jamesbornholt, line 40: Not going to block on it because we can refactor it later, but the `Client` type is a little weird here -- it's saying that you can pass any client into any prefetcher, but in reality it probably makes sense for a prefetcher to be coupled to a particular client once it's constructed.

I think in practice we want `trait Prefetch<Client>` or something, but that's a bit of annoying change to make.

mountpoint-s3/src/prefetch.rs
```diff
@@ -80,55 +130,63 @@ impl Default for PrefetcherConfig {
...
 }
 
-impl<Client, Runtime> Debug for PrefetcherInner<Client, Runtime> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("PrefetcherInner").field("config", &self.config).finish()
+impl<Stream> Prefetcher<Stream>
+where
+    Stream: ObjectPartStream,
+{
+    /// Create a new [Prefetcher] that will make requests to the given client.
```
jamesbornholt, line 142: nit: no client any more

jamesbornholt (commented):
Still not really sure we have the abstraction right here, but it's better than before and not worth blocking on, so just a couple notes.

jamesbornholt:
The send constraint might just be a nightly regression (https://github.com/rust-lang/rust/issues/117602), maybe check again tomorrow.

passaro:
> The send constraint might just be a nightly regression ([rust-lang/rust#117602](https://github.com/rust-lang/rust/issues/117602)), maybe check again tomorrow.

Right. It's been fixed now. I reverted the changes and rebased.

jamesbornholt (approved):

PR #598, opened 2023-11-06
Add caching Prefetch implementation

passaro:
## Description of change

Follow-up to #595  . Introduce a new implementation of the `Prefetch` trait which uses a `DataCache`. This change also adds CLI args to Mountpoint to enable the on-disk data cache. The new args are only available under the "caching" feature and are subject to change.

Relevant issues: #255

No. User-visible changes under the `caching` flag.

---

### DIFF
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -7,6 +7,7 @@
 //! we increase the size of the GetObject requests up to some maximum. If the reader ever makes a
 //! non-sequential read, we abandon the prefetching and start again with the minimum request size.
 
+mod caching_stream;
 mod part;
 mod part_queue;
 mod part_stream;
@@ -27,6 +28,8 @@ use thiserror::Error;
 use tracing::trace;
 
 use crate::checksums::{ChecksummedBytes, IntegrityError};
+use crate::data_cache::DataCache;
+use crate::prefetch::caching_stream::CachingPartStream;
 use crate::prefetch::part_stream::{ClientPartStream, ObjectPartStream, RequestRange};
 use crate::prefetch::seek_window::SeekWindow;
 use crate::prefetch::task::RequestTask;
@@ -81,6 +84,21 @@ where
     Runtime: Spawn + Send + Sync + 'static,
 {
     let part_stream = ClientPartStream::new(runtime);
+    Prefetcher::new(part_stream, prefetcher_config)
+}
+
+pub type CachingPrefetcher<Cache, Runtime> = Prefetcher<CachingPartStream<Cache, Runtime>>;
+
+pub fn caching_prefetch<Cache, Runtime>(
+    cache: Cache,
+    runtime: Runtime,
+    prefetcher_config: PrefetcherConfig,
+) -> CachingPrefetcher<Cache, Runtime>
+where
+    Cache: DataCache + Send + Sync + 'static,
+    Runtime: Spawn + Send + Sync + 'static,
+{
+    let part_stream = CachingPartStream::new(runtime, cache);
     Prefetcher::new(part_stream, prefetcher_config)
 }
 
@@ -504,6 +522,10 @@ mod tests {
     // It's convenient to write test constants like "1 * 1024 * 1024" for symmetry
     #![allow(clippy::identity_op)]
 
+    use crate::data_cache::InMemoryDataCache;
+    use crate::prefetch::part_stream::ClientPartStream;
+
+    use super::caching_stream::CachingPartStream;
     use super::*;
     use futures::executor::{block_on, ThreadPool};
     use mountpoint_s3_client::error::{GetObjectError, ObjectClientError};
@@ -514,6 +536,8 @@ mod tests {
     use proptest_derive::Arbitrary;
     use std::collections::HashMap;
     use test_case::test_case;
+
+    const MB: usize = 1024 * 1024;
 
     #[derive(Debug, Arbitrary)]
     struct TestConfig {
@@ -534,6 +558,12 @@ mod tests {
     fn default_stream() -> ClientPartStream<ThreadPool> {
         let runtime = ThreadPool::builder().pool_size(1).create().unwrap();
         ClientPartStream::new(runtime)
+    }
+
+    fn caching_stream(block_size: usize) -> CachingPartStream<InMemoryDataCache, ThreadPool> {
+        let runtime = ThreadPool::builder().pool_size(1).create().unwrap();
+        let cache = InMemoryDataCache::new(block_size as u64);
+        CachingPartStream::new(runtime, cache)
     }
 
     fn run_sequential_read_test<Stream: ObjectPartStream + Send + Sync + 'static>(
@@ -578,8 +608,12 @@ mod tests {
         assert_eq!(next_offset, size);
     }
 
-    #[test]
-    fn sequential_read_small() {
+    #[test_case(default_stream())]
+    #[test_case(caching_stream(1 * MB))]
+    fn sequential_read_small<Stream>(part_stream: Stream)
+    where
+        Stream: ObjectPartStream + Send + Sync + 'static,
+    {
         let config = TestConfig {
             first_request_size: 256 * 1024,
             max_request_size: 1024 * 1024 * 1024,
@@ -588,11 +622,15 @@ mod tests {
             max_forward_seek_distance: 16 * 1024 * 1024,
             max_backward_seek_distance: 2 * 1024 * 1024,
         };
-        run_sequential_read_test(default_stream(), 1024 * 1024 + 111, 1024 * 1024, config);
+        run_sequential_read_test(part_stream, 1024 * 1024 + 111, 1024 * 1024, config);
     }
 
-    #[test]
-    fn sequential_read_medium() {
+    #[test_case(default_stream())]
+    #[test_case(caching_stream(1 * MB))]
+    fn sequential_read_medium<Stream>(part_stream: Stream)
+    where
+        Stream: ObjectPartStream + Send + Sync + 'static,
+    {
         let config = TestConfig {
             first_request_size: 256 * 1024,
             max_request_size: 64 * 1024 * 1024,
@@ -601,11 +639,15 @@ mod tests {
             max_forward_seek_distance: 16 * 1024 * 1024,
             max_backward_seek_distance: 2 * 1024 * 1024,
         };
-        run_sequential_read_test(default_stream(), 16 * 1024 * 1024 + 111, 1024 * 1024, config);
+        run_sequential_read_test(part_stream, 16 * 1024 * 1024 + 111, 1024 * 1024, config);
     }
 
-    #[test]
-    fn sequential_read_large() {
+    #[test_case(default_stream())]
+    #[test_case(caching_stream(1 * MB))]
+    fn sequential_read_large<Stream>(part_stream: Stream)
+    where
+        Stream: ObjectPartStream + Send + Sync + 'static,
+    {
         let config = TestConfig {
             first_request_size: 256 * 1024,
             max_request_size: 64 * 1024 * 1024,
@@ -614,7 +656,8 @@ mod tests {
             max_forward_seek_distance: 16 * 1024 * 1024,
             max_backward_seek_distance: 2 * 1024 * 1024,
         };
-        run_sequential_read_test(default_stream(), 256 * 1024 * 1024 + 111, 1024 * 1024, config);
+
+        run_sequential_read_test(part_stream, 256 * 1024 * 1024 + 111, 1024 * 1024, config);
     }
 
     fn fail_sequential_read_test<Stream: ObjectPartStream + Send + Sync + 'static>(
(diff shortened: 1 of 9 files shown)


mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -0,0 +1,422 @@
...
+            .instrument(span)
+        };
+
+        let task_handle = self.runtime.spawn_with_handle(request_task).unwrap();
+
+        RequestTask::from_handle(task_handle, size, start, part_queue)
+    }
+}
+
+fn get_aligned_request_range(range: RequestRange, block_size: u64) -> RequestRange {
```
jamesbornholt, line 82: Sort of frustrating that this tricky part alignment logic is almost, but not quite, copied and pasted from the non-caching version. Actually, isn't it exactly the same except for the last case? Maybe we should write a common version that returns `Option<RequestRange>` and then have different fallback logic in the two cases?

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -0,0 +1,422 @@
...
+        Self {
+            client,
+            cache,
+            bucket,
+            cache_key,
+            part_queue_producer,
+        }
+    }
+
+    async fn get_from_cache(&self, range: RequestRange) {
```
jamesbornholt, line 143: nit: maybe to make more clear the usage pattern?
```suggestion
    async fn get_from_cache(self, range: RequestRange) {
```

(and then the callsite can just be
```rust
request.get_from_cache(range).instrument(span)
```
)

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -0,0 +1,422 @@
...
+        let key = &self.cache_key.s3_key;
+        let block_size = self.cache.block_size();
+        let block_range = self.block_indices_for_byte_range(&range);
+
+        // TODO: consider starting GetObject requests pre-emptively if cache blocks are missing
```
jamesbornholt, line 148: Just to make sure I understand: this stream is _behind_ the prefetcher, so not doing this doesn't really affect how good a job of prefetching we do, just possibly slightly delays some requests starting in the partially cached case.

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -0,0 +1,422 @@
...
+        for block_index in block_range.clone() {
+            match self.cache.get_block(&self.cache_key, block_index) {
```
jamesbornholt, line 150: Not for this commit: we'll want some metrics on cache hit/miss latency and cache hit rate.

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -0,0 +1,422 @@
...
+                        .await;
+                }
+                Err(error) => {
+                    trace!(?key, ?range, block_index, ?error, "error reading block from cache");
+                    return self
+                        .get_from_client(range.trim_start(block_index * block_size), block_index..block_range.end)
+                        .await;
+                }
+            }
+        }
```
jamesbornholt, line 170: Maybe this version makes the logic a little clearer? The point being that we fail out immediately if we fail to read a cache block or if a block isn't cached.
```rust
// TODO: consider starting GetObject requests pre-emptively if cache blocks are missing
for block_index in block_range.clone() {
    match self.cache.get_block(&self.cache_key, block_index) {
        Ok(Some(block)) => {
            trace!(?key, ?range, block_index, "cache hit");
            let part = self.make_part(block, block_index, &range);
            metrics::counter!("cache.total_bytes", part.len() as u64, "type" => "read");
            self.part_queue_producer.push(Ok(part));
            continue;
        }
        Ok(None) => trace!(?key, ?range, block_index, "cache miss - no data for block"),
        Err(error) => trace!(?key, ?range, block_index, ?error, "error reading block from cache"),
    }

    // If a block is uncached or reading it fails, we don't try to recover and just fall back to S3
    // for the entire stream
    return self
        .get_from_client(range.trim_start(block_index * block_size), block_index..block_range.end)
        .await;
}
```

jamesbornholt: also, probably `error!` in the error case, it's not something we expect in normal operation?

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -0,0 +1,422 @@
...
+                Some(Ok((offset, body))) => {
+                    trace!(offset, length = body.len(), "received GetObject part");
+
+                    // Split the body into blocks.
+                    let mut body: Bytes = body.into();
+                    while !body.is_empty() {
+                        let remaining = (block_size as usize).saturating_sub(buffer.len()).min(body.len());
+                        let chunk = body.split_to(remaining);
+                        if let Err(e) = buffer.extend(chunk.into()) {
+                            error!(key, error=?e, "Integrity check failed");
```
jamesbornholt, line 220:
```suggestion
                            error!(key, error=?e, "integrity check for body part failed");
```

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -0,0 +1,422 @@
...
+                return;
+            }
+        };
+
+        pin_mut!(get_object_result);
+        let mut block_index = block_range.start;
+        let mut buffer = ChecksummedBytes::default();
+        loop {
+            match get_object_result.next().await {
+                Some(Ok((offset, body))) => {
```
jamesbornholt, line 211: For paranoia, I think we want to keep the `offset` around and do some arithmetic with it to make sure it's always in sync with the `block_index`. Right now I can't see anything stopping us from flipping a bit in `block_index` and then returning the wrong bytes to a future read. Keeping `offset` around would give us a second source of truth for location (one that comes from S3 rather than being computed from `block_index`) that we can plumb all the way down into the cache -- the cache blocks should record their byte offset.

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -0,0 +1,422 @@
...
+    fn update_cache(&self, block_index: u64, block: &ChecksummedBytes) {
+        // TODO: consider updating the cache asynchronously
+        let start = Instant::now();
+        match self.cache.put_block(self.cache_key.clone(), block_index, block.clone()) {
+            Ok(()) => {
+                metrics::histogram!("cache.write_duration_us", start.elapsed().as_micros() as f64);
+                metrics::counter!("cache.total_bytes", block.len() as u64, "type" => "write");
+            }
+            Err(error) => {
+                warn!(key=?self.cache_key.s3_key, block_index, ?error, "failed to update cache");
```
jamesbornholt, line 265: Not for this PR: metric on cache failures? We should take a pass over all our cache metrics once we're happy with the implementation.

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -0,0 +1,422 @@
...
+                Some(Err(e)) => {
+                    error!(key, error=?e, "GetObject body part failed");
+                    self.part_queue_producer
+                        .push(Err(PrefetchReadError::GetRequestFailed(e)));
+                    break;
+                }
+                None => {
+                    if !buffer.is_empty() {
+                        if buffer.len() + (block_index * block_size) as usize == range.object_size() {
+                            // Write last block to the cache.
```
jamesbornholt, line 243: Took me a second to parse this: this only applies if we're at the end of the entire object, not the end of the request? What happens if the last part of this request isn't block-sized -- we just drop it on the floor? Isn't that going to break caching the first 8MB of an object, where the requests won't be nicely MB-aligned?

mountpoint-s3/src/prefetch/caching_stream.rs
passaro: Yes, I like it.

mountpoint-s3/src/prefetch/caching_stream.rs
passaro: That's correct. The TODO is there to remind of the possible optimization, but we'd need to measure if it would matter. I'll rephrase it.

mountpoint-s3/src/prefetch/caching_stream.rs
passaro: It needs a comment. The `if` handles exactly that case: the last part of the request is not block-sized but it reaches the end of the object. We do want to cache the (smaller) block.
In practice, this is the only case. Requests either end on a block boundary or at the end of the object, because they come from `get_aligned_request`. But we do no enforce it.

dannycjones: I also struggled to grok this bit at first, but wasn't sure how we can make it clearer.

In what cases do we need to worry about a non-empty buffer where its not the end of the object? Should we add `unreachable!` or a warning in this case?

jamesbornholt: I think "requests either end on a block boundary or at the end of the object" is the part I was missing. Can we just assert that? Or at least `debug_assert!` it if we're paranoid about not crashing if we got the math wrong, since it is recoverable.

mountpoint-s3/src/prefetch.rs
```diff
@@ -703,6 +751,23 @@ mod tests {
...
+        ) {
+            run_sequential_read_test(caching_stream(block_size), size, read_size, config);
+        }
+
+        #[test]
+        fn proptest_sequential_read_small_read_size_with_cache(size in 1u64..1 * 1024 * 1024, read_factor in 1usize..10,
+            block_size in 16usize..1 * 1024 * 1024, config: TestConfig) {
+            let read_size = (size as usize / read_factor).max(1);
+            run_sequential_read_test(caching_stream(block_size), size, read_size, config);
+        }
```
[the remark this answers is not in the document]
jamesbornholt: We have both because it's annoying to write "choose a random read size _up to_ the object size" with the nice proptest syntax. It's just laziness on my part when I wrote the tests these new versions are copied and pasted from. Let's fix it separately.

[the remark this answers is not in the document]
dannycjones: Ah, got it! I'm happy with a comment that points the reader in that direction.

mountpoint-s3/src/prefetch/caching_stream.rs
passaro: Yes. I also had gone back and forth on this. Another option to consider - maybe later - is for `get_block` to return an error on cache miss.

mountpoint-s3/tests/fuse_tests/mod.rs
```diff
@@ -145,6 +146,44 @@ mod mock_session {
...
+                &prefix,
+                mount_dir.path(),
+                test_config.filesystem_config,
+            );
+            let test_client = create_test_client(client, &prefix);
+
+            (mount_dir, session, test_client)
+        }
+    }
+
```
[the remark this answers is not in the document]
passaro: It's more complicated because they need 2 different types of prefetchers. We may need to pass a "prefetcher creator" function. Not sure it's worth to avoid the code duplication at this time.

mountpoint-s3/src/prefetch/caching_stream.rs
passaro: Added the check but not stored the byte offset in the cache blocks.

mountpoint-s3/src/prefetch/caching_stream.rs
passaro: Here we are handling the response from `GetObject` though, so I don't think an assertion is appropriate.

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -0,0 +1,422 @@
...
+            block_size as usize
+        } else {
+            // if it exceeds block boundaries, trim it to the block boundaries
+            let request_boundary = offset + preferred_size as u64;
+            let remainder = request_boundary % block_size;
+            if remainder != 0 {
+                preferred_size + (block_size - remainder) as usize
+            } else {
+                preferred_size
+            }
```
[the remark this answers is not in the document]
passaro: This code is now in `RequestRange::align`

mountpoint-s3/tests/fuse_tests/mod.rs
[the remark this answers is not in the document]
dannycjones: Can't be done due to prefetcher type, no worries.

mountpoint-s3/src/prefetch/caching_stream.rs
passaro: I've come around: added comments and asserts to clarify what is in the buffer throughout the loop but also the assert on the last block. It can only fail if we introduce a logic bug in this function or if the client starts returning the wrong range.

mountpoint-s3/src/prefetch/caching_stream.rs
passaro: Will complete in a separate change

jamesbornholt (approved):

PR #603, opened 2023-11-14
Update PyTorch example to support single files and different models

jamesbornholt:
We'd like to expand our testing to single files rather than only sharded representations. This change adds the ability to create and train against that style of dataset. It also adds some other useful configurations for Mountpoint training and for configuring a different model to train rather than hardcoding ResNet-50.

## Description of change

Relevant issues: 

---


monthonk (commented):
Nice improvement! left a few minor comments.

I also got error `ModuleNotFoundError: TorchData must be built with BUILD_S3=1 to use this datapipe` while testing s3io job, not sure if it's related to this change though.

jamesbornholt:
> I also got error `ModuleNotFoundError: TorchData must be built with BUILD_S3=1 to use this datapipe` while testing s3io job, not sure if it's related to this change though.

No, not related -- the torchdata install in the DLAMI isn't built with S3IO support, you have to uninstall and reinstall it from pip. Kinda annoying.

monthonk (approved):

PR #608, opened 2023-11-15
Add instance and config metadata to user agent

jamesbornholt:
## Description of change

We'd like to record the same metadata that AWS SDKs gather so we can
better understand how different Mountpoint features are used. This
change adds support for detecting platform and instance metadata and
including it in HTTP User-agents. We follow the SDK template for
serializing this metadata.

To make this cleaner, I moved the instance info logic into the client
crate so that all users can get this kind of user agent. The new
`UserAgent` struct supports addings arbitrary key/value pairs, and we
use that in Mountpoint to record basic configurations.

User agents are always a bit annoying to test, but I manually verified
in a few cases (caching enabled/disabled) that this change was sending
the expected headers.

No.

---


mountpoint-s3/src/autoconfigure.rs
```diff
@@ -0,0 +1,48 @@
...
+pub fn network_throughput(instance_info: &InstanceInfo) -> anyhow::Result<f64> {
+    let instance_type = instance_info
+        .instance_type()
+        .map_err(|e| anyhow!("failed to get instance type: {e}"))?;
+    let throughput = get_maximum_network_throughput(instance_type).context("failed to get network throughput")?;
+    Ok(throughput)
+}
+
+fn get_maximum_network_throughput(ec2_instance_type: &str) -> anyhow::Result<f64> {
+    const INSTANCE_THROUGHPUT: &str = "instance_throughput";
```
passaro, line 17: Not for this PR, but should we consider using the recommended configuration from https://github.com/awslabs/aws-c-s3/pull/367?

dannycjones (approved):
LGTM, nice.

dannycjones (changes requested):
Almost LGTM, one typo I forgot I commented

dannycjones (approved):

mountpoint-s3/src/autoconfigure.rs
jamesbornholt: Right now that API only supports a few hard-coded instance types: https://github.com/awslabs/aws-c-s3/blob/5fe1c3b116d4b2ac69304af2d2d3031e4b6d6d14/source/s3_platform_info.c#L328-L335

I think probably we should consult it first, and then fall back to our version if we don't get a match from it.

passaro: I was hoping we could remove our list. Not really, unfortunately.

PR #609, opened 2023-11-16
Fix bug in ChecksummedBytes::into_inner

passaro:
## Description of change

`ChecksummedBytes::into_inner()` was returning data from `self` rather than from the `shrink_to_fit` result. Added regression tests for `ChecksummedBytes` and for `DiskDataCache` (only caller of the `into_inner()`).

No. 

---


dannycjones (approved):
Nice catch! LGTM, thanks for adding plenty of tests here.

PR #611, opened 2023-11-17
Store block offsets in disk data cache

passaro:
## Description of change

Add the block offset to the header stored in each cache block. This allows for additional consistency checks when reading or writing to the cache, where the offset can be compared with the block index.

It does change the on disk cache format, but it is still currently under the `caching` feature flag.

---


dannycjones (dismissed):
In the interest of stabilizing the format on disk, I'm inclined to merge as is.

I do like the idea we discussed of having some kind of guard type that wraps something and requires a certain set of metadata to be provided before allowing code access to the contents. Let's do this at a later point.

dannycjones (approved):

PR #612, opened 2023-11-17
Simplify and rename cache configuration flags

passaro:
## Description of change

Reduce the number of flags to configure the object metadata and content cache. 

This is how they will be displayed in `mount-s3 --help`:
```
Caching options:
      --cache <DIRECTORY>       Enable caching of object metadata and content to the given directory
      --metadata-ttl <SECONDS>  Time-to-live (TTL) for cached metadata in seconds [default: 1s]
      --max-cache-size <MiB>    Maximum size of the cache directory in MiB [default: preserve 5% of available space]
```

Relevant issues: #255 

No. All flags under the `caching` feature.

---

### DIFF
--- a/mountpoint-s3/src/data_cache/disk_data_cache.rs
+++ b/mountpoint-s3/src/data_cache/disk_data_cache.rs
@@ -26,11 +26,28 @@ const HASHED_DIR_SPLIT_INDEX: usize = 2;
 
 /// On-disk implementation of [DataCache].
 pub struct DiskDataCache {
-    block_size: u64,
     cache_directory: PathBuf,
-    limit: CacheLimit,
+    config: DiskDataCacheConfig,
     /// Tracks blocks usage. `None` when no cache limit was set.
     usage: Option<Mutex<UsageInfo<DiskBlockKey>>>,
+}
+
+/// Configuration for a [DiskDataCache].
+#[derive(Debug)]
+pub struct DiskDataCacheConfig {
+    /// Size of data blocks.
+    pub block_size: u64,
+    /// How to limit the cache size.
+    pub limit: CacheLimit,
+}
+
+impl Default for DiskDataCacheConfig {
+    fn default() -> Self {
+        Self {
+            block_size: 1024 * 1024,                               // 1 MiB block size
+            limit: CacheLimit::AvailableSpace { min_ratio: 0.01 }, // Preserve 1% available space
+        }
+    }
 }
 
 /// Limit the cache size.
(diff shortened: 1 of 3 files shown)


mountpoint-s3/src/main.rs
```diff
@@ -597,34 +561,25 @@ fn mount(args: CliArgs) -> anyhow::Result<FuseSession> {
...
-                }
-            };
 
-            let cache = DiskDataCache::new(path, args.data_cache_block_size, limit);
+            let mut cache_config = DiskDataCacheConfig::default();
+            if let Some(max_size_in_mb) = args.max_cache_size {
+                cache_config.limit = CacheLimit::TotalSize {
+                    max_size: (max_size_in_mb * 1_000_000) as usize,
+                };
+            }
```
dannycjones, line 581: Why megabytes over mebibytes?

dannycjones (commented):
Two comments.

I think we should add a few CLI tests in a follow-up change also.

mountpoint-s3/src/main.rs
passaro: To match the flag description. Happy to change, but then would we also have something like this?
```
--max-cache-size <MiB>     Maximum size of the cache directory in MiB [default: preserve 1% of available space]
```

dannycjones: Yeah, we'd need to change both

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -26,13 +26,30 @@ const HASHED_DIR_SPLIT_INDEX: usize = 2;
...
+    pub block_size: u64,
+    /// How to limit the cache size.
+    pub limit: CacheLimit,
+}
+
+impl Default for DiskDataCacheConfig {
+    fn default() -> Self {
+        Self {
+            block_size: 1024 * 1024,                               // 1 MiB block size
+            limit: CacheLimit::AvailableSpace { min_ratio: 0.01 }, // Preserve 1% available space
```
jamesbornholt, line 48: default should be larger because of FS metadata overhead (inodes etc), let's do 5%

mountpoint-s3/src/main.rs
```diff
@@ -223,68 +223,33 @@ struct CliArgs {
     #[cfg(feature = "caching")]
     #[clap(
         long,
-        help = "Enable caching of file and directory metadata for a configurable time-to-live (TTL)",
+        help = "Enable caching of object metadata and content. Directory provided will be used for caching object content. If running multiple Mountpoint processes concurrently, use a unique value for this argument",
```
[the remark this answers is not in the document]
dannycjones: I just suggested to go the other way 😄 I don't mind, let's just ensure we are on the same page with future changes to CLI args.

mountpoint-s3/src/main.rs
dannycjones: Let's use MiB.

dannycjones (approved):
LGTM, pending anything from James

jamesbornholt (approved):

PR #613, opened 2023-11-17
Use top-level file in write benchmark

passaro:
The write throughput benchmarks were failing when running the second iteration because fio would not re-create the nested path for the benchmark file. This appears to have started since #584. Further investigation required to determine if there is an issue with Mountpoint.
This change only unblocks the benchmarks by configuring fio to use a file at the top-level in the mount directory, which was likely the intention in the first place (fio's `--filename` path is relative to `--directory`).

---

### DIFF
--- a/mountpoint-s3/scripts/fs_bench.sh
+++ b/mountpoint-s3/scripts/fs_bench.sh
@@ -153,7 +153,7 @@ write_benchmark () {
     fi
 
     # set bench file
-    bench_file=${mount_dir}/${job_name}_${RANDOM}.dat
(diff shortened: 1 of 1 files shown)


jamesbornholt (approved):
From the log, it looks like this is implicit directories working as intended: once the last object in the directory was unlinked, the directory went away, but fio doesn't try to recreate the entire directory hierarchy, just the file. I assume it worked before #584 because the local directory's dentry was cached forever by the kernel, so it never needed to re-lookup the directory to do the `creat`.

PR #614, opened 2023-11-20
Implement O_DIRECT for open to bypass metadata cache

dannycjones:
## Description of change

This is part of metadata and data caching feature currently in implementation: #581.

Some applications may want to opt-out of caching, maybe aware or not aware of Mountpoint itself. Applications use the `O_DIRECT` flag when opening a file to indicate that file content should be read directly, bypassing the Kernel caches.

We want to also skip Mountpoint's metadata cache when this flag is provided, so that open calls using it will see the latest object in S3 at open - as it works today without opting into caching.

Relevant issues: #255

No breaking changes, changes only to a feature has not been released yet.

---


mountpoint-s3/tests/fuse_tests/consistency_test.rs
```diff
@@ -83,3 +83,98 @@ fn page_cache_sharing_test_mock_with_cache(prefix: &str) {
...
+                .expect("should be OK as result is cached");
+            assert_eq!(
+                buf,
+                &old_contents[..buf.len()],
+                "bytes read should be old object from cache"
+            );
+        }
+
+        // Open the file w/ O_DIRECT, which should see the new file on S3 despite the old file being cached
+        let mut buf = [0u8; OBJECT_SIZE];
```
dannycjones, line 147: If I use `vec![0u8; OBJECT_SIZE]` here, I sometimes find that the buffer hasn't been written to (even though `read_exact_at` has returned `Ok`). This makes me super uncomfortable (which is why this PR is still draft, not ready to merge).

jamesbornholt: `O_DIRECT` is really hard to test under the Rust test runner because it has undefined behavior if run concurrently with a `fork`. We've seen it before in #114. Do you still see the issue if you filter to run only these tests?

dannycjones: Ah, I saw those limitations with `O_DIRECT` and `fork` but wasn't expecting us to be using it in the tests. That sounds like a probable cause!

I don't see the issue when running the test on its own.

passaro (approved):
Change in `open` looks good to me.

I've re-organized the integration tests so that we can force your O_DIRECT tests to execute serially.

mountpoint-s3/tests/common/mod.rs
```diff
@@ -1,3 +1,8 @@
+#![allow(unused)]
```
dannycjones, line 1: Do we need this if hiding `fuse` behind the feature?

mountpoint-s3/tests/direct_io.rs
```diff
@@ -0,0 +1,104 @@
+//! O_DIRECT tests are grouped in a separate integration test module so they can be
+//! run serially. This is because O_DIRECT has undefined behavior if run concurrently
+//! with a fork, and FUSE tests involve a fork to spawn fusermount.
+
+#![cfg(target_os = "linux")]
+#![cfg(feature = "s3_tests")]
```
dannycjones, line 6: Tests in this module also run with mock client, we should drop this condition?

dannycjones:
LGTM, but its my PR so also needs your approval @passaro  😄

passaro (approved):

PR #615, opened 2023-11-20
Use a unique target for request spans

jamesbornholt:
## Description of change

These spans are emitted at a high level (WARN), which is annoying for clients that want to use an adapter to the `log` facade, where `tracing` will emit an event whenever a span is created. To help these clients filter out these span events, let's use a unique target for them.

I verified that log messages for requests are still emitting this span metadata when Mountpoint requests log, so this only affects clients that aren't using `tracing`.

No.

---


passaro (approved):
LGTM

PR #618, opened 2023-11-21
Improve logging for ETags and FUSE operation flags

dannycjones:
## Description of change

Small logging improvements I wanted to make while debugging #614.

- It was useful to include the ETag at different points, but the named-field struct for ETag meant the `Debug` format was noisy. Moving to a tuple struct is simple and ETag is now formatted as `ETag("etag")`.
- FUSE operations which take flags were originally formatting the flags as decimal numbers. It's appears easier to interpret the flags when using binary usually, so we update the logs to format those numbers as binary like `0b10010101`.
- Some of the logs on their own weren't useful. For example, we know an Inode lookup count was incremented but we only know the parent Inode. Fix that by including the Inode number that actually had its count incremented.
- We never emitted the new file handle number before use - now we do that.

Relevant issues: N/A

No breaking changes.

---


passaro (approved):
LGTM

PR #633, opened 2023-11-27
Update CRT submodules to latest releases

dannycjones:
## Description of change

Pulling in the latest changes.

```
9e11ad8 Update CRT submodules to latest releases
Submodule mountpoint-s3-crt-sys/crt/aws-c-auth 71bad38..50f3e0d:
  > Add an Option to Disable v2 to v1 Fallback (#218)
Submodule mountpoint-s3-crt-sys/crt/aws-c-common fefbf4b..80f21b3:
  > Fixed directory creation to ignore umask, and allow cross process loc… (#1078)
Submodule mountpoint-s3-crt-sys/crt/aws-c-io c9cb777..df64f57:
  > Use 500ms rather than 25ms for the exponential backoff defaults. (#612)
  > Remove null check for aws_pkcs11_lib_login_user (#611)
Submodule mountpoint-s3-crt-sys/crt/aws-c-s3 83008e5..cc6ba34:
  > Report S3 operation name of specific request that failed. (#377)
  > Auto - Update S3 Ruleset & Partition (#376)
  > Mem limiter (#368)
  > fix stale bot permissions (#371)
Submodule mountpoint-s3-crt-sys/crt/aws-lc a8d06de..80f3f33:
  > Bump release version number string to 1.17.4 (#1319)
  > Improve Codecov reporting (#1313)
  > Merge pull request #1314 from torben-hansen/aws-lc-s2n-bignum-update-2023-11-19
  > Merge pull request #1308 from torben-hansen/aws-lc-s2n-bignum-update-2023-11-15
  > Add BNAssertTest (#1267)
  > Silence warning on CMP0116 (#1284)
  > Use OPENSSL_DEPRECATED (#1285)
  > Add integration ci for Monit (#1286)
  > Include Codecov in CI (#1307)
  > Fix bn_assert_fits_in_bytes for big-endian (#1258)
  > Fix-up curve25519 code to prepare for s2n-bignum (#1306)
  > Merge pull request #1302 from samuel40791765/upstream-merge-2023-11-10
  > Add check for sk_X509_push (#1304)
  > Refactor ED25519_verify into hw and nohw backend (#1305)
  > Add back OCSP integration test executing (#1303)
  > Refactor ED25519_sign into hw and nohw backend (#1276)
  > Add HybridKeyShare support for SecP256r1Kyber768Draft00 and X25519Kyber768Draft00 (#1201)
```

Relevant issues: N/A

## Does this change impact existing behavior?

[...]

PR #635, opened 2023-11-27
Fix mountpoint-s3-crt RequestType conversion

dannycjones:
## Description of change

This is mainly used in metrics, so impact should have been minimal.

What was happening before this change was that the match arms were binding the matched pattern to a new variable `AWS_S3_REQUEST_TYPE_DEFAULT`, rather than matching the Enum. This is because `AWS_S3_REQUEST_TYPE_DEFAULT` is actually a constant representing an Enum value.

I'll follow up this change with a fix to Clippy in CI to actually fail on warnings, which I believe we wanted but it does not appear to work correctly. (We should deny `warnings`, not just `clippy::all`.)

Relevant issues: N/A

It may fix some metric reporting, if any.

---

### DIFF
--- a/mountpoint-s3-crt/src/s3/client.rs
+++ b/mountpoint-s3-crt/src/s3/client.rs
@@ -1029,17 +1029,16 @@ pub enum RequestType {
 
 impl From<aws_s3_request_type> for RequestType {
     fn from(value: aws_s3_request_type) -> Self {
-        use aws_s3_request_type::*;
         match value {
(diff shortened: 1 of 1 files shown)


dannycjones:
We need to review this commit as I think we should update bindings: https://github.com/awslabs/aws-c-s3/commit/cc6ba346b55ef012f7131d98dcc68e16acc16d95

dannycjones:
This is related to the CRT update #633. I'll include this fix in that PR.

PR #637, opened 2023-11-28
Update cache directory to create content with MP owner access only

dannycjones:
## Description of change

Even though this directory is within a path provided by the user, we can provide a better default permission mode. No-one but the current Mountpoint process needs access to the contents of this directory, so we default to 700/600 for cache directories/files we are creating and deleting.

We also simplify our directory cleanup logic by outright deleting the whole sub-directory rather than just emptying it. This will also surface issues earlier (at mount time) since we'll know if we cannot write to this directory.

Relevant issues: #255.

The inner cache directory is an internal of Mountpoint and should not be depended on by user applications.

The default permission mode will be made more restrictive. Any user running the version containing this change will have the cache directory removed entirely at mount time (rather than just the contents as it is today), and the directory will be created with the new permission mode.

Before this change, any permission errors when using the cache directory may have instead been surfaced after mount time even if we wouldn't have been able to write there in the first place (assuming the directory already existed). Now, it will fail at Mount time.

---

### DIFF
--- a/mountpoint-s3/src/data_cache/cache_directory.rs
+++ b/mountpoint-s3/src/data_cache/cache_directory.rs
@@ -8,6 +8,7 @@
 
 use std::fs;
 use std::io;
+use std::os::unix::fs::DirBuilderExt;
 use std::path::{Path, PathBuf};
 
 use thiserror::Error;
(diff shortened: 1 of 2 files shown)


passaro (approved):
LGTM

PR #639, opened 2023-11-28
Update Clippy CI to deny warnings

dannycjones:
## Description of change

Our intention is for there to be no Clippy warnings, and our CI should enforce this. Today, it only enforces Clippy's own warnings but not more generic ones. This change fixes that!

It would have caught a regression we discovered in #633 earlier, which we only noticed by reviewing warnings when debugging another issue.

Relevant issues: #633

No change.

---


passaro (commented):
Let's rebase on `main` first.

passaro (approved):

PR #642, opened 2023-11-28
Support S3 Express One Zone

monthonk:
## Description of change

This change configure `mountpoint-s3-client` to use the new SigV4-Express signing algorithm when we detected that the bucket is an S3 Express One Zone bucket.

The tests are modified to support testing against S3 Express One Zone buckets, we can enable them by providing environment variables `S3_BUCKET_NAME`, `S3_EXPRESS_ONE_ZONE_ENDPOINT` and run the test with `s3express_tests` feature flag. For example. if the S3 Express One Zone bucket name is `doc-example-bucket--use1-az5--x-s3`, we can set environment variables with the following values
- `S3_BUCKET_NAME=doc-example-bucket--use1-az5--x-s3`
- `S3_EXPRESS_ONE_ZONE_ENDPOINT=https://doc-example-bucket--use1-az5--x-s3.s3express-use1-az5.us-east-1.amazonaws.com`

This PR also includes @jamesbornholt's change to fix a problem where ListObjectsV2 is unordered on S3 Express. This is an issue for our shadowing semantics in readdir, which does a merge-sort of the local and remote keys, assuming they're both ordered.

We are going to give up on readdir being ordered on Express -- it's not required by VFS, and the alternative would be to buffer all directory entries in memory before returning any, which would be too expensive on large directories. So this change introduces a new unordered variant of the readdir implementation. It just returns directory entries verbatim from ListObjectsV2, and then returns all otherwise-unreturned local entries at the end of the iterator.

Doing this requires Mountpoint to know whether the bucket is a directory bucket or not, and so I added both a command-line flag and auto-detection for that.

No breaking changes.

---


dannycjones (approved):
LGTM! 🚀

PR #646, opened 2023-11-28
Exclude unused files from mountpoint-s3-crt-sys crate

dannycjones:
## Description of change

The size of the `mountpoint-s3-crt-sys` create has exceeded 10MiB. We exclude from the crate some files we know we don't use. We should also follow up and eliminate more to ensure the crate is smaller again.

Relevant issues: N/A

No breaking change.

---


passaro (approved):
LGTM

PR #649, opened 2023-11-29
Fix crash on invalid part size

passaro:
## Description of change

The CRT changed the allowed values for `part_size` in a recent update (merged in #633). Because of that change, launching Mountpoint with previously accepted part sizes may result in a crash, e.g.:

```
> mount-s3 <BUCKET> <DIR> --foreground --part-size 10000000 
[...] panic on Thread { id: ThreadId(1), name: Some("main"), .. } at mountpoint-s3-client/src/s3_crt_client.rs:304:64: called `Result::unwrap()` on an `Err` value: Error(34, "aws-c-common: AWS_ERROR_INVALID_ARGUMENT, An invalid argument was passed to a function.")
```

This change prevents the panic and instead reports an error. Note that the error message is currently not too informative. We plan to improve it in a follow up change.

Example:
```
> mount-s3 <BUCKET> <DIR> --foreground --part-size 10000000 --debug-crt
[...] ERROR awscrt::S3Client: Failed to initialize buffer pool. Chunk size must be either 0 or more than 1 KB and size must be 1 KB aligned.    
Caused by:
    0: Unknown CRT error
    1: CRT error 34: aws-c-common: AWS_ERROR_INVALID_ARGUMENT, An invalid argument was passed to a function.
```

N/A

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -301,7 +301,7 @@ impl S3CrtClientInner {
         let user_agent = config.user_agent.unwrap_or_else(|| UserAgent::new(None));
         let user_agent_header = user_agent.build();
 
-        let s3_client = Client::new(&allocator, client_config).map_err(NewClientError::construction_error)?;
+        let s3_client = Client::new(&allocator, client_config).map_err(NewClientError::CrtError)?;
 
         Ok(Self {
             allocator,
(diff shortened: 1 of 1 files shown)


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -769,10 +769,24 @@ pub enum NewClientError {
...
     InvalidConfiguration(String),
+    /// An internal error from within the AWS Common Runtime
+    #[error("Unknown CRT error")]
+    CrtError(#[source] mountpoint_s3_crt::common::error::Error),
+}
+
+impl NewClientError {
+    fn construction_error(e: mountpoint_s3_crt::common::error::Error) -> Self {
+        const AWS_ERROR_INVALID_ARGUMENT: i32 = 34;
+        if e.raw_error() == AWS_ERROR_INVALID_ARGUMENT {
```
monthonk, line 784: I'm a bit uncomfortable about using hard-coded number. Should we add a function `is_invalid_argument_error` in `mountpoint_s3_crt::common::error::Error` and use the enum from Rust binding directly?

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -769,10 +769,24 @@ pub enum NewClientError {
...
+            Self::InvalidConfiguration("see crt logs for details".into())
```
monthonk, line 785: Would using `e.to_string()` gives a bit better hint for the error?

monthonk (commented):
Thanks for fixing this @passaro , added a few comments.

mountpoint-s3-client/src/s3_crt_client.rs
passaro: I've removed this method and just wrapped the CRT error for now. The error message may be even less informative, but we can fix that at a later time.

monthonk (approved):
LGTM!

PR #634, opened 2023-11-27
Add FileName/Object key in error logging

sauraank:
## Description of change
`InodeError` is logged mutliple times directly at error level. So, added Filename in FileDoesNotExist type. Could not add InodeErrorInfo for InodeDoesNotExist error type because it is called by [get()](https://github.com/awslabs/mountpoint-s3/blob/801e4c175cb85e21f713cf919e0f18ac8d2a9188/mountpoint-s3/src/inode.rs#L527) of `SuperBlockInner`. It only has Inode number as a parameter. 
Also added object key in UploadWriteError types. 
For rest of the errors added object key wherever it was available. 

Relevant issues: 
#555 
## Does this change impact existing behavior?
No, it does not impact existing behaviour.

No breaking changes

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

Issue #555, opened 2023-10-12
Consistently show file names in log entries


passaro:
Log entries in Mountpoint, especially those reporting warnings or errors, should report the file name (or object key) involved. That would be helpful for users trying to understand Mountpoint behavior or track down issues.

Currently, not all entries include the file name. For example, this is the warning logged when a write operation fails because it is out of order:
```
WARN write{req=52 ino=49 fh=3 offset=512 length=512}: mountpoint_s3::fuse: write failed: upload error: out of order write; expected offset 0 but got 512
```

This is related to #533, but focuses only on surfacing file names.

PR #665, opened 2023-12-05
Add filename at Error and Warn level in logs

sauraank:
## Description of change
Added filename (object key where filename was not available) at the error level and warn level logs since these are shown by default.

Used span to add the filename on the first occurence of filename in the callstack of the method. This results in logging like -
```
getattr{req=3 ino=34 name=“hello_new1.txt”}: mountpoint_s3::fuse: getattr failed: inode error: inode 34 (full key “hello_new1.txt”) for remote key “hello_new1.txt” is stale, replaced by inode 35 (full key “hello_new1.txt”)
```
In the PR #634 , it was getting logged as -
```
getattr{req=7 ino=2}: mountpoint_s3::fuse: getattr failed: inode error: inode 2 (full key “hello_new.txt”) for remote key “hello_new.txt” is stale, replaced by inode 3 (full key “hello_new.txt”)
```
Keeping both the PRs for now, in case we dont need the way it is done in this PR.

Also, added filename in all cases of `InodeError` as it was getting multiple times.

Relevant issues: 
#555 

No, it does not change existing behaviour of any functionalities. Only change the error logs.

No breaking changes.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -6,7 +6,7 @@ use std::ffi::{OsStr, OsString};
 use std::str::FromStr;
 use std::time::{Duration, UNIX_EPOCH};
 use time::OffsetDateTime;
-use tracing::{debug, error, trace};
+use tracing::{debug, error, trace, Span};
 
 use fuser::consts::FOPEN_DIRECT_IO;
 use fuser::{FileAttr, KernelConfig};
@@ -666,6 +666,7 @@ where
                 None => return reply.error(err!(libc::EBADF, "invalid file handle")),
             }
         };
+        Span::current().record("name", handle.inode.name());
         let file_etag: ETag;
         let mut request = match &handle.typ {
             FileHandleType::Write { .. } => return reply.error(err!(libc::EBADF, "file handle is not open for reads")),
@@ -770,6 +771,7 @@ where
                 None => return Err(err!(libc::EBADF, "invalid file handle")),
             }
         };
+        Span::current().record("name", handle.inode.name());
 
         let len = {
             let mut request = match &handle.typ {
@@ -971,6 +973,7 @@ where
                 None => return Err(err!(libc::EBADF, "invalid file handle")),
             }
         };
+        Span::current().record("name", file_handle.inode.name());
         let mut request = match &file_handle.typ {
             FileHandleType::Write(request) => request.lock().await,
             FileHandleType::Read { .. } => return Ok(()),
@@ -1019,6 +1022,7 @@ where
                 .remove(&fh)
                 .ok_or_else(|| err!(libc::EBADF, "invalid file handle"))?
         };
+        Span::current().record("name", file_handle.inode.name());
 
         // Unwrap the atomic reference to have full ownership.
         // The kernel should make a release call when there is no more references to the file handle,
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -167,7 +167,7 @@ impl<Client: ObjectClient> UploadState<Client> {
                     UploadState::InProgress { handle, .. } => {
                         if let Err(err) = handle.finish_writing() {
                             // Log the issue but still return the write error.
-                            error!(?err, "error updating the inode status");
+                            error!(?err, "error updating the inode status (key={:?})", key);
                         }
                     }
                     Self::Failed(_) | Self::Completed => unreachable!("checked above"),
@@ -231,7 +231,7 @@ impl<Client: ObjectClient> UploadState<Client> {
         };
         if let Err(err) = handle.finish_writing() {
             // Log the issue but still return put_result.
-            error!(?err, "error updating the inode status");
+            error!(?err, "error updating the inode status (key={:?})", key);
         }
         put_result
     }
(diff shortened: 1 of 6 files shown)


sauraank:
Closing this PR as have another PR #6

mountpoint-s3/src/fuse.rs
```diff
@@ -92,28 +92,28 @@ where
         }
     }
 
-    #[instrument(level="warn", skip_all, fields(req=_req.unique(), ino=ino))]
+    #[instrument(level="warn", skip_all, fields(req=_req.unique(), ino=ino, name=field::Empty))]
```
passaro, line 95: Why `name` and not `key`? Is there any case where we do not have the full key?

mountpoint-s3/src/fs.rs
```diff
@@ -666,6 +666,7 @@ where
                 None => return reply.error(err!(libc::EBADF, "invalid file handle")),
             }
         };
+        Span::current().record("name", handle.inode.name());
```
passaro, line 669: Should we refactor all these calls into a function? Maybe in the `logging` module?
So that the call here would look something like (assuming we do want the key instead of the name):
```suggestion
        logging::record_key(&handle.full_key);
```

mountpoint-s3/src/inode.rs
```diff
@@ -838,7 +842,7 @@ impl SuperblockInner {
             InodeKindData::Directory { children, .. } => children.get(name).cloned(),
         };
         match (remote, inode) {
-            (None, None) => Err(InodeError::FileDoesNotExist),
+            (None, None) => Err(InodeError::FileDoesNotExist(name.into())),
```
passaro, line 845: Should we also provide the parent key?

mountpoint-s3/src/inode.rs
```diff
@@ -1239,7 +1243,7 @@ impl Inode {
     fn get_inode_state(&self) -> Result<RwLockReadGuard<InodeState>, InodeError> {
         let inode_state = self.inner.sync.read().unwrap();
         match &inode_state.kind_data {
-            InodeKindData::Directory { deleted, .. } if *deleted => Err(InodeError::InodeDoesNotExist(self.ino())),
+            InodeKindData::Directory { deleted, .. } if *deleted => Err(InodeError::DirectoryDoesNotExist(self.err())),
```
passaro, line 1246: I'm not too convinced about this change. We would return a different error depending on whether the inode was actually removed or just flagged with `deleted`. So far, this has been an implementation detail.

mountpoint-s3/src/inode.rs
```diff
@@ -1485,8 +1489,10 @@ impl InodeStat {
 pub enum InodeError {
     #[error("error from ObjectClient")]
     ClientError(#[source] anyhow::Error),
-    #[error("file does not exist")]
-    FileDoesNotExist,
+    #[error("file does not exist {0:?}")]
+    FileDoesNotExist(OsString),
+    #[error("directory does not exist at inode {0}")]
```
passaro, line 1494: See previous comment. But also the error message may be confusing: we know the given `ino` corresponds to the directory in the info, but are also saying that it does not exist.

passaro (commented):
I like the log entries with the field added to the span. Just a little concerned it may be difficult to enforce that we always fill it consistently.

sauraank:
> I like the log entries with the field added to the span. Just a little concerned it may be difficult to enforce that we always fill it consistently.

Yes, I am filling it at the first occurence of the filename in its call stack. So, any error logged before that wont have filename filled for it.

mountpoint-s3/src/fuse.rs
sauraank: Because, in the issue it was mentioned to add filename. So, I thought of adding filename wherever it is possible.

mountpoint-s3/src/inode.rs
sauraank: I thought to do is to cover as much case as possible to log filename. But you are right, deleted is only supposed to be implementation detail.

mountpoint-s3/src/inode.rs
sauraank: Yes, I should change it to only log Directory name if we distinguish this case from normal deleted inode case.
But, I am going ahead with removing this error case as you mentioned in the previous comment.

mountpoint-s3/src/inode.rs
sauraank: I am not sure because in half of the only parent InodeNo. is present. And I dont think that would give much information.

mountpoint-s3/src/fuse.rs
passaro: Yes, you are right that this should be from the filesystem point of view. So ideally we could have the "full" path (from the mount point), but okay to stick to just the name for consistency and even more for brevity (the span is getting a bit busy).

mountpoint-s3/src/inode.rs
```diff
@@ -1485,8 +1490,8 @@ impl InodeStat {
 pub enum InodeError {
     #[error("error from ObjectClient")]
     ClientError(#[source] anyhow::Error),
-    #[error("file does not exist")]
-    FileDoesNotExist,
+    #[error("file does not exist {0:?} in parent {1}")]
```
passaro, line 1493: Maybe:
```suggestion
    #[error("file {0:?} does not exist in parent inode {1}")]
```

passaro (approved):

Issue #550, opened 2023-10-11
mount-s3 --no-log flag still sends log to the console


miguelmachacon:
### Mountpoint for Amazon S3 version

mount-s3 1.0.2

### AWS Region

eu-central-1

### Describe the running environment

Running in Amazon EC2 with Amazon Linux 2023 connected to the same S3 account. 

![Screenshot 2023-10-10 at 16 30 14](https://github.com/awslabs/mountpoint-s3/assets/147605870/1f55da83-6bec-4ca5-8c1e-fbdd84f84f5d)

### What happened?

I tried to mount the s3 without any log and 'bucket incomingmedia-migchag is mounted at /media/s3' appeared in the console even if launched it with --no-log flag enabled

### Relevant log output

```shell
'bucket incomingmedia-migchag is mounted at /media/s3'
```

arsh:
Hello @miguelmachacon, did this behavior cause any issues or was it just a surprise?

We intentionally want to print to stdout after a successful mount as a way to provide a feedback to the user and don’t think `--no-log` option should prevent it. To prevent anything being printed to the stdout, redirects can be used. For example 

```
mount-s3 <BUCKET_NAME> ~/mnt 1>/dev/null
```

dannycjones:
Please let us know if there's still an issue here, I'll close this issue for now.

PR #666, opened 2023-12-05
Replace println! with tracing::debug! to honor --no-log flag.

arsh:
## Description of change

Replace `println!` with `tracing::debug!` to honor `--no-log` flag.

Relevant issues: [550](https://github.com/awslabs/mountpoint-s3/issues/550)

When using `--no-log` no messages would be printed.

---

### DIFF
--- a/mountpoint-s3/src/main.rs
+++ b/mountpoint-s3/src/main.rs
@@ -447,7 +447,7 @@ fn main() -> anyhow::Result<()> {
                 let status = receiver.recv_timeout(timeout);
                 match status {
                     Ok('0') => {
-                        println!(
+                        tracing::debug!(
                             "{} is mounted at {}",
                             args.bucket_description(),
                             args.mount_point.display()


mountpoint-s3/src/main.rs
```diff
@@ -447,7 +447,7 @@ fn main() -> anyhow::Result<()> {
                 let status = receiver.recv_timeout(timeout);
                 match status {
                     Ok('0') => {
-                        println!(
```
arsh, line 450: I could have deleted this line instead but decided to keep it as it could bring more clarity when debugging an issue. Happy to remove if we think that is better.

dannycjones: We actually do want to log this explicitly to stdout for positive confirmation to the user when the mount completes.

Instead, we can probably gate this `println!` behind an if condition based on if `--no-log` was provided.

passaro: Do we really want to do that? I don't think `--no-log` should affect stdout. Unless we are retrofitting it to also mean something like `--quiet`.

dannycjones: Good point. Either we make the change or we can close out that bug report.

arsh: I agree with @passaro that this isn't a `--quiet` option. I will discard this PR and work on closing that bug.

arsh:
This is behaving as intended and we have no plan to fix it.

Issue #651, opened 2023-11-29
S3 Express One Zone documentation updates


dlakhaws:
### Tell us more about this new feature.

The [permission recommendation](https://github.com/awslabs/mountpoint-s3/blob/099638eede5fc34756e46b9eb6af6582c8fe1a55/doc/CONFIGURATION.md#iam-permissions) for Mountpoint weren't updated to include [S3 Express One Zone IAM permissions](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-express-security-iam-identity-policies.html). Would like to see a section with this added for customers using S3 Express One Zone!

jamesbornholt:
There's a few places in the docs we need to update to include Express One Zone:
* IAM permissions (`s3express:CreateSession`)
* Storage classes (don't use `--storage-class EXPRESS_ONEZONE`, instead just specify a directory bucket)
* Region detection (not sure this works as described for Express)

We should probably also update the semantics documentation to note that `readdir` is unordered on an S3 Express One Zone bucket.

PR #667, opened 2023-12-05
Update documentation for S3 Express One Zone

monthonk:
## Description of change

Update the configuration and semantics documents to address differences on using Mountpoint with S3 Express One Zone buckets.

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/651

No, only document updates.

---

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -79,7 +79,7 @@ Here is an example least-privilege policy document to add to an IAM user or role
 }
 ```
 
-Directory bucket, or S3 Express One Zone storage class in other words, has different authentication mechanism from general purpose buckets. The policy `s3:*` doesn't apply to directory buckets, you will need `s3express:CreateSession` policy in order to access them. Here is an example of least-privilege policy document.
+Directory buckets, introduced with the S3 Express One Zone storage class, use a different authentication mechanism from general purpose buckets. Instead of using `s3:*` actions, you should allow the `s3express:CreateSession` action. Here is an example of least-privilege policy document.
 
 ```
 {
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -79,6 +79,21 @@ Here is an example least-privilege policy document to add to an IAM user or role
 }
 ```
 
+Directory bucket, or S3 Express One Zone storage class in other words, has different authentication mechanism from general purpose buckets. The policy `s3:*` doesn't apply to directory buckets, you will need `s3express:CreateSession` policy in order to access them. Here is an example of least-privilege policy document.
+
+```
+{
+	"Version": "2012-10-17",
+	"Statement": [
+        {
+            "Effect": "Allow",
+            "Action": "s3express:CreateSession",
+            "Resource": "arn:aws:s3express:REGION:ACCOUNT-ID:bucket/DOC-EXAMPLE-BUCKET--az_id--x-s3"
+        }
+    ]
+}
+```
+
 Mountpoint also respects access control lists (ACLs) applied to objects in your S3 bucket, but does not allow you to automatically attach ACLs to objects created with Mountpoint. A majority of modern use cases in Amazon S3 no longer require the use of ACLs. We recommend that you keep ACLs disabled for your S3 bucket, and instead use bucket policies to control access to your objects.
 
 ## S3 bucket configuration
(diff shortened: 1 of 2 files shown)


doc/CONFIGURATION.md
```diff
@@ -175,6 +190,10 @@ Amazon S3 offers a [range of storage classes](https://aws.amazon.com/s3/storage-
 * `GLACIER` for [S3 Glacier Flexible Retrieval](https://aws.amazon.com/s3/storage-classes/glacier/)
 * `DEEP_ARCHIVE` for [S3 Glacier Deep Archive](https://aws.amazon.com/s3/storage-classes/glacier/)
 
+> [!IMPORTANT]
+> Do not set the storage class to `EXPRESS_ONEZONE` as it is a distinct storage class and cannot be set for general purpose
+buckets. If you want to use S3 Express One Zone storage class, just specify a directory bucket name when mounting.
```
arsh, line 195: Is this something that could be enforced by code instead of relying on this documentation?

monthonk: This value is still valid for directory buckets, but it's the same as not setting storage class at all. So, maybe we can just don't allow people to set `--storage-class` to `EXPRESS_ONEZONE`.

monthonk: Users also cannot set storage class other than `EXPRESS_ONEZONE` for directory buckets. I think the right time to do the validation is after we have resolved the endpoint and know whether this is a directory bucket or not.

dannycjones: FWIW, I'd caution against setting limits like this that would normally be enforced service-side in S3 but could feasibly be relaxed in the future. In adding this restriction to the client, users are forced to upgrade version to adopt new features (like a new storage class for directory buckets).

Maybe forcing updates has its upsides from maintainers' point of view, but I'd lean against it for now. 😄

dannycjones: What's important is that the user gets clear feedback into what went wrong. If the wrong storage class is set, is the error clear in the Mountpoint logs? Of course we're always limited by the error responses we can send via filesystem APIs.

doc/CONFIGURATION.md
```diff
@@ -79,6 +79,21 @@ Here is an example least-privilege policy document to add to an IAM user or role
 }
 ```
 
+Directory bucket, or S3 Express One Zone storage class in other words, has different authentication mechanism from general purpose buckets. The policy `s3:*` doesn't apply to directory buckets, you will need `s3express:CreateSession` policy in order to access them. Here is an example of least-privilege policy document.
```
dannycjones, line 82: Maybe we can tweak this wording a bit.

```suggestion
Directory buckets, introduced with the S3 Express One Zone storage class, use a different authentication mechanism from general purpose buckets. Instead of using `s3:*` actions, you should allow the `s3express:CreateSession` action. Here is an example of least-privilege policy document.
```

doc/CONFIGURATION.md
```diff
@@ -93,7 +108,7 @@ When constructing the directory structure for your mount, Mountpoint removes the
 
 ### Region detection
 
-Amazon S3 buckets are associated with a single AWS Region. Mountpoint attempts to automatically detect the region for your S3 bucket at startup time and directs all S3 requests to that region. However, in some scenarios this region detection may fail, preventing your bucket from being mounted and displaying Access Denied or No Such Bucket errors. You can override Mountpoint's automatic bucket region detection with the `--region` command-line argument or `AWS_REGION` environment variable.
+Amazon S3 buckets are associated with a single AWS Region. Mountpoint attempts to automatically detect the region for your S3 bucket at startup time and directs all S3 requests to that region. However, in some scenarios like cross-region mount with a directory bucket, this region detection may fail, preventing your bucket from being mounted and displaying Access Denied or No Such Bucket errors. You can override Mountpoint's automatic bucket region detection with the `--region` command-line argument or `AWS_REGION` environment variable.
```
dannycjones, line 111: @monthonk, in what cases would this fail? We should detect the correct region and endpoint based on the name of the directory bucket, right?

dannycjones (changes requested):
Thanks Monthon, just a bit more feedback. Most is inline.

On storage classes, can we update the main body to be a bit clearer on default behavior. Mountpoint doesn't specify a storage class and respects the default from S3 - which is `STANDARD` for general purpose buckets or `EXPRESS_ONEZONE` for directory buckets.

```
- By default, Mountpoint uses the S3 Standard storage class, which is appropriate for a wide variety of use cases.
+ Mountpoint respects the default storage class from S3 unless otherwise configured, which is appropriate for a wide variety of use cases.
```

dannycjones (approved):
LGTM, thanks @monthonk.

PR #668, opened 2023-12-05
Print successful mount for both foreground and background modes. Also update documentation for the `--no-log` argument.

arsh:
## Description of change

Print successful mount for both foreground and background modes. Also update documentation for the `--no-log` argument.

Relevant issues: N/A

Yes. When running on foreground mode, we print a message to stdout after a successful mount. As noted in the documentation, the `--no-log` argument has no effects on this or any messages sent to stdout.

---

### DIFF
--- a/mountpoint-s3/src/main.rs
+++ b/mountpoint-s3/src/main.rs
@@ -218,7 +218,7 @@ struct CliArgs {
 
     #[clap(
         long,
-        help = "Disable all logging",
+        help = "Disable all logging. You will still see stdout messages.",
         help_heading = LOGGING_OPTIONS_HEADER,
         conflicts_with_all(["log_directory", "debug", "debug_crt", "log_metrics"])
     )]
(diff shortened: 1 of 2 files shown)


mountpoint-s3/src/main.rs
```diff
@@ -218,7 +218,7 @@ struct CliArgs {
 
     #[clap(
         long,
-        help = "Disable all logging",
+        help = "Disable all logging excluding standard output. If no output is desired, consider redirecting it.",
```
passaro, line 221: Could this be confusing when using `--foreground` (which logs to stdout)?

arsh: Yes, it could. This is because the behavior of this argument is different depending on whether it is running in the foreground or background.

What do you think of this:

> Disable all logging if running in the foreground. When running in the background, this argument do not prevent messages sent to stdout from being displayed. If no output is desired, consider redirecting it.

I'll have to update the MD file too.

jamesbornholt:
This seems fine. Though I kind of liked the idea of `--no-log` implying "no output at all" even if, pedantically, this message isn't a "log message".

jamesbornholt (approved):

passaro (approved):

PR #669, opened 2023-12-06
Add CRT error to CRT request logs, move CRT per-request logs to DEBUG

dannycjones:
## Description of change

At the moment, errors for individual requests made by the CRT client (inside of a meta request) can fail and we log that they failed, but only with the HTTP Status (or lack thereof as "-1").

For example, these three requests failed for different reasons.

```
WARNING request failed request_type=GetObject http_status=-1 range=None duration=3.048089489s ttfb=None request_id=<unknown>
WARNING request failed request_type=GetObject http_status=500 range=None duration=2.968672313s ttfb=Some(2.713102782s) request_id=0H2C1MFGVVR9ARMV
WARNING request failed request_type=GetObject http_status=-1 range=None duration=290.737317ms ttfb=None request_id=<unknown>
```

We have access to a CRT error code here, so we should surface that. These also log on success, so I've wrapped it in an `Option<>`.

I've also updated the message itself to be "CRT request" to try and highlight that these are inner requests made by the CRT that can be retried or be part of a much larger _meta_ request.

An error message may look like this now:

```
WARNING CRT request failed request_type=GetObject crt_error=Some(Error(1048, "aws-c-io: AWS_IO_SOCKET_TIMEOUT, socket operation timed out.")) http_status=-1 range=None duration=3.126663123s ttfb=None request_id=<unknown>
```

I almost wonder if we should be logging these as warning, as it has caused some confusion around what has actually happened and if the data was really retrieved or the application failed to handle that error. For now, I'll leave them as warning.

Relevant issues: N/A

No behavior change, logs contain new information. Message changes slightly.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -449,6 +449,7 @@ impl S3CrtClientInner {
 
                 let http_status = metrics.status_code().unwrap_or(-1);
                 let request_failure = !(200..299).contains(&http_status);
+                let crt_error = Some(metrics.error()).filter(|e| e.is_err());
                 let request_type = request_type_to_metrics_string(metrics.request_type());
                 let request_id = metrics.request_id().unwrap_or_else(|| "<unknown>".into());
                 let duration = metrics.total_duration();
@@ -458,12 +459,12 @@ impl S3CrtClientInner {
                 let log_level = status_code_to_log_level(http_status);
 
                 let message = if request_failure {
-                    "request failed"
+                    "CRT request failed"
                 } else {
-                    "request finished"
+                    "CRT request finished"
                 };
-                event!(log_level, %request_type, http_status, ?range, ?duration, ?ttfb, %request_id, "{}", message);
-                trace!(detailed_metrics=?metrics, "request completed");
+                event!(log_level, %request_type, ?crt_error, http_status, ?range, ?duration, ?ttfb, %request_id, "{}", message);
+                trace!(detailed_metrics=?metrics, "CRT request completed");
 
                 let op = span_telemetry.metadata().map(|m| m.name()).unwrap_or("unknown");
                 if let Some(ttfb) = ttfb {
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -447,8 +447,8 @@ impl S3CrtClientInner {
             .on_telemetry(move |metrics| {
                 let _guard = span_telemetry.enter();
 
-                let http_status = metrics.status_code().unwrap_or(-1);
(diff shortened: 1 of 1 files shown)


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -449,6 +449,7 @@ impl S3CrtClientInner {
 
                 let http_status = metrics.status_code().unwrap_or(-1);
```
passaro, line 450: Should we leave this as Option, rather than introduce the confusing `-1` status?

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -458,12 +459,12 @@ impl S3CrtClientInner {
                 let log_level = status_code_to_log_level(http_status);
```
passaro, line 459: Does this make sense for "CRT requests"? Maybe we should log all of them at DEBUG, regardless of the status. Failures at the meta-request level will be logged at WARN anyway in `on_finish`.

passaro: At the moment this also logs `http_status==-1` at WARN.

passaro (commented):
The change looks good. We could go a bit further, but happy to do that in a separate PR if you prefer.

mountpoint-s3-client/src/s3_crt_client.rs
dannycjones: Yeah, I think that makes sense. I had the same thoughts, wasn't sure whether to leave it as is (since `-1` can have its own convention outside of Rust) but maybe not worth it.

I'm thinking we switch to debug anyway, so let's just go with consistency at this point. I'll change to `None`.

dannycjones: I'd still be keeping it as `-1` in the failure metric at the bottom for now.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -447,23 +447,22 @@ impl S3CrtClientInner {
...
-                    "request failed"
+                    "CRT request failed"
                 } else {
-                    "request finished"
+                    "CRT request finished"
                 };
-                event!(log_level, %request_type, http_status, ?range, ?duration, ?ttfb, %request_id, "{}", message);
-                trace!(detailed_metrics=?metrics, "request completed");
+                debug!(%request_type, ?crt_error, http_status, ?range, ?duration, ?ttfb, %request_id, "{}", message);
+                trace!(detailed_metrics=?metrics, "CRT request completed");
```
jamesbornholt, line 465: it's just bikeshedding but i don't think "CRT request" is the right way to describe these. it's an S3 request that failed. if we're going to push these down to debug level, i don't see the need to also write "CRT" here.

jamesbornholt (approved):

PR #673, opened 2023-12-11
Introduce ObjectId type to reference specific versions of S3 objects

passaro:
## Description of change

The new `ObjectId` type holds an S3 object key and etag to identify a specific version of an object. It is used in the data cache, where it is a 1-1 replacement for `CacheKey`, and also in the prefetcher, where it replaces the S3 key previously used in `Part`.

No.

---

### DIFF
--- a/mountpoint-s3/src/prefetch/caching_stream.rs
+++ b/mountpoint-s3/src/prefetch/caching_stream.rs
@@ -8,7 +8,8 @@ use mountpoint_s3_client::{types::ETag, ObjectClient};
 use tracing::{debug_span, trace, warn, Instrument};
 
 use crate::checksums::ChecksummedBytes;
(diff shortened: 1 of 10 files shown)


passaro:
This PR splits `ObjectId` off #644.

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -81,7 +82,7 @@ struct CachingRequest<Client: ObjectClient, Cache> {
     client: Client,
     cache: Arc<Cache>,
     bucket: String,
-    cache_key: CacheKey,
+    cache_key: ObjectId,
```
dannycjones, line 85: it's likely worth renaming this argument to match, but not blocking on this

Issue #516, opened 2023-09-14
Amazon S3 on Outposts support for write workloads


knichwitz:
### Tell us more about this new feature.

I would like to have the possibility to use S3 Mountpoint on AWS Outpost.

This would help to reduce costs for EBS on configuration Outpost Racks.

EBS is up to 3x more expensive than S3 Storage on AWS Outpost. Running Machine Learning use cases at scale with S3 Mountpoint, would support many users in configuring cost efficient AWS Outpost Racks

ahmarsuhail:
Hey, thanks for opening this issue. As of `v1.0.1`, Mountpoint should work with outposts in read only mode. You can enable this by passing in  `--read-only` during mount time. Currently Mountpoint does not work with Outposts for writes as Outposts does not support additional checksums. We plan to enable write support for Outposts and will share here when we have an update.

jamesbornholt:
Writes to Outposts buckets should work in the next Mountpoint release!

MrFishFinger:
hi @jamesbornholt - amazing, love your work!

`--upload-checksums=off` also seems to fix writing to a B2 backblaze bucket:

```
❯ mount-s3 --endpoint-url=https://s3.eu-central-003.backblazeb2.com --upload-checksums=off  test-b2-bucket /mnt/s3/test-b2-bucket
bucket test-b2-bucket is mounted at /mnt/s3/test-b2-bucket

❯ touch /mnt/s3/test-b2-bucket/test2.txt

❯ ls -lh /mnt/s3/test-b2-bucket/
insgesamt 0
-rw-r--r-- 1 user user 0 Mai 24 21:34 test1.txt
-rw-r--r-- 1 user user 0 Mai 24 21:36 test2.txt
```

(i used an [unofficial deb package](https://github.com/awslabs/mountpoint-s3/actions/runs/9189184243/artifacts/1526387350) built in a recent `main` [CI job](https://github.com/awslabs/mountpoint-s3/actions/runs/9189184243/job/25270674663))

metadaddy:
I just verified that [Mountpoint for Amazon S3 version 1.7](https://github.com/awslabs/mountpoint-s3/releases/tag/mountpoint-s3-1.7.0) works with B2:

```console
# mount-s3 --endpoint-url=https://s3.us-west-004.backblazeb2.com/ --upload-checksums=off metadaddy-mountpoint ./tester
bucket metadaddy-mountpoint is mounted at ./tester
[...]

PR #677, opened 2023-12-14
Add mount option to disable additional checksum for S3 Outposts write support

sauraank:
## Description of change
Added the disabling of checksums based on signing name as "s3-outposts".
Commented out the mount option of Disabling additional checksums to support write in Outposts since the decision is not yet made. Users of Outposts will have to pass `--disable-checksum` to support write, if we go this path.
This is draft PR, since we have not finalised our decision on the way to do it. But, whatever decision we take we will need to make similar changes in `upload.rs` module. 

Passed the trailing checksum boolean to S3FileSystemConfig and then to uploader. Using this config, set the trailing checksum as false  and bypassing the checksum verification for the case above.
Manually tested it works with write on Outposts. I will need to add tests.

Relevant issues: 
#516 

Yes, it disables additional checksums in case of writing on Outpost bucket (in this way, set by user). 

---

### DIFF
--- a/mountpoint-s3/src/upload.rs
+++ b/mountpoint-s3/src/upload.rs
@@ -206,8 +206,10 @@ mod tests {
     };
     use test_case::test_case;
 
+    #[test_case(true; "trailing_checksum_on")]
+    #[test_case(false; "trailing_checksum_off")]
     #[tokio::test]
-    async fn complete_test() {
+    async fn complete_test(trailing_checksum: bool) {
         let bucket = "bucket";
         let name = "hello";
         let key = name;
@@ -217,7 +219,7 @@ mod tests {
             part_size: 32,
             ..Default::default()
         }));
-        let uploader = Uploader::new(client.clone(), None, true);
+        let uploader = Uploader::new(client.clone(), None, trailing_checksum);
         let request = uploader.put(bucket, key).await.unwrap();
 
         assert!(!client.contains_key(key));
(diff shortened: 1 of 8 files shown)


mountpoint-s3/src/upload.rs
```diff
@@ -64,6 +69,7 @@ pub struct UploadRequest<Client: ObjectClient> {
     key: String,
     next_request_offset: u64,
     hasher: Hasher,
+    set_checksum: bool,
```
passaro, line 72: Use consistent name here

mountpoint-s3/src/upload.rs
```diff
@@ -88,6 +94,7 @@ impl<Client: ObjectClient> UploadRequest<Client> {
             key: key.to_owned(),
             next_request_offset: 0,
             hasher: Hasher::new(),
```
passaro, line 96: The `trailing_checksums` flag should also affect hashing: if disabled, we don't need the `Hasher`, or at least we don't need to calculate the checksum in `write`.

mountpoint-s3/src/upload.rs
```diff
@@ -125,7 +132,7 @@ impl<Client: ObjectClient> UploadRequest<Client> {
         let size = self.size();
         let checksum = self.hasher.finalize();
         self.request
-            .review_and_complete(move |review| verify_checksums(review, size, checksum))
+            .review_and_complete(move |review| verify_checksums(review, size, checksum, self.set_checksum))
```
passaro, line 135: You could skip the review completely and just call `PutObjectRequest::complete()` here.

passaro: Alternatively, you could still use a review to verify just the size. In either cases, you do not want to calculate the checksum if not using it.

passaro (changes requested):
While we discuss whether this is the right approach, I've commented on a couple of implementation details.

chrsmia:
I believe this also will be required for support with s3-snow compatible storage on Snowball Edge.  So there's greater need than just for Outposts.
SHA checksums are not matching causing write activities to fail.

jamesbornholt:
Moved to #849.

PR #687, opened 2024-01-02
Rework ChecksummedBytes internals to use a Range instead of a Bytes slice

passaro:
## Description of change

Preliminary refactor to prepare for adding integrity checks on the range itself. No changes in behavior.

No.

---

### DIFF
--- a/mountpoint-s3/src/checksums.rs
+++ b/mountpoint-s3/src/checksums.rs
@@ -14,9 +14,9 @@ use thiserror::Error;
 pub struct ChecksummedBytes {
     /// Underlying buffer
     buffer: Bytes,
-    /// Range over `buffer`
+    /// Range over [Self::buffer]
     range: Range<usize>,
-    /// Checksum for `orig_bytes`
+    /// Checksum for [Self::buffer]
     checksum: Crc32c,
 }
 
--- a/mountpoint-s3/src/checksums.rs
+++ b/mountpoint-s3/src/checksums.rs
@@ -1,4 +1,4 @@
-use std::ops::RangeBounds;
+use std::ops::{Bound, Range, RangeBounds};
 
 use bytes::{Bytes, BytesMut};
 use mountpoint_s3_crt::checksums::crc32c::{self, Crc32c};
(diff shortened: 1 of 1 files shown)


mountpoint-s3/src/checksums.rs
```diff
@@ -275,26 +316,69 @@ mod tests {
...
+    fn slice(original: Range<usize>, range: impl RangeBounds<usize>) -> ChecksummedBytes {
+        let buffer = Bytes::copy_from_slice(&vec![0; original.len()]);
+        let checksum = crc32c::checksum(&buffer);
+        let bytes = ChecksummedBytes {
+            buffer,
+            range: original,
+            checksum,
+        };
+        bytes.slice(range)
+    }
```
dannycjones, line 336: I'd prefer we just include this setup inline in the test. We can simplify the code a bit too:

```rust
let buffer = Bytes::from_static([0; original.len()]);
let bytes = ChecksummedBytes::from_bytes(buffer);
let slice = bytes.slice(range);
```

dannycjones: better yet, we could just add (maybe test only) `ChecksummedBytes::from_static`? test setup / 'boilerplate' becomes:

```rust
let bytes = ChecksummedBytes::from_static([0; original.len()]);
let slice = bytes.slice(range);
```

dannycjones (changes requested):
about there, just two minor things

mountpoint-s3/src/checksums.rs
passaro: Your suggestion is missing the `original` range, but you are definitely right about the `slice` helper function not being clear! 
In fact, I think the issue is that it is both setting up the instance to test and invoking the `slice` method on it. I've fixed that by using the helper function just for the setup and calling `slice` in each test function. I hope that it's clearer now.

dannycjones: thanks, yep clearer!

dannycjones (approved):
lgtm!

Issue #670, opened 2023-12-07
Mountpoint_s3::fuse: release failed


djhenry2:
### Mountpoint for Amazon S3 version

mount-s3 1.1.1

### AWS Region

us-east-2

### Describe the running environment

Ubuntu 22.04 mounting S3 bucket in the same account.

### What happened?

Many errors with "Mountpoint_s3::fuse: release failed"

Mount command:
mount-s3 --allow-root  --read-only --log-directory /var/log/mountpoint-s3/ sftpgw-i-***** wn-s3   (bucket name hidden)

### Relevant log output

[mountpoint-s3-2023-12-07T13-52-09Z.log](https://github.com/awslabs/mountpoint-s3/files/13601607/mountpoint-s3-2023-12-07T13-52-09Z.log)

dannycjones:
Thanks for reporting this. I see its similar to the issue in #582. Thanks for sharing the logs, that should help with a reproduction I hope.

@djhenry2 Can you share any impact to your workload? Is it causing issues or currently just warnings in the logs (which we should fix).

dannycjones:
We released a possible fix for this in v1.3.2. @djhenry2, can you give it a try and check its solved for your use case?

The fix is to remove a race condition related to file handle de-allocation (https://github.com/awslabs/mountpoint-s3/commit/6e7252dd2e54932e277e5b5ee7000f9bc816a682). There was a chance that before file handles references were released from an operation like `read` or `fsync`, the `release` request would arrive and be unable to de-allocate the file handle. This meant that the file handle would be left in the file handle table, and could lead to increased memory usage since the file handles hold on to prefetched file content.

sauraank:
I am closing this issue due to inactivity. If you any further information or question, please re-open this issue or create a new issue. Thanks.

PR #691, opened 2024-01-05
Replace callback in FS read with simple Result instead

dannycjones:
## Description of change

This chance simplifies `fs::read` by returning a result rather than invoking a callback to `fuse::read`.

We're making this change primarily due to the risk of a race condition introduced. Before this change, we reply directly to the FUSE driver via a callback (`ReadReplier`) before exiting the FS module code. The risk here is that we've already replied to the driver before we drop things like the file handle guard. Albeit small, this is a race condition and we intend to remove it to avoid any risk from it.

This race condition is suspected to be the root cause for this issue where FUSE release fails unable to unwrap the file handle reference: https://github.com/awslabs/mountpoint-s3/issues/670. I haven't yet reproduced that issue but the code before this change did not appear correct.

I also reviewed other usages of file handles to ensure that they are dropped at the appropriate time: `open`, `write`, `fsync`, `flush`, and `release` all look OK and should drop the reference before returning the response to the FUSE driver. Note, we may have a similar behavior in the directory handles which I'll review separately.

When triggered, this race condition could have a large impact since the file handle holds a reference to prefetched data and this would never be freed as file handles are put back in the file handle table when `release` fails.

Relevant issues:
- #670

Fixes possible bug where `read` completes early / asynchronously.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -1,5 +1,6 @@
 //! FUSE file system types and operations, not tied to the _fuser_ library bindings.
 
+use bytes::Bytes;
 use nix::unistd::{getgid, getuid};
 use std::collections::HashMap;
 use std::ffi::{OsStr, OsString};
@@ -642,7 +643,7 @@ where
     }
 
     #[allow(clippy::too_many_arguments)] // We don't get to choose this interface
-    pub async fn read<R: ReadReplier>(
+    pub async fn read(
         &self,
         ino: InodeNo,
         fh: u64,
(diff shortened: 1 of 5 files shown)


mountpoint-s3/src/fs.rs
```diff
@@ -642,16 +643,15 @@ where
     }
 
     #[allow(clippy::too_many_arguments)] // We don't get to choose this interface
-    pub async fn read<R: ReadReplier>(
```
jamesbornholt, line 645: can delete the `ReadReplier` trait now?

dannycjones: true! not sure why it didn't warn about it being unused, dropping it now

jamesbornholt (approved):

PR #696, opened 2024-01-10
Introduce negative metadata cache entries

passaro:
## Description of change

Reduce latency when repeatedly looking up non-existing files or directories (when cache is enabled). 

This change adds negative metadata cache entries: whenever a lookup fails because an object does not exist, we cache a “negative” entry with the same TTL as for successful lookups and use it to reply to subsequent kernel requests for the same name.

The negative entries are maintained separately from the inode tree using the new `NegativeCache` type, which enforces an upper limit to the number of entries and handles their expiration.

Repeated lookups for non-existing entries will not trigger S3 requests for the configured TTL.

---

### DIFF
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -26,7 +26,7 @@ use std::ffi::{OsStr, OsString};
 use std::fmt::{Debug, Display};
 use std::os::unix::prelude::OsStrExt;
 use std::sync::atomic::AtomicBool;
-use std::time::{Duration, Instant, SystemTime};
+use std::time::{Duration, SystemTime};
 
 use anyhow::anyhow;
 use fuser::FileType;
@@ -46,6 +46,9 @@ use crate::sync::atomic::{AtomicU64, Ordering};
 use crate::sync::RwLockReadGuard;
 use crate::sync::RwLockWriteGuard;
 use crate::sync::{Arc, RwLock};
+
+mod expiry;
+use expiry::Expiry;
 
 mod readdir;
 pub use readdir::ReaddirHandle;
@@ -1058,7 +1061,7 @@ pub struct LookedUp {
 impl LookedUp {
     /// How much longer this lookup will be valid for
     pub fn validity(&self) -> Duration {
-        self.stat.expiry.saturating_duration_since(Instant::now())
+        self.stat.expiry.validity()
     }
 }
 
@@ -1446,7 +1449,7 @@ impl InodeKindData {
 #[derive(Debug, Clone)]
 pub struct InodeStat {
     /// Time this stat becomes invalid and needs to be refreshed
-    expiry: Instant,
+    expiry: Expiry,
 
     /// Size in bytes
     pub size: usize,
@@ -1478,7 +1481,7 @@ pub enum WriteStatus {
 
 impl InodeStat {
     fn is_valid(&self) -> bool {
-        self.expiry >= Instant::now()
+        self.expiry.is_valid()
     }
 
     /// Objects in flexible retrieval storage classes can't be accessed via GetObject unless they are
@@ -1511,12 +1514,9 @@ impl InodeStat {
         restore_status: Option<RestoreStatus>,
         validity: Duration,
     ) -> InodeStat {
-        let expiry = Instant::now()
-            .checked_add(validity)
-            .expect("64-bit time shouldn't overflow");
         let is_readable = Self::is_readable(storage_class, restore_status);
         InodeStat {
-            expiry,
+            expiry: Expiry::from_now(validity),
             size,
             atime: datetime,
             ctime: datetime,
@@ -1528,11 +1528,8 @@ impl InodeStat {
 
     /// Initialize an [InodeStat] for a directory, given some metadata.
     fn for_directory(datetime: OffsetDateTime, validity: Duration) -> InodeStat {
-        let expiry = Instant::now()
-            .checked_add(validity)
-            .expect("64-bit time shouldn't overflow");
         InodeStat {
-            expiry,
+            expiry: Expiry::from_now(validity),
             size: 0,
             atime: datetime,
             ctime: datetime,
@@ -1543,9 +1540,7 @@ impl InodeStat {
     }
 
     fn update_validity(&mut self, validity: Duration) {
-        self.expiry = Instant::now()
-            .checked_add(validity)
-            .expect("64-bit time shouldn't overflow");
+        self.expiry = Expiry::from_now(validity);
     }
 }
 
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -50,6 +50,9 @@ use crate::sync::{Arc, RwLock};
 mod expiry;
 use expiry::Expiry;
 
+mod negative_cache;
+use negative_cache::NegativeCache;
+
 mod readdir;
 pub use readdir::ReaddirHandle;
 
@@ -83,6 +86,7 @@ pub struct Superblock {
 struct SuperblockInner {
     bucket: String,
     inodes: RwLock<HashMap<InodeNo, Inode>>,
+    negative_cache: NegativeCache,
     next_ino: AtomicU64,
     mount_time: OffsetDateTime,
     config: SuperblockConfig,
@@ -120,9 +124,12 @@ impl Superblock {
         let mut inodes = HashMap::new();
         inodes.insert(ROOT_INODE_NO, root);
 
+        let negative_cache = NegativeCache::new(config.cache_config.negative_cache_size, config.cache_config.file_ttl);
+
         let inner = SuperblockInner {
             bucket: bucket.to_owned(),
             inodes: RwLock::new(inodes),
+            negative_cache,
             next_ino: AtomicU64::new(2),
             mount_time,
             config,
@@ -614,7 +621,7 @@ impl SuperblockInner {
         };
 
         let lookup = match lookup {
-            Some(lookup) => lookup,
+            Some(lookup) => lookup?,
             None => {
                 let remote = self.remote_lookup(client, parent_ino, name).await?;
                 self.update_from_remote(parent_ino, name, remote)?
@@ -627,22 +634,33 @@ impl SuperblockInner {
 
     /// Lookup an [Inode] against known directory entries in the parent,
     /// verifying any returned entry has not expired.
-    fn cache_lookup(&self, parent_ino: InodeNo, name: &str) -> Option<LookedUp> {
-        fn do_cache_lookup(parent: Inode, name: &str) -> Option<LookedUp> {
+    /// If no record for the given `name` is found, returns [None].
+    /// If an entry is found in the negative cache, returns [Some(Err(InodeError::FileDoesNotExist))].
+    fn cache_lookup(&self, parent_ino: InodeNo, name: &str) -> Option<Result<LookedUp, InodeError>> {
+        fn do_cache_lookup(
+            superblock: &SuperblockInner,
+            parent: Inode,
+            name: &str,
+        ) -> Option<Result<LookedUp, InodeError>> {
             match &parent.get_inode_state().ok()?.kind_data {
                 InodeKindData::File { .. } => unreachable!("parent should be a directory!"),
                 InodeKindData::Directory { children, .. } => {
-                    let inode = children.get(name)?;
-                    let inode_stat = &inode.get_inode_state().ok()?.stat;
-                    if inode_stat.is_valid() {
-                        let lookup = LookedUp {
-                            inode: inode.clone(),
-                            stat: inode_stat.clone(),
-                        };
-                        return Some(lookup);
+                    if let Some(inode) = children.get(name) {
+                        let inode_stat = &inode.get_inode_state().ok()?.stat;
+                        if inode_stat.is_valid() {
+                            let lookup = LookedUp {
+                                inode: inode.clone(),
+                                stat: inode_stat.clone(),
+                            };
+                            return Some(Ok(lookup));
+                        }
                     }
                 }
             };
+
+            if cfg!(feature = "negative_cache") && superblock.negative_cache.contains(parent.ino(), name) {
+                return Some(Err(InodeError::FileDoesNotExist(name.to_owned(), parent.err())));
+            }
 
             None
         }
@@ -650,7 +668,7 @@ impl SuperblockInner {
         let lookup = self
             .get(parent_ino)
             .ok()
-            .and_then(|parent| do_cache_lookup(parent, name));
+            .and_then(|parent| do_cache_lookup(self, parent, name));
 
         match &lookup {
             Some(lookup) => trace!("lookup returned from cache: {:?}", lookup),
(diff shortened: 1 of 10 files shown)


arsh (commented):
Just some questions.

mountpoint-s3/src/inode/expirable_set.rs
```diff
@@ -0,0 +1,100 @@
...
+pub struct ExpirableSet {
+    inner: RwLock<ExpirableSetInner>,
+    max_size: usize,
+    validity: Duration,
+}
+
+#[derive(Debug, Default)]
+struct ExpirableSetInner {
+    map: LinkedHashMap<Key, Expiry>,
+    size: usize,
```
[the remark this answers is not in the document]
passaro: Obviously, yes! Not sure how I had missed it in the first place. Fixed now.
Thanks for noticing it!

mountpoint-s3/src/inode.rs
```diff
@@ -80,6 +86,7 @@ pub struct Superblock {
 struct SuperblockInner {
     bucket: String,
     inodes: RwLock<HashMap<InodeNo, Inode>>,
+    known_missing_keys: ExpirableSet,
```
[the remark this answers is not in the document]
passaro: Fair question. In fact, it had started that way, but I decided to move the mutex into `ExpirableSet` so that the "inode.negative_cache.operation_duration_us" metrics could also include its contention.
The alternative I see would be to add wrapper methods to `SuperblockInner` and move the locking (and the metrics) there. `ExpirableSet` would still be quite specific though, so I'm not convinced it would be worth it.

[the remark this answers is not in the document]
arsh: To clarify, this is not a blocker. Just thinking out loud.

mountpoint-s3/tests/fuse_tests/lookup_test.rs
dannycjones, line 1: Do we need a new test here just to give one example of negative caching working end-to-end with FUSE?

mountpoint-s3/src/inode/expiry.rs
dannycjones, line 1: Nice!

mountpoint-s3/src/inode/expirable_set.rs
dannycjones, line 1: Do we want a unit test or two on `insert` method?

dannycjones (commented):
Looking good! I had just two questions related to testing and one on the metric I think we need to move around a little.

mountpoint-s3/src/inode.rs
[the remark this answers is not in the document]
passaro: Good point. I'll try to rename to something more specific.

mountpoint-s3/src/inode/expirable_set.rs
```diff
@@ -0,0 +1,90 @@
...
+        let start = Instant::now();
+        let Some(value) = self.map.read().unwrap().get(&key).cloned() else {
+            return false;
+        };
+        metrics::histogram!(
+            "inode.negative_cache.operation_duration_us",
+            start.elapsed().as_micros() as f64,
+            "op" => "contains",
+        );
+        value.is_valid()
```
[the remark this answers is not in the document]
passaro: Good spot!

mountpoint-s3/src/inode/expirable_set.rs
```diff
@@ -0,0 +1,90 @@
...
+            map: RwLock::new(Default::default()),
+            max_size,
+            validity,
+        }
+    }
+
+    pub fn contains(&self, parent_ino: InodeNo, child_name: &str) -> bool {
+        let key = Key {
+            parent_ino,
+            child_name: child_name.to_owned(),
```
passaro, line 36: This copy annoys me quite a bit. Probably not the most critical optimization, but is there a simple way to avoid it?

mountpoint-s3/src/inode.rs
```diff
@@ -1013,6 +1046,8 @@ impl SuperblockInner {
                 }
                 if let Some(existing_inode) = existing_inode {
                     writing_children.remove(&existing_inode.ino());
+                } else {
+                    self.known_missing_keys.remove(parent.ino(), name);
```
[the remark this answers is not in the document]
passaro: Yes. Mentioned it in `remove` doc.

mountpoint-s3/src/inode/expiry.rs
```diff
@@ -0,0 +1,24 @@
+use std::time::{Duration, Instant};
```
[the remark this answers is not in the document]
passaro: We extracted creation and validation into the `Expiry` type (and module), which we can use for both `Inode`s and the negative cache.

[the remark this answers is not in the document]
dannycjones: Even leaving out the re-use: it lets us review what these three (small) methods are doing in isolation, be happy with them, and then use them around the code base as single line method calls. The method calls clearly show what they do by their method name and accompanying RustDoc.

mountpoint-s3/src/inode.rs
```diff
@@ -822,7 +843,13 @@ impl SuperblockInner {
             InodeKindData::Directory { children, .. } => children.get(name),
         };
         match (remote, inode) {
-            (None, None) => Err(InodeError::FileDoesNotExist(name.to_owned(), parent.err())),
+            (None, None) => {
+                if self.config.cache_config.serve_lookup_from_cache {
+                    // Insert or update validity of negative cache entry.
+                    self.known_missing_keys.insert(parent.ino(), name);
```
vladem, line 849: so since it's taking a global write lock for cache, is it still `try_update_fast_path`? I believe you've mentioned some experiments around the performance previously

mountpoint-s3/src/inode.rs
```diff
@@ -80,6 +86,7 @@ pub struct Superblock {
 struct SuperblockInner {
     bucket: String,
     inodes: RwLock<HashMap<InodeNo, Inode>>,
+    known_missing_keys: NegativeCache,
```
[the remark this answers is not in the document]
passaro: Not really. It's a leftover from a previous iteration. Will rename.

mountpoint-s3/src/inode/expiry.rs
```diff
@@ -0,0 +1,24 @@
...
+impl Expiry {
+    /// Create a new instance with the given validity from now.
+    pub fn from_now(validity: Duration) -> Self {
+        let expiry = Instant::now()
+            .checked_add(validity)
+            .expect("64-bit time shouldn't overflow");
+        Self(expiry)
+    }
+
+    /// How much longer this instance will be valid for.
```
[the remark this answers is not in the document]
passaro: Yes, will rename this and related methods

mountpoint-s3/src/inode/negative_cache.rs
```diff
@@ -0,0 +1,204 @@
...
+use super::{expiry::Expiry, InodeNo};
+
+use crate::sync::RwLock;
+
+/// A caches for negative lookups.
+/// Maintains a bounded set of (parent_ino, child_name) entries that expire after a fixed time.
+#[derive(Debug)]
+pub struct NegativeCache {
+    /// Holds keys in expiration order from oldest to newest.
+    map: RwLock<LinkedHashMap<Key, Expiry>>,
```
vladem, line 14: so this is going to be `100'000` entities max and we expect it to use no more than `20MB` of space, any particular estimations on how much memory a single entity could take? the key is generally unbounded, right?

mountpoint-s3/tests/fuse_tests/lookup_test.rs
```diff
@@ -198,3 +199,49 @@ fn lookup_unicode_keys_s3() {
...
+    };
+    let (mount_point, _session, mut test_client) = creator_fn("lookup_with_negative_cache", config);
+
+    // Check negative caching
+    let file_path = mount_point.path().join(FILE_NAME);
+    metadata(&file_path).expect_err("should fail as no object exists");
+
+    test_client.put_object(FILE_NAME, b"hello").unwrap();
+
+    metadata(&file_path).expect_err("should fail as mountpoint should use negative cache");
```
vladem, line 228: could the cache entity expire at this point? or it relies on the delete-on-insert implementation detail?

mountpoint-s3/tests/fuse_tests/lookup_test.rs
```diff
@@ -198,3 +199,49 @@ fn lookup_unicode_keys_s3() {
...
+
+    // Use read dir to discover the new file
+    let dir_entry_names = read_dir_to_entry_names(read_dir(mount_point.path()).unwrap());
```
vladem, line 231: just for my understanding, so the `readdir(parent)` will call `create_inode_locked(child)` which will remove the child from negative cache by `self.known_missing_keys.remove`, right?

mountpoint-s3/tests/fuse_tests/lookup_test.rs
passaro: No, it relies on the 10 min TTL (line 213).

mountpoint-s3/tests/fuse_tests/lookup_test.rs
passaro: Correct

mountpoint-s3/src/inode/negative_cache.rs
```diff
@@ -0,0 +1,204 @@
...
+    pub fn insert(&self, parent_ino: InodeNo, child_name: &str) {
+        let expiry = Expiry::from_now(self.validity);
+        let key = Key {
+            parent_ino,
+            child_name: child_name.to_owned(),
+        };
+        let start = Instant::now();
+        let mut map = self.map.write().unwrap();
+        if map.insert(key, expiry).is_none() {
+            // Remove entries that are no longer valid.
```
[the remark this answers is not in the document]
passaro: Good point. This is definitely something we could add later. I'm leaning towards leaving the logic inline for now. We can extract it to a separate method when we need it.

mountpoint-s3/src/inode/negative_cache.rs
passaro: The rough estimation is based on indirectly measuring Mountpoint process memory while trying to lookup random entries (I used UUIDs, 37 characters): the result of the test is a marginal cost of <200 bytes per entry.
In general, it will depend on the length of the key, but it is not really unbounded: consider that the max file name length is 255 bytes.

mountpoint-s3/src/inode.rs
```diff
@@ -624,30 +635,39 @@ impl SuperblockInner {
...
     /// Lookup an [Inode] against known directory entries in the parent,
     /// verifying any returned entry has not expired.
-    fn cache_lookup(&self, parent_ino: InodeNo, name: &str) -> Option<LookedUp> {
-        fn do_cache_lookup(parent: Inode, name: &str) -> Option<LookedUp> {
+    fn cache_lookup(&self, parent_ino: InodeNo, name: &str) -> Option<Result<LookedUp, InodeError>> {
+        fn do_cache_lookup(
+            superblock: &SuperblockInner,
+            parent: Inode,
+            name: &str,
+        ) -> Option<Result<LookedUp, InodeError>> {
```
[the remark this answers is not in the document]
passaro: Added comment

mountpoint-s3/src/inode.rs
passaro: That's fair. It is a different lock, but the method names may be misleading. Let me think about it.

mountpoint-s3/src/inode/negative_cache.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn test_insert_resets_validity() {
+        let cache = NegativeCache::new(100, Duration::from_millis(50));
+
+        cache.insert(1, "child1");
+        sleep(Duration::from_millis(100));
+        assert!(!cache.contains(1, "child1"));
+
+        cache.insert(1, "child1");
+        assert!(cache.contains(1, "child1"));
+    }
```
dannycjones: This test doesn't quite test the validity reset. I think we want something more like:

```suggestion
    #[test]
    fn test_insert_resets_validity() {
        let cache = NegativeCache::new(100, Duration::from_millis(100));

        cache.insert(1, "child1");
        sleep(Duration::from_millis(50));
        cache.insert(1, "child1"); // refresh
        sleep(Duration::from_millis(75)); // original 100ms exceeded
        assert!(cache.contains(1, "child1"));
    }
```

dannycjones (commented):
LGTM. I think we need to land on what the next steps are for this change.

mountpoint-s3/src/inode/negative_cache.rs
dannycjones: S3 supports keys up to 1024 bytes. Since this is a negative cache, I'm sure that a user could ask for something even longer but we can assume 1024 as a sensible upper bound.

I think it would help to document how we chose 100,000 as the default max size, and maybe some of the reasoning here can go into it.

mountpoint-s3/tests/cli.rs
dannycjones, line 1: Thanks for adding these

dannycjones (commented):
Also reviewed the latest changes, looking good.

arsh (commented):
Just a few comments / questions.

mountpoint-s3/src/inode/negative_cache.rs
vladem: > I'm sure that a user could ask for something even longer but we can assume 1024 as a sensible upper bound.

seems like they can not given the linux limitation for the file size, right?

adjusting Alex's estimate (with 37 chars for names) for names max of 255 chars, we'll have `< 420 bytes` per entity => `420 * 100'000 / 1024 / 1024 ~< 40 MB` in total would be more precise, but yeah I see the values matching in order of magnitude :)

mountpoint-s3/src/inode/negative_cache.rs
```diff
@@ -0,0 +1,204 @@
...
+/// A caches for negative lookups.
+/// Maintains a bounded set of (parent_ino, child_name) entries that expire after a fixed time.
+#[derive(Debug)]
+pub struct NegativeCache {
+    /// Holds keys in expiration order from oldest to newest.
+    map: RwLock<LinkedHashMap<Key, Expiry>>,
+    /// Upper bound for the cache.
+    max_size: usize,
+    /// TTL of a key at insertion.
+    ttl: Duration,
```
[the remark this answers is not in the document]
passaro: Yes, that is the assumption for this structure. In fact, the implementation relies on it because it means that the insertion order is also the expiration order.

mountpoint-s3/src/inode/negative_cache.rs
```diff
@@ -0,0 +1,204 @@
...
+            }
+
+            // Remove entries that exceed the limit.
+            while map.len() > self.max_size {
+                let Some((_, e)) = map.pop_front() else {
+                    break;
+                };
+                // Report how many entries are evicted while still current.
+                metrics::counter!(
+                    "metadata_cache.negative_cache.entries_evicted_before_expiry",
```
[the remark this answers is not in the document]
passaro: This metric will help detect cases where the negative cache is full. I don't currently see much value in monitoring the expirations. Besides, we track the current total, insertions, and deletions, so it could be derived.

arsh (dismissed):
LGTM

arsh (approved):

Issue #533, opened 2023-09-25
Improve logging to have clearer explanations or actions


dannycjones:
It should be easy for users of mountpoint-s3 to determine why mountpoint-s3 behaved in some way, such as when operations are failed when an unsupported file operation is used.

For example, random writes are not supported by mountpoint-s3. A random write results in the following error message only:

```
WARN write{req=50 ino=2 fh=1 offset=36 length=4}: mountpoint_s3::fuse: write failed: upload error: out of order write; expected offset 3461864 but got 36
```

It would be more clearer to users if this was called out specifically as a file operation unsupported by mountpoint-s3, so that the next action to take is clear.

sauraank:
Completed all the tasks of this story. So, closing this issue as completed.

PR #699, opened 2024-01-12
Improve error logs for unsupported operations: File Overwrite, Random Write, Directory Shadowing, Unlink (without mount option)

sauraank:
## Description of change
Added clear message for the operations that are unsupported but did not had easy to understand logs. Added links to semantics for deletion and allowed filenames to be visible using Mountpoint. But, with links there is a risk that they can get broken once the content is moved somewhere else.
I will add another PR for troubleshooting page and down level the error noise to `debug!` in other Pull Requests.

With this new change, following are error logs that we receive:
For directory shadowing of file with same name, `ls` give following output -
```
WARN readdir{req=5 ino=1 fh=2 offset=17}: mountpoint_s3::inode::readdir::ordered: file 'out' (full key "out") is omitted because another directory 'out' exist with the same name.
```
For file deletion without using --allow-delete flag, give following in logs -
```
WARN unlink{req=8 parent=1 name="ini.txt"}: mountpoint_s3::fuse: unlink failed: Deletes are disabled. Use '--allow-delete' mount option to enable it.
```
Out of Write gives -
```
WARN write{req=52 ino=49 fh=3 offset=512 length=512 name="new.txt"}: mountpoint_s3::fuse: write failed: upload error: out of order write NOT supported by Mountpoint, aborting the upload; expected offset 0 but got 512
```
UPDATE: Overwriting a file is now supported, so the error message is the one set in PR #487  -
```
WARN file overwrite is disabled by default, you need to remount with --allow-overwrite flag and open the file in truncate mode (O_TRUNC) to overwrite it
```
Also changed the logging in case of finish writing to update status of Inode, but the Inode WriteStatus is in Invalid state. Earlier it was `InodeNotWritable` which is same error for File Overwrite when its flag is not enabled. Also, it did not completely explained that WriteStatus of the Inode is not valid for writing.

Relevant issues: 
#533 

No. It only improves the error logging as mentioned above.

No, there is no breaking change.
---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -1198,7 +1198,10 @@ where
 
     pub async fn unlink(&self, parent_ino: InodeNo, name: &OsStr) -> Result<(), Error> {
         if !self.config.allow_delete {
-            return Err(err!(libc::EPERM, "deletes are disabled"));
+            return Err(err!(
+                libc::EPERM,
+                "Deletes are disabled. Use '--allow-delete' mount option to enable it."
+            ));
         }
         Ok(self.superblock.unlink(&self.client, parent_ino, name).await?)
     }
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -1174,7 +1174,7 @@ impl WriteHandle {
 
                 Ok(())
             }
-            _ => Err(InodeError::InodeNotWritable(inode.err())),
+            _ => Err(InodeError::InodeInvalidWriteStatus(inode.err())),
         }
     }
 }
@@ -1555,6 +1555,8 @@ pub enum InodeError {
     FileAlreadyExists(InodeErrorInfo),
     #[error("inode {0} is not writable")]
     InodeNotWritable(InodeErrorInfo),
+    #[error("Invalid state of inode {0} to be written")]
+    InodeInvalidWriteStatus(InodeErrorInfo),
     #[error("inode {0} is already being written")]
     InodeAlreadyWriting(InodeErrorInfo),
     #[error("inode {0} is not readable while being written")]
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -1555,7 +1555,7 @@ pub enum InodeError {
     FileAlreadyExists(InodeErrorInfo),
     #[error("inode {0} is not writable")]
     InodeNotWritable(InodeErrorInfo),
-    #[error("Invalid state of inode {0} to be written")]
+    #[error("Invalid state of inode {0} to be written. Aborting the write.")]
     InodeInvalidWriteStatus(InodeErrorInfo),
     #[error("inode {0} is already being written")]
     InodeAlreadyWriting(InodeErrorInfo),
--- a/mountpoint-s3/src/inode/readdir.rs
+++ b/mountpoint-s3/src/inode/readdir.rs
@@ -458,7 +458,7 @@ mod ordered {
                     (Some(entry), Some(last_entry)) => {
                         if last_entry.name() == entry.name() {
                             warn!(
-                                "{} is omitted because another {} exist with the same name.\n",
+                                "{} is omitted because another {} exist with the same name",
                                 entry.description(),
                                 last_entry.description(),
                             );
--- a/mountpoint-s3/src/inode/readdir.rs
+++ b/mountpoint-s3/src/inode/readdir.rs
@@ -198,7 +198,7 @@ impl ReaddirHandle {
 
 /// A single entry in a readdir stream. Remote entries have not yet been converted to inodes -- that
 /// should be done lazily by the consumer of the entry.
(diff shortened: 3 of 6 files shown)


mountpoint-s3/src/inode.rs
```diff
@@ -1502,8 +1509,8 @@ pub enum InodeError {
     IsDirectory(InodeErrorInfo),
     #[error("file already exists at inode {0}")]
     FileAlreadyExists(InodeErrorInfo),
-    #[error("inode {0} is not writable")]
-    InodeNotWritable(InodeErrorInfo),
+    #[error("inode {0} is not writable: {1:?}")]
+    InodeNotWritable(InodeErrorInfo, String),
```
monthonk, line 1513: I think we should name the new variable so we don't have to guess what it is, something like
```suggestion
    #[error("inode {inode_error_info} is not writable because {error_reason:?}")]
    InodeNotWritable {
        inode_error_info: InodeErrorInfo,
        error_reason: String,
    },
```

mountpoint-s3/src/inode.rs
```diff
@@ -1130,7 +1133,11 @@ impl WriteHandle {
 
                 Ok(())
             }
-            _ => Err(InodeError::InodeNotWritable(inode.err())),
+            WriteStatus::Remote => Err(InodeError::InodeNotWritable(
+                inode.err(),
+                "File already uploaded to S3.".to_owned(),
+            )),
+            WriteStatus::LocalUnopened => unreachable!("Unopened local file cannot be written"),
```
monthonk, line 1140: I don't think we know this for sure, what if there is a bug and write status becomes `LocalUnopened`? We should also return a reasonable error for this case.

mountpoint-s3/src/upload.rs
```diff
@@ -49,7 +49,7 @@ pub enum UploadWriteError<E: std::error::Error> {
     #[error("put request failed")]
     PutRequestFailed(#[from] E),
 
-    #[error("out of order write; expected offset {expected_offset:?} but got {write_offset:?}")]
+    #[error("out of order write NOT supported by Mountpoint, Aborting the Upload; expected offset {expected_offset:?} but got {write_offset:?}")]
```
monthonk, line 52: nit:
```suggestion
    #[error("out of order write is NOT supported by Mountpoint, aborting the upload; expected offset {expected_offset:?} but got {write_offset:?}")]
```

monthonk (changes requested):
@sauraank thanks for improving the logs! I have put a few comments.

mountpoint-s3/src/inode.rs
```diff
@@ -1069,7 +1069,10 @@ impl WriteHandle {
                 })
             }
             WriteStatus::LocalOpen => Err(InodeError::InodeAlreadyWriting(inode.err())),
-            WriteStatus::Remote => Err(InodeError::InodeNotWritable(inode.err())),
+            WriteStatus::Remote => Err(InodeError::InodeNotWritable(
+                inode.err(),
+                "Writing to existing file is NOT supported by Mountpoint.".to_owned(),
```
passaro, line 1074: Rather than adding a message here, I'd prefer having a specific `InodeError` case for each situation.

mountpoint-s3/src/fs.rs
```diff
@@ -1057,7 +1057,10 @@ where
 
     pub async fn unlink(&self, parent_ino: InodeNo, name: &OsStr) -> Result<(), Error> {
         if !self.config.allow_delete {
-            return Err(err!(libc::EPERM, "deletes are disabled"));
+            return Err(err!(
+                libc::EPERM,
+                "Deletes are disabled. Use '--allow-delete' mount option to enable it. \nFor more information on deletion, visit: https://github.com/awslabs/mountpoint-s3/blob/main/doc/SEMANTICS.md#deletes"
```
monthonk, line 1062: I'm kind of not sure if we want to add documentation links in the logs. It's probably hard to maintain between different versions. As you mentioned, it could be broken if we move the docs to somewhere else or it's probably not true anymore when we decided to change the semantics.

mountpoint-s3/src/inode.rs
monthonk: Let's discuss this offline, also include @passaro in the discussion.

mountpoint-s3/src/inode/readdir.rs
```diff
@@ -456,13 +456,16 @@ mod ordered {
                 // Deduplicate the entry we want to return
                 match next {
                     Some(entry) => {
-                        if self.last_name.as_deref() == Some(entry.name()) {
+                        let last_name = self.last_entry.as_ref().map(|entry| entry.name());
+                        let last_description = self.last_entry.as_ref().map(|entry| entry.description());
+                        if last_name == Some(entry.name()) {
```
monthonk, line 461: Probably changing this check to `match &self.last_entry` would be clearer, so we know that if last_entry exists then last_name and last_description must have some values.

monthonk (approved):
LGTM

Issue #706, opened 2024-01-19
Mountpoint operations hang indefinitely


SamStudio8:
### Mountpoint for Amazon S3 version

mount-s3 1.3.2

### AWS Region

eu-west-1

### Describe the running environment

Mountpoint installed to EC2 instances and a single client is started to share a bucket mount with ECS tasks.
```
2024-01-18T19:19:04.080064Z DEBUG mountpoint_s3_client::instance_info: got instance info from IMDS identity_document=IdentityDocument { instance_type: "m5.12xlarge", region: "eu-west-1" }
2024-01-18T19:19:04.080323Z  INFO mount_s3: target network throughput 12 Gbps
2024-01-18T19:19:04.080338Z DEBUG mount_s3: using instance region eu-west-1
```

### Mountpoint options

```shell
mount-s3 --debug -l /home/ec2-user/mnt-log --allow-other --uid 1000 --gid 1000 --part-size 33554432 --cache /home/ec2-user/mnt-cache --metadata-ttl 15 <bucket> /home/ec2-user/mnt
```

### What happened?

Occasionally, mount-s3 enters what appears to be a deadlock and will hang indefinitely. Any process attempting to access the mount will be stuck in an uninterruptible sleep. I don't have a lot to go on but attach the relevant logs in the hope that somebody who knows the internals of mount-s3 might be able to see what might be going on. I've no idea if useful but I've also backtraced the main client-s3 thread and one of the workers.

As our ECS tasks are stuck I can access the machine to read logs and run commands. The instance is otherwise usable: EBS volume is responsive and the mount-s3 client continues to log (the same lines repeatedly). I do not have a repro but I am happy to collect additional data with any further debugging suggestions, as I am able to unreliably reproduce the issue.

### Relevant log output

```shell
## Tail of mountpoint-s3 log
2024-01-18T19:19:04.076711Z  INFO mount_s3: mount-s3 1.3.2
[...]
2024-01-18T19:35:28.265228Z DEBUG fuser::request: FUSE(721680) ino 0x0000000000000941 WRITE fh FileHandle(26906), offset 16384, size 4096, write flags 0x0
2024-01-18T19:35:28.265244Z DEBUG fuser::request: FUSE(721681) ino 0x0000000000000925 WRITE fh FileHandle(26881), offset 184320, size 4096, write flags 0x0
[...]

PR #711, opened 2024-01-24
Fix a race condition on create and forget operations

monthonk:
## Description of change

There is a possibility of deadlock if `create` and `forget` are called on files under the same directory. This is because they are both trying to hold the locks on `inodes` in Superblock and the state of the parent directory at the same time.

This commit prevents `create` and `forget` from holding two locks simultaneously.

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/706

No, only a bug fix.

---

### DIFF
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -131,64 +131,62 @@ impl Superblock {
     /// The kernel may forget a number of references (`n`) in one forget message to our FUSE implementation.
     /// If the lookup count reaches zero, it is safe for the [Superblock] to delete the [Inode].
     pub fn forget(&self, ino: InodeNo, n: u64) {
-        let inodes = self.inner.inodes.read().unwrap();
-        match inodes.get(&ino) {
-            None => {
+        let inode = {
+            if let Some(inode) = self.inner.inodes.read().unwrap().get(&ino).cloned() {
+                inode
+            } else {
                 debug_assert!(
                     false,
                     "forget should not be called on inode already removed from superblock"
                 );
                 error!("forget called on inode {ino} already removed from the superblock");
+                return;
             }
-            Some(inode) => {
-                // We have to clone the inode so it will not hold reference to the lock which we want to drop as soon as possible.
-                let inode = inode.clone();
-                drop(inodes);
-                logging::record_name(inode.name());
-                let new_lookup_count = inode.dec_lookup_count(n);
-                if new_lookup_count == 0 {
-                    // Safe to remove, kernel no longer has a reference to it.
-                    trace!(ino, "removing inode from superblock");
-                    let Some(inode) = self.inner.inodes.write().unwrap().remove(&ino) else {
-                        error!("forget called on inode {ino} already removed from the superblock");
-                        return;
-                    };
+        };
 
-                    let inodes = self.inner.inodes.read().unwrap();
+        logging::record_name(inode.name());
+        let new_lookup_count = inode.dec_lookup_count(n);
+        if new_lookup_count == 0 {
+            // Safe to remove, kernel no longer has a reference to it.
+            trace!(ino, "removing inode from superblock");
+            let Some(inode) = self.inner.inodes.write().unwrap().remove(&ino) else {
+                error!("forget called on inode {ino} already removed from the superblock");
+                return;
+            };
+
+            let parent = {
+                if let Some(parent) = self.inner.inodes.read().unwrap().get(&inode.parent()).cloned() {
+                    parent
+                } else {
                     // Should be impossible for this to fail (VFS inodes reference their parent, so
                     // children need to be freed first), but let's not crash in a `forget` function...
-                    let Some(parent) = inodes.get(&inode.parent()) else {
-                        debug_assert!(false, "children should be forgotten before parents");
-                        return;
-                    };
-                    // Clone the parent inode so we can drop the lock on inodes.
-                    let parent = parent.clone();
-                    drop(inodes);
-                    let mut parent_state = parent.inner.sync.write().unwrap();
-                    let InodeKindData::Directory {
-                        children,
-                        writing_children,
-                        ..
-                    } = &mut parent_state.kind_data
-                    else {
-                        unreachable!("parent is always a directory");
-                    };
-                    if let Some(child) = children.get(inode.name()) {
-                        // Don't accidentally remove a newer inode (e.g. remote shadowing local)
-                        if child.ino() == ino {
-                            children.remove(inode.name());
-                        }
-                    }
-                    writing_children.remove(&ino);
-
-                    if let Ok(state) = inode.get_inode_state() {
-                        metrics::counter!(
-                            "metadata_cache.inode_forgotten_before_expiry",
-                            state.stat.is_valid().into(),
-                        );
-                    };
+                    debug_assert!(false, "children should be forgotten before parents");
+                    return;
+                }
+            };
+            let mut parent_state = parent.inner.sync.write().unwrap();
+            let InodeKindData::Directory {
+                children,
+                writing_children,
+                ..
+            } = &mut parent_state.kind_data
+            else {
+                unreachable!("parent is always a directory");
+            };
+            if let Some(child) = children.get(inode.name()) {
+                // Don't accidentally remove a newer inode (e.g. remote shadowing local)
+                if child.ino() == ino {
+                    children.remove(inode.name());
                 }
             }
+            writing_children.remove(&ino);
+
+            if let Ok(state) = inode.get_inode_state() {
+                metrics::counter!(
+                    "metadata_cache.inode_forgotten_before_expiry",
+                    state.stat.is_valid().into(),
+                );
+            };
         }
     }
 
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -149,13 +149,10 @@ impl Superblock {
                 if new_lookup_count == 0 {
                     // Safe to remove, kernel no longer has a reference to it.
                     trace!(ino, "removing inode from superblock");
-                    let inode = self
-                        .inner
-                        .inodes
-                        .write()
-                        .unwrap()
-                        .remove(&ino)
-                        .expect("we just retrieved it");
+                    let Some(inode) = self.inner.inodes.write().unwrap().remove(&ino) else {
+                        error!("forget called on inode {ino} already removed from the superblock");
+                        return;
+                    };
 
                     let inodes = self.inner.inodes.read().unwrap();
                     // Should be impossible for this to fail (VFS inodes reference their parent, so
--- a/mountpoint-s3/src/inode.rs
+++ b/mountpoint-s3/src/inode.rs
@@ -131,63 +131,68 @@ impl Superblock {
     /// The kernel may forget a number of references (`n`) in one forget message to our FUSE implementation.
     /// If the lookup count reaches zero, it is safe for the [Superblock] to delete the [Inode].
     pub fn forget(&self, ino: InodeNo, n: u64) {
-        // Put inode removal in a block so we don't hold the lock on the inodes table longer than needed.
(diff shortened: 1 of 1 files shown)


mountpoint-s3/src/inode.rs
```diff
@@ -131,55 +131,63 @@ impl Superblock {
...
-                    trace!(ino, "removing inode from superblock");
-                    let inode = inodes.remove(&ino).expect("we just retrieved it");
-
-                    // Should be impossible for this to fail (VFS inodes reference their parent, so
-                    // children need to be freed first), but let's not crash in a `forget` function...
-                    let Some(parent) = inodes.get(&inode.parent()) else {
-                        debug_assert!(false, "children should be forgotten before parents");
+        // Put inode removal in a block so we don't hold the lock on the inodes table longer than needed.
+        let (inode, parent) = {
+            let mut inodes = self.inner.inodes.write().unwrap();
```
passaro, line 136: We should avoid holding a write lock on the Inodes map for long in the first place. Especially while we need to lock other mutexes, like in `dec_lookup_count`.

What if we took just a read lock here (and released immediately), and then, only if required, took a write lock before line 152?

mountpoint-s3/src/inode.rs
```diff
@@ -2797,4 +2809,53 @@ mod tests {
         assert_eq!(file_inodestat.ctime, ts);
         assert_eq!(file_inodestat.mtime, ts);
     }
+
+    #[cfg(feature = "shuttle")]
+    mod shuttle_tests {
```
arsh, line 2814: Does it make sense to add a test for a file X that the kernel calls `forget` on but simultaneously we get a `create` request? I'm thinking we are in the middle of processing a request for `forget` while in parallel we get a request to create a file of the same name and same directory.

mountpoint-s3/src/inode.rs
```diff
@@ -131,55 +131,63 @@ impl Superblock {
...
-                        metrics::counter!(
-                            "metadata_cache.inode_forgotten_before_expiry",
-                            state.stat.is_valid().into(),
-                        );
-                    };
                 }
             }
+        };
+
+        let mut parent_state = parent.inner.sync.write().unwrap();
```
arsh, line 168: Is it possible for the parent of the inode to change from the moment you capture it in the inner block?

mountpoint-s3/src/inode.rs
```diff
@@ -343,44 +351,48 @@ impl Superblock {
...
-        // racing lookup. (It would be nice to lock the parent and *then* lookup, but we'd have to
-        // hold that lock across the remote API calls).
-        let InodeKindData::Directory { children, .. } = &mut parent_state.kind_data else {
-            return Err(InodeError::NotADirectory(parent_inode.err()));
-        };
-        if let Some(inode) = children.get(name) {
-            return Err(InodeError::FileAlreadyExists(inode.err()));
-        }
+        // Put inode creation in a block so we don't hold the lock on the parent state longer than needed.
+        let (inode, stat) = {
```
passaro, line 355: nit: this is `LookedUp`

mountpoint-s3/src/inode.rs
```diff
@@ -2797,4 +2809,53 @@ mod tests {
...
+                    .await
+                    .unwrap();
+
+                forget_task.join().unwrap();
+                let lookup_count = lookup.inode.inner.sync.read().unwrap().lookup_count;
+                assert_eq!(lookup_count, 0);
+            }
+
+            check_random(|| block_on(test_helper()), 1000);
+            check_pct(|| block_on(test_helper()), 1000, 3);
```
dannycjones, line 2858: I don't know enough about PCT to know if its worth doing both, but I see we do this elsewhere so LGTM :)

Nice that this test case catches the deadlock and now passes.

dannycjones (commented):
Generally looks good to me, with some tweaks to how long we hold the write lock as mentioned by @passaro.

I like the idea we discussed in-person with @passaro around hiding the inode table lock with interior mutability, but let's think about that as a future enhancement. Mentioning it here to keep the idea around.

mountpoint-s3/src/inode.rs
arsh: Although it will probably be different inodes, so maybe it does not make sense.

mountpoint-s3/src/inode.rs
```diff
@@ -141,19 +141,32 @@ impl Superblock {
...
                     trace!(ino, "removing inode from superblock");
-                    let inode = inodes.remove(&ino).expect("we just retrieved it");
-
+                    let inode = self
+                        .inner
+                        .inodes
+                        .write()
+                        .unwrap()
+                        .remove(&ino)
+                        .expect("we just retrieved it");
```
vladem, line 158: well, technically, 2 concurrent `forget` calls on a node with `lookup_count == 1` may result in this failure, since we're not holding the write lock on the collection (but it seems that kernel should never call forget like that, right?)

mountpoint-s3/src/inode.rs
monthonk: It is possible, but I think that is true even when we locked the `inodes` because parent is actually a shared reference.

mountpoint-s3/src/inode.rs
monthonk: That's true, let's make sure we do not crash here even if there are 2 concurrent forgets.

mountpoint-s3/src/inode.rs
monthonk: Yeah, it will be different inodes. I will leave the test as it is for now.

mountpoint-s3/src/inode.rs
```diff
@@ -131,63 +131,65 @@ impl Superblock {
...
+                debug_assert!(
+                    false,
+                    "forget should not be called on inode already removed from superblock"
+                );
+                error!("forget called on inode {ino} already removed from the superblock");
+            }
+            Some(inode) => {
+                // We have to clone the inode so it will not hold reference to the lock which we want to drop as soon as possible.
+                let inode = inode.clone();
+                drop(inodes);
```
passaro, line 146: I'd recommend something like this (untested):

```suggestion
        match self.inner.inodes.read().unwrap().get(&ino).cloned() {
            None => {
                debug_assert!(
                    false,
                    "forget should not be called on inode already removed from superblock"
                );
                error!("forget called on inode {ino} already removed from the superblock");
            }
            Some(inode) => {
```

mountpoint-s3/src/inode.rs
```diff
@@ -131,63 +131,65 @@ impl Superblock {
...
+                    trace!(ino, "removing inode from superblock");
+                    let Some(inode) = self.inner.inodes.write().unwrap().remove(&ino) else {
+                        error!("forget called on inode {ino} already removed from the superblock");
+                        return;
+                    };
+
+                    let inodes = self.inner.inodes.read().unwrap();
+                    // Should be impossible for this to fail (VFS inodes reference their parent, so
+                    // children need to be freed first), but let's not crash in a `forget` function...
+                    let Some(parent) = inodes.get(&inode.parent()) else {
```
passaro, line 160: Same trick with `cloned()` here.

mountpoint-s3/src/inode.rs
monthonk: Tried that but it didn't work. I'm afraid it is still locking the `inodes`.

mountpoint-s3/src/inode.rs
```diff
@@ -131,7 +131,7 @@ impl Superblock {
     /// The kernel may forget a number of references (`n`) in one forget message to our FUSE implementation.
     /// If the lookup count reaches zero, it is safe for the [Superblock] to delete the [Inode].
     pub fn forget(&self, ino: InodeNo, n: u64) {
-        let mut inodes = self.inner.inodes.write().unwrap();
+        let inodes = self.inner.inodes.read().unwrap();
         match inodes.get(&ino) {
```
dannycjones, line 135: Sorry for churn, but would it be safest/clearest to just have:

```suggestion
        let inode = {
            let inodes = self.inner.inodes.read().unwrap();
            inodes.get(&ino).cloned()
        }
        match inode {
```

This would avoid the need to drop `inodes` explicitly later.

We can even pull the `None` case into this scope and then just return an unwrapped non-option inode.

mountpoint-s3/src/inode.rs
```diff
@@ -141,19 +141,29 @@ impl Superblock {
...
                 if new_lookup_count == 0 {
                     // Safe to remove, kernel no longer has a reference to it.
                     trace!(ino, "removing inode from superblock");
-                    let inode = inodes.remove(&ino).expect("we just retrieved it");
+                    let Some(inode) = self.inner.inodes.write().unwrap().remove(&ino) else {
+                        error!("forget called on inode {ino} already removed from the superblock");
+                        return;
+                    };
 
+                    let inodes = self.inner.inodes.read().unwrap();
```
dannycjones, line 157: Let's use the scoped lock pattern here.

```suggestion
                    let parent = {
                        let inodes = self.inner.inodes.read().unwrap();
                        inodes.get(&inode.parent()).cloned()
                    }
```

mountpoint-s3/src/inode.rs
dannycjones: I put a similar comment elsewhere, we can avoid the locking with a scoped block.

dannycjones (changes requested):
Hey, almost ready to ship.

I want to make sure we use the right pattern on how we take and drop the lock using a scope instead of relying on a drop call later.

dannycjones (approved):
LGTM!

mountpoint-s3/src/inode.rs
```diff
@@ -131,54 +131,62 @@ impl Superblock {
...
-                    trace!(ino, "removing inode from superblock");
-                    let inode = inodes.remove(&ino).expect("we just retrieved it");
+        };
+
+        logging::record_name(inode.name());
+        let new_lookup_count = inode.dec_lookup_count(n);
+        if new_lookup_count == 0 {
+            // Safe to remove, kernel no longer has a reference to it.
+            trace!(ino, "removing inode from superblock");
+            let Some(inode) = self.inner.inodes.write().unwrap().remove(&ino) else {
```
sauraank, line 152: Not blocking this PR to get merged. Just want to ask how scoping down the write lock is preventing the race condition?

dannycjones: Do you mind to share the exact (or one of the) sequence of locks that could cause the deadlock, @monthonk?

monthonk: So, there are two things
1. The prevent deadlock, whenever we want to hold the lock for more than one resources we have to do it in the same sequence. But we did not follow this rule in previous code since 
    - `create` would lock `parent_state` first to add a child inode, then lock `inodes` to add the new inode to the hash map.
    - `forget` would lock `inodes` first because we want to remove a given inode from the hash map, then we lock `parent_state` to remove it from the parent inode.

a possible sequence of deadlock could be `create::lock::parent_state -> forget::lock::inodes -> create::lock::inodes -> forget::lock::parent_state`. We can fix this problem by holding one lock at a time for each operation.

2. Even if we can prevent the race condition from 1. we also don't want to hold the write lock on `inodes` longer than needed because this hash map is used across all file system operations. That's why we also scoped down the write lock in `forget` into multiple read/write locks. Basically, it's not necessary thing to do but a best practice.

mountpoint-s3/src/inode.rs
```diff
@@ -2797,4 +2809,53 @@ mod tests {
...
+    #[cfg(feature = "shuttle")]
+    mod shuttle_tests {
+        use mountpoint_s3_client::mock_client::{MockClient, MockClientConfig};
+        use shuttle::{check_pct, check_random, thread};
+        use shuttle::{future::block_on, sync::Arc};
+
+        use super::*;
+
+        #[test]
+        fn test_create_and_forget_race_condition() {
```
jamesbornholt, line 2822: Cool test!

PR #684, opened 2023-12-29
Update `--profile` flag to override default chain's profile rather than replacing default chain

dannycjones:
## Description of change

_A previous PR was open for this change but was very stale: #415._

This change means that a profile can be specified that does not configure any credentials, but may configure other things. **Mountpoint does not read any configuration from an AWS Profile other than credentials today** but may do in the future.

Before this change, the entire credential chain was replaced with one that would supply profile credentials or error - no fallback. This was not in line with AWS CLI or SDKs.

This should fix #389 since the default chain also allows credential process credentials to be used.

## Does this change impact existing behavior?

Yes, it fixes the behavior of the `--profile <AWS_PROFILE>` flag to be in line with `AWS_PROFILE` environment variable, as well as `--profile <AWS_PROFILE>` flag of the AWS CLI.

It also fixes bug #389.

**Note:** It does not validate that the AWS profile exists, like the AWS CLI does. If the profile does not exist, the parameter will be ignored. Mountpoint already behaves this way today when using the `AWS_PROFILE` environment variable. Mountpoint previously reporting an error was a side effect of replacing the default chain with a profile provider.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

PR #715, opened 2024-01-25
Add support for --sse, --sse-kms-key-id flags under a feature flag

vladem:
## Description of change

1. Pass `--sse`, `--sse-kms-key-id` to the S3CrtClient (changes are added under a compile-time feature flag); 
3. Add happy test to `fuse_tests/write_test.rs` that uses IAM session policies;
4. Add error tests to `fuse_tests/fork_test.rs` that check error logs.

*Thinking of adding the checksumming logic in a separate PR.*

Relevant issues: **not linking an issue to avoid accidental closing**

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -378,6 +378,8 @@ pub struct S3FilesystemConfig {
     pub storage_class: Option<String>,
     /// S3 personality (for different S3 semantics)
     pub s3_personality: S3Personality,
+    /// Server side encryption configuration to be used when creating new S3 object
+    pub server_side_encryption: ServerSideEncryption,
 }
 
 impl Default for S3FilesystemConfig {
@@ -396,6 +398,7 @@ impl Default for S3FilesystemConfig {
             allow_overwrite: false,
             storage_class: None,
             s3_personality: S3Personality::Standard,
+            server_side_encryption: Default::default(),
         }
     }
 }
@@ -418,6 +421,34 @@ impl S3Personality {
             Self::Standard => true,
             Self::ExpressOneZone => false,
         }
+    }
+}
+
+/// Server-side encryption configuration for newly created objects
+#[derive(Debug, Clone, Default)]
+pub struct ServerSideEncryption {
+    sse_type: Option<String>,
+    sse_kms_key_id: Option<String>,
+}
+
+impl ServerSideEncryption {
+    /// Construct SSE settings from raw values provided via CLI
+    pub fn new(sse_type: Option<String>, sse_kms_key_id: Option<String>) -> Self {
+        // TODO: compute checksum
+        Self {
+            sse_type,
+            sse_kms_key_id,
+        }
+    }
+
+    /// String representation of the SSE type as it is expected by S3 API
+    pub fn sse_type(&self) -> Option<String> {
+        self.sse_type.clone()
+    }
+
+    /// AWS KMS Key ID, if provided
+    pub fn key_id(&self) -> Option<String> {
+        self.sse_kms_key_id.clone()
     }
 }
 
@@ -462,7 +493,11 @@ where
 
         let client = Arc::new(client);
 
-        let uploader = Uploader::new(client.clone(), config.storage_class.to_owned());
+        let uploader = Uploader::new(
+            client.clone(),
+            config.storage_class.to_owned(),
+            config.server_side_encryption.clone(),
+        );
 
         Self {
             config,
--- a/mountpoint-s3/tests/common/fuse.rs
+++ b/mountpoint-s3/tests/common/fuse.rs
@@ -9,8 +9,11 @@ use mountpoint_s3::fuse::S3FuseFilesystem;
 use mountpoint_s3::prefetch::{Prefetch, PrefetcherConfig};
 use mountpoint_s3::prefix::Prefix;
 use mountpoint_s3::S3FilesystemConfig;
+use mountpoint_s3_client::config::S3ClientAuthConfig;
 use mountpoint_s3_client::types::PutObjectParams;
 use mountpoint_s3_client::ObjectClient;
+use mountpoint_s3_crt::auth::credentials::{CredentialsProvider, CredentialsProviderStaticOptions};
+use mountpoint_s3_crt::common::allocator::Allocator;
 use tempfile::TempDir;
 
 pub trait TestClient: Send {
@@ -46,6 +49,7 @@ pub struct TestSessionConfig {
     pub part_size: usize,
     pub filesystem_config: S3FilesystemConfig,
     pub prefetcher_config: PrefetcherConfig,
+    pub auth_config: S3ClientAuthConfig,
 }
 
 impl Default for TestSessionConfig {
@@ -54,7 +58,21 @@ impl Default for TestSessionConfig {
             part_size: 8 * 1024 * 1024,
             filesystem_config: Default::default(),
             prefetcher_config: Default::default(),
+            auth_config: Default::default(),
         }
+    }
+}
+
+impl TestSessionConfig {
+    pub fn with_credentials(mut self, credentials: aws_sdk_s3::config::Credentials) -> Self {
+        let auth_config = CredentialsProviderStaticOptions {
+            access_key_id: credentials.access_key_id(),
+            secret_access_key: credentials.secret_access_key(),
+            session_token: credentials.session_token(),
+        };
+        let credentials_provider = CredentialsProvider::new_static(&Allocator::default(), auth_config).unwrap();
+        self.auth_config = S3ClientAuthConfig::Provider(credentials_provider);
+        self
     }
 }
 
@@ -256,7 +274,8 @@ pub mod s3_session {
 
         let client_config = S3ClientConfig::default()
             .part_size(test_config.part_size)
-            .endpoint_config(EndpointConfig::new(&region));
+            .endpoint_config(EndpointConfig::new(&region))
+            .auth_config(test_config.auth_config);
         let client = S3CrtClient::new(client_config).unwrap();
         let runtime = client.event_loop_group();
         let prefetcher = default_prefetch(runtime, test_config.prefetcher_config);
--- a/mountpoint-s3/tests/fuse_tests/fork_test.rs
+++ b/mountpoint-s3/tests/fuse_tests/fork_test.rs
@@ -564,7 +564,7 @@ fn get_mount_from_source_and_mountpoint(source: &str, mount_point: &str) -> Opti
     let stdout_reader = BufReader::new(stdout);
     let stdout_lines = stdout_reader.lines();
 
-    for line in stdout_lines.flatten() {
+    for line in stdout_lines.map_while(Result::ok) {
         let str: Vec<&str> = line.split_whitespace().collect();
         let source_rec = str[0];
         let mount_point_rec = str[2];
--- a/mountpoint-s3/tests/fuse_tests/fork_test.rs
+++ b/mountpoint-s3/tests/fuse_tests/fork_test.rs
@@ -4,10 +4,14 @@
 use assert_cmd::prelude::*;
 #[cfg(not(feature = "s3express_tests"))]
 use aws_config::BehaviorVersion;
+#[cfg(all(feature = "sse_kms", not(feature = "s3express_tests")))]
+use aws_sdk_s3::primitives::ByteStream;
 #[cfg(not(feature = "s3express_tests"))]
 use aws_sdk_sts::config::Region;
 use std::fs;
 use std::io::{BufRead, BufReader};
+#[cfg(all(feature = "sse_kms", not(feature = "s3express_tests")))]
+use std::io::{Read, Write};
 use std::path::Path;
 use std::process::{Child, ExitStatus, Stdio};
 use std::time::{Duration, Instant};
@@ -16,6 +20,8 @@ use test_case::test_case;
 
 use crate::common::fuse::read_dir_to_entry_names;
 use crate::common::s3::{create_objects, get_test_bucket_and_prefix, get_test_bucket_forbidden, get_test_region};
+#[cfg(all(feature = "sse_kms", not(feature = "s3express_tests")))]
+use crate::common::s3::{get_scoped_down_credentials, get_test_kms_key_id, get_test_sdk_client};
 #[cfg(not(feature = "s3express_tests"))]
 use crate::common::s3::{get_subsession_iam_role, tokio_block_on};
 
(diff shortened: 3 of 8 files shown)


mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -481,3 +627,40 @@ fn unmount(mount_point: &Path) {
...
     panic!("failed to unmount");
 }
+
+#[cfg(all(feature = "sse_kms", not(feature = "s3express_tests")))]
+struct ChildProcess(std::process::Child);
+
+#[cfg(all(feature = "sse_kms", not(feature = "s3express_tests")))]
+impl Drop for ChildProcess {
+    fn drop(&mut self) {
+        if !self
```
passaro, line 637: This may need some comments. Also, is there no way to avoid changing the signature of `wait_for_exit`?

vladem: I was thinking of adding a `wait_for_exit_not_owning`, but was not able to come up with a good name, which rang a bell in me, that probably that is not a good idea :)

it could be something like this: 
```
fn wait_for_exit_not_owning(child: &mut Child) -> ExitStatus {}

fn wait_for_exitg(mut child: Child) -> ExitStatus {
    wait_for_exit(&mut child)
}
```

otherwise -- this `drop` method can not move out its `self.0`:
```
cannot move out of `self` which is behind a mutable reference
```

(I can remove the whole `ChildProcess` thing, but it has a convenient functionality of printing MP's logs on test failure)

passaro: Another option is to use `Option<Child>` (sorry) for the field and `take()` it in `drop`.

passaro: But the main question was: why are you killing the process? Other tests call `unmount` at the end, any reason for the difference? If so, it would need a comment.

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -380,6 +384,130 @@ fn mount_scoped_credentials() -> Result<(), Box<dyn std::error::Error>> {
...
+        .arg(format!("--region={region}"))
+        .arg(format!("--prefix={prefix}"))
+        .arg("--sse=aws:kms:dsse")
+        .arg(format!("--sse-kms-key-id={key_id}"))
+        .arg("--auto-unmount")
+        .arg("--foreground");
+    if let Some(credentials) = credentials {
+        cmd.env("AWS_ACCESS_KEY_ID", credentials.access_key_id())
+            .env("AWS_SECRET_ACCESS_KEY", credentials.secret_access_key())
+            .env("AWS_SESSION_TOKEN", credentials.session_token().unwrap());
```
dannycjones, line 409: should we assume the session token is always set?

i'd recommend we just check if its set and optionally apply it.

https://github.com/awslabs/mountpoint-s3/blob/16cb2ed1baf218083aa48e96c1059c53e6983680/mountpoint-s3/tests/auth-cli.rs#L34-L36

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -380,6 +384,130 @@ fn mount_scoped_credentials() -> Result<(), Box<dyn std::error::Error>> {
...
+    }
+    let child = cmd.spawn().expect("unable to spawn child");
+    wait_for_mount("mountpoint-s3", mount_point.to_str().unwrap());
+    child
+}
+
+#[cfg(all(feature = "sse_kms", not(feature = "s3express_tests")))]
+fn erroneous_write(mount_point: &Path) {
+    let mut f = fs::File::create(mount_point.join("file.txt")).expect("can not open file for write");
```
dannycjones, line 418: nit: these `expect` messages should state why you expected the value to be `Ok`/`Some`.

```suggestion
    let mut f = fs::File::create(mount_point.join("file.txt")).expect("should be able to open file for writing");
```

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -380,6 +384,130 @@ fn mount_scoped_credentials() -> Result<(), Box<dyn std::error::Error>> {
...
+            .env("AWS_SECRET_ACCESS_KEY", credentials.secret_access_key())
+            .env("AWS_SESSION_TOKEN", credentials.session_token().unwrap());
+    }
+    let child = cmd.spawn().expect("unable to spawn child");
+    wait_for_mount("mountpoint-s3", mount_point.to_str().unwrap());
+    child
+}
+
+#[cfg(all(feature = "sse_kms", not(feature = "s3express_tests")))]
+fn erroneous_write(mount_point: &Path) {
```
dannycjones, line 417: this method specifically is for erroneous writes related to KMS, let's rename it

same for other tests, lets make it clearer its for kms

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -380,6 +384,130 @@ fn mount_scoped_credentials() -> Result<(), Box<dyn std::error::Error>> {
...
+    assert!(
+        write_result.is_err(),
+        "should not be able to write to the file without proper sse"
+    );
+}
+
+#[cfg(all(feature = "sse_kms", not(feature = "s3express_tests")))]
+#[test]
+fn write_with_inexistent_key() {
+    let (bucket, prefix) = get_test_bucket_and_prefix("mount_and_check_log");
```
dannycjones, line 430: it's annoying / boilerplate-y, but this argument should be the name of the test

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -380,6 +384,130 @@ fn mount_scoped_credentials() -> Result<(), Box<dyn std::error::Error>> {
...
+        {"Effect": "Allow", "Action": ["s3:*"], "Resource": "*"}
+    ]}"#;
+    let credentials = tokio_block_on(get_scoped_down_credentials(policy_with_no_kms_perms));
+    let (bucket, prefix) = get_test_bucket_and_prefix("mount_and_check_log");
+
+    let mount_point = assert_fs::TempDir::new().expect("can not create a mount dir");
+    let child = mount_with_sse(&bucket, mount_point.path(), &prefix, &sse_key, Some(credentials));
+    erroneous_write(mount_point.path());
+    unmount_and_check_log(child, mount_point.path(), &expected_log_line);
+}
```
dannycjones, line 456: I'd recommend using a consistent ordering/sequence in these tests. Feel free to pick whatever order but something like...

- setup 
  - get bucket and prefix
  - prepare mount directory
  - get client, credentials, whatever
  - mount?
- prepare expected log line
- test/assertions

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -380,6 +384,130 @@ fn mount_scoped_credentials() -> Result<(), Box<dyn std::error::Error>> {
...
+
+#[cfg(all(feature = "sse_kms", not(feature = "s3express_tests")))]
+#[test]
+fn read_with_no_permissions_for_a_key() {
+    // create file
+    let (bucket, prefix) = get_test_bucket_and_prefix("mount_and_check_log");
+    let mut test_client = crate::common::fuse::s3_session::create_test_client(&get_test_region(), &bucket, &prefix);
```
dannycjones, line 463: We should use Rust AWS SDK here.

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -380,6 +384,130 @@ fn mount_scoped_credentials() -> Result<(), Box<dyn std::error::Error>> {
...
+    let mount_point = assert_fs::TempDir::new().expect("can not create a mount dir");
+    let child = mount_with_sse(&bucket, mount_point.path(), &prefix, &sse_key, Some(credentials));
+    erroneous_write(mount_point.path());
+    unmount_and_check_log(child, mount_point.path(), &expected_log_line);
+}
+
+#[cfg(all(feature = "sse_kms", not(feature = "s3express_tests")))]
+#[test]
+fn read_with_no_permissions_for_a_key() {
+    // create file
```
dannycjones, line 461: when separating the test into sections, we could use scopes so its clear what set of code that the comment is relevant for.

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -380,6 +384,130 @@ fn mount_scoped_credentials() -> Result<(), Box<dyn std::error::Error>> {
...
+        );
+
+        let unencrypted_object = mount_point.join(unencrypted_object);
+        let mut f = fs::File::open(unencrypted_object).expect("can not open file for read");
+        let read_result = f.read_to_end(&mut data);
+        assert!(
+            read_result.is_ok(),
+            "should not able to read a default-encrypted file after the first read failure"
+        );
+    }
```
dannycjones, line 507: why is this in a scoped block? not against it, just the reason isn't clear

I guess to close the files? let's add a comment to the scope

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -12,8 +12,12 @@ use tempfile::TempDir;
 use test_case::test_case;
 
 use mountpoint_s3::S3FilesystemConfig;
+#[cfg(all(feature = "s3_tests", not(feature = "s3express_tests")))]
+use mountpoint_s3::ServerSideEncryption;
 
 use crate::common::fuse::{self, read_dir_to_entry_names, TestClientBox, TestSessionConfig};
+#[cfg(all(feature = "s3_tests", not(feature = "s3express_tests")))]
+use crate::common::{get_scoped_down_credentials, s3::get_test_kms_key_id, s3::tokio_block_on};
```
dannycjones, line 20: please separate different module imports

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -807,6 +811,76 @@ where
     assert!(hello_contents.is_empty());
 }
 
+// This test checks that a write can be performed when IAM session policy inforces the usage of the specific SSE type and a KMS key ID
+// This test also contains error cases, that check that IAM session policy actually rejects writes with wrong SSE
+#[cfg(all(feature = "s3_tests", not(feature = "s3express_tests")))]
+#[test_case(ServerSideEncryption::new(None, None), true)]
+#[test_case(ServerSideEncryption::new(Some("aws:kms".to_owned()), None), true)]
+#[test_case(ServerSideEncryption::new(Some("aws:kms".to_owned()), Some(get_test_kms_key_id())), false)]
+fn write_with_sse_settings_test(sse: ServerSideEncryption, should_fail: bool) {
```
dannycjones, line 820: typo

```suggestion
// This test checks that a write can be performed when IAM session policy enforces the usage of the specific SSE type and a KMS key ID
// This test also contains error cases, that check that IAM session policy actually rejects writes with wrong SSE
#[cfg(all(feature = "s3_tests", not(feature = "s3express_tests")))]
#[test_case(ServerSideEncryption::new(None, None), true)]
#[test_case(ServerSideEncryption::new(Some("aws:kms".to_owned()), None), true)]
#[test_case(ServerSideEncryption::new(Some("aws:kms".to_owned()), Some(get_test_kms_key_id())), false)]
fn write_with_sse_settings_test(sse: ServerSideEncryption, should_fail: bool) {
```

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -807,6 +811,76 @@ where
...
+            write_result.is_err(),
+            "should not be able to write to the file without proper sse"
+        );
+        return;
+    }
+
+    assert!(
+        write_result.is_ok(),
+        "should be able to write to the file with proper sse"
+    );
```
dannycjones, line 871: prefer expect, it should hopefully provide better information to see why it wasn't OK?

mountpoint-s3/tests/common/mod.rs
```diff
@@ -97,3 +102,36 @@ fn init_tracing_subscriber() {
...
+        .credentials()
+        .expect("credentials should be present if assume_role succeeded")
+        .to_owned();
+    Credentials::new(
+        credentials.access_key_id().unwrap(),
+        credentials.secret_access_key().unwrap(),
+        credentials.session_token().map(|s| s.to_owned()),
+        None,
+        "scoped_down_sts_creds",
+    )
```
dannycjones, line 136: I really should get #684 merged.

We need to redact these credentials if you're going to use them with `Command`, since it prints out the environment on failure.

For example:

https://github.com/awslabs/mountpoint-s3/blob/16cb2ed1baf218083aa48e96c1059c53e6983680/mountpoint-s3-client/tests/common/creds.rs#L25-L38

dannycjones (dismissed):
A few comments (sorry), a lot of nitpicky comments anyway. Generally looks good!

mountpoint-s3/src/main.rs
```diff
@@ -584,7 +603,10 @@ fn mount(args: CliArgs) -> anyhow::Result<FuseSession> {
     filesystem_config.allow_overwrite = args.allow_overwrite;
     filesystem_config.s3_personality =
         get_s3_personality(args.bucket_type, &args.bucket_name, client.endpoint_config());
-
+    #[cfg(feature = "sse_kms")]
+    {
+        filesystem_config.server_side_encryption = ServerSideEncryption::new(args.sse, args.sse_kms_key_id);
+    }
```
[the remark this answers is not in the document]
dannycjones: ah, it is needed! this isn't stable yet, booo

mountpoint-s3/src/fs.rs
```diff
@@ -411,6 +414,30 @@ impl S3Personality {
...
+            sse_type,
+            sse_kms_key_id,
+        }
+    }
+
+    /// Verfify the checksum and return a the SSE type and the AWS KMS Key ID if provided
+    pub fn into_inner(&self) -> (Option<String>, Option<String>) {
+        // TODO: verify checksum
+        (self.sse_type.clone(), self.sse_kms_key_id.clone())
+    }
```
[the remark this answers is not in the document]
passaro: A good reason to implement an `into_...` method and consume `self` is that we want to validate the content and return it directly, that is avoid any `clone` **after** verifying the checksum.

However, since this method is only useful once the checksum verification is implemented, we could also discuss it in that future PR.

mountpoint-s3/src/main.rs
```diff
@@ -266,6 +268,23 @@ struct CliArgs {
...
     )]
     pub user_agent_prefix: Option<String>,
+
+    #[cfg(feature = "sse_kms")]
+    #[clap(
+        long,
+        help = "Set the server-side encryption algorithm for new objects",
+        help_heading = BUCKET_OPTIONS_HEADER,
+        value_parser = clap::builder::PossibleValuesParser::new(["aws:kms", "aws:kms:dsse"]))]
+    pub sse: Option<String>,
```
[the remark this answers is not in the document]
passaro: Sorry, I pushed in that direction in a previous comment. You are right, we have not been consistent at all.

mountpoint-s3/tests/fuse_tests/fork_test.rs
passaro: +1, and not make `create_test_client` public.

mountpoint-s3/src/main.rs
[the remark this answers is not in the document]
dannycjones: Maybe we should go back and make things consistent, but that's a discussion for another PR :)

mountpoint-s3/tests/common/mod.rs
vladem: to make it work we will need to add `--nocapture` [here](https://github.com/awslabs/mountpoint-s3/blob/main/.github/workflows/integration.yml#L98), right?

mountpoint-s3/src/fs.rs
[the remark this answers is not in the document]
vladem: I thought about adding a enum here but after all it just looked redundant to me (correct me if I am wrong). I see the interface of this struct something similar to what we have in current revision, ie receiving strings (provided as CLI args) and returning strings to construct a `PutObject` request. That works well for KMS/KMS-DSSE, because we do not do anything different depending on KMS/KMS-DSSE/default, so we do not need pattern matching on enum variants.

Now SSE-C. Its true that it won't fit very well in current version of this struct. `sse_kms_key_id` argument of `new` will need to be something like `key_material` or `key_material_path_on_disk`. We will probably need to do some pattern matching depending on `sse_type` **inside** implementation of this struct, e.g. to read the material from disk in case of SSE-C. But still it seems like it better to be encapsulated by this struct, right? Meaning the interface will stay the same: `new` and `into_inner`. Thus even if we will want to add a enum in future, it will be a very straightforward change, affecting these 20 lines of `ServerSideEncryption` implementation only.

mountpoint-s3/tests/fuse_tests/fork_test.rs
vladem: exactly for that, will add a comment

mountpoint-s3/tests/common/fuse.rs
```diff
@@ -363,6 +382,18 @@ pub mod s3_session {
...
+                None => (),
+                Some("aws:kms") => {
+                    request = request.set_server_side_encryption(Some(aws_sdk_s3::types::ServerSideEncryption::AwsKms))
+                }
+                Some("aws:kms:dsse") => {
+                    request =
+                        request.set_server_side_encryption(Some(aws_sdk_s3::types::ServerSideEncryption::AwsKmsDsse))
+                }
+                _ => panic!("unexpected sse type was used in a test"),
+            }
```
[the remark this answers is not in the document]
vladem: i am removing this bit completely, since I do not see any use for it right now (will use SDK client to upload objects from tests)

mountpoint-s3/tests/fuse_tests/fork_test.rs
vladem: added an `sse` suffix in all of them

mountpoint-s3/src/fs.rs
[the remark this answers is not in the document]
passaro: I struggle to see how you would implement checksum verification in the `unpack` method (who copies the Strings and when matters).

May I suggest reverting to the two getter methods in this PR? We can think about the best approach when we add the checksums.

mountpoint-s3/tests/common/mod.rs
dannycjones: That's a good thing to question...

I believe that it shouldn't matter. If the test fails, then the captured output should then be flushed to stdout and the masking pattern should be processed.

I wonder if it matters if Cargo switches it to `stderr`...

mountpoint-s3/src/fs.rs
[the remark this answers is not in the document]
dannycjones: Yep, let's just use two getters and tackle the checksumming in its own PR.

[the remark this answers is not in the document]
vladem: yeah, a little of back and forth, but two getters seem better then returning a tuple

mountpoint-s3/tests/fuse_tests/write_test.rs
vladem: they are now in the same module, so done :)

mountpoint-s3/tests/fuse_tests/write_test.rs
vladem: looks nicer for sure, will print an error on failure, thats true

mountpoint-s3/tests/common/mod.rs
vladem: Not much about output streams in the doc: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#masking-a-value-in-a-log

To be completely safe we can print this both to stderr and stdout; thus we are making sure that any other occurrence of creds in either of the streams (e.g. because of panic) will come after a masking line `::add-mask::..`. 

Currently, it *may* happen, depending on github's runner implementation, that an unmasked value in stderr will be processed earlier than the `::add-mask::..` line in stdout

vladem: I am leaving this as it is, only printing to `stdout`.

First of all, currently, all output of the test (even of `eprintln!()`) goes to `stdout` of cargo process, e.g:
```
# looking at stderr only
$ cargo test -- test_cmd_env 1>/dev/null
    Finished test [unoptimized + debuginfo] target(s) in 0.30s
     Running unittests src/lib.rs (target/debug/deps/fuser-7c62d545ed9ded77)
     Running tests/integration_tests.rs (target/debug/deps/integration_tests-c4309ec71a67c1eb)
     Running unittests src/lib.rs (target/debug/deps/mountpoint_s3-621df098ed460e0c)
     Running unittests src/main.rs (target/debug/deps/mount_s3-04d2109b3c02fbc6)
     Running tests/cli.rs (target/debug/deps/cli-74459daf07d48e44)
error: test failed, to rerun pass `-p mountpoint-s3 --test cli`
```

```
# looking at stdout only
$ cargo test -- test_cmd_env 2>/dev/null
<...>
failures:

---- test_cmd_env stdout ----
::add-mask::extraodinary_secret_value_stdout
::add-mask::extraodinary_secret_value_stderr
printing this once again: extraodinary_secret_value_stdout
printing this once again: extraodinary_secret_value_stderr
thread 'test_cmd_env' panicked at mountpoint-s3/tests/cli.rs:222:5:
fail the test


failures:
    test_cmd_env

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 14 filtered out; finished in 0.00s
```

There were a couple of suggestions of printing that to stderr, but all closed with explanation that "machine-redable output will always go to stdout", output of the test falls into that category:
https://github.com/actions/runner/issues/891
https://github.com/rust-lang/cargo/issues/5641

passaro (dismissed):
Looks good. Just a completely optional suggestion.

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -380,6 +386,133 @@ fn mount_scoped_credentials() -> Result<(), Box<dyn std::error::Error>> {
...
+    let child = cmd.spawn().expect("unable to spawn child");
+    wait_for_mount("mountpoint-s3", mount_point.to_str().unwrap());
+    child
+}
+
+#[cfg(all(feature = "sse_kms", not(feature = "s3express_tests")))]
+fn erroneous_write_sse(mount_point: &Path) {
+    let mut f = fs::File::create(mount_point.join("file.txt")).expect("should be able to open file for writing");
+    let data = vec![0xaa; 32];
+    let write_result = f.write_all(&data);
```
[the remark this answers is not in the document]
vladem: sounds good to me, thanks for the review! renamed this to be `write_to_file` as well

passaro (approved):

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -432,7 +564,7 @@ fn get_mount_from_source_and_mountpoint(source: &str, mount_point: &str) -> Opti
     let stdout_reader = BufReader::new(stdout);
     let stdout_lines = stdout_reader.lines();
 
-    for line in stdout_lines.flatten() {
+    for line in stdout_lines.map_while(Result::ok) {
```
passaro, line 567: This is in #742, why include it here?

passaro (approved):

PR #716, opened 2024-01-25
Fix decrement of FH gauge for RW handles that try to write to existing files

dannycjones:
## Description of change

In the logs, I was seeing that the gauge for unassigned file handles was not being decremented properly and increased over time with each attempt to write to a RW file handle where the file already exists.

For example:

```
echo 'hello-world' > /mnt/s3/existing-file.txt
```

```
2024-01-23T15:44:32.700092Z  INFO mountpoint_s3::metrics: fs.current_handles[type=unassigned]: 2
2024-01-23T15:44:37.700419Z  INFO mountpoint_s3::metrics: fs.current_handles[type=unassigned]: 2
2024-01-23T15:44:42.700741Z  INFO mountpoint_s3::metrics: fs.current_handles[type=unassigned]: 2
```

This change quickly fixes the issue.

In future, we may consider treating file handles more like a state machine which manages its own gauges and reduces the amount of places where we have to think about incrementing and decrementing these gauges.

No breaking change. Fixes a counter.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -1145,6 +1145,7 @@ where
         let mut state = file_handle.state.into_inner();
         let request = match state {
             FileHandleState::Created { lookup, flags, pid } => {
+                metrics::decrement_gauge!("fs.current_handles", 1.0, "type" => "unassigned");
                 // This happens when release is called before any read() or write(),
                 // since we don't know what type of handle it would be we need to consider
                 // what to do next for both cases.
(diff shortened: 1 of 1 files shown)


mountpoint-s3/src/fs.rs
```diff
@@ -1168,8 +1168,8 @@ where
             }
             FileHandleState::Read { .. } => {
                 // TODO make sure we cancel the inflight PrefetchingGetRequest. is just dropping enough?
-                file_handle.inode.finish_reading()?;
                 metrics::decrement_gauge!("fs.current_handles", 1.0, "type" => "read");
+                file_handle.inode.finish_reading()?;
```
dannycjones, line 1172: Swapped these two lines since we have a small concern that in the future, `finish_reading` could fail and we wouldn't decrement the gauge. It currently doesn't do anything though.

passaro (approved):
LGTM!

arsh (approved):
It would have been nice to see the before and after in your description of the change.

dannycjones:
> It would have been nice to see the before and after in your description of the change.
https://github.com/awslabs/mountpoint-s3/pull/716#pullrequestreview-1844041904

Makes sense, the after is no metric being emitted. Gauges aren't emitted as metrics when they have the value zero.

PR #718, opened 2024-01-25
Reduce error logging noise for lookup and throughput configuration

sauraank:
## Description of change
- Downlevel the `FileDoesNotExist` inode error in case of lookup to debug level as it s default behaviour in cases of creation of a new file. It should not be logged as error at default level. Also downlevel the `create` fuse operation not supported to debug level. `create` is replaced by `mknod` and `open` in Mountpoint and is not an error.
- For the condition of overwrite, setattr is called with size having `Some` value, but for modifying metadata setattr is called with size `None`. So, I have used that to distinguish file overwrite message from modifying metadata in `setattr`.
- Downlevel the EC2 instance info not retrieved error to debug. It was not useful in case of non-EC2 instance. And it still appears if maximum trhoughput is not set using CLI flag because mountpoint need it to configure throughput. Also added information that it sets to default throughput if not able auto-configure. Earlier, the failure appeared as an error. But, we still set it to default value.

I changed the macro to log `fuse_error` messages. Other than than replaced the error messages as follows -
In case of failing to detect network thorughput, we get warn message like:

```
 WARN mount_s3: failed to detect network throughput: failed to get instance type: IMDS query failed: Invalid response from Imds: . Using 10 gbps as throughput.
Use --maximum-throughput-gbps CLI flag to configure a target throughput appropriate for the instance.
```

If trying to overwrite a file without setting `--allow-overwrite` flag, following is the warn message recieved:

```
WARN setattr{req=11 ino=2 name="twin.txt"}: mountpoint_s3::fuse: setattr failed: file overwrite is disabled by default, you need to remount with --allow-overwrite flag and open the file in truncate mode (O_TRUNC) to overwrite it
```

Also, the lookup failure for file not existing is downlevel to DEBUG:

```
2024-01-30T19:07:21.643021Z DEBUG lookup{req=4 ino=1 name="404"}: mountpoint_s3::fuse: lookup failed: file does not exist: file "404" does not exist in parent inode 1 (full key "")
```

Relevant issues: 
#533 
No, it does not impact existing behaviour.

No breaking change.

---


passaro (commented):
A few questions, mostly around macros.

mountpoint-s3/src/fs.rs
```diff
@@ -604,15 +612,28 @@ where
...
             _flags,
             atime,
-            mtime
+            mtime,
+            size
         );
-        let lookup = self.superblock.setattr(&self.client, ino, atime, mtime).await?;
+        let lookup = match (self.superblock.setattr(&self.client, ino, atime, mtime).await, size) {
+            (Ok(lookup), _) => lookup,
+            (Err(e), Some(_)) => {
```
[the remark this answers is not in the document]
sauraank: I need to update it in comments. Actually, for the condition of overwrite, setattr is called with size having `Some` value, but for modifying metadata setattr is called with size `None`. So, I have used that to distinguish file overwrite message from modifying metadata in setattr.

mountpoint-s3/src/fs/error.rs
```diff
@@ -32,25 +34,32 @@ use crate::upload::UploadWriteError;
 #[macro_export]
 macro_rules! err {
     // Base case -- don't use directly
-    ($errno:expr, __source:$source:expr, $message:literal, $($args:tt)*) => {
+    ($errno:expr, __source:$source:expr, $level:path, $message:literal, $($args:tt)*) => {
```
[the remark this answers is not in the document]
sauraank: I tried on using expr but it was giving error. Let me try again.

mountpoint-s3/src/fs/error.rs
```diff
@@ -61,6 +70,13 @@ pub struct Error {
     pub(crate) errno: libc::c_int,
     pub(crate) message: String,
     pub(crate) source: Option<anyhow::Error>,
+    pub(crate) level: Level,
+}
+
+impl Error {
+    pub(crate) fn get_log_level(&self) -> tracing::Level {
```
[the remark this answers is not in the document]
sauraank: I think I should make it private. Because, we keep the fields of a struct private right?

mountpoint-s3/src/inode.rs
```diff
@@ -1571,7 +1571,8 @@ pub enum InodeError {
     UnlinkNotPermittedWhileWriting(InodeErrorInfo),
     #[error("corrupted metadata for inode {0}")]
     CorruptedMetadata(InodeErrorInfo),
-    #[error("inode {0} is a remote inode and its attributes cannot be modified")]
+    // Cannot distinguish the case of overwrite from other cases based on any flags or mode for setattr()
```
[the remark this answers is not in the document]
sauraank: Yes. It was before I identified the way to differentiate the overwrite from modifying metadata. Let me update it.

mountpoint-s3/src/fuse.rs
```diff
@@ -25,23 +25,38 @@ pub mod session;
...
     }};
     ($name:literal, $reply:expr) => {
-        fuse_unsupported!($name, $reply, libc::ENOSYS);
-    }
+        fuse_unsupported!($name, $reply, libc::ENOSYS, tracing::Level::WARN)
+    };
+    ($name:literal, $reply:expr, $err:expr) => {
+        fuse_unsupported!($name, $reply, $err,tracing::Level::WARN)
+    };
 }
```
dannycjones, line 60: Same feedback as `fuse_error!(...)` above about using `event!(...)` macro.

mountpoint-s3/src/fuse.rs
```diff
@@ -25,23 +25,38 @@ pub mod session;
...
         ::metrics::counter!("fuse.op_unimplemented", 1, "op" => $name);
         $reply.error($err);
     }};
     ($name:literal, $reply:expr) => {
-        fuse_unsupported!($name, $reply, libc::ENOSYS);
-    }
+        fuse_unsupported!($name, $reply, libc::ENOSYS, tracing::Level::WARN)
+    };
+    ($name:literal, $reply:expr, $err:expr) => {
+        fuse_unsupported!($name, $reply, $err,tracing::Level::WARN)
```
dannycjones, line 58: nit:

```suggestion
        fuse_unsupported!($name, $reply, $err, tracing::Level::WARN)
```

mountpoint-s3/src/fs.rs
```diff
@@ -604,15 +612,31 @@ where
         atime: Option<OffsetDateTime>,
         mtime: Option<OffsetDateTime>,
         _flags: Option<u32>,
+        size: Option<u64>,
```
dannycjones, line 615: nit: keep flags as the last argument

mountpoint-s3/src/fs.rs
```diff
@@ -604,15 +612,31 @@ where
...
+            "fs:setattr with ino {:?} flags {:?} atime {:?} mtime {:?} size {:?}",
             ino,
             _flags,
             atime,
-            mtime
+            mtime,
+            size
         );
-        let lookup = self.superblock.setattr(&self.client, ino, atime, mtime).await?;
+        let lookup = match (self.superblock.setattr(&self.client, ino, atime, mtime).await, size) {
```
dannycjones, line 625: please move the `setattr` call above the match, and match on a variable like `setattr_result`

mountpoint-s3/src/fs.rs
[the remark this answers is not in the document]
dannycjones: I think we need to crisp up the comment a bit.

What we're doing is ignoring all other arguments if size is set, to return a specific error that overwrite isn't supported. Maybe customer's application actually wants to set two things, one of which is size, and we'll just tell them to set allow overwrite to fix...

Maybe that's OK, but we need to consciously make that decision. (I need to think a bit more on whether this is the right thing to do, I don't have context on this decision.)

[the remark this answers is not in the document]
dannycjones: Please also use `_size` so its clear to reader what we're ignoring without needing to look back up at the match.

mountpoint-s3/src/fs.rs
```diff
@@ -604,15 +612,31 @@ where
...
+            // For file overwrite, setattr is called with its size field having Some value.
+            // but for other cases like modifying atime, mtime etc the size field have None value.
+            // So, using that as distinguishing factor for both the errors.
+            (Err(e), Some(_)) => {
+                match e {
+                    InodeError::SetAttrNotPermittedOnRemoteInode(_) => return Err(
+                        err!(libc::EPERM, "file overwrite is disabled by default, you need to remount with --allow-overwrite flag and open the file in truncate mode (O_TRUNC) to overwrite it"),
+                    ),
+                    _ => return Err(e.into()),
+                }
```
dannycjones, line 636: You can probably put `InodeError::SetAttrNotPermittedOnRemoteInode` into the pattern match arm, and then let it fall through to the next arm if it doesn't match.

dannycjones (changes requested):
Thanks Ankit!

The macros are tough and I think we need to revisit them a bit. I also put a few comments inline on style or just general discussion on the change.

mountpoint-s3/src/fuse.rs
```diff
@@ -25,23 +25,38 @@ pub mod session;
...
+            ::tracing::Level::ERROR => ::tracing::error!("{} failed: {:#}", $name, err),
+            ::tracing::Level::WARN => ::tracing::warn!("{} failed: {:#}", $name, err),
+            ::tracing::Level::INFO => ::tracing::info!("{} failed: {:#}", $name, err),
+            ::tracing::Level::DEBUG => ::tracing::debug!("{} failed: {:#}", $name, err),
+            ::tracing::Level::TRACE => ::tracing::trace!("{} failed: {:#}", $name, err),
+        }
         ::metrics::counter!("fuse.op_failures", 1, "op" => $name);
         $reply.error(err.to_errno());
     }};
 }
```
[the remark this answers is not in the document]
jamesbornholt: I think this is the same idea as we already have here:

https://github.com/awslabs/mountpoint-s3/blob/e52d68ede985ecdb31f68e8db2beef29b528c8b1/mountpoint-s3-client/src/s3_crt_client.rs#L68-L78

The problem is that while tracing's `event!` lets you provide a level, it needs to be known at compile time. It's a bit annoying.

That said, we should totally factor this out into an `event!` macro, especially since you repeat the same idea twice in this file now (once for `fuse_error` and once for `fuse_unsupported`).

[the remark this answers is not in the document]
dannycjones: Ah, my bad. That's annoying!

OK, then this looks good. Let's factor as an `event!` macro with a short rustdoc explaining why its needed.

mountpoint-s3/src/main.rs
```diff
@@ -507,7 +507,12 @@ fn mount(args: CliArgs) -> anyhow::Result<FuseSession> {
         match autoconfigure::network_throughput(&instance_info) {
             Ok(throughput) => throughput,
             Err(e) => {
-                tracing::warn!("failed to detect network throughput: {:?}", e);
+                tracing::warn!(
+                    "failed to detect network throughput: {:?}. Using {} gbps as throughput.
+                    Use --maximum-throughput-gbps CLI flag to configure a target throughput appropriate for the instance.",
+                    e,
+                    DEFAULT_TARGET_THROUGHPUT
+                    );
```
dannycjones, line 515: how's this? I add `\` since the current string would add a newline and white-space/indentation to the logs.

```suggestion
                tracing::warn!(
                    "Failed to detect network throughput, using default {DEFAULT_TARGET_THROUGHPUT} Gbps. \
                    Use '--maximum-throughput-gbps' mount option to configure a different value. Detection failed due to: {e:?}"
                );
```

mountpoint-s3/src/fs/error.rs
```diff
@@ -32,25 +34,32 @@ use crate::upload::UploadWriteError;
...
     };
     ($errno:expr, $message:literal, $($args:tt)*) => {
-        err!($errno, __source:None, $message, $($args)*)
+        err!($errno, __source:None, ::tracing::Level::WARN, $message, $($args)*)
     };
     ($errno:expr, $message:literal) => {
-        err!($errno, __source:None, $message,)
+        err!($errno, __source:None, ::tracing::Level::WARN, $message,)
     };
 }
```
dannycjones, line 64: What's the benefit of this being a macro? Would we be better served with using `Error` directly, and maybe setters for `source` and `level`?

Usage example...

```rust
let mut fs_err = fs::Error::new(libc::ENOENT, "file does not exist"); // may need `format!(...)` if we want args in there
fs_err.source(err);
fs_err.level(tracing::Level::DEBUG);
return Err(fs_err);
```

mountpoint-s3/src/main.rs
sauraank: Yes. this would be better. In this case I think we can surface the [error](https://github.com/awslabs/mountpoint-s3/blob/e52d68ede985ecdb31f68e8db2beef29b528c8b1/mountpoint-s3-client/src/instance_info.rs#L31) "EC2 instance not retrieved" at warn level. Earlier, I was thinking to change it to `debug!` as we are again logging it here.

mountpoint-s3/src/fs.rs
```diff
@@ -577,7 +577,15 @@ where
...
-        let lookup = self.superblock.lookup(&self.client, parent, name).await?;
+        let lookup = match self.superblock.lookup(&self.client, parent, name).await {
+            Ok(lookup) => lookup,
+            Err(error) => match error {
+                InodeError::FileDoesNotExist(_, _) => {
+                    return Err(err!(libc::ENOENT, source: error, Level::DEBUG, "file does not exist",))
+                }
+                _ => return Err(error.into()),
+            },
+        };
```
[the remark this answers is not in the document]
dannycjones: Sorry - this suggestion had a bunch of typos, I updated them.

`err` was typo'd as `error`, second arm needs an `into()`. I also forgot the `?` operator at the end to return the errors.

mountpoint-s3/src/fs/error.rs
dannycjones: Any thoughts/preference @passaro / @monthonk?

sauraank: Yes. It can work by using setters. I thought of having the macro as it was already there for other cases. And maybe using macro, we can directly define the error at one go with the number of parameters we want to define.
Also, in the above way, we will have to set a default `source` and `level`. Should we do that?

dannycjones: Let's hold off on it for now.

arsh: My 2 cents. If using setters, how do you ensure, for example that both source and level are set (if that is desired)?

If feels that using the macro also allows more flexibility with the types, for example if we need to change them later.

mountpoint-s3/src/fs.rs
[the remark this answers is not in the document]
sauraank: Here, I am trying to surface file overwrite error in the logs. At present, to overwrite a file, call `setatttr` is called first when `--allow-overwrite` flag is not used. So, `SetAttrNotPermittedOnRemoteInode` error is logged. 
In case of file overwrite, `setattr` is called size parameter having `Some` value. When we try to modify metadata we  have size as `None` in setattr. So, I was trying to use that to distinguish file overwrite error logging from general metadata modification on remote file error.
Note that I am only checking for size only  if we get the `SetAttrNotPermittedOnRemoteInode` error, so we do not falsely include other case of error.

mountpoint-s3/src/fs.rs
```diff
@@ -603,16 +613,29 @@ where
...
+        let setattr_result = self.superblock.setattr(&self.client, ino, atime, mtime).await;
+        let lookup = match (setattr_result, _size) {
+            (Ok(lookup), _) => lookup,
+            (Err(InodeError::SetAttrNotPermittedOnRemoteInode(_)), Some(_)) => {
+                // For file overwrite(in case of not having `--file-overwrite` flag), first `setattr()` is called with `Some` value of size, whereas for only metadata modification size is `None`.
+                // In both cases, `setattr()` fails if the call is made on remote file.
+                // So, distinguishing file overwrite error from general metadata modification for remote file error.
+                return Err(
+                    err!(libc::EPERM, "file overwrite is disabled by default, you need to remount with --allow-overwrite flag and open the file in truncate mode (O_TRUNC) to overwrite it"));
+            }
```
dannycjones, line 636: What I want to call out here is that we are explicitly making the trade-off: if we see that size is set, then we assume that it is an attempt to make a truncate overwrite.

Maybe something like...

```rust
// We want to provide better feedback to users to prompt them to opt-in to file overwrites if it looks like what the application needs.
// Instead of complex logic to match `setattr` truncation only, we just check for the error and if the size was set in the request.
// If so, we assume its probably a truncation.
```

Two questions on this logic though:

- Does the value of the size matter? Will size ever be non-zero for overwrites that we support? We could explicitly match on the value too if it makes sense.
- Why do we not check if overwrites is actually enabled or not? This error doesn't make sense if customer has it enabled.

dannycjones (commented):
Looking good, I have just some questions remaining on our thinking for what cases we want to intercept the `setattr` error.

mountpoint-s3/src/fs.rs
sauraank: Unfortunately, I could find any good documentation of `setattr()` method of fuser and what value is set for size for different cases. Through my observations, it is always `Some(0)` for file overwrite and `None` for other cases. 
We can add the check if the overwrite is enabled or not. I did not add it, because when enabled, it does not call setattr. Fuse follows lookup -> open -> getxattr (not supported) -> write -> flush.

mountpoint-s3/src/inode.rs
```diff
@@ -1571,7 +1571,7 @@ pub enum InodeError {
     UnlinkNotPermittedWhileWriting(InodeErrorInfo),
     #[error("corrupted metadata for inode {0}")]
     CorruptedMetadata(InodeErrorInfo),
-    #[error("inode {0} is a remote inode and its attributes cannot be modified")]
+    #[error("cannot modify metadata for remote file at inode {0}")]
```
monthonk, line 1574: Is it always remote file? I think inode can also be a directory.

mountpoint-s3/src/fuse.rs
```diff
@@ -454,7 +471,7 @@ where
         _flags: i32,
         reply: ReplyCreate,
     ) {
-        fuse_unsupported!("create", reply);
+        fuse_unsupported!("create", reply, libc::ENOSYS, tracing::Level::DEBUG);
```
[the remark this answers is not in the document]
sauraank: I tried that. Issue is there is already a macro with 3 parameters, i.e., `$name:literal, $reply:expr, $err:expr`. So, it is not able to distinguish which macro rule do we want to use when it is called with 3 parameters.

mountpoint-s3/src/fs.rs
monthonk: There is a subtle detail here. If overwrite is enabled mountpoint will be initialized with `FUSE_ATOMIC_O_TRUNC` capability, that means the kernel will not call setattr on truncate. But at least we should add a comment why we don't need to check if overwrite is enabled here. https://github.com/awslabs/mountpoint-s3/blob/e52d68ede985ecdb31f68e8db2beef29b528c8b1/mountpoint-s3/src/fs.rs#L523-L532

mountpoint-s3/src/fuse.rs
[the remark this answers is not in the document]
monthonk: I see, because the level is also expression. I'm ok to leave it as is then.

mountpoint-s3/src/fs.rs
sauraank: Thanks @monthonk @dannycjones . I have added the check for overwrite in `setattr` as well just to be on the safe side even when there is any implementation change in future.

mountpoint-s3/src/fuse.rs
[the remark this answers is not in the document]
dannycjones: While I don't think we need that extra pattern (level but no errno), I did have a similar thought which is why I opened this thread: https://github.com/awslabs/mountpoint-s3/pull/718#discussion_r1469417290.

mountpoint-s3/src/fs.rs
```diff
@@ -603,16 +613,29 @@ where
...
+            (Ok(lookup), _, _) => lookup,
+            (Err(InodeError::SetAttrNotPermittedOnRemoteInode(_)), Some(0), false) => {
+                // We want to provide better feedback to users to prompt them to opt-in to file overwrites if it looks like what the application needs.
+                // Instead of complex logic to match `setattr` truncation only, we just check for the error and if the size was set in the request.
+                // If so, we assume its probably a truncation.
+                return Err(
+                    err!(libc::EPERM, "file overwrite is disabled by default, you need to remount with --allow-overwrite flag and open the file in truncate mode (O_TRUNC) to overwrite it"));
+            }
+            (Err(e), _, _) => return Err(e.into()),
+        };
```
dannycjones, line 638: There's a cool thing with the match statement you can do which is add an if condition using a 'match guard'. https://doc.rust-lang.org/reference/expressions/match-expr.html#match-guards

Try it out for the `allow_overwrite` check, I think it'll make the match neater and easier to follow.

mountpoint-s3/src/fs.rs
dannycjones: Makes sense, thanks Ankit! That sounds right.

mountpoint-s3/src/fs.rs
```diff
@@ -577,7 +577,17 @@ where
...
 
-        let lookup = self.superblock.lookup(&self.client, parent, name).await?;
+        let lookup = self
+            .superblock
+            .lookup(&self.client, parent, name)
+            .await
+            .map_err(|err| match err {
+                InodeError::FileDoesNotExist(_, _) => {
+                    // Lookup returning ENOENT is common case, and we dont want to warn in case `FileDoesNotExist` within ENOENT
+                    err!(libc::ENOENT, source: err, Level::DEBUG, "file does not exist")
```
dannycjones, line 587: For reference, this outputs like this (with debug logs on):

```
2024-01-30T19:07:21.643021Z DEBUG lookup{req=4 ino=1 name="404"}: mountpoint_s3::fuse: lookup failed: file does not exist: file "404" does not exist in parent inode 1 (full key "")
```

dannycjones (commented):
Looks good! Try out the match guard, but otherwise I think this is good to go.

dannycjones:
For reference, adding what a file not found looks like in the logs right now:

```
2024-01-30T19:07:21.643021Z DEBUG lookup{req=4 ino=1 name="404"}: mountpoint_s3::fuse: lookup failed: file does not exist: file "404" does not exist in parent inode 1 (full key "")
```

dannycjones:
@sauraank I encourage you to name the pull requests as you would the final commit (since that's how it gets merged).

We loosely follow the seven rules in this post at the moment. (The post is a good read another time!)

https://cbea.ms/git-commit/#seven-rules

mountpoint-s3/src/fs.rs
```diff
@@ -603,16 +613,29 @@ where
         ino: InodeNo,
         atime: Option<OffsetDateTime>,
         mtime: Option<OffsetDateTime>,
+        _size: Option<u64>,
```
dannycjones, line 616: We're using `_size` now, remove the underscore.

dannycjones (approved):
LGTM, thanks Ankit!

mountpoint-s3/src/fs.rs
```diff
@@ -603,16 +613,29 @@ where
...
-        let lookup = self.superblock.setattr(&self.client, ino, atime, mtime).await?;
+        let setattr_result = self.superblock.setattr(&self.client, ino, atime, mtime).await;
+        let lookup = match (setattr_result, size) {
+            (Ok(lookup), _) => lookup,
+            (Err(InodeError::SetAttrNotPermittedOnRemoteInode(_)), Some(0)) if !self.config.allow_overwrite => {
+                // We want to provide better feedback to users to prompt them to opt-in to file overwrites if it looks like what the application needs.
+                // Instead of complex logic to match `setattr` truncation only, we just check for the error and if the size was set in the request.
+                // If so, we assume its probably a truncation.
+                return Err(
+                    err!(libc::EPERM, "file overwrite is disabled by default, you need to remount with --allow-overwrite flag and open the file in truncate mode (O_TRUNC) to overwrite it"));
```
passaro, line 635: nit: the immediate error here is still that we do not support `setattr`, we are just guessing that the root cause is an attempt to overwrite. Maybe the message could reflect that, for example something like:

"{original_settattr_error}. If you want to overwrite an existing file, you need to remount ..."

sauraank: With this change the error we get is:
```
WARN setattr{req=11 ino=2 name="twin.txt"}: mountpoint_s3::fuse: setattr failed: file overwrite is disabled by default, you need to remount with --allow-overwrite flag and open the file in truncate mode (O_TRUNC) to overwrite it
```
It mentions the error is in setattr. And we are checking all the conditions for the root cause to be file overwrite. I think if we want to show both file attributes and file overwrite as root cause, we should not check the condition and keep it for all the `SetAttrOnRemoteInodeNotPermitted` as I was doing it earlier in this PR.

PR #723, opened 2024-01-31
Add a new mock client that simulates GET throughput

jamesbornholt:
## Description of change

For performance testing and microbenchmarking, we'd like to be able to separate the S3 service and the CRT datapath from our own client and file system. This mock client wraps our existing in-memory `MockClient`, but can simulate a target network throughput by rate-limiting the `get_object` stream. The goal is to be able to use this client in place of a regular `S3CrtClient` when we want to isolate performance questions. It doesn't yet rate-limit `put_object`.

Along the way, I realized we're including the mock client in our release builds, because it's an always-on feature of the client crate. This change therefore also does a little dependency refactoring to remove the mock and failure clients from the non-test dependency closure, by adding a new `mock` feature to the client crate that is off by default. I checked this works by seeing that the `mount-s3` release binary is a few MBs smaller, and that `strings mount-s3 | grep mock` no longer includes the mock client's symbols.

No.

---

### DIFF
--- a/mountpoint-s3-client/Cargo.toml
+++ b/mountpoint-s3-client/Cargo.toml
@@ -46,7 +46,7 @@ aws-sdk-s3 = "0.29.0"
 aws-sdk-sts = "0.29.0"
 aws-smithy-runtime-api = "0.56.1"
 bytes = "1.2.1"
-clap = "4.1.9"
+clap = { version = "4.1.9", features = ["derive"] }
 ctor = "0.1.23"
 proptest = "1.0.0"
 rusty-fork = "0.3.0"
--- a/mountpoint-s3-client/Cargo.toml
+++ b/mountpoint-s3-client/Cargo.toml
@@ -24,8 +24,6 @@ once_cell = "1.16.0"
 percent-encoding = "2.2.0"
 pin-project = "1.0.12"
 platform-info = "2.0.2"
-rand = "0.8.5"
-rand_chacha = "0.3.1"
 regex = "1.7.1"
 static_assertions = "1.1.0"
 thiserror = "1.0.34"
@@ -33,6 +31,12 @@ time = { version = "0.3.17", features = ["formatting", "parsing"] }
 tracing = { version = "0.1.35", default-features = false, features = ["std", "log"] }
 xmltree = "0.10.3"
 serde_json = "1.0.104"
+
+# Dependencies for the mock client only
+async-io = { version = "2.3.1", optional = true }
+async-lock = { version = "2.6.0", optional = true }
+rand = { version = "0.8.5", optional = true }
+rand_chacha = { version = "0.3.1", optional = true }
 
 [dev-dependencies]
 anyhow = { version = "1.0.64", features = ["backtrace"] }
@@ -51,6 +55,11 @@ test-case = "2.2.2"
 tokio = { version = "1.24.2", features = ["rt", "macros"] }
 tracing-subscriber = { version = "0.3.14", features = ["fmt", "env-filter"] }
 
+# HACK: we want our own tests to use the mock client, but don't want to enable it for consumers by
+# default, so we take a dev-dependency on ourself with that feature enabled.
+# https://github.com/rust-lang/cargo/issues/2911#issuecomment-749580481
+mountpoint-s3-client = { path = ".", features = ["mock"] }
+
 [build-dependencies]
 built = { version = "0.6.0", features = ["git2"] }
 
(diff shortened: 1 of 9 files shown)


jamesbornholt:
The culmination of this change plus #724 is a new `mock-mount-s3` binary that is identical to the real binary, but with a different backend that uses this new mock throughput client: https://github.com/jamesbornholt/mountpoint-s3/commit/dcce6bbade849a2d4e1528f35ca43bea125235f6

mountpoint-s3-client/src/mock_client/leaky_bucket.rs
```diff
@@ -0,0 +1,235 @@
+//! A token-based rate limiter.
+//!
+//! This is adapted from https://github.com/Gelbpunkt/leaky-bucket-lite (Apache 2.0) to remove the
+//! Tokio dependency and instead work with the generic async-io and async-lock crates that we use.
```
arsh, line 4: This is an interesting tradeoff. Since `mountpoint-s3-client` already has a dependency on `tokio`, I'm curious about what else led to the decision of reimplementing the leaky bucket instead of using that other crate? Obviously, this now is code we need to review, test, maintain, etc.

mountpoint-s3-client/src/mock_client/throughput_client.rs
```diff
@@ -0,0 +1,214 @@
...
+                });
+                let actual = start.elapsed().as_secs_f64();
+
+                assert_eq!(num_bytes, OBJECT_SIZE, "didn't stream entire object");
+
+                let expected = OBJECT_SIZE as f64 / (rate_gbps / 8.0 * 1.0e9);
+                assert!(
+                    actual > expected * 0.9,
+                    "too fast: rate_gbps={rate_gbps} actual={actual}s expected={expected}s"
+                );
```
arsh, line 205: Should we also check that the client isn't _too_ slow?

arsh (approved):
LGTM. Left some questions for my general understanding.

mountpoint-s3/Cargo.toml
```diff
@@ -7,7 +7,7 @@ publish = false
 
 [dependencies]
 fuser = { path = "../vendor/fuser", version = "0.14.0", features = ["abi-7-28"] }
-mountpoint-s3-client = { path = "../mountpoint-s3-client", version = "0.6.2" }
+mountpoint-s3-client = { path = "../mountpoint-s3-client", version = "0.6.2", default-features = false }
```
passaro, line 10: Is `default-features=false` required? Or just for extra caution?

passaro (commented):
Just a couple of questions.

mountpoint-s3-client/src/mock_client/throughput_client.rs
jamesbornholt: That's the check right below this one, but checking for "too slow" is pretty sensitive to slow hosts, so I'm leaving it commented out to avoid it becoming flaky in CI.

mountpoint-s3/Cargo.toml
jamesbornholt: Nope, just a leftover from when I had mock enabled by default at some point in writing this. Fixed.

mountpoint-s3-client/src/mock_client/leaky_bucket.rs
jamesbornholt: `mountpoint-s3-client` and `mountpoint-s3` both have dev-dependencies on tokio, but this would require a non-dev dependency (at least when the `mock` feature is enabled). In general we've tried to take an approach of not hardcoding an async runtime, because (for example) the CRT provides its own, and so all of our code should work on any async runtime. So while we could just say that users of this client need to use tokio, I felt like this code is simple and stable enough that it wouldn't hurt to pull it in like this.

mountpoint-s3-client/Cargo.toml
```diff
@@ -51,13 +55,20 @@ test-case = "2.2.2"
 tokio = { version = "1.24.2", features = ["rt", "macros"] }
 tracing-subscriber = { version = "0.3.14", features = ["fmt", "env-filter"] }
 
+# HACK: we want our own tests to use the mock client, but don't want to enable it for consumers by
+# default, so we take a dev-dependency on ourself with that feature enabled.
+# https://github.com/rust-lang/cargo/issues/2911#issuecomment-749580481
+mountpoint-s3-client = { path = ".", features = ["mock"] }
```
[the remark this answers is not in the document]
jamesbornholt: The `put_object` tests use it through `object_client_test!`, and the `client_benchmark` example now uses it too. At some point we should go through and make all the tests use `object_client_test!`.

passaro (approved):
LGTM

PR #731, opened 2024-02-05
Update and prune some dependencies

jamesbornholt:
## Description of change

This is a collection of four commits (probably best read individually) that aim to clean up our dependency closure a bit:
1. Take just enough version updates to remove several duplicate crates from our closure
2. Remove some crates and features from the release closure that were only being used in test code
3. Remove the dependency on `fs2`, which is old and unmaintained and has an equivalent function in `nix` anyway
4. Sorted the dependencies in Cargo.toml because I'm a pedant

The `fs2` change is slightly subtle as `nix`'s interface is different (it doesn't return space in bytes but space in blocks). But since we only care about the ratio, we don't need to do anything fancy.

I reviewed the changelogs for all the version updates. The only interesting one was `metrics`, which has had a big API refactor in 0.22, so I only updated to 0.21 for now. We can follow up on that later.

I also didn't update to the GA versions of the Rust SDK as that's quite a large change. I have a separate commit for that that I'll PR later.

No.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1694,16 +1694,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9c198f91728a82281a64e1f4f9eeb25d82cb32a5de251c6bd1b5154d63a8e7bd"
 
 [[package]]
-name = "libc-stdhandle"
-version = "0.1.0"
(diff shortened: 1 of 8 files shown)


arsh (commented):
Neat!

PR #741, opened 2024-02-07
Fix packaging workflow on AL2023

jamesbornholt:
## Description of change

This is prep for moving the CI to AL2023. That OS sets a very high ulimit for open files, and that seems to interact badly with `yum` on Centos 7. The net result is that our packaging workflow takes hours to install OS dependencies. There's an issue: https://github.com/docker/buildx/issues/379. A workaround suggested there is to just lower the ulimit, which works in my testing.

No.

---

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -196,7 +196,7 @@ jobs:
       with:
         ref: ${{ inputs.ref }}
         submodules: true
-    - name: Install nightly Rust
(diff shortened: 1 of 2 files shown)


passaro (approved):
Looks good! Thanks for fixing this and great timing: I started a run just yesterday and noticed it seemed slow, but only this morning I found out it had taken hours!

arsh (approved):

PR #743, opened 2024-02-08
Expose max_backoff_secs in exponential backoff retry options, increase default retries

larroy:
## Description of change

Using mount-s3 for distributed DDP training of ML models. I'm observing AWS_ERROR_S3_SLOW_DOWN throttling errors and system calls such as open failing.

Will increase the number of retries

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -245,7 +245,8 @@ impl S3CrtClientInner {
 
         let mut retry_strategy_options = StandardRetryOptions::default(&mut event_loop_group);
         // Match the SDK "legacy" retry strategies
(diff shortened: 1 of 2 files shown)


dannycjones:
Thanks for the contribution, @larroy!

I was hoping to understand a bit more about the problem first. If you can [open a bug report](https://github.com/awslabs/mountpoint-s3/issues/new?assignees=&labels=bug&projects=&template=bug-report.yml) and share some more details about your workload, that would help us understand the problem and the best way to solve it.

Things like this will be useful:
- How many files you're reading at the same time?
- How many files in total?
- What sizes are the files?
- Is there any repeated access to the same files / parts of files?
- Read pattern: sequential or random?
- How much you're seeing this error vs. successful reads?

PR #745, opened 2024-02-12
Verify SSE settings checksum before and after uploading an object

vladem:
## Description of change
1. Compute checksum of SSE settings and verify it before and after an upload request, in both cases we **panic** if the check failed;
   - [s3_crt_client/put_object.rs](https://github.com/awslabs/mountpoint-s3/compare/main...vladem:checksums?expand=1#diff-3cc8a34fe5c2b4ea653d83a6343362f790632f67f0dc12767fad37466f957e5e) now returns SSE headers instead of doing check within the client;
1. Disallow empty string for `--sse-kms-key-id` CLI arg;
1. Log SSE settings with INFO level to allow debugging;
1. Document new CLI args in CONFIGURATION.md; 

Relevant issues: *not linking yet*

This is not a breaking change, since SSE settings are under a feature flag.

---


doc/CONFIGURATION.md
```diff
@@ -158,7 +158,9 @@ If necessary, you can use the `--endpoint-url` command-line argument to fully ov
...
+
+If you would like to encrypt new objects written with Mountpoint using a different server-side encryption settings than the bucket's default, you can use `--sse <aws:kms|aws:kms:dsse>` and `--sse-kms-key-id example-key-id` command-line arguments. You can only override bucket's default SSE type with either `aws:kms` or `aws:kms:dsse`. When you are using one of those types, server-side encryption may be further configured to use a specific KMS key ID with a `--sse-kms-key-id` argument. When an `--sse` argument is used without a KMS key ID, S3 will use an [AWS managed KMS key](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#key-mgmt), which is created automatically. Please note, that these command line arguments do only configure server-side encryption for *new* objects, all of the *existing* objects will remain encrypted with settings specified on their creation.
```
dannycjones, line 163: A few small nits. We can also use more passive voice here.

```suggestion
New objects can be encrypted using different SSE settings than the bucket's default. To use a different setting, the argument `--sse <aws:kms|aws:kms:dsse>` should be used to specify the SSE encryption type, and `--sse-kms-key-id <KEY_ARN>` may be used to optionally specify a KMS key ID. When the `--sse` argument is used without providing a KMS key ID, S3 will use an [AWS managed KMS key](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#key-mgmt), which is created automatically. Please note that these command line arguments only configure server-side encryption for *new* objects created with Mountpoint, all of the *existing* objects will remain encrypted with settings specified at the time of their creation.
```

dannycjones (commented):
Just some quick feedback on the docs change, I'll need a bit more time to look over the code changes.

mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -197,63 +168,23 @@ impl PutObjectRequest for S3PutObjectRequest {
...
-            self.response_headers
-                .lock()
-                .expect("must be able to acquire headers lock")
+        let response_headers_lock = self
+            .response_headers
+            .lock()
+            .expect("must be able to acquire headers lock");
+        let response_headers = response_headers_lock
+            .as_ref()
+            .expect("PUT response headers must be available at this point");
```
passaro, line 177: nit: you can just take the value out here, like this:
```suggestion
        let response_headers = self
            .response_headers
            .lock()
            .expect("must be able to acquire headers lock")
            .take()
            .expect("PUT response headers must be available at this point");
```

mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -197,63 +168,23 @@ impl PutObjectRequest for S3PutObjectRequest {
...
+        let try_get_header_value = |headers: &Headers, key: &str| -> Option<String> {
+            headers
+                .get(key)
+                .ok()
```
passaro, line 181: nit: prefer a function if you are not closing over anything. Also, can be simplified a bit:
```suggestion
        fn try_get_header_value(headers: &Headers, key: &str) -> Option<String> {
            headers.get(key).ok()?.value().clone().into_string().ok()
        }
```

mountpoint-s3/src/fs.rs
```diff
@@ -425,30 +428,92 @@ impl S3Personality {
...
-    /// String representation of the SSE type as it is expected by S3 API
-    pub fn sse_type(&self) -> Option<String> {
-        self.sse_type.clone()
+    /// Computes the checksum of SSE settings by combining two strings containing the type and the key
+    /// Note, that this implementation yields the same result for Some("") and None, but we may safely
+    /// assume that it will never be called with an empty string as one of its parameters.
+    fn compute_checksum(sse_type: Option<&str>, sse_kms_key_id: Option<&str>) -> Crc32c {
+        let mut hasher: Hasher = Hasher::new();
+        if let Some(maybe_sse_type) = sse_type {
+            hasher.update(maybe_sse_type.as_bytes().as_ref());
```
passaro, line 465: Here and below:
```suggestion
            hasher.update(maybe_sse_type.as_bytes());
```

mountpoint-s3/src/fs.rs
```diff
@@ -425,30 +428,92 @@ impl S3Personality {
...
+
+    /// Checks that values provided as arguments to this function match the checksum stored in the object.
+    /// S3 will return some values for sse type and key even if they were not set on our side. We want to check
+    /// only the values which we set, thus response values are only added to checksum if they are present in this object.
+    pub fn verify_response(&self, sse_type: Option<&str>, sse_kms_key_id: Option<&str>) -> Result<(), IntegrityError> {
+        // we first validate in-memory values, since we are using them to compute a response's checksum
+        let computed = Self::compute_checksum(self.sse_type.as_deref(), self.sse_kms_key_id.as_deref());
+        if computed != self.checksum {
+            return Err(IntegrityError::ChecksumMismatch(self.checksum, computed));
+        }
```
passaro, line 493: nit: we could extract this into a `validate` methods (used here and in `into_inner`).

doc/CONFIGURATION.md
vladem: fixed

mountpoint-s3/src/fs.rs
```diff
@@ -425,30 +428,92 @@ impl S3Personality {
...
+        // computing a response's checksum, adding values only if they were present in a request
+        let sse_type = self.sse_type.as_ref().and(sse_type);
+        let sse_kms_key_id = self.sse_kms_key_id.as_ref().and(sse_kms_key_id);
+        let computed = Self::compute_checksum(sse_type, sse_kms_key_id);
+        if computed == self.checksum {
+            Ok(())
+        } else {
+            Err(IntegrityError::ChecksumMismatch(self.checksum, computed))
+        }
```
passaro, line 502: Using `.and` is a clever trick, maybe too clever? I'd rather revert to an explicit `if` block.

But more importantly, why compute and compare the checksum? Why not compare the actual strings like in the original `check_response_headers`?

doc/CONFIGURATION.md
```diff
@@ -158,7 +158,9 @@ If necessary, you can use the `--endpoint-url` command-line argument to fully ov
...
+
+New objects can be encrypted using different SSE settings than the bucket's default. To use a different setting, the argument `--sse <aws:kms|aws:kms:dsse>` should be used to specify the SSE encryption type, and `--sse-kms-key-id <KEY_ARN>` may be used to optionally specify a KMS key ID. When the `--sse` argument is used without providing a KMS key ID, S3 will use an [AWS managed KMS key](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#key-mgmt), which is created automatically. Please note that these command line arguments only configure server-side encryption for *new* objects created with Mountpoint, all of the *existing* objects will remain encrypted with settings specified at the time of their creation.
```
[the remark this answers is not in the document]
dannycjones: Yep, that's important. Let's move these docs to a follow-up PR.

mountpoint-s3/src/fs.rs
vladem: That's a good point, we can directly compare strings here, since we already validated that in-memory settings (`self.sse_type`, `self. sse_kms_key_id`) [match the checksum](https://github.com/awslabs/mountpoint-s3/pull/745/files/1a4cfc2e0656f203d916beddcbdd7453ce9c0321#diff-5397f952b2dcde7127c30d9a585c4e5a8491a0c70804e4b1e457c633637c63cdR489), so basically only compute the checksum once in `verify_response`. 

**We still need to compute the checksum once in this method though.** We will not do the check here, if an in-memory value is `None` (e.g. if `self.sse_kms_key_id == None`). Computing the checksum for in-memory values is the only way to ensure that the check won't be mistakenly skipped.

doc/CONFIGURATION.md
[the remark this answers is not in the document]
vladem: makes sense, good that we crisped that already though

mountpoint-s3-client/src/s3_crt_client/put_object.rs
vladem: looks more neat! we consume headers though and making them unavailable for further use from `self.response_headers`, but since it is in `review_and_complete` I believe there is no reason to expect we will ever need them after this point

mountpoint-s3/src/upload.rs
```diff
@@ -136,9 +145,19 @@ impl<Client: ObjectClient> UploadRequest<Client> {
...
-        self.request
+        let result = self
+            .request
             .review_and_complete(move |review| verify_checksums(review, size, checksum))
-            .await
+            .await?;
+        self.sse
+            .verify_response(result.sse_type.as_deref(), result.sse_kms_key_id.as_deref())
+            .unwrap_or_else(|err| {
+                panic!(
```
vladem, line 155: I've found surprising that panicing here does not result in Mountpoint being stopped (the same applies for `panic!` on line 95 in `UploadRequest::new`), it just prints the stacktrace and logs the failure for a single thread:
```
thread 'fuse-worker-3' panicked at /local/home/vlaad/local/mountpoint-s3/mountpoint-s3/src/upload.rs:155:17:
SSE settings were corrupted, object c0cfb6fe22bc40c8b0c8c35ecbb931c8 WAS uploaded to S3, panicing, error: SSE type mismatch. expected: Some("aws:kms"), actual: Some("aws:kms")
2024-02-14T16:50:27.421974Z  WARN fuser::reply: Reply not sent for operation 50, replying with I/O error
```

user application fails on `flush()` as expected, here is the output of a python script:
```
Traceback (most recent call last):
  File "/local/home/vlaad/local/mountpoint-s3/writer.py", line 24, in <module>
    f.flush()
OSError: [Errno 5] Input/output error
```

The behaviour which we were discussing previously (and I assumed it here) is [aborting on panic](https://doc.rust-lang.org/book/ch09-01-unrecoverable-errors-with-panic.html). I believe we don't want to enable this for all `panic!`'s in our code. 

Keeping the status quo of this PR, when a thread panicked while the rest of the system keeps working, seems like not a good choice (e.g. is the panicked thread even restarted afterwards?). I think, we need to return an error instead of panicking (will likely require changing signature at least in `UploadRequest::new` case). And then 2 options:
1. accept that we keep MP running in those cases (as you can see `flush` returns an error to the user in any case, so this is a good signal that things went wrong);
2. work on a mechanism that gracefully stops Mountpoint on certain errors (somewhere at [fs.rs](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/fs.rs#L243) level?)

I'd go with 1/, what do you think is the best here?

vladem: by changing signature of `UploadRequest::new` I mean something like this:
https://github.com/vladem/mountpoint-s3/pull/5/files

jamesbornholt: `flush` returns an error, and that makes it to the user as a return code from `close(fd)`, but people rarely check that return code.

The other option is `std::os::exit(1)` if you want to abort the process.

vladem: Hm, I see the point about unreliability of `flush` in error signalling. What I do not like about `exit()` is that it doesn't do any clean up. For example if there were simultaneous MPU uploads, they may be [left](https://aws.amazon.com/blogs/aws-cloud-financial-management/discovering-and-deleting-incomplete-multipart-uploads-to-lower-amazon-s3-costs/) incomplete on S3 side.

I think we still don't want to panic on `UploadRequest::new` as returning an error from there will also make `write()` to [return](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/fs.rs#L137) an `EIO`/`EPERM`, which I expect most users to check. And no data was uploaded yet, so its not a high risk to overlook the error.

For a "nuclear" case of `UploadRequest::complete` (when the data is known to be uploaded already and the corruption happened in transit or on S3 side) I think it's acceptable to do `std::os::exit(1)`.

vladem: Implemented error handling as described in the last comment in a separate branch. Can merge it to this one if we agree its better than `panic`ing: https://github.com/vladem/mountpoint-s3/pull/5

vladem: Discussed with @passaro, and decided to exit in both cases: on `UploadRequest::new` and on `UploadRequest::complete`. In a follow-up PR I am going to change `UploadRequest::new` to return an error, in a similar way as it is done here: https://github.com/vladem/mountpoint-s3/pull/5

mountpoint-s3/src/upload.rs
```diff
@@ -136,9 +149,25 @@ impl<Client: ObjectClient> UploadRequest<Client> {
...
+        self.sse
+            .verify_response(result.sse_type.as_deref(), result.sse_kms_key_id.as_deref())
+            .unwrap_or_else(|err| {
+                error!(
+                    "SSE settings were corrupted, object {} WAS uploaded to S3, terminating, error: {}",
+                    self.key, err
+                );
+                // Reaching this point is very unlikely and means that SSE settings were corrupted in transit or on S3 side, this may be a sign of a bug
+                // in CRT code or S3. Thus, we terminate Mountpoint to send the most noticeable signal to customer about the issue. We conciously choose
+                // exiting instead of returning an error because:
```
passaro, line 165: typo: consciously

But maybe reword like: 
```suggestion
                // in CRT code or S3. Thus, we terminate Mountpoint to send the most noticeable signal to customer about the issue.
                // We prefer exiting instead of returning an error because:
```

passaro (commented):
Looks good. Just one typo to fix.

jamesbornholt (commented):
Just small stuff.

mountpoint-s3/src/cli.rs
```diff
@@ -630,6 +631,10 @@ where
     #[cfg(feature = "sse_kms")]
     {
         filesystem_config.server_side_encryption = ServerSideEncryption::new(args.sse, args.sse_kms_key_id);
+        tracing::info!(
+            "using server-side encryption - {}",
+            filesystem_config.server_side_encryption
+        );
```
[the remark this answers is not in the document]
vladem: interesting suggestion, I got this log with just deriving `Debug` and using `{:?}` formatting:
> 2024-03-08T13:50:28.718985Z DEBUG mountpoint_s3::cli: CLI arguments: CliArgs { bucket_name: "my-bucket", mount_point: "/home/vlaad/mounted_s3", prefix: None, region: None, endpoint_url: None, force_path_style: false, transfer_acceleration: false, dual_stack: false, requester_pays: false, bucket_type: None, no_sign_request: false, profile: None, read_only: false, storage_class: None, allow_delete: false, allow_overwrite: false, auto_unmount: false, allow_root: false, allow_other: false, maximum_throughput_gbps: None, max_threads: 16, part_size: 8388608, uid: None, gid: None, dir_mode: None, file_mode: None, foreground: true, expected_bucket_owner: None, log_directory: None, log_metrics: false, debug: true, debug_crt: false, no_log: false, cache: None, metadata_ttl: None, max_cache_size: None, user_agent_prefix: None, sse: Some("aws:kms"), sse_kms_key_id: Some("arn:...") }

Is it what you mean? 

The only concern that I have -- how to minimise the risk of printing sensitive data. Currently CLI args do not seem to contain any of those, but what if we add something in future? May be we should bother about that only when we'll actually add a CLI flag holding sensitive data. But also we may make the printing opt-in for each field, i.e explicitly implement `Display` for `CliArgs` struct.

mountpoint-s3/src/fs.rs
```diff
@@ -397,30 +400,111 @@ impl S3Personality {
...
+            None => "bucket_default",
+            Some(sse_type) => sse_type.as_str(),
+        };
+        let sse_key_log = match &self.sse_kms_key_id {
+            None => "default",
+            Some(sse_kms_key_id) => sse_kms_key_id.as_str(),
+        };
+        write!(f, "type: {}, key id: {}", sse_type_log, sse_key_log)
     }
 }
```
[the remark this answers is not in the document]
vladem: I think, the only place is `cli.rs`, when printing this to logs I use `{}` for non-debug formatting. But given the suggestion in your previous comment of printing whole `CliArgs` I should be able to remove this.

mountpoint-s3/src/upload.rs
```diff
@@ -136,9 +149,25 @@ impl<Client: ObjectClient> UploadRequest<Client> {
...
         let checksum = self.hasher.finalize();
-        self.request
+        let result = self
+            .request
             .review_and_complete(move |review| verify_checksums(review, size, checksum))
-            .await
+            .await?;
+        self.sse
+            .verify_response(result.sse_type.as_deref(), result.sse_kms_key_id.as_deref())
+            .unwrap_or_else(|err| {
```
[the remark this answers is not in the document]
vladem: closure with side effects also looks like a bad pattern to me, at least in case of `unwrap_or_else` it makes sense that a closure should solely process unwrapped values

jamesbornholt (approved):

PR #750, opened 2024-02-13
Refactor error handling on read requests

passaro:
## Description of change

Minor refactor to simplify how errors returned from `Prefetch::read()` are handled. It implements the error conversion in the `fs::error` module, to align it to errors on the write path.

Preliminary work for #644 

No change in behavior.

---


mountpoint-s3/src/fs/error.rs
```diff
@@ -110,6 +112,22 @@ impl<E: std::error::Error + Send + Sync + 'static> From<UploadWriteError<E>> for
     }
 }
 
+impl<E: std::error::Error + Send + Sync + 'static> From<PrefetchReadError<E>> for Error {
```
sauraank, line 115: So the errors logged are still the same, right?

sauraank (commented):
LGTM. I had one question. Why are we directly returning `Bytes` from prefetcher rather than converting them from `ChecksummedBytes` at the `fs` level?

mountpoint-s3/src/prefetch.rs
```diff
@@ -331,7 +324,7 @@ where
             to_read -= part_len;
         }
 
-        Ok(response)
+        Ok(response.into_bytes()?)
```
vladem, line 327: not relevant to this PR changes, but for the sake of understanding: so we validate the checksum here, at which point is it computed? I see that we combine checksums of `part_bytes`, is it the `S3CrtClient` (our code) who computes checksum of the `part_bytes` in the first place? or it's provided by S3?

vladem (approved):

mountpoint-s3/src/prefetch.rs
passaro: We do it in`mountpoint-s3`, here when downloading from S3: https://github.com/awslabs/mountpoint-s3/blob/608b921cb7b81d2012cd196087aa6aeaeaa1b6b1/mountpoint-s3/src/prefetch/part_stream.rs#L222

passaro:
> LGTM. I had one question. Why are we directly returning `Bytes` from prefetcher rather than converting them from `ChecksummedBytes` at the `fs` level?

* the prefetcher can also return a `PrefetchReadError::Integrity` (in case the integrity checks failed at some other point in the pipeline), so it is simpler to merge the two and return an already validated `Bytes`
* Before the change, we were immediately validating the `ChecksummedBytes` and extracting the `Bytes` anyway, so this does not actually moves when the validation happens.

jamesbornholt:
I kind of think we might prefer to keep the prefetcher returning `ChecksummedBytes`. Today we just convert immediately, but there's nothing stopping future us from inserting more work between the prefetcher and the FUSE response, and the type at least makes that decision clear.

passaro:
@jamesbornholt, @sauraank, I reverted the change to the return type of `read`. It brings back 2 routes to get a form of integrity error, but overall I agree we want to propagate our "checked" type as much as possible, even if right now we immediately extract the buffer.

jamesbornholt:
Ok, but now I'm not sure why we need this change — does moving the error parsing logic out into a `From` impl buy us anything? I reverted this after checking out #644 and it didn't seem to break anything.

passaro:
> Ok, but now I'm not sure why we need this change — does moving the error parsing logic out into a `From` impl buy us anything? I reverted this after checking out #644 and it didn't seem to break anything.

Fair question. You're right, this is not really tied to #644 anymore, but I do see moving the `PrefetchReadError` mapping logic to the `error` module, as done for `UploadWriteError`, as a marginal improvement.

dannycjones (approved):
lgtm!

Issue #749, opened 2024-02-13
Files unexpectedly released leading to `Bad file descriptor`


jchorl:
### Mountpoint for Amazon S3 version

1.4.0

### AWS Region

ca-central-1

### Describe the running environment

EC2, amazon linux 2. Mountpoint is running inside a privileged ubuntu docker container.

### Mountpoint options

```shell
MOUNTPOINT_LOG=trace mount-s3 my-bucket mnt --prefix my-prefix/ --foreground -l mplogs
```

### What happened?

Summary:
- Calling a bash script that calls itself, served from mountpoint mount
- It invokes a heavy disk/mem operation, **unrelated to mountpoint**
- When the script returns to running, it is no longer accessible, leading to `Bad file descriptor` errors

The setup is a bit complicated. I'll put the files at the end because some are a bit large.
1. A bash script that should be at least a few pages in length. It's important the whole thing not fit into a single page[0].
    - This script should go in `my-bucket/my-prefix/`
2. An application that hammers disk + ram[1]. I wrote a simple C binary that reads a large file into memory.
3. A large file. I used `head -c 12G </dev/urandom > ~/big-file` to generate.

It's important to note I was running on a `r6i.large`, with a `gp3` vol, 250gb, 3000 iops, 125 throughput.

Here's how I ran it:

In one terminal, get mountpoint running:
```
docker run -it --rm --privileged ubuntu:23.10 bash
apt-get update; apt-get install -y gcc wget vim

# install mountpoint
wget https://s3.amazonaws.com/mountpoint-s3-release/latest/x86_64/mount-s3.deb
apt-get install -y ./mount-s3.deb

# prepare bin to stress disk/ram
cd
mkdir joshtest
vim main.c # contents later in report
gcc -o josh-mem-test main.c

# prep a big file to thrash disk/ram
head -c 12G </dev/urandom > ~/big-file

# copy script to bucket/prefix, contents later in report
aws s3 cp testscript.sh s3://my-bucket/my-prefix/

# get mountpoint running
mount-s3 my-bucket mnt --prefix my-prefix/ --foreground
```

In a separate terminal, exec in to the container and run the script

```
docker exec -it 040b2bcbb33c bash
bash ./mnt/testscript.sh f1
# outputs
# in f1
# in f2
# allocating a bunch of ram
# Tue Feb 13 06:26:43 GMT 2024
# Buffer size:
# 8
# done allocating a bunch of ram
[...]

PR #751, opened 2024-02-14
Fix issue preventing reads after flush on a file handle

passaro:
## Description of change

Fix a regression introduced in https://github.com/awslabs/mountpoint-s3/commit/0030b0a527638a76b29fe387cfa6ae22b0ad1c92 where closing (`flush`) a file descriptor would prevent subsequent reads on the corresponding file handle. The issue was caused by `flush` setting the internal state of the file handle to `Closed` and decrementing the number of readers for the inode, which is used to disallow overwrites when there are open read file handles.

This PR includes 2 main changes:
* it removes the `Closed` state and reverts `flush` to a no op for read file handles. The number of readers for an inode is decremented only on `release`.
* it makes `open` eagerly discriminate between READ and WRITE, rather than on first read or write (or flush or release). This reverts the opposite change in #487 and allows to report error conditions on `open` when appropriate rather than on subsequent operations.

Relevant issues:  #749

Because `release` is async, there is a race condition when trying to overwrite a file immediately after closing a read file handle on it, which results in the overwrite to occasionally fail on open.

---

### DIFF
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+### Other changes
+* Fix an issue where read file handles could be closed too early, leading to bad file descriptor errors on subsequent reads. As a consequence of this fix, opening an existing file to overwrite it **immediately** after closing a read file handle may occasionally fail with an "Operation not permitted" error. In such cases, Mountpoint logs will also report that the file is "not writable while being read". ([#751](https://github.com/awslabs/mountpoint-s3/pull/751))
+
 ## v1.4.0 (January 26, 2024)
 
 ### New features
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Other changes
 * Fix an issue where read file handles could be closed too early, leading to bad file descriptor errors on subsequent reads. As a consequence of this fix, opening an existing file to overwrite it **immediately** after closing a read file handle may occasionally fail with an "Operation not permitted" error. In such cases, Mountpoint logs will also report that the file is "not writable while being read". ([#751](https://github.com/awslabs/mountpoint-s3/pull/751))
+* File handles are no longer initialized lazily. Lazy initialization was introduced in version 1.4.0 but is reverted in this change. If upgrading from 1.4.0, you may see errors that were previously deferred until read/write now raised at open time. ([#751](https://github.com/awslabs/mountpoint-s3/pull/751))
 
 ## v1.4.0 (January 26, 2024)
 
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -69,14 +69,9 @@ where
     /// A state where the file handle is created but the type is not yet determined
     Created { lookup: LookedUp, flags: i32, pid: u32 },
     /// The file handle has been assigned as a read handle
-    Read {
-        request: Prefetcher::PrefetchResult<Client>,
-        pid: u32,
-    },
+    Read(Prefetcher::PrefetchResult<Client>),
     /// The file handle has been assigned as a write handle
     Write(UploadState<Client>),
-    /// The file handle is already closed, currently only used to tell that the read is finished
-    Closed,
 }
 
 impl<Client, Prefetcher> std::fmt::Debug for FileHandleState<Client, Prefetcher>
@@ -92,9 +87,8 @@ where
                 .field("flags", flags)
                 .field("pid", pid)
                 .finish(),
-            FileHandleState::Read { request: _, pid } => f.debug_struct("Read").field("pid", pid).finish(),
+            FileHandleState::Read(_) => f.debug_struct("Read").finish(),
             FileHandleState::Write(arg0) => f.debug_tuple("Write").field(arg0).finish(),
-            FileHandleState::Closed => f.debug_struct("Closed").finish(),
         }
     }
 }
@@ -147,7 +141,6 @@ where
     async fn new_read_handle(
         lookup: &LookedUp,
         flags: i32,
-        pid: u32,
         fs: &S3Filesystem<Client, Prefetcher>,
     ) -> Result<FileHandleState<Client, Prefetcher>, Error> {
         if flags & libc::O_WRONLY != 0 {
@@ -169,7 +162,7 @@ where
         let request = fs
             .prefetcher
             .prefetch(fs.client.clone(), &fs.bucket, &full_key, object_size, etag.clone());
-        let handle = FileHandleState::Read { request, pid };
+        let handle = FileHandleState::Read(request);
         metrics::increment_gauge!("fs.current_handles", 1.0, "type" => "read");
         Ok(handle)
     }
@@ -766,19 +759,18 @@ where
         logging::record_name(handle.inode.name());
         let mut state = handle.state.lock().await;
         let request = match &mut *state {
-            FileHandleState::Created { lookup, flags, pid, .. } => {
+            FileHandleState::Created { lookup, flags, .. } => {
                 metrics::decrement_gauge!("fs.current_handles", 1.0, "type" => "unassigned");
 
-                *state = FileHandleState::new_read_handle(lookup, *flags, *pid, self).await?;
-                if let FileHandleState::Read { request, .. } = &mut *state {
+                *state = FileHandleState::new_read_handle(lookup, *flags, self).await?;
+                if let FileHandleState::Read(request) = &mut *state {
                     request
                 } else {
                     unreachable!("handle type always be assigned above");
                 }
             }
-            FileHandleState::Read { request, .. } => request,
+            FileHandleState::Read(request) => request,
             FileHandleState::Write(_) => return Err(err!(libc::EBADF, "file handle is not open for reads")),
-            FileHandleState::Closed => return Err(err!(libc::EBADF, "file handle is already closed")),
         };
 
         match request.read(offset as u64, size as usize).await {
@@ -881,7 +873,6 @@ where
                 }
                 FileHandleState::Read { .. } => return Err(err!(libc::EBADF, "file handle is not open for writes")),
                 FileHandleState::Write(request) => request,
-                FileHandleState::Closed => return Err(err!(libc::EBADF, "file handle is already closed")),
             };
 
             request.write(offset, data, &handle.full_key).await?
@@ -1122,7 +1113,6 @@ where
             }
             FileHandleState::Read { .. } => return Ok(()),
             FileHandleState::Write(request) => request,
-            FileHandleState::Closed => return Ok(()),
         };
         self.complete_upload(request, &file_handle.full_key, false, None).await
     }
@@ -1141,10 +1131,6 @@ where
         //   process. In many cases, the child will then immediately close (flush) the duplicated
         //   file descriptors. We will not complete the upload if we can detect that the process
         //   invoking flush is different from the one that originally opened the file.
-        //
-        // The same for read path. We want to stop the prefetcher and decrease the reader count
-        // as soon as users close a file descriptor so that we don't block users from doing other
-        // operation like overwrite the file.
         let file_handle = {
             let file_handles = self.file_handles.read().await;
             match file_handles.get(&fh) {
@@ -1156,29 +1142,11 @@ where
         let mut state = file_handle.state.lock().await;
         match &mut *state {
             FileHandleState::Created { .. } => Ok(()),
-            FileHandleState::Read { pid: open_pid, .. } => {
-                if !are_from_same_process(*open_pid, pid) {
-                    trace!(
-                        file_handle.full_key,
-                        pid,
-                        open_pid,
-                        "not stopping prefetch because current pid differs from pid at open"
-                    );
-                    return Ok(());
-                }
-                // TODO make sure we cancel the inflight PrefetchingGetRequest. is just dropping enough?
-                file_handle.inode.finish_reading()?;
-
-                // Mark the file handle state as closed so we only update the reader count once
-                *state = FileHandleState::Closed;
-                metrics::decrement_gauge!("fs.current_handles", 1.0, "type" => "read");
-                Ok(())
-            }
+            FileHandleState::Read { .. } => Ok(()),
             FileHandleState::Write(request) => {
                 self.complete_upload(request, &file_handle.full_key, true, Some(pid))
                     .await
             }
-            FileHandleState::Closed => Ok(()),
         }
     }
 
@@ -1241,7 +1209,6 @@ where
                 return Ok(());
             }
             FileHandleState::Write(request) => request,
-            FileHandleState::Closed => return Ok(()),
         };
 
         let result = request.complete_if_in_progress(&file_handle.full_key).await;
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -702,11 +702,11 @@ where
             let is_truncate = flags & libc::O_TRUNC != 0;
             if !remote_file || (self.config.allow_overwrite && is_truncate) {
                 // If the file is new or opened in truncate mode, we know it must be a write handle.
-                trace!("fs:open choosing write handle for O_RDWR");
+                debug!("fs:open choosing write handle for O_RDWR");
                 FileHandleState::new_write_handle(&lookup, lookup.inode.ino(), flags, pid, self).await?
             } else {
                 // Otherwise, it must be a read handle.
-                trace!("fs:open choosing read handle for O_RDWR");
+                debug!("fs:open choosing read handle for O_RDWR");
                 FileHandleState::new_read_handle(&lookup, self).await?
             }
         } else if flags & libc::O_WRONLY != 0 {
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -66,8 +66,6 @@ where
     Client: ObjectClient + Send + Sync + 'static,
     Prefetcher: Prefetch,
 {
-    /// A state where the file handle is created but the type is not yet determined
-    Created { lookup: LookedUp, flags: i32, pid: u32 },
     /// The file handle has been assigned as a read handle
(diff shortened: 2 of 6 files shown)


mountpoint-s3/tests/fuse_tests/read_test.rs
```diff
@@ -309,3 +309,41 @@ fn read_errors_test_s3() {
...
+
+    let path = mount_point.path().join(KEY);
+    let mut f = open_for_read(path, true).unwrap();
+
+    let mut content = vec![0; 128];
+    f.read_exact(&mut content).unwrap();
+
+    let mut f_dup = f.try_clone().unwrap();
+
+    drop(f);
```
dannycjones, line 333: A small comment mentioning we're intentionally closing the file and triggering a flush here would be useful

dannycjones (approved):
LGTM.

Acknowledging trade-off that we mitigate the regression preventing reads after flush but introducing a race condition into file overwrites (available since v1.4.0 released this month) which could prevent opening a file for writing immediately after closing a file. (@passaro should we add an ignored test for this? I can add one in a follow-up PR.)

dannycjones (approved):
LGTM!

mountpoint-s3/src/fs.rs
```diff
@@ -1156,29 +1146,11 @@ where
         let mut state = file_handle.state.lock().await;
         match &mut *state {
             FileHandleState::Created { .. } => Ok(()),
-            FileHandleState::Read { pid: open_pid, .. } => {
```
vladem, line 1159: is the comment on line 1135 still relevant?
```
// The same for read path. We want to stop the prefetcher and decrease the reader count
// as soon as users close a file descriptor so that we don't block users from doing other
// operation like overwrite the file.
```

passaro: good spot

mountpoint-s3/src/fs.rs
```diff
@@ -1156,29 +1146,11 @@ where
...
-                    trace!(
-                        file_handle.full_key,
-                        pid,
-                        open_pid,
-                        "not stopping prefetch because current pid differs from pid at open"
-                    );
-                    return Ok(());
-                }
-                // TODO make sure we cancel the inflight PrefetchingGetRequest. is just dropping enough?
-                file_handle.inode.finish_reading()?;
```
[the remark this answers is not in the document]
passaro: Yes, release will call `finish_reading`. But consider that the kernel will only call `release` once, so that is not a concern.

mountpoint-s3/tests/fuse_tests/read_test.rs
```diff
@@ -309,3 +309,44 @@ fn read_errors_test_s3() {
...
+    let mut content = vec![0; 128];
+    f.read_exact(&mut content).unwrap();
+
+    let mut f_dup = f.try_clone().unwrap();
+
+    // Close the file. Triggers a flush on the file handle.
+    drop(f);
+
+    // Read using the duplicated instance (same underlying handle).
+    // Seek to the end of the file to avoid relying on the kernel cache.
```
arsh, line 337: Out of curiosity, how do we know how much data will the kernel cache in this case? Do we know that it will cache only those 128 bytes? Should you use `O_DIRECT` and `O_SYNC` when opening the file?

arsh (commented):
Which state will the file overwrites feature be in if this change is pushed? Is the only behavior change the one you listed?

mountpoint-s3/src/fs.rs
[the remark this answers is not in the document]
vladem: Got it, thanks. So file_handle can only once enter a `FileHandleState::Read` state, where we increment the `InodeState::reader_count`, we rely on `release` being called only once, so the `InodeState` will be updated appropriately.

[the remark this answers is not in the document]
vladem: A little curious about the Danny's comment, would appreciate an explanation:
> introducing a race condition into file overwrites which could prevent opening a file for writing immediately after closing a file

Does not `close(fd)` in customer code wait for our `release` to finish? Or kernel calls `release` asynchronously **after** returning from `close(fd)` (where `fd` is the last one for the given file handle)?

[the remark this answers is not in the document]
dannycjones: >Or kernel calls release asynchronously after returning from close(fd) (where fd is the last one for the given file handle)?

It's roughly this. It's not clear to me when release is invoked but we believe that the Kernel will not wait for that to complete before invoking further file system operations over FUSE.

I think `close(fd)` is effectively `flush`, and errors thrown in flush will be visible there. Release will be invoked at some point after that, we don't know how strict that relationship is. It's also important to recall that the Kernel calling `flush` isn't even guaranteed, which is why we also ensure we perform vital cleanup / completion in `release`.

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -763,10 +763,60 @@ where
...
+    };
+    let (mount_point, _session, mut test_client) = creator_fn(prefix, test_config);
+
+    // Make sure there's an existing directory and a file
+    test_client.put_object("dir/hello.txt", b"hello world").unwrap();
+
+    let _subdir = mount_point.path().join("dir");
+    let path = mount_point.path().join("dir/hello.txt");
+
+    // File should be empty when opened with O_WRONLY and O_TRUNC even without any write
```
monthonk, line 814: nit: not O_WRONLY anymore depending on the param.

mountpoint-s3/src/fs.rs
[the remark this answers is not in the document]
vladem: I see, so basically if there are 2 processes, where:
1. first one has a read fd to `mounted/a.txt`;
2. second one wants to open `mounted/a.txt` for write.

Even if the first process has already finished execution there is no guarantee that the second one will be able to open file for write (even if e.g. the process 2 is the parent of process 1 and it calls `waitpid` on it). 

Thanks for explanation!

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -672,7 +672,7 @@ fn multi_thread_test() {
     .unwrap();
 }
 
-fn overwrite_test<F>(creator_fn: F, prefix: &str)
+fn overwrite_test<F>(creator_fn: F, prefix: &str, write_only: bool)
```
dannycjones, line 675: What's the purpose of this test? Overwrite with truncate, no fsync?

dannycjones:
`read_overwrite_read_test` is failing as expected, nice.

We need a change log entry for the bug fix, but also to call out the edge case race condition that will be introduced by the bug fix.

mountpoint-s3/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## Unreleased
 
+### Other changes
+* Fix an issue where a file handle may be closed too early and result in `bad file descriptor` errors on subsequent reads. This change resolves the issue but introduces a race condition when using `--allow-overwrite`: in some cases, overwriting an existing file immediately after reading it may fail. ([#751](https://github.com/awslabs/mountpoint-s3/pull/751))
```
arsh, line 4: What should users do if they face this race condition?

mountpoint-s3/tests/fuse_tests/write_test.rs
passaro: This is a basic test that overwrite works. Truncate is required to overwrite a file and fsync is used in the test so that the subsequent read is guaranteed to be consistent.

dannycjones: Ah, got it. Happy path.

mountpoint-s3/tests/fuse_tests/read_test.rs
dannycjones: We're using big numbers far larger than we expect the Kernel to cache.

I think we'd avoid `O_DIRECT` since it also has semantic changes in MP itself (skipping caches, always checking with S3). `O_SYNC` is not something we've used, I think its rejected which is fine as I believe its used when writing only - and we cannot fulfil the guarantees for `O_SYNC`, since a write cannot be guaranteed to be persisted to S3 - only `fsync`.

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -89,17 +89,6 @@ where
...
-    if append {
-        // We can't append existing files
-        let err = open_for_write(&path, append, write_only).expect_err("can't append existing file");
-        assert_eq!(err.kind(), ErrorKind::InvalidInput);
-    } else {
-        // We shouldn't be allowed to write the file again
-        let mut f = open_for_write(&path, append, write_only).expect("should be able to open the file");
-        let err = f.write(b"hello world").expect_err("can't write existing file");
-        assert_eq!(err.kind(), ErrorKind::PermissionDenied);
-    }
```
dannycjones, line 102: already tested elsewhere, we're removing this to simplify the test

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -168,11 +156,18 @@ where
...
         .expect_err("write to an existing file should fail");
+    assert_eq!(err.raw_os_error(), Some(libc::EBADF));
+
+    // For default config, existing files cannot be opened with O_TRUNC
+    let err = File::options()
+        .read(true)
+        .write(true)
+        .truncate(true)
+        .open(&path)
+        .expect_err("existing file cannot be opened with O_TRUNC");
```
dannycjones, line 170: this was missing before, confirming this works as we expected

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -836,55 +839,6 @@ fn overwrite_fail_on_write_without_truncate_test_mock(write_only: bool) {
...
-    let write_fh = options.write(true).open(&path).expect("open should succeed");
-    drop(write_fh);
-
-    // File content should not be changed
-    let mut options = File::options();
-    let mut fh = options.read(true).write(true).open(&path).unwrap();
-    let mut hello_contents = String::new();
-    fh.read_to_string(&mut hello_contents).unwrap();
-    assert_eq!(hello_contents, "hello world");
-}
```
dannycjones, line 873: we do not believe we want to allow this behavior. i.e. open for writing only, closing file, no change.

cc @jamesbornholt

mountpoint-s3/tests/fs.rs
```diff
@@ -700,22 +695,13 @@ async fn test_duplicate_write_fails() {
     assert_eq!(dentry.attr.size, 0);
     let file_ino = dentry.attr.ino;
 
-    let opened = fs
+    let _opened = fs
         .open(file_ino, libc::S_IFREG as i32 | libc::O_WRONLY, 0)
         .await
         .unwrap();
-    _ = fs
```
vladem, line 707: why removing this? "first write should succeed", right?

mountpoint-s3/tests/fuse_tests/read_test.rs
```diff
@@ -273,7 +273,7 @@ where
     assert_eq!(dir_entry_names, vec!["hello.txt"]);
 
     // Overwrite the test file and verify that we can't read from a file opened in O_WRONLY
-    let mut write_fh = File::options().write(true).open(&file_path).unwrap();
+    let mut write_fh = File::options().write(true).truncate(true).open(&file_path).unwrap();
```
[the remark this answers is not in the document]
vladem: oh I see, this restriction actually makes sense, probably it works now the same way, but reports the error on `write` and if we didn't add `truncate` here we would see it on `open`

mountpoint-s3/src/fs.rs
```diff
@@ -726,10 +698,29 @@ where
...
+                FileHandleState::new_write_handle(&lookup, lookup.inode.ino(), flags, pid, self).await?
+            } else {
+                // Otherwise, it must be a read handle.
+                debug!("fs:open choosing read handle for O_RDWR");
+                FileHandleState::new_read_handle(&lookup, self).await?
+            }
+        } else if flags & libc::O_WRONLY != 0 {
+            FileHandleState::new_write_handle(&lookup, lookup.inode.ino(), flags, pid, self).await?
+        } else {
+            FileHandleState::new_read_handle(&lookup, self).await?
```
vladem, line 715: does prefetching starts on `open` or on `read` with this change? when prefetching starts in current MP version?

mountpoint-s3/src/fs.rs
```diff
@@ -726,10 +698,29 @@ where
             return Err(err!(libc::EINVAL, "O_SYNC and O_DSYNC are not supported"));
         }
 
-        // All file handles will be lazy initialized on first read/write.
-        let state = FileHandleState::new(lookup, flags, pid).await.into();
+        let state = if flags & libc::O_RDWR != 0 {
```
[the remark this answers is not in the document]
dannycjones: This is worth a change log entry, you're right.

We made changes to tests to accommodate this change.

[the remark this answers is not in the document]
passaro: Yes, most of the changes in the tests are due to this change. Note that this is reverting a change made when introducing overwrite support, which was not explicitly documented in the first place.

mountpoint-s3/src/fs.rs
```diff
@@ -726,10 +698,29 @@ where
...
+            let is_truncate = flags & libc::O_TRUNC != 0;
+            if !remote_file || (self.config.allow_overwrite && is_truncate) {
```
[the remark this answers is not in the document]
passaro: The current behavior on open of an existing file is (with allow_overwrite):
* O_WRONGLY: succeed with O_TRUNC, fail otherwise,
* O_RDWR: open as write with O_TRUNC, as read without.

Before this change, the intended behavior was same, but enforced on first read/write (or flush or release!). This allowed for the edge case of open without O_TRUNC and no writes (or reads), which would result in a no op. After this change, this is resolved on open and it's either an error or a read.

mountpoint-s3/src/fs.rs
passaro: We now initialize the prefetcher on open, but we still only spawn requests on the first read.

mountpoint-s3/src/fs.rs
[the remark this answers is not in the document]
dannycjones: Looks like this test was failing before and now passes:

```rust


fn overwrite_empty_file<F>(creator_fn: F, prefix: &str)
where
    F: FnOnce(&str, TestSessionConfig) -> (TempDir, BackgroundSession, TestClientBox),
{
    let filesystem_config = S3FilesystemConfig {
        allow_overwrite: true,
        ..Default::default()
    };
    let test_config = TestSessionConfig {
        filesystem_config,
        ..Default::default()
    };
    let (mount_point, _session, mut test_client) = creator_fn(prefix, test_config);

    // Make sure there's an existing directory and a file
    test_client.put_object("dir/hello.txt", b"").unwrap();

    let _subdir = mount_point.path().join("dir");
    let path = mount_point.path().join("dir/hello.txt");

    // Open the file in non-truncate mode and do nothing
    let mut options = File::options();
    let _err = options.write(true).open(&path).expect_err("open should not succeed without O_TRUNC");
}

#[cfg(feature = "s3_tests")]
#[test]
fn overwrite_empty_file_s3() {
    overwrite_empty_file(fuse::s3_session::new, "overwrite_empty_file");
}

#[test]
fn overwrite_empty_file_mock() {
    overwrite_empty_file(fuse::mock_session::new, "overwrite_empty_file");
}
```

mountpoint-s3/tests/fs.rs
passaro: yes, but it's not needed for the test anymore. The open above already resolves this to be a "write" handle, while before we needed an actual write call. What we are testing here, though, is the failure of the next open.

mountpoint-s3/src/fs.rs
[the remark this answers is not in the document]
vladem: I see, so there is no place in existing documentation which we could reference and say "this is changing now, errors will be reported earlier". But probably worth bridging this gap in new CHANGELOG.md entry? Something like:
> Release v1.4.0 made some errors reported on first access instead of `open()`, changes in v1.4.1 reverts this to reporting on `open()`

mountpoint-s3/src/fs.rs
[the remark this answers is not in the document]
vladem: > the intended behavior was same
sounds good and thanks for explanation why making initialisation eager, I definitely lost the original intention of fixing read-stale-after-no-op-truncate

mountpoint-s3/src/fs.rs
vladem: that resolves my concern 👍

mountpoint-s3/tests/fs.rs
vladem: ah, I see, the `write` would be redundant for the test with current `open` semantics, thanks

mountpoint-s3/src/fs.rs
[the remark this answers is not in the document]
dannycjones: I've pushed a test similar to above which verifies the new behavior. I think we should adopt the new behavior, and shouldn't have allowed opening a write handle on an empty file.

This change brings the behavior in line both with and without `--allow-overwrites`.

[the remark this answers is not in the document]
vladem: > Looks like this test was failing before and now passes:

Wow, you've quickly assembled a test case for that! So you're opening  in`O_WRONLY`-non-truncate mode and it's failing fast. A little different from my original thought of `O_RDWR`-non-truncate mode. In both cases if the error was previously reported on `write` and now it's just reported earlier, we're good in my opinion.

mountpoint-s3/src/fs.rs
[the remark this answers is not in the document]
dannycjones: I've added a note to the change log in the next push.

vladem (approved):
LGTM:
1. flush() is no-op for read;
2. inode readers counter is decremented on `release` (accepted behaviour that `open` for write following a `close` may fail for short period of time);
3. eager initialisation of file handles to avoid FH being readable when we already have a scheduled work to write to it

arsh (approved):
LGTM

PR #756, opened 2024-02-16
Split up the overwrite test into multiple, more specific tests

passaro:
## Description of change

Split up the overwrite test into multiple, more specific tests.

N/A

---

### DIFF
--- a/mountpoint-s3/tests/fuse_tests/write_test.rs
+++ b/mountpoint-s3/tests/fuse_tests/write_test.rs
@@ -672,7 +672,7 @@ fn multi_thread_test() {
     .unwrap();
 }
 
-fn overwrite_test<F>(creator_fn: F, prefix: &str)
+fn overwrite_test<F>(creator_fn: F, prefix: &str, write_only: bool)
 where
     F: FnOnce(&str, TestSessionConfig) -> (TempDir, BackgroundSession, TestClientBox),
 {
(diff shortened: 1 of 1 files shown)


mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -732,41 +766,151 @@ where
...
+    overwrite_no_truncate_test(fuse::s3_session::new, "overwrite_no_truncate_test", write_only);
+}
+
+#[test_case(true; "write_only")]
+#[test_case(false; "read_write")]
+fn overwrite_no_truncate_test_mock(write_only: bool) {
+    overwrite_no_truncate_test(fuse::mock_session::new, "overwrite_no_truncate_test", write_only);
+}
+
+fn overwrite_truncate_test<F>(creator_fn: F, prefix: &str, write_only: bool)
```
dannycjones, line 888: Does this test still have the race condition of file handle being opened for reading too soon after a write, resulting in bad FD?

mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -778,21 +922,47 @@ where
...
+}
 
-    // Check the new file content
+#[test_case(true; "write_only")]
+#[test_case(false; "read_write")]
+fn overwrite_truncate_test_mock(write_only: bool) {
+    overwrite_truncate_test(fuse::mock_session::new, "overwrite_truncate_test", write_only);
+}
+
+fn read_overwrite_read_test<F>(creator_fn: F, prefix: &str)
```
dannycjones, line 940: Similarly, do we expect to see bad FD if the read occurs too soon after the write or write too soon after read?

dannycjones: This is currently passing on `main`, we'll review again when changing behavior.

mountpoint-s3/tests/fuse_tests/write_test.rs
dannycjones: This is currently passing on `main`, we'll review again when changing behavior.

dannycjones (approved):

mountpoint-s3/tests/fuse_tests/write_test.rs
passaro: Yes, you are right. But nothing has changed in this PR.

mountpoint-s3/tests/fuse_tests/write_test.rs
passaro: read after write, yes. write after read is safe on `main` (may not be if we merge #751)

PR #778, opened 2024-02-23
Add a client metric for number of resolved IPs for the endpoint

jamesbornholt:
## Description of change

This should help us identify issues where we're not getting enough IP diversity, limiting throughput.

No.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -19,7 +19,7 @@ use mountpoint_s3_crt::http::request_response::{Header, Headers, Message};
 use mountpoint_s3_crt::io::async_stream::AsyncInputStream;
 use mountpoint_s3_crt::io::channel_bootstrap::{ClientBootstrap, ClientBootstrapOptions};
 use mountpoint_s3_crt::io::event_loop::EventLoopGroup;
-use mountpoint_s3_crt::io::host_resolver::{HostResolver, HostResolverDefaultOptions};
+use mountpoint_s3_crt::io::host_resolver::{AddressKinds, HostResolver, HostResolverDefaultOptions};
 use mountpoint_s3_crt::io::retry_strategy::{ExponentialBackoffJitterMode, RetryStrategy, StandardRetryOptions};
 use mountpoint_s3_crt::s3::client::{
     init_signing_config, ChecksumConfig, Client, ClientConfig, MetaRequestOptions, MetaRequestResult, MetaRequestType,
@@ -219,6 +219,7 @@ struct S3CrtClientInner {
     part_size: usize,
     bucket_owner: Option<String>,
     credentials_provider: Option<CredentialsProvider>,
+    host_resolver: HostResolver,
 }
 
 impl S3CrtClientInner {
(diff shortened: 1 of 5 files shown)


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -437,6 +439,10 @@ impl S3CrtClientInner {
         let span_body = request_span.clone();
         let span_finish = request_span;
 
+        let endpoint = options.get_endpoint().expect("S3Message always has an endpoint");
```
jamesbornholt, line 442: I don't love this — it's possible to reach this code by constructing a `MetaRequestOptions` yourself — but the APIs here don't make that clear right now. I'd like to refactor a bit to make more explicit that we expect this `options` to always have been constructed by one of our own methods, but I'll defer that for now.

mountpoint-s3-crt/src/common/string.rs
```diff
@@ -0,0 +1,41 @@
...
+
+/// An immutable string holding either text or binary data.
+// We call this `AwsString` just because Rust's `String` type is in the prelude and we don't want to
+// confuse the two.
+#[derive(Debug)]
+pub struct AwsString {
+    inner: NonNull<aws_string>,
+}
+
+impl AwsString {
```
arsh, line 18: Can we unit test this and validate we're dropping, etc.?

mountpoint-s3-crt/src/io/host_resolver.rs
```diff
@@ -43,6 +44,15 @@ impl HostResolver {
 
         Ok(Self { inner })
     }
+
+    /// Get the current number of known host addresses for a given hostname
+    pub fn get_host_address_count(&self, hostname: &AwsString, kinds: AddressKinds) -> Result<usize, Error> {
```
arsh, line 49: It would be nice to have a test that use this with a fixed S3 endpoint and validate we get some addresses back. Thoughts?

arsh (commented):
LGTM. Asked some questions about testing.

mountpoint-s3-crt/src/common/string.rs
jamesbornholt: Not easily -- the undering `aws_string` gets deallocated as part of `aws_string_destroy`, and so the `NonNull<aws_string>` becomes a dangling pointer after that point. We could try to peek inside the allocator to check it was actually deallocated, but that would be racy with other threads running at test time since they all share the same allocator.

mountpoint-s3-crt/src/io/host_resolver.rs
jamesbornholt: We don't currently have a way to actually look up a host using `HostResolver` because we don't do any resolutions ourselves (the S3 client does them), and so we haven't bound the lookup functions. I think I'd rather not add new bindings for those just to be able to write tests.

Instead, what we can probably do is write a unit test on `S3CrtClient` that validates the new metric gets emitted correctly. Writing metrics tests is hard right now, but I have a followup PR to this that updates us to the newest [`metrics` crate](https://docs.rs/metrics), which makes those tests much simpler to write. So I think we should tackle testing this as a followup.

arsh (approved):

PR #779, opened 2024-02-23
Add benchmarks for S3 express one zone bucket

sauraank:
## Description of change
Added a separate workflow for s3 express one bucket benchmark and running that workflow on separate job. We have got successful result on wf-changes/s3-express-benchmark: https://github.com/awslabs/mountpoint-s3/actions/runs/8007392433
Pages for s3 express benchmark: [Throughput](https://awslabs.github.io/mountpoint-s3/dev/s3-express/bench/), [Latency](https://awslabs.github.io/mountpoint-s3/dev/s3-express/latency_bench/)

Also added the step of creation objects required for benchmarking within the scripts, so that if it required to setup benchmarking again for some other case, we dont need tohe manual setup.

Relevant issues: 

---

### DIFF
--- a/.github/workflows/bench_main.yml
+++ b/.github/workflows/bench_main.yml
@@ -14,3 +14,8 @@ jobs:
     uses: ./.github/workflows/bench.yml
     with:
       ref: ${{ github.event.after }}
+  s3express-integration:
+    name: Benchmarks (s3express)
+    uses: ./.github/workflows/bench_s3express.yml
+    with:
+      ref: ${{ github.event.after }}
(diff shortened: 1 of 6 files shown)


sauraank:
Benchmark result is failing for `readdir_100000` job because it has once benchmarked using none files within the prefix `dir_100000` and the result was 0.03 s but when read with 10000 files within it, it returned result of 8.97s which is expected. But. this exceeds the threshold of 2 times that we have for our benchmarks.

arsh (approved):
LGTM

sauraank:
Thanks Andres!

PR #781, opened 2024-02-23
Add metrics for number of inodes

jamesbornholt:
## Description of change

We track the total number of inodes as well as tracking them by kind. I did this by updating how we use gauges to work better for our current metrics stack, so that we don't just emit the same value every 5 seconds forever.

No.

---


arsh (approved):
LGTM

PR #783, opened 2024-02-24
Adding benchmarks that use caching.

arsh:
## Description of change

_NOTE: This is a new PR replacing https://github.com/awslabs/mountpoint-s3/pull/761 because through this branch `wf-changes/caching-benchmarks` I was able to actually run the new job for caching: https://github.com/awslabs/mountpoint-s3/actions/runs/8031555207._

This change is adding benchmarks that use mountpoint caching with files up to 1 GB in size because the current workers have only 30 GB of disk. If we want to support bigger sizes, we'll likely need to make changes to provision new workers.

The fio jobs were copied from the `read` benchmarks and renamed accordingly. I ran some of the tests locally by using the _filter_ functionality. We might not need all of these tests for the cache benchmarks but defaulted to have the same as the read ones and change later if needed.

```bash
# filter to only run this test for both for normal reads and cached ones
export JOB_NAME_FILTER=seq_read_small

./mountpoint-s3/scripts/fs_cache_bench.sh
Will only run fio jobs which match seq_read_small
Skipping job mountpoint-s3/scripts/fio/read/rand_read_4t_direct.fio because it does not match seq_read_small
Skipping job mountpoint-s3/scripts/fio/read/rand_read_4t_direct_small.fio because it does not match seq_read_small
...

# list the mount and caching directories (sample)
ls -d1 /tmp/fio*

/tmp/fio-ZaNYWdZaBK3W
/tmp/fio-ZaNYWdZaBK3W-cache-O8X4T2w4UDnx

# benchmark results for this run
[
  {
    "name": "cache_sequential_read_small_file",
    "value": 1184.8017578125,
    "unit": "MiB/s"
  },
  {
    "name": "sequential_read_small_file",
    "value": 16.4091796875,
    "unit": "MiB/s"
  }
]
```

There is opportunity of doing some refactor between the different bench scripts but I'm prioritizing getting this out first and do refactor later if needed.

The only impact should be to our CI benchmarks that now will run additional tests that use caching (in a separate job).

---

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -161,3 +161,52 @@ jobs:
         auto-push: ${{ inputs.environment && 'false' || 'true' }}
         comment-on-alert: true
         max-items-in-chart: 20
+
+  cache-bench:
+    name: Benchmark (Cache)
+    runs-on: [self-hosted, linux, x64, nvme-high-performance]
+
+    environment: ${{ inputs.environment }}
+
+    steps:
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v3
+      with:
+        role-to-assume: ${{ vars.ACTIONS_IAM_ROLE }}
+        aws-region: ${{ vars.S3_REGION }}
+        role-duration-seconds: 21600
+    - name: Checkout code
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ inputs.ref }}
+        submodules: true
+        persist-credentials: false
+    - name: Install operating system dependencies
+      uses: ./.github/actions/install-dependencies
+      with:
+        fuseVersion: 2
+        libunwind: true
+        fio: true
+    - name: Set up stable Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        override: true
+    - name: Build
+      run: cargo build --release
+    - name: Run Benchmark
+      run: mountpoint-s3/scripts/fs_cache_bench.sh
+    - name: Check benchmark results
+      uses: benchmark-action/github-action-benchmark@v1
+      with:
+        tool: 'customBiggerIsBetter'
+        output-file-path: results/output.json
+        benchmark-data-dir-path: dev/cache_bench
+        alert-threshold: "200%"
+        fail-on-alert: true
+        # GitHub API token to make a commit comment
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        # Store the results and deploy GitHub pages automatically if the results are from main branch
+        auto-push: ${{ inputs.environment && 'false' || 'true' }}
+        comment-on-alert: true
+        max-items-in-chart: 20
--- a/mountpoint-s3/scripts/fs_cache_bench.sh
+++ b/mountpoint-s3/scripts/fs_cache_bench.sh
@@ -51,6 +51,7 @@ run_fio_job() {
   job_name=$(basename "${job_file}")
   job_name="${job_name%.*}"
 
+
   echo -n "Running job ${job_name} for ${iterations} iterations... "
 
   for i in $(seq 1 $iterations);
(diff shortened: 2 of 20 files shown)


dannycjones (commented):
The code itself looks good.

The concerns I have right now are:
- Is it worthwhile to still run these benchmarks at 1GiB rather than the 100GiB being used in other benchmarks? Should we merge now or rather just get it right in the first place by increasing the available storage?
- Do we want to update the EC2 instances to ensure they have a sensible EBS volume configuration or SSD prior to merging and gathering results? The current EBS configuration used by the runners, while good for many use cases, are not provisioned well for the use of a cache.

arsh:
I think this can be incremental and moves us in the right direction. I'd prefer to push this and tackle the host in a separate PR.

jamesbornholt:
For this one I'd rather we get the config right before merging. We've seen that customers look at our benchmark results to guide them on architecture decisions. We don't want to mislead them with the results that we publish if we're using a platform/config/etc that we actually don't recommend.

arsh:
Ok. I'll reach out to discuss what we think the right hosts should be.

Thanks!

.github/workflows/bench.yml
```diff
@@ -160,3 +160,52 @@ jobs:
...
+
+    environment: ${{ inputs.environment }}
+
+    steps:
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v3
+      with:
+        role-to-assume: ${{ vars.ACTIONS_IAM_ROLE }}
+        aws-region: ${{ vars.S3_REGION }}
+        role-duration-seconds: 21600
```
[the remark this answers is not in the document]
sauraank: I think session length and timeout on the job should be different as in 1 session, we run multiple jobs. If we have high value of job timeout as session length (say 6 hours) 1 job can take entire time of session starving other jobs if the job run into some unexpected error.

arsh:
The caching benchmarks has been updated to use a `m5dn.24xlarge` instance with local storage and test using 100GB files.

mountpoint-s3/scripts/fs_cache_bench.sh
```diff
@@ -0,0 +1,198 @@
...
+
+cache_benchmark () {
+  jobs_dir=mountpoint-s3/scripts/fio/read_cache
+
+  if should_setup_storage; then
+    local_storage=/localstorage
+    sudo mkfs -t ext4 /dev/nvme1n1 #/dev/nvme0n1 is used by root EBS vol
+    sudo mkdir $local_storage
+    current_user_id=$(id -u)
+    sudo mount /dev/nvme1n1 $local_storage
```
sauraank, line 119: was curious to know why are we using NVMe block for mount dir?

arsh: This is the instance local storage which is bigger than the root volume. The root volume is an EBS (over the network disk) with only 30 GB of capacity. For caching, we need at least 100GB of capacity to cache the S3 data locally.

sauraank (approved):

mountpoint-s3/scripts/fs_cache_bench.sh
```diff
@@ -0,0 +1,198 @@
...
+  fi
+
+  for job_file in "${jobs_dir}"/*.fio; do
+
+    if ! should_run_job "${job_file}"; then
+      echo "Skipping job ${job_file} because it does not match ${JOB_NAME_FILTER}"
+      continue
+    fi
+
+    mount_dir=$(mktemp -d $local_storage/fio-XXXXXXXXXXXX)
```
sauraank, line 133: Here `mount_dir` is using the local storage right?

mountpoint-s3/scripts/fs_cache_bench.sh
```diff
@@ -0,0 +1,198 @@
...
+    # creates a cache directoy with the suffix of the mount directory
+    cache_dir=$(mktemp -d -t `basename "${mount_dir}"`-cache-XXXXXXXXXXXX)
```
sauraank, line 135: This creates another directory for cache which is not on local storage. I was thinking cache dir would require local storage/

sauraank (changes requested):

mountpoint-s3/scripts/fs_cache_bench.sh
arsh: Yeah, that is incorrect and now fixed.

mountpoint-s3/scripts/fs_cache_bench.sh
arsh: Yeah. This is not a problem but I reverted it to use `/tmp` for the mount point.

sauraank (approved):
LGTM

Issue #722, opened 2024-01-31
Add the expected behaviours of Mountpoint in the troubleshooting page that customers face


sauraank:
### Tell us more about this new feature.

There are cases which are expected behaviour for Mountpoint but customers might find it surprising. 
For example -

Directory(S3 prefix) disappear when we delete all the files contained within than prefix.
This is an expected behaviour as S3 do not support directories and objects are just grouped using prefix appearing like Directory. We have explained it in [Semantics document](https://github.com/awslabs/mountpoint-s3/blob/83fa7c654623cba34b44f99c3b8930b693ebd0b1/doc/SEMANTICS.md#directories)

We would like these cases to be reported in **troubleshooting page** as well so the customer know about these cases easily and be able to distinguish it from actual errors.

Issue #784, opened 2024-02-26
GitHub Actions credentials appear to expire early for benchmark runs exceeding 1 hour


sauraank:
### Mountpoint for Amazon S3 version

mountpoint-s3 v1.4.1

### AWS Region

us-east-1

### Describe the running environment

Running on Self hosted runner for Github Actions.
Operating system:
Distribution: Ubuntu 22.04.3 LTS
Architecture: x86_64
EC2 instance:
Instance type: m5n.24xlarge
Availability zone: us-east-1b

### Mountpoint options

```shell
target/release/mount-s3 BUCKET-NAME /tmp/mount-s3-g6N4H9xvWgzl --debug --allow-delete --log-directory=logs/readdir_100000 --prefix=benchmark/
```

### What happened?

Github OIDC assume IAM role (i.e. ask AWS STS for a set of short-term credentials) and use them for running the github action. When a bucket is mounted by mountpoint using Assume Role and workload is run for longer than an hour, it starts throwing `ExpiredToken` error. It might be happening with other Web Identity which Assume Role as well. 

### Relevant log output

```shell
Our benchmark logs: https://github.com/awslabs/mountpoint-s3/actions/runs/8002443498/job/21855663125#step:8:190.
```

sauraank:
We can probably increase the session duration to upto [12 hours](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html#API_AssumeRole_RequestParameters).
But, I think we would like the credentials to refreshed after 1 session and not throw error?

dannycjones:
Just sharing what we discussed this morning together here:

Our CI is getting a set of session credentials from `aws-actions/configure-aws-credentials` and populating this in the environment for the job. We don't expect those credentials to be refreshed, as they are static environment credentials and it appears to be behave as expected if - for example - you used short-term credentials in the environment with the AWS CLI.

What we do need to figure out is why our session is 1 hour rather than the expected 6 hours.

arsh:
One hour is the default. I think @sauraank was seeing this issue because he was not using `role-duration-seconds` to override the default.

sauraank:
I understood that part and I have already added its fix in my PR #779 . I created this issue initially because I was thinking we would want the credentials to refreshed after 1 session and not having mountpoint failing. But, that is the expected behaviour. So, we can close this issue once my PR is merged.

sauraank:
[...]

PR #789, opened 2024-02-27
Add new cases of expected behaviours in troubleshooting doc

sauraank:
## Description of change
Updated troubleshooting documentation with 4 new cases. It includes the expected behaviour in case of all file deletion, credential expiry while bucket is still mounted, throttling error and invalid hostname for storage provider not supporting virtual style.
NOTE: I am not adding the the cases of invalid credentials at the time of mounting as the error message in those cases are super clear and user don't need to go to logs to understand the issue.

Relevant issues: 
#722 

No. It is just documentation change.

---

### DIFF
--- a/doc/TROUBLESHOOTING.md
+++ b/doc/TROUBLESHOOTING.md
@@ -169,15 +169,17 @@ In this case, try using `--force-path-style` CLI option when you are mounting th
 
 NOTE - Third party storage provider are not officially supported by Mountpoint for Amazon S3.
 
+For more details on how Mountpoint handles endpoint, please see our [configuration documentation](https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#endpoints-and-aws-privatelink).
+
 ## Directory disappear after deleting all the files within it
 
 Amazon S3 does not support directories and objects are just grouped using prefix.
 Mountpoint automatically infers a directory structure for your bucket by treating the `/` separator after prefix in your object keys as a delimiter between directories.
 
-So, if all the files within a prefix is deleted, the prefix itself cease to exist.
+So, if all the files within a prefix are deleted, the prefix itself and the corresponding directory cease to exist.
 In this case, it is expected that mountpoint will not be able to show the directory for listing or other file system operation.
 
-Workaround to persist a directory could be creating a hidden empty file (for example, `.keep`).
+Workaround to persist a directory could be creating an empty file (for example, `.keep`). These files can be hidden as `ls` filters out files with prefix `.` without `-a` option.
 
 For more details on how Mountpoint maps S3 object keys to files and directories, see the [semantics documentation](https://github.com/awslabs/mountpoint-s3/blob/main/doc/SEMANTICS.md#mapping-s3-object-keys-to-files-and-directories).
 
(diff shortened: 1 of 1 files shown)


doc/TROUBLESHOOTING.md
```diff
@@ -145,3 +145,69 @@ Mountpoint logs should contain an error similar to below:
...
+
+Caused by:
+    0: initial ListObjectsV2 failed for bucket my-bucket in region us-east-1
+    1: Client error
+    2: Unknown CRT error
+    3: CRT error 1059: aws-c-io: AWS_IO_DNS_INVALID_NAME, Host name was invalid for dns resolution.
+Error: Failed to create mount process
+```
+
+In this case, try using `--force-path-style` CLI option when you are mounting the bucket using Mountpoint.
```
vladem, line 165: Basically, _virtual hosted style_ is `https://docexamplebucket.example.com`, compared to _path style_ being `https://example.com/docexamplebucket`. I was only able to comprehend it after reading through our [CONFIGURATION.md](https://github.com/awslabs/mountpoint-s3/blob/82a5fe606c047fee7e7015537cde0b4aabcf0133/doc/CONFIGURATION.md#endpoints-and-aws-privatelink). 

Do you think adding an example (or a link to `CONFIGURATION.md` with such example) will help?

doc/TROUBLESHOOTING.md
```diff
@@ -145,3 +145,69 @@ Mountpoint logs should contain an error similar to below:
...
+
+NOTE - Third party storage provider are not officially supported by Mountpoint for Amazon S3.
+
+## Directory disappear after deleting all the files within it
+
+Amazon S3 does not support directories and objects are just grouped using prefix.
+Mountpoint automatically infers a directory structure for your bucket by treating the `/` separator after prefix in your object keys as a delimiter between directories.
+
+So, if all the files within a prefix is deleted, the prefix itself cease to exist.
```
vladem, line 174:
```suggestion
So, if all the files within a prefix are deleted, the prefix itself and the corresponding directory cease to exist.
```

doc/TROUBLESHOOTING.md
```diff
@@ -145,3 +145,69 @@ Mountpoint logs should contain an error similar to below:
...
+In this case, it is expected that mountpoint will not be able to show the directory for listing or other file system operation.
+
+Workaround to persist a directory could be creating a hidden empty file (for example, `.keep`).
```
vladem, line 177: I'd like documentation to be clear here that MP does not hide or treat in any special way those `.keep` files (nor does S3). It's `ls` who filter them out without an `-a` option, right?

doc/TROUBLESHOOTING.md
```diff
@@ -145,3 +145,69 @@ Mountpoint logs should contain an error similar to below:
...
+[WARN] lookup{req=104 ino=1 name="Input"}:
+list_objects{id=20 bucket=plutodemo continued=false delimiter=/ max_keys=1 prefix=Input/}: mountpoint_s3_client::s3_crt_client:
+meta request failed duration=10.023623ms error=ClientError(Forbidden("The provided token has expired."))
+
+[WARN] lookup{req=106 ino=1 name="Input"}:
+head_object{id=21 bucket=plutodemo key=Input}: mountpoint_s3_client::s3_crt_client:
+meta request failed duration=11.087781ms error=ClientError(Forbidden("<no message>"))
+```
+
+You might need to have a fresh mount of your bucket with valid credentials.
```
vladem, line 206: What could we possibly suggest to avoid this in future? Extending credentials life-time -- probably not, but MP should support mechanisms which support credentials renewal, e.g. if provided through file and the file is updated will MP fetch them?

doc/TROUBLESHOOTING.md
```diff
@@ -145,3 +145,69 @@ Mountpoint logs should contain an error similar to below:
...
+For more details to configure AWS credentials, see the [configuration documentation](https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#aws-credentials).
+
+## Throttling Errors
+
+When looking at the logs, these will appear as failed requests with `http_status=503` or `http_status=429` .
```
vladem, line 211: Log examples, please! Also, as a developer apart from realising that S3 is throttling my app, what should I do? Will MP retry those automatically?

sauraank: The problem with log example here is that I have never experienced throttling from S3. I added it as it was there in support runbook. Probably, we should not be adding this case here.

doc/TROUBLESHOOTING.md
sauraank: I think it is expected behaviour if the credential expire, the aws requests should fail (its the same for CLI). I discussed this issue a bit in #784 . But, we have a bit tricky situation that workload might still be running.

doc/TROUBLESHOOTING.md
sauraank: I have already added the link to aws documentation of virtual hosted style addressing. In that difference is explained well with example.

doc/TROUBLESHOOTING.md
vladem: I see, so the main bit here is that Mountpoint does not do throttling itself. Thanks for describing a way of mitigating S3 throttling errors, but where does this come from? You mention there `3,500 PUT/COPY/POST/DELETE or 5,500 GET/HEAD`, these values may become outdated, I'd add a link to the source of that information.

Otherwise LGTM!

doc/TROUBLESHOOTING.md
```diff
@@ -145,3 +145,86 @@ Mountpoint logs should contain an error similar to below:
...
+list_objects{id=16643 bucket=*** continued=false delimiter=/ max_keys=1 prefix=***}: mountpoint_s3_client::s3_crt_client:
+request failed request_type=Default http_status=503 range=None duration=314.021865ms ttfb=Some(8.180981ms) request_id=***
+```
+
+The 503 or 429 status codes means the request limits have been exceeded.
+Mountpoint itself does not do any throttling, so any throttling will be from S3 or possibly dependent services, like STS which is used to provide credentials.
+
+You can try to mitigate  throttling (503 Slow Down) by distributing objects across multiple prefixes if your use case allows it.
+Since, you can send 3,500 PUT/COPY/POST/DELETE or 5,500 GET/HEAD requests per second per prefix in an S3 bucket, increasing the prefix in the bucket would allow more requests to be processed by S3.
+Amazon S3 gradually scales up to handle requests for each of the prefixes separately.
```
jamesbornholt, line 230:
```suggestion
Amazon S3 automatically scales to high request rates.
Your application can achieve at least 3,500 PUT/COPY/POST/DELETE or 5,500 GET/HEAD requests per second per partitioned Amazon S3 prefix.
You can reduce the impact of throttling errors by distributing objects across multiple prefixes in your bucket.
```

jamesbornholt (approved):

PR #790, opened 2024-02-28
Add request ID to meta request failures and add tests

jamesbornholt:
## Description of change

A side effect of https://github.com/awslabs/mountpoint-s3/pull/669 was that there's now no way to get request IDs for failed requests at the default logging settings, as only DEBUG-level messages include the request IDs. This change adds request IDs to the meta request failure message when available, so that these WARN-level messages still include request IDs.

I also added some new infrastructure to test metrics and log messages. For metrics, we build a new `metrics::Recorder` that collects all the metrics and can then be searched to find them. For log messages, we build a `tracing_subscriber::Layer` that collects all tracing events emitted while enabled. In both cases, the new objects aren't thread safe, as both `Recorder`s and `Layer`s are global state. So these tests need to continue to use `rusty_fork` to split into a new process per test.

Relevant issues: followup to #781.

No.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2000,6 +2000,7 @@ dependencies = [
  "metrics",
  "mountpoint-s3-client",
  "mountpoint-s3-crt",
+ "mountpoint-s3-crt-sys",
  "once_cell",
  "percent-encoding",
  "pin-project",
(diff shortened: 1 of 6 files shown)


monthonk (approved):
LGTM

arsh (approved):
Left a question but not a blocker.

PR #794, opened 2024-03-04
Cancel S3 requests when dropped 

### LINKED ISSUE #510: Dropped GetObject requests are not cancelled in CRT bindings
This issue is opened to track known issue that the current Rust bindings for the CRT do not abort the GetObject request. This means that they may continue to consume memory and bandwidth for some time on data that ultimately won't be used.

Not to be confused with #505 which instead updates the prefetcher to discard the Rust future pulling data from the GetObject request, but doesn't solve the missing plumbing to the CRT.

jamesbornholt:
## Description of change

Today we don't cancel S3 requests when dropped. For our prefetcher that
means we keep streaming (up to) 2GB of data that will never be used.
This change cancels in-flight requests when dropped, so that the CRT
will stop streaming them. Some bytes might still be in flight or
delivered, which is fine. Canceling requests is a no-op if they've
already completed.

The tricky case for this change is PutObject. Our current implementation
of `PutObjectRequest::write` blocks until the bytes it provides are
consumed by the client. But sometimes the client might stop reading from
the stream because the request has failed. That case happens to work
today because we don't retain a reference to the meta request ourselves,
and so the failed request's destructors run immediately after the
failure, which unblocks the writer and returns it an error. But now we do
hold onto a reference, and the destructors can't run until the last
reference is released, so the writer is never unblocked. To fix this, we
make the `write` and `complete` methods of the `PutObjectRequest` poll
_both_ the write stream and the request itself in parallel. If the request
completes, this gives us a chance to bail out of the write rather than
blocking forever.

Relevant issues: fixes #510.

No.

---

### DIFF
--- a/mountpoint-s3-client/tests/get_object.rs
+++ b/mountpoint-s3-client/tests/get_object.rs
@@ -152,3 +152,48 @@ async fn test_get_object_412_if_match() {
         Err(ObjectClientError::ServiceError(GetObjectError::PreconditionFailed))
     ));
 }
+
+#[test_case(false; "early")]
+#[test_case(true; "after read")]
+#[tokio::test]
+async fn test_get_object_cancel(read: bool) {
+    const OBJECT_SIZE: usize = 30_000_000;
+
+    let sdk_client = get_test_sdk_client().await;
+    let (bucket, prefix) = get_test_bucket_and_prefix("test_get_object_cancel");
+
+    // Create one large object named "hello"
+    let key = format!("{prefix}/hello");
+    let body = vec![0x42; OBJECT_SIZE];
+    sdk_client
+        .put_object()
+        .bucket(&bucket)
+        .key(&key)
+        .body(ByteStream::from(body))
+        .send()
+        .await
+        .unwrap();
+
+    let client: S3CrtClient = get_test_client();
+
+    let mut request = client
+        .get_object(&bucket, &key, None, None)
+        .await
+        .expect("get_object should succeed");
+
+    if read {
+        let mut bytes = 0;
+        while let Some(next) = request.next().await {
+            let (_offset, body) = next.expect("part download should succeed");
+            bytes += body.len();
+        }
+        assert_eq!(bytes, OBJECT_SIZE);
+    } else {
+        // Wait a bit for the request to make some progress. This is very racy, but should be short
+        // enough that there are still requests outstanding.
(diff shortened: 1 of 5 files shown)


mountpoint-s3-client/tests/get_object.rs
```diff
@@ -152,3 +152,48 @@ async fn test_get_object_412_if_match() {
...
+        }
+        assert_eq!(bytes, OBJECT_SIZE);
+    } else {
+        // Wait a bit for the request to make some progress. This is very racy, but should be short
+        // enough that there are still requests outstanding.
+        tokio::time::sleep(std::time::Duration::from_millis(500)).await;
+    }
+
+    // Explicitly cancel the request.
+    drop(request);
```
passaro, line 198: I guess here we are just exercising the cancel on drop, since testing that outstanding requests are dropped may be tricky. Maybe it needs a comment?

mountpoint-s3-client/tests/put_object.rs
```diff
@@ -167,9 +169,51 @@ async fn test_put_object_abort() {
 
     drop(request); // Drop without calling complete().
 
-    // Allow for the AbortMultipartUpload to complete.
+    // Try to wait a while for the async abort to complete. For the larger object, this might be
```
passaro, line 172: This seems to still not be enough: https://github.com/awslabs/mountpoint-s3/actions/runs/8134885229/job/22228500040?pr=794#step:8:1228

jamesbornholt: I'm still messing with it to confirm, but it looks like a CRT-side issue — the AbortMultipartUpload can race with in-flight UploadParts, which might succeed after the Abort succeeds and so re-create the upload. I'm not sure if that's worth fixing here (since we're not actually changing how PUT works), so I might just disable the test for large objects.

mountpoint-s3-client/tests/get_object.rs
jamesbornholt: Yeah, I'll update the comment. It would be nice to be able to test this; maybe a little mock S3 server whose response ordering we can control, like the CRT S3 client uses for testing. But that's a problem for future us.

passaro (approved):
LGTM!

monthonk (approved):
LGTM! I was planning to put a PR up for this but the trick with `PutObjectRequest` is what I was still figuring out.

PR #795, opened 2024-03-04
Proof-of-concept prefetcher optimization for some non-sequential read patterns

passaro:
## Description of change

Mountpoint prefetcher is built to maximize throughput on sequential reads. It works by downloading increasingly larger blocks of content in advance while Mountpoint receives sequential read requests. On out-of-order requests (outside a small tolerance window), a new download is started and the pre-fetched content is dropped.

One example of an access pattern where the current logic results in sub-optimal performance is a quasi-sequential pattern, where read requests are sequential for the most part, but can occasionally seek forward (i.e. skip some content) further than the current 16MiB tolerance window. 

A small test using Mountpoint 1.4.1 on a c5.4xlarge instance in eu-west-1 records these timings:

```
Sequential reads (total 10518925312 bytes): 9.7s
Skip pattern, read 399MB, skip 99MB (total 8439925312 bytes): 44.4s
```

This PR demonstrates a change in the prefetcher which can significantly improve throughput for similar access patterns by not dropping in-flight downloads when they "cover" the forward seek. Once applied, the same test gives:

```
Skip pattern, read 399MB, skip 99MB (total 8439925312 bytes): 10.0s
```

Note that this is only a proof of concept to demonstrate the potential improvement. In order to adopt it in Mountpoint, it will require at least a few modifications, e.g.:

* When examining in-flight GetObject requests, we should first look at the range already downloaded, rather than the whole requested range. One way to do it could be to have a "non-blocking" path to read/consume data only if already available.

* `config.max_forward_seek_distance` could then be used in the heuristic to determine whether to wait for more data to be downloaded or cancel the request.

> [!WARNING]
> **Not to be merged in its current form.**

---

### DIFF
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -460,16 +460,6 @@ where
             // Can't seek if there's no requests in flight at all
             return Ok(false);
         };
-        let future_remaining = self.future_tasks.iter().map(|task| task.remaining()).sum::<usize>() as u64;
-        if total_seek_distance
-            >= (current_task.remaining() as u64 + future_remaining).min(self.config.max_forward_seek_distance)
-        {
-            // TODO maybe adjust the next_request_size somehow if we were still within
-            // max_forward_seek_distance, so that strides > first_request_size can still get
-            // prefetched.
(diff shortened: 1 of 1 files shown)


arsh:
Can you share the test code as well? Meaning the application that is reading the data through MP.

passaro:
Superseded by #797

PR #797, opened 2024-03-04
Account already downloaded data when resetting a prefetcher

vladem:
## Description of change

This is a development if an idea from this PR: https://github.com/awslabs/mountpoint-s3/pull/795

The idea is to avoid discarding prefetched data if it is already downloaded. This optimises a certain read pattern, which reads data sequentially but skips chunks between reads.

Here is some benchmarking with `fio` job included in [the pr](https://github.com/awslabs/mountpoint-s3/pull/799) (no cache usage, remounted in between `fio` invocations):
```
# MP built from main
$ fio --directory=/home/vlaad/mounted_s3 /home/vlaad/local/mountpoint-s3/mountpoint-s3/scripts/fio/read/seq_read_skip_read.fio | grep "bw="
  read : io=5888.0MB, bw=20035KB/s, iops=0, runt=300944msec
$ !!
  read : io=5888.0MB, bw=19642KB/s, iops=0, runt=306958msec
$ !!
  read : io=6400.0MB, bw=20691KB/s, iops=0, runt=316740msec
```

```
# MP built from this branch 
$ fio --directory=/home/vlaad/mounted_s3 /home/vlaad/local/mountpoint-s3/mountpoint-s3/scripts/fio/read/seq_read_skip_read.fio
  read : io=47616MB, bw=160757KB/s, iops=0, runt=303308msec
$ !!
  read : io=41728MB, bw=142388KB/s, iops=0, runt=300092msec
$ !!
  read : io=51712MB, bw=172735KB/s, iops=0, runt=306557msec
```

**upd.** updated benchmark results with longer runtime `300s` and larger file `10gb`, lowest observed improvement is `6.8x`

Relevant issues: No

This is not a functional change and thus not a breaking one. We should reason about possible performance regressions in certain read patterns before merging.    

---


mountpoint-s3/src/prefetch.rs
```diff
@@ -461,9 +461,7 @@ where
             return Ok(false);
         };
         let future_remaining = self.future_tasks.iter().map(|task| task.remaining()).sum::<usize>() as u64;
-        if total_seek_distance
-            >= (current_task.remaining() as u64 + future_remaining).min(self.config.max_forward_seek_distance)
-        {
+        if total_seek_distance >= (current_task.remaining() as u64 + future_remaining) {
```
jamesbornholt, line 464: Might be cleaner to just make all the seeking decisions here. I think it's something like this, imagining tasks have a `available()` method that returns how many bytes are available to read right now:
```rust
let mut available = 0;
for task in std::iter::once(&*current_task).chain(self.future_tasks.iter()) {
    available += task.available();
    // If this request isn't fully downloaded then we can't seek to bytes in any later ones
    if task.available() != task.remaining() {
        break;
    }
};

let future_remaining = self.future_tasks.iter().map(|task| task.remaining()).sum::<usize>() as u64;
let remaining = current_task.remaining() + future_remaining;

// If we have enough bytes already downloaded (`available`) to skip straight to this read, then do
// it. Otherwise, we're willing to skip up to `max_forward_seek_distance` ahead without aborting the
// inflight requests.
let max_seek_distance = remaining.min(self.max_forward_seek_distance).max(available);

if total_seek_distance > max_seek_distance {
    return Ok(false);
}
```

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -84,13 +90,21 @@ impl<E: std::error::Error + Send + Sync> PartQueue<E> {
             Ok(part)
         }
     }
+
+    pub fn downloaded(&self) -> usize {
```
jamesbornholt, line 94: Maybe instead implement `available()`, the number of bytes currently in the queue. Saves callers from having to do the math themselves, since "how many bytes can I read right now" is the question they care about.

jamesbornholt (commented):
I like this idea better than #795. Just a couple notes.

mountpoint-s3/src/prefetch/part_queue.rs
passaro: I'm not sure `PartQueue` has enough context for that, but `RequestTask` could do the math and expose `available()` (or equivalent in terms of absolute offset/range, which I think would be clearer to use in prefetch).

mountpoint-s3/src/prefetch.rs
passaro: >  // If this request isn't fully downloaded then we can't seek to bytes in any later ones

It may be an edge case (we only have 2 requests max atm), but I don't see why introduce this limitation: if we are seeking to the range of a `future_task` request, why would it matter if the one in `current_task` isn't complete?

passaro: I think we can rewrite this logic to only have a single loop through the tasks to find the one where the **requested** range contains the seek offset. Then, we can apply the check for the `available()` bytes with the `max_forward_seek_distance` tolerance window.

Also, I suspect it would be clearer if we reframe everything in terms of absolute offsets (or ranges), rather than `reamaining` or `available` bytes.

passaro: Something roughly like this:

```rust
// Jump ahead to the right request
if offset >= current_task.end_offset() {
    self.next_sequential_read_offset = current_task.end_offset();
    self.current_task = None;
    while let Some(next_request) = self.future_tasks.pop_front() {
        if next_request.end_offset() > offset {
            self.current_task = Some(next_request);
            break;
        } else {
            self.next_sequential_read_offset = next_request.end_offset();
        }
    }
    if self.current_task.is_none() {
        // No inflight task containing the target offset.
        trace!(current_offset=?self.next_sequential_read_offset, requested_offset=?offset, "seek failed: not enough inflight data");
        return Ok(false);
    }
    // We could try harder to preserve the backwards seek buffer if we're near the
    // request boundary, but it's probably not worth the trouble.
    self.backward_seek_window.clear();
}

let current_task = self
    .current_task
    .as_mut()
    .expect("a request existed that covered this seek offset");

// If we have enough bytes already downloaded (`available`) to skip straight to this read, then do 
// it. Otherwise, we're willing to wait for the bytes to download only if they're coming "soon", where 
// soon is defined as up to `max_forward_seek_distance` bytes ahead of the available offset. 
if offset > current_task.available_offset() + self.config.max_forward_seek_distance { 
    return Ok(false); 
}
```

vladem: I see 2 decisions which affect functionality, making them explicit: 
1. We seek through tasks which are not fully completed (e.g. `current_task` has downloaded 0.5/1GB, `future_tasks[0]` has downloaded 0.2/1GB, seeked position belongs to `future_tasks[0]` => we drop current and jump to the next)
2. Max offset to wait for is based on `available_offset`: `current_task.available_offset() + self.config.max_forward_seek_distance`, which will make us to wait in more cases.

vladem: Reasoning in absolute offsets also looks clearer to me. I'll update the PR shortly to do that

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -17,27 +17,32 @@ pub struct PartQueue<E: std::error::Error> {
...
 pub fn unbounded_part_queue<E: std::error::Error>() -> (PartQueue<E>, PartQueueProducer<E>) {
     let (sender, receiver) = unbounded();
-    let downloaded = Arc::new(AtomicUsize::new(0));
+    let bytes_counter = Arc::new(AtomicUsize::new(0));
     let part_queue = PartQueue {
         current_part: AsyncMutex::new(None),
         receiver,
         failed: AtomicBool::new(false),
-        downloaded: Arc::clone(&downloaded),
+        bytes_arrived_total: Arc::clone(&bytes_counter),
```
passaro, line 40: nit: maybe `bytes_received`?

vladem: well, I put some thinking in that tbh! `bytes_received` may be misleading because of `PartQueue.receiver`, which may imply that those bytes **were received** (i.e read through receiver), but what it actually tracks the number of bytes arrived to the queue (which may also not have the clearest meaning, naming is complicated)

mountpoint-s3/src/prefetch.rs
```diff
@@ -483,19 +473,33 @@ where
...
 
         let current_task = self
             .current_task
             .as_mut()
             .expect("a request existed that covered this seek offset");
+        // If we have enough bytes already downloaded (`available`) to skip straight to this read, then do
+        // it. Otherwise, we're willing to wait for the bytes to download only if they're coming "soon", where
+        // soon is defined as up to `max_forward_seek_distance` bytes ahead of the available offset.
+        let available_offset = current_task.available_offset();
+        if offset > available_offset.saturating_add(self.config.max_forward_seek_distance) {
```
passaro, line 494: `max_forward_seek_distance` is now the maximum number of bytes we are happy to wait to be downloaded when seeking forward. Can we think of a more appropriate name?

mountpoint-s3/src/prefetch/part_queue.rs
passaro: Well, they are the bytes received by the receiver, so it still sounds good to be. I'd rather clarify in the rustdoc than add a different name.

mountpoint-s3/src/prefetch.rs
jamesbornholt: `max_forward_seek_wait_distance`?

Also, there's a subtle thing here that it doesn't matter if `available_offset + max_forward_seek_wait_distance > current_task.end_offset()`, because the loop above guarantees that the `offset` lands within this task in the loop above. Maybe worth a comment on the line above this one that selects `current_task`?

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -1,3 +1,5 @@
+use std::sync::atomic::AtomicUsize;
+use std::sync::Arc;
```
jamesbornholt, line 2: These should be `crate::sync::...`

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -84,13 +95,21 @@ impl<E: std::error::Error + Send + Sync> PartQueue<E> {
...
+    pub fn bytes_arrived_total(&self) -> usize {
+        self.bytes_arrived_total.load(Ordering::SeqCst)
+    }
 }
 
 impl<E: std::error::Error + Send + Sync> PartQueueProducer<E> {
     /// Push a new [Part] onto the back of the queue
     pub fn push(&self, part: Result<Part, PrefetchReadError<E>>) {
         // Unbounded channel will never actually block
+        let part_size = part.as_ref().ok().map(|part| part.len());
```
jamesbornholt, line 108: nit: reverse these lines, comment is about `send_blocking`

mountpoint-s3/src/prefetch/part_queue.rs
jamesbornholt: I liked `bytes_downloaded`, but also not worth bikeshedding over the name.

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -84,13 +95,21 @@ impl<E: std::error::Error + Send + Sync> PartQueue<E> {
             Ok(part)
         }
     }
+
+    pub fn bytes_arrived_total(&self) -> usize {
```
jamesbornholt, line 99: nit: if we're sticking with the naming above, then `total_bytes_arrived`

jamesbornholt (commented):
Yeah I like this. We should think about how we want metrics for the prefetcher to work now, but let's do that in a separate PR.

mountpoint-s3/src/prefetch/part_queue.rs
vladem: I'll replace with `bytes_received`, may be not completely accurate, but symmetrical to the sending part and there is a comment with a clarification

mountpoint-s3/src/prefetch.rs
vladem: oh, yeah, not regarding this particular line, but I also thought about the case when `offset` is beyond the last task's range, I'll add a comment and will try to add a test (if possible without too much hacks) 

`max_forward_seek_wait_distance` looks good to me, can not think about anything less wordy

mountpoint-s3/src/prefetch/part_queue.rs
vladem: chosen `bytes_received` as a less wordy alternative

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -84,13 +93,21 @@ impl<E: std::error::Error + Send + Sync> PartQueue<E> {
...
-        // Unbounded channel will never actually block
+        let part_size = part.as_ref().ok().map(|part| part.len());
         let send_result = self.sender.send_blocking(part);
+        if let Some(part_size) = part_size {
+            self.bytes_sent.fetch_add(part_size, Ordering::SeqCst);
+        }
+        // Unbounded channel will never actually block
         if send_result.is_err() {
             trace!("closed channel");
         }
```
passaro, line 113: The comment refers to `send_blocking`, so reorder like this:

```suggestion
        let part_size = part.as_ref().ok().map(|part| part.len());
        // Unbounded channel will never actually block
        let send_result = self.sender.send_blocking(part);
        if send_result.is_err() {
            trace!("closed channel");
        }
        if let Some(part_size) = part_size {
            self.bytes_sent.fetch_add(part_size, Ordering::SeqCst);
        }
```

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -102,12 +102,11 @@ impl<E: std::error::Error + Send + Sync> PartQueue<E> {
 impl<E: std::error::Error + Send + Sync> PartQueueProducer<E> {
     /// Push a new [Part] onto the back of the queue
     pub fn push(&self, part: Result<Part, PrefetchReadError<E>>) {
-        let part_size = part.as_ref().ok().map(|part| part.len());
-        let send_result = self.sender.send_blocking(part);
-        if let Some(part_size) = part_size {
-            self.bytes_sent.fetch_add(part_size, Ordering::SeqCst);
+        if let Ok(ref part) = part {
```
passaro, line 105: nit: you can also use:
```suggestion
        if let Ok(part) = &part {
```

passaro (approved):

jamesbornholt (approved):

PR #800, opened 2024-03-07
Record seek_distance metrics whether or not the seek triggers a reset

passaro:
## Description of change

By always recording the length of a seek attempt, we should get a better picture of the read pattern. The `out_of_order` metric can already be used to determine whether or not the seek could be performed without resetting the prefetcher.

N/A

---


vladem (approved):
LGTM! I think, it would be also interesting to record how much downloaded data we're discarding (we know it's more than `max_forward_seek_wait_distance`, but is it 1mb or 10gb?)

PR #700, opened 2024-01-12
Release new crate versions

dannycjones:
## Description of change

There's not many changes here, but let's get what we have out.

Relevant issues: N/A

## Does this change impact existing behavior?

This change itself? No.

The crate releases however include one breaking change for `mountpoint-s3-crt`: "Return type of `common::Uri::host_port` changed from `u16` to `u32`".

Open question: should this be a minor rather than patch update for `mountpoint-s3-crt` version?

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

PR #803, opened 2024-03-07
Remove crate check workflow for non -sys crates

jamesbornholt:
## Description of change

This workflow exists because we're worried about the 10MiB limit. The
    sys crate is the only one that's in any danger of hitting that limit
    because it's packaging a ton of C code, including an entire TLS
    implementation. This workflow is also broken (#700) when updating crate
    versions for multiple crates at the same time, which we really want to
    do.

Since we only care about the size of the sys crate, let's only check
    that one.

No.

---

### DIFF
--- a/.github/workflows/crates.yml
+++ b/.github/workflows/crates.yml
@@ -16,8 +16,6 @@ jobs:
     strategy:
       matrix:
         crate:
-          - mountpoint-s3-client
-          - mountpoint-s3-crt
           - mountpoint-s3-crt-sys
 
     steps:


passaro (approved):
Happy to unblock #700 for now, but I'd like to understand if the comment on line 32 is wrong:
```
 # `--no-verify` avoids building using crates.io dependencies, which for local packages may not be updated yet
```

jamesbornholt:
> Happy to unblock #700 for now, but I'd like to understand if the comment on line 32 is wrong:
> 
> ```
>  # `--no-verify` avoids building using crates.io dependencies, which for local packages may not be updated yet
> ```

`--no-verify` just turns off the final step of `cargo package` where it extracts the final `.tar` file to a new directory and tries to build it. But the dependency resolution still needs to succeed.

Issue #719, opened 2024-01-25
Caching doesn't seem to work


raykrueger:
### Mountpoint for Amazon S3 version

mount-s3 1.3.2

### AWS Region

us-west-2

### Describe the running environment

Ubuntu 22.04 and Amazon Linux 2023

### Mountpoint options

```shell
sudo mount-s3 --cache /mnt/raid0 kruraymo20240124--usw2-az1--x-s3 /mnt/s3
```

### What happened?

Using the --cache arg provides no performance benefit.

I have my cache dir set to /mnt/raid0, which is a 2 nvme raid0 mount. When I read a 10G file from my S3 mount it takes 9-10 seconds. When I read the file every time after the first, I expect that time to be the same speed as reading from that /mnt/raid0 location. Every subsequent read takes the same amount of time as the first. Also, it takes the same amount of time if I have no --cache at all.

### Relevant log output

```shell
# mount-s3 --cache /mnt/raid0 kruraymo20240124--usw2-az1--x-s3 /mnt/s3
bucket kruraymo20240124--usw2-az1--x-s3 is mounted at /mnt/s3

# time cat /mnt/s3/data/file_1 > /dev/null

real	0m9.108s
user	0m0.007s
sys	0m4.287s
# time cat /mnt/s3/data/file_1 > /dev/null

real	0m7.780s
user	0m0.012s
sys	0m4.683s
# time cat /mnt/s3/data/file_1 > /dev/null

real	0m7.410s
user	0m0.014s
sys	0m4.509s

```

The 2s drop is margin for error variation.

The cache is populated.
```
# du -sh /mnt/raid0/mountpoint-cache/
11G	/mnt/raid0/mountpoint-cache/
```

This is the cache performance I expect.
```
# time cat /mnt/raid0/mountpoint-cache/V1/9c/eaf5ecb13463aa2b4834c069b935949355d31e8b9c201fe4f3b461ad5d0132/* > /dev/null

real	0m1.587s
user	0m0.010s
sys	0m1.576s
```
```

**Comment by raykrueger** [comment]
Based on looking at metrics in the debug logs it looks like the cache is "working", blocks are pulled and read from cache. The performance is not close to what I would expect for reading from disk. The same 10G file copied to my /mnt/raid0 disk comes off as such...

```
# time cat /mnt/raid0/file_1 > /dev/null

real	0m1.150s
user	0m0.000s
sys	0m1.150s
```

[...]

PR #808, opened 2024-03-10
Serialization benchmarks.

arsh:
## Description of change

1. Using Criterion benchmarks, I want to show how we see a performance gain (~33% faster) by switching from `bincode` to `bincode2` when reading the cached blocks from disk.
2. I want to get feedback on changes made to `DiskDataCache` to abstract out the serialization library. For example, am I doing the right abstraction? Am I using the trait boundaries in the right way? So, mostly on the Rust-side of things.

### Benchmarks results
In summary, using `bincode2` gets a better performance than `bincode` and closer to the base line of reading the contents of a file to memory (`read_file` in the graphs). From flamegraphs, we noticed that `bincode` was zeroing the vectors in its `fill_buffer` [implementation](https://github.com/bincode-org/bincode/blob/v1.3.3/src/de/read.rs#L143-L149). `bincode2` has a separate [implementation](https://github.com/pravega/bincode2/blob/171ea73400c98e82c3315b91154bde61243b1530/src/de/read.rs#L138-L165) which is more efficient.

#### Reading a file (base line)

Screenshot 2024-03-10 at 10 19 56 AM

#### Reading cache through bincode

Screenshot 2024-03-10 at 10 20 10 AM

#### Reading cache through bincode2

Screenshot 2024-03-10 at 10 20 23 AM

Relevant issues: N/A

No

---

### DIFF
--- a/mountpoint-s3/Cargo.toml
+++ b/mountpoint-s3/Cargo.toml
@@ -16,6 +16,7 @@ async-channel = "2.1.1"
 async-lock = "3.3.0"
 async-trait = "0.1.57"
 bincode = "1.3.3"
+bincode2 = "2.0.1"
 bytes = { version = "1.2.1", features = ["serde"] }
 clap = { version = "4.1.9", features = ["derive"] }
 const_format = "0.2.30"
@@ -54,6 +55,7 @@ aws-config = "1.1.4"
 aws-sdk-s3 = "1.14.0"
 aws-sdk-sts = "1.12.0"
 base16ct = { version = "0.1.1", features = ["alloc"] }
+criterion = "0.5"
 ctor = "0.2.6"
 filetime = "0.2.21"
 futures = { version = "*", features = ["thread-pool"] }
@@ -92,3 +94,10 @@ path = "src/main.rs"
 name = "mock-mount-s3"
 path = "src/bin/mock-mount-s3.rs"
 required-features = ["mountpoint-s3-client/mock"]
+
+[[bench]]
+name = "cache_serialization"
+harness = false
+
+[profile.bench]
+debug = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -623,6 +623,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "bincode2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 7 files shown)


passaro (commented):
This is great! I'll add a few specific comments, but first:

> I want to get feedback on changes made to DiskDataCache to abstract out the serialization library

Do we need to abstract the serialization library? Any reason not to just replace it?

jamesbornholt:
This is cool, but `bincode2` hasn't been updated for 4 years and looks unmaintained, so not sure we should be switching to it.

mountpoint-s3/Cargo.toml
```diff
@@ -16,6 +16,7 @@ async-channel = "2.1.1"
 async-lock = "3.3.0"
 async-trait = "0.1.57"
 bincode = "1.3.3"
+bincode2 = "2.0.1"
```
jamesbornholt, line 19: I don't like that we have to take a non-dev-dependency on "all the versions we want to benchmark" here. I'd rather put it behind a feature flag, or just not commit the versions we're not using.

arsh:
> This is cool, but `bincode2` hasn't been updated for 4 years and looks unmaintained, so not sure we should be switching to it.

True. We are using `bincode 1.3.3` (3 years old) and `bincode` has evolved to `2.x` (a complete rewrite).

I'll try to upgrade to `bincode 2.x` and measure things.

arsh:
> This is cool, but `bincode2` hasn't been updated for 4 years and looks unmaintained, so not sure we should be switching to it.

`bincode 1.x` is in a similar boat. The version we use, `1.3.3`, was released 3 years ago. After that, `bincode` switched to `2.x`. These `2.x` versions are marked with `rc` which I presume means release candidate and not production ready? Since `2.x` claims to be a re-write, I did the same benchmarks using it and did not see the improvements we got from `bincode2`.

One way forward could be to switch to `bincode = 2.x`, do `perf` analysis to determine why its performance isn't as good as `bincode2`, and potentially submit a PR to fix it.

More importantly, I was hoping this micro-optimization would help with the overall cache performance as described in https://github.com/awslabs/mountpoint-s3/issues/719 but when I tested it following what is described there, I did not see an improvement, which looks like something else (to be determined) is the bottleneck.

passaro:
Closing in favor of #1426

PR #809, opened 2024-03-11
Return error from `Uploader::put` on checksum mismatch

vladem:
## Description of change

This changes the way MP treats SSE settings checksum mismatch in case when it occurred **before** the start of the request. Previously MP paniced, the new behaviour is to return an error. 

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/534

No, SSE-KMS is behind a feature flag.

---

### DIFF
--- a/mountpoint-s3/src/upload.rs
+++ b/mountpoint-s3/src/upload.rs
@@ -384,4 +384,49 @@ mod tests {
         assert!(!client.contains_key(key));
         assert!(!client.is_upload_in_progress(key));
     }
+
+    #[test_case(Some("aws:kmr"), Some("some_key_alias"))]
+    #[test_case(Some("aws:kms"), Some("some_key_ali`s"))]
+    #[test_case(None, Some("some_key_alias"))]
+    #[test_case(Some("aws:kms"), None)]
+    #[tokio::test]
+    async fn sse_corruption_error_test(sse_type_corrupted: Option<&str>, key_id_corrupted: Option<&str>) {
+        let client = Arc::new(MockClient::new(Default::default()));
+        let mut uploader = Uploader::new(
+            client,
+            None,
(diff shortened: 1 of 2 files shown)


jamesbornholt:
Seems like we can probably write a unit test for this case now? We could add a `#[cfg(test)]` method to `fs::ServerSideEncryption` that corrupts the checksum, and then put a unit test in `upload.rs` that calls that method at the right time (might need `Arc::get_mut` for that) to inject the corruption.

(With a little refactoring we could probably also do the same for the complete case? But that might be harder).

mountpoint-s3/src/upload.rs
```diff
@@ -49,9 +57,17 @@ impl<Client: ObjectClient> Uploader<Client> {
         &self,
         bucket: &str,
         key: &str,
-    ) -> ObjectClientResult<UploadRequest<Client>, PutObjectError, Client::ClientError> {
+    ) -> Result<UploadRequest<Client>, UploadPutError<PutObjectError, Client::ClientError>> {
         UploadRequest::new(Arc::clone(&self.inner), bucket, key).await
     }
+
+    #[cfg(test)]
+    pub fn corrupt_sse(&mut self, sse_type: Option<String>, sse_kms_key_id: Option<String>) {
```
vladem, line 65: not sure if we want failure-injection api at the Uploader level; if not we can omit the `test_open_with_corrupted_sse` test

jamesbornholt: Yeah, this one we probably don't need the method, since you can write the test in the same module and just touch the private fields directly.

jamesbornholt: Oh I see, each test uses the method from the other file. Seems fine then.

jamesbornholt (approved):

PR #820, opened 2024-03-20
Expose memory consumption metrics

monthonk:
## Description of change

Collect and expose some memory consumption metrics for the prefetcher, CRT buffer pool, and total usage.

No, only metrics

---

### DIFF
--- a/mountpoint-s3/src/metrics.rs
+++ b/mountpoint-s3/src/metrics.rs
@@ -8,6 +8,7 @@ use std::time::Duration;
 
 use dashmap::DashMap;
 use metrics::{Key, Metadata, Recorder};
+use sysinfo::{get_current_pid, MemoryRefreshKind, ProcessRefreshKind, System};
 
 use crate::sync::mpsc::{channel, RecvTimeoutError, Sender};
 use crate::sync::Arc;
@@ -31,6 +32,7 @@ pub const TARGET_NAME: &str = "mountpoint_s3::metrics";
 /// Panics if a sink has already been installed.
 pub fn install() -> MetricsSinkHandle {
     let sink = Arc::new(MetricsSink::new());
+    let mut sys = System::new();
 
     let (tx, rx) = channel();
 
@@ -40,12 +42,16 @@ pub fn install() -> MetricsSinkHandle {
             loop {
                 match rx.recv_timeout(AGGREGATION_PERIOD) {
                     Ok(()) | Err(RecvTimeoutError::Disconnected) => break,
-                    Err(RecvTimeoutError::Timeout) => inner.publish(),
+                    Err(RecvTimeoutError::Timeout) => {
+                        poll_process_metrics(&mut sys);
+                        inner.publish()
+                    }
                 }
             }
             // Drain metrics one more time before shutting down. This has a chance of missing
             // any new metrics data after the sink shuts down, but we assume a clean shutdown
             // stops generating new metrics before shutting down the sink.
+            poll_process_metrics(&mut sys);
             inner.publish();
         })
     };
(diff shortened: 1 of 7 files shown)


mountpoint-s3/src/fuse/session.rs
```diff
@@ -89,8 +92,29 @@ impl FuseSession {
...
+                }
+                Err(RecvTimeoutError::Disconnected) => {
+                    trace!("fuse session's channel is disconnected, closing filesystem session");
+                    break;
+                }
+                Err(RecvTimeoutError::Timeout) => {
+                    // Report process level stats
+                    sys.refresh_process_specifics(pid, ProcessRefreshKind::new().with_memory());
+                    if let Some(process) = sys.process(pid) {
+                        metrics::gauge!("mountpoint_s3.total_memory_usage").set(process.memory() as f64);
```
vladem, line 113: Cool, we'll have a memory usage reported!

What bothers me is the time when we record this value: it is not synchronised with the recording times of other memory metrics,  e.g. `prefetch.bytes_in_queue`. This `total_memory_usage` will likely be queried from procfs in the beginning of [the aggregation period](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/metrics.rs#L22), but published with others at the end of the period.

I am suggesting to query `process.memory()` just before [publishing](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/metrics.rs#L49) the metrics. Though we must be sure that it does not delay the publishing significantly. Does`refresh_process_specifics` takes nanoseconds or higher order of magnitude?

vladem (commented):
overall LGTM, have a single concern regarding the difference in time at which we record new metrics, which may make it difficult to compare those values

for crt metrics, I believe its hard to fix, given the existing code; for procfs probably we could make it more alligned in time?

mountpoint-s3/src/fuse/session.rs
```diff
@@ -89,8 +92,29 @@ impl FuseSession {
     /// Block until the file system is unmounted or this process is interrupted via SIGTERM/SIGINT.
     /// When that happens, unmount the file system (if it hasn't been already unmounted).
     pub fn join(mut self) -> anyhow::Result<()> {
-        let msg = self.receiver.recv();
-        trace!("received message {msg:?}, closing filesystem session");
+        let mut sys =
+            System::new_with_specifics(RefreshKind::new().with_processes(ProcessRefreshKind::new().with_memory()));
+        let pid = Pid::from(process::id() as usize);
```
passaro, line 97: Why not [sysinfo](https://docs.rs/sysinfo/latest/sysinfo/index.html)::[get_current_pid](https://docs.rs/sysinfo/latest/sysinfo/fn.get_current_pid.html#)?

mountpoint-s3/src/fuse/session.rs
```diff
@@ -89,8 +92,29 @@ impl FuseSession {
     /// Block until the file system is unmounted or this process is interrupted via SIGTERM/SIGINT.
     /// When that happens, unmount the file system (if it hasn't been already unmounted).
     pub fn join(mut self) -> anyhow::Result<()> {
-        let msg = self.receiver.recv();
-        trace!("received message {msg:?}, closing filesystem session");
+        let mut sys =
+            System::new_with_specifics(RefreshKind::new().with_processes(ProcessRefreshKind::new().with_memory()));
```
passaro, line 96: Is this retrieving memory info for all processes? In any case, it seems to be redundant if you then use `refresh_process_specifics` on line 111.

mountpoint-s3/src/fuse/session.rs
monthonk: Good point. I was thinking about putting the `total_memory_usage` in the metric sink itself too, but wasn't really sure. But this might be a good reason.

The time it takes for refreshing should be in nanoseconds, we only have to make sure not to initialise the `System` struct every time.

mountpoint-s3/src/fuse/session.rs
monthonk: Yes, this will pre-fill memory info for all processes then we can refresh a specific one later. The performance might be a bit different, but creating an empty `System` should be fine too.

monthonk:
Seems like there is a new update for clippy. Opened https://github.com/awslabs/mountpoint-s3/pull/824 for that.

mountpoint-s3/src/metrics.rs
```diff
@@ -61,6 +67,16 @@ pub fn install() -> MetricsSinkHandle {
     handle
 }
 
+/// Report process level metrics
+fn poll_process_metrics(sys: &mut System) {
+    if let Ok(pid) = get_current_pid() {
+        sys.refresh_process_specifics(pid, ProcessRefreshKind::new().with_memory());
+        if let Some(process) = sys.process(pid) {
+            metrics::gauge!("mountpoint_s3.total_memory_usage").set(process.memory() as f64);
```
jamesbornholt, line 75: nit: something like `process.memory_usage`

also, not sure how useful this is without also knowing how much memory the system has, is that something we can easily get?

vladem: +1, I've contemplated for a while about this method:
https://docs.rs/sysinfo/latest/sysinfo/struct.System.html#method.available_memory

it likely won't be accurate inside a container (without additional effort from a user), but still seems useful

monthonk: yeah, will add this

jamesbornholt (approved):

PR #823, opened 2024-03-21
Add `s3.client.total_bytes` metric

vladem:
## Description of change

Add a `s3.client.total_bytes` metric counting total bytes received from CRT per 5 sec periods:
- It follows the same idea as the existing `fuse.total_bytes` and makes it possible to reason about how much data was downloaded vs how much was consumed;
- It differs from `s3.meta_requests.throughput_mibs`, which:
  - Is only emitted at the end of the request;
  - Is emitted **per-request** thus not providing data about total bandwidth.

Relevant issues: None

---

### DIFF
--- a/mountpoint-s3/src/prefetch/caching_stream.rs
+++ b/mountpoint-s3/src/prefetch/caching_stream.rs
@@ -195,6 +195,7 @@ where
             match get_object_result.next().await {
                 Some(Ok((offset, body))) => {
                     trace!(offset, length = body.len(), "received GetObject part");
+                    metrics::counter!("s3.client.total_bytes", "type" => "read").increment(body.len() as u64);
 
                     let expected_offset = block_offset + buffer.len() as u64;
                     if offset != expected_offset {
--- a/mountpoint-s3/src/prefetch/part_stream.rs
+++ b/mountpoint-s3/src/prefetch/part_stream.rs
@@ -207,6 +207,7 @@ where
                     match get_object_result.next().await {
                         Some(Ok((offset, body))) => {
                             trace!(offset, length = body.len(), "received GetObject part");
+                            metrics::counter!("s3.client.total_bytes", "type" => "read").increment(body.len() as u64);
                             // pre-split the body into multiple parts as suggested by preferred part size
                             // in order to avoid validating checksum on large parts at read.
                             let mut body: Bytes = body.into();


mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -195,6 +195,7 @@ where
             match get_object_result.next().await {
                 Some(Ok((offset, body))) => {
                     trace!(offset, length = body.len(), "received GetObject part");
+                    metrics::counter!("s3.client.total_bytes", "type" => "read").increment(body.len() as u64);
```
jamesbornholt, line 198: maybe `"type" => "read_cached"` so we can distinguish the two?

passaro: This is bytes from S3, not from the cache, though. I don't think we want to report it any differently than when the cache is not enabled.

jamesbornholt: Oh. It's very weird that we have two places that do that.

vladem: It is, I think, @passaro also mentioned earlier that this may benefit from some refactoring. We have 2 similar loops in [part_stream.rs](https://github.com/awslabs/mountpoint-s3/blob/1b61d010a38bc637121fd0a7fb84be287e951d05/mountpoint-s3/src/prefetch/part_stream.rs#L206) and in [caching_stream.rs](https://github.com/awslabs/mountpoint-s3/blob/1b61d010a38bc637121fd0a7fb84be287e951d05/mountpoint-s3/src/prefetch/caching_stream.rs#L190) which pump data from `get_object_result` to `part_queue_producer`. It seems that these 2 blocks have more differences than in-common now.

Common logic:
- reading from `get_object_result`, constructing `ChecksummedBytes`, writing to `part_queue_producer`;
- handling `Err` from  `get_object_result`;
- updating the `s3.client.total_bytes` counter now.

Difference:
- `caching_stream` constructs `block`-size parts, possibly consuming multiple `get_object_result.next()` return-values; 
   - `part_stream` constructs `preferred_part_size`-size parts;
- `caching_stream` updates the cache before writing to `part_queue_producer`;
- `caching_stream` does offset validation for `get_object_result`;
- `caching_stream` may need to flush data on the end of stream (`None`) from `get_object_result`;

One possible refactoring is to perform the construction of a cache block from parts produced by the original `part_stream.rs`. This may imply an additional checksum computation though: at first we checksum parts at `part_stream.rs`, and then we checksum assembled cache blocks. This is still a raw idea though, aligning block offsets may be tricky. Also this would mean having 2 queues: 
1. "get_object_result -> cache writer" queue
2. "cache writer -> prefetcher" queue

passaro: We could consider some refactoring here, but I'd say it's out of scope for this PR.

PR #824, opened 2024-03-22
Fix failing CI jobs

monthonk:
## Description of change

Fix clippy errors from recent updates including
- https://rust-lang.github.io/rust-clippy/master/index.html#/question_mark
- https://rust-lang.github.io/rust-clippy/master/index.html#/suspicious_open_options

Also update cargo about to latest version to make it work with Rust v1.77

No

---

### DIFF
--- a/mountpoint-s3/tests/fuse_tests/lookup_test.rs
+++ b/mountpoint-s3/tests/fuse_tests/lookup_test.rs
@@ -177,12 +177,7 @@ where
 
     // Not really a "lookup" test, but since we're playing with Unicode, may as well do it here
     let new_path = mount_point.path().join("🐈/👻");
-    let mut f = OpenOptions::new()
-        .write(true)
-        .create(true)
-        .truncate(false)
-        .open(&new_path)
-        .unwrap();
+    let mut f = OpenOptions::new().write(true).create_new(true).open(&new_path).unwrap();
     f.write_all("hello world 4".as_bytes()).unwrap();
     f.sync_all().unwrap();
     drop(f);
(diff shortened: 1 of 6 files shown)


jamesbornholt (approved):

mountpoint-s3/tests/fuse_tests/lookup_test.rs
```diff
@@ -177,7 +177,12 @@ where
 
     // Not really a "lookup" test, but since we're playing with Unicode, may as well do it here
     let new_path = mount_point.path().join("🐈/👻");
-    let mut f = OpenOptions::new().write(true).create(true).open(&new_path).unwrap();
+    let mut f = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(false)
```
passaro, line 183: nit: why use `truncate(false)` here? Mountpoint accepts it but does not really support it. Shouldn't we use `truncate(true)` most of the time, or (better yet?) `create_new`?

monthonk: For this PR, I just wanted to fix clippy errors and the default value when we don't specify them is false. So, I'm only matching the behavior.

dannycjones (dismissed):

mountpoint-s3/tests/fuse_tests/lookup_test.rs
dannycjones: >the default value when we don't specify them is false

I think I'd lean on setting `truncate` to false throughout the tests for this reason, except of course where we explicitly test truncate behaviour.

Or `create_new` works too in this test, since it is equivalent. 😄

passaro (approved):
LGTM!

Issue #819, opened 2024-03-18
`ls` fails every second run | `rewinddir` support


luiscape:
### Mountpoint for Amazon S3 version

1.5.0

### AWS Region

us-east-1

### Describe the running environment

EC2 instance (`m6id.2xlarge`) running Ubuntu 20.04.6 LTS.

### Mountpoint options

```shell
mount-s3 --user-agent-prefix "my-agent" --allow-other --log-metrics --debug --cache /tmp/cache/ --metadata-ttl 5 my-bucket /my-mount
```

### What happened?

Calling `ls /my-mount` one time correctly returns the listed files. Calling it a second time fails with:

```
ls: reading directory '/my-mount/': Invalid argument
```

Debug logs show a FUSE error. 

Calling `ls` in any other nested directory doesn't show this error. 

### Relevant log output

```shell
Mar 18 19:58:20 ip-10-1-0-87 mount-s3[1638219]: [DEBUG] fuser::request: FUSE( 36) ino 0x0000000000000001 GETATTR
Mar 18 19:58:20 ip-10-1-0-87 mount-s3[1638219]: [DEBUG] fuser::request: FUSE( 38) ino 0x0000000000000001 READDIRPLUS fh FileHandle(4), offset 0, size 4096
Mar 18 19:58:20 ip-10-1-0-87 mount-s3[1638219]: [WARN] readdirplus{req=38 ino=1 fh=4 offset=0}: mountpoint_s3::fuse: readdirplus failed: out-of-order readdir, expected=166, actual=0
```

arsh:
Hello @luiscape, thanks for reporting this. I was unable to reproduce this issue. Can you provide the full log file?

Thanks!

luiscape:
@arsh I missed a clarification. We run `mount-s3` on the host but expose the mount as a bind mount inside a container (aka "sandbox") in gVisor. gVisor will reuse and rewind dir handles. Maybe mountdir isn't handling `rewinddir` correctly?

I am able to reproduce the issue on the issue with the following C++ program (`listdir.cpp`):

```cpp
#include <stdio.h>
#include <dirent.h>

int main () {
  DIR *dh = opendir("/my-mount");
  if (dh == NULL) {
    printf("failed to open dir\n");
    return -1;
  }

  struct dirent *dirent;
  while ((dirent = readdir(dh)) != NULL) {
    printf("%s\n", dirent->d_name);
  }
  
  printf("rewinding\n");
  rewinddir(dh);
[...]

PR #825, opened 2024-03-23
Adding support for rewinddir by restarting readdir if offset is zero.

arsh:
## Description of change

If we get a `readdir` call with offset 0, we restart the directory stream. In https://github.com/awslabs/mountpoint-s3/pull/581, we added support for consecutive calls with same offset by storing the last response. However, we're not using that feature when offset is 0 because we want any files that were added or removed between the `opendir` and the `rewinddir` to reflect in subsequent calls to `readdir`.

If all the use cases for which https://github.com/awslabs/mountpoint-s3/pull/581 was added are for offset 0, we could remove it. However, since https://github.com/awslabs/mountpoint-s3/pull/581 allows repeated offsets other than 0, it could break code that is relying on this behavior.

It is worth calling out that we cannot distinguish from a `rewinddir` or `seekdir`. We know that `rewinddir` will always call with offset 0, while `seekdir` could be with any offset. Regardless, `seekdir` is still not supported.

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/819

The behavior of `readdir` is changing: now, calling it with an offset of 0 will restart the directory stream, ensuring that new or removed files are accurately reflected. Previously, it was possible to miss new files or still observe removed files if `readdir` returned the last response due to consecutive calls with the same offset.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -50,6 +50,10 @@ impl DirHandle {
     fn next_offset(&self) {
         self.offset.fetch_add(1, Ordering::SeqCst);
     }
+
+    fn rewind_offset(&self) {
+        self.offset.store(0, Ordering::SeqCst);
+    }
 }
 
 #[derive(Debug)]
@@ -995,7 +999,7 @@ where
         // special case where we need to rewind and restart the streaming
         if offset == 0 && dir_handle.offset() != 0 {
             // only do this if this is not the first call with offset 0
-            dir_handle.offset.store(0, Ordering::SeqCst);
+            dir_handle.rewind_offset();
             let new_handle = self.superblock.readdir(&self.client, parent, 1000).await?;
             *dir_handle.handle.lock().await = new_handle;
         }
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -1004,7 +1004,7 @@ where
         if offset == 0 && dir_handle.offset() != 0 {
             // only do this if this is not the first call with offset 0
             dir_handle.rewind_offset();
-            let new_handle = self.default_handle( parent).await?;
+            let new_handle = self.default_handle(parent).await?;
             *dir_handle.handle.lock().await = new_handle;
         }
 
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -939,10 +939,14 @@ where
         Ok(len)
     }
 
+    async fn default_handle(&self, parent: InodeNo) -> Result<ReaddirHandle, InodeError> {
+        self.superblock.readdir(&self.client, parent, 1000).await
+    }
+
     pub async fn opendir(&self, parent: InodeNo, _flags: i32) -> Result<Opened, Error> {
         trace!("fs:opendir with parent {:?} flags {:#b}", parent, _flags);
 
-        let inode_handle = self.superblock.readdir(&self.client, parent, 1000).await?;
+        let inode_handle = self.default_handle(parent).await?;
 
         let fh = self.next_handle();
         let handle = DirHandle {
@@ -1000,7 +1004,7 @@ where
         if offset == 0 && dir_handle.offset() != 0 {
             // only do this if this is not the first call with offset 0
             dir_handle.rewind_offset();
-            let new_handle = self.superblock.readdir(&self.client, parent, 1000).await?;
+            let new_handle = self.default_handle( parent).await?;
             *dir_handle.handle.lock().await = new_handle;
         }
 
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -992,12 +992,20 @@ where
                 .ok_or_else(|| err!(libc::EBADF, "invalid directory handle"))?
         };
 
+        // special case where we need to rewind and restart the streaming
+        if offset == 0 && dir_handle.offset() != 0 {
+            // only do this if this is not the first call with offset 0
+            dir_handle.offset.store(0, Ordering::SeqCst);
+            dir_handle.handle.rewind().await;
+        }
+
         if offset != dir_handle.offset() {
             // POSIX allows seeking an open directory. That's a pain for us since we are streaming
             // the directory entries and don't want to keep them all in memory. But one common case
             // we've seen (https://github.com/awslabs/mountpoint-s3/issues/477) is applications that
             // request offset 0 twice in a row. So we remember the last response and, if repeated,
             // we return it again.
+
             let last_response = dir_handle.last_response.lock().await;
             if let Some((last_offset, entries)) = last_response.as_ref() {
                 if offset == *last_offset {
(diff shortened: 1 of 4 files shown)


mountpoint-s3/src/inode/readdir.rs
```diff
@@ -186,6 +186,11 @@ impl ReaddirHandle {
         self.inner.update_from_remote(self.dir_ino, entry.name(), remote_lookup)
     }
 
+    pub async fn rewind(&self) {
+        self.readded.lock().unwrap().take();
+        self.iter.lock().await.rewind();
+    }
```
jamesbornholt, line 192: Instead of having to reason about this stateful thing, can we just recreate the whole `ReaddirHandle` by calling `self.superblock.readdir` again? It's a little bit wasteful because it needs to recreate the local entries and the ListObjects stream, but it's also much easier to get correct. 

We can go do something fancier if it turns out we don't like the performance of the simpler version. But "something fancy" might be as simple as just keeping the `last_response` code around.

mountpoint-s3/src/inode/readdir.rs
```diff
@@ -491,6 +516,10 @@ mod ordered {
         fn next(&mut self) -> Option<ReaddirEntry> {
             self.entries.pop_front()
         }
+
+        fn rewind(&mut self) {
+            self.entries.clear();
+        }
```
jamesbornholt, line 522: Not sure this is right — we need to include the local entries again if we're going back to 0, but this throws them all away. This kind of subtlety is why I think it might be better to just throw the whole handle away and start again.

jamesbornholt: Either way, let's make sure we have a test covering this case. Something like a directory whose contents are just one local file, iterate the handle once and then rewind it.

mountpoint-s3/src/fs.rs
```diff
@@ -1016,6 +1024,7 @@ where
                     return Ok(reply);
                 }
             }
```
jamesbornholt, line 1026: We don't need this `last_response` case any more, right?

mountpoint-s3/src/fs.rs
```diff
@@ -992,12 +992,20 @@ where
                 .ok_or_else(|| err!(libc::EBADF, "invalid directory handle"))?
         };
 
+        // special case where we need to rewind and restart the streaming
+        if offset == 0 && dir_handle.offset() != 0 {
+            // only do this if this is not the first call with offset 0
+            dir_handle.offset.store(0, Ordering::SeqCst);
```
jamesbornholt, line 998: nit: put this in a method on `DirHandle` to match its `next_offset()` method

mountpoint-s3/src/inode/readdir.rs
arsh: There is a test that adds a local and a remote file after `opendir` and those are reflected in `readdir` after rewinding. I'll add one additional test that covers just the local case as you mentioned.

Regardless, let me try to just recreate everything which should simplify things.

mountpoint-s3/src/inode/readdir.rs
arsh: I'll give this a try!

mountpoint-s3/src/fs.rs
arsh: Decided to leave it since removing it might break existing application that were relying on this behavior.

mountpoint-s3/src/inode/readdir.rs
arsh: Changed to replacing the handled.

mountpoint-s3/src/fs.rs
arsh: Done!

mountpoint-s3/src/fs.rs
```diff
@@ -992,12 +1000,21 @@ where
                 .ok_or_else(|| err!(libc::EBADF, "invalid directory handle"))?
         };
 
+        // special case where we need to rewind and restart the streaming
+        if offset == 0 && dir_handle.offset() != 0 {
+            // only do this if this is not the first call with offset 0
+            dir_handle.rewind_offset();
+            let new_handle = self.default_handle(parent).await?;
```
jamesbornholt, line 1007: I think we need to `rewind_offset` _after_ creating the new handle — if handle creation fails we want to leave the old one in place (which `?` is doing here) and the offset unchanged.

mountpoint-s3/src/fs.rs
```diff
@@ -935,15 +939,19 @@ where
         Ok(len)
     }
 
+    async fn default_handle(&self, parent: InodeNo) -> Result<ReaddirHandle, InodeError> {
+        self.superblock.readdir(&self.client, parent, 1000).await
+    }
```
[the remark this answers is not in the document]
arsh: done

mountpoint-s3/src/fs.rs
```diff
@@ -1070,11 +1088,11 @@ where
         if dir_handle.offset() < 2 {
             let lookup = self
                 .superblock
-                .getattr(&self.client, dir_handle.handle.parent(), false)
+                .getattr(&self.client, dir_handle.handle.lock().await.parent(), false)
```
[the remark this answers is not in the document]
arsh: Moved the lock to be at the top after the rewind logic.

I'm not sure what prevents (if anything) concurrent calls to `readdir`. I wrote C code that calls it on the same `fd` and it worked fine. Reading `readdir` man page it says this is not expected to be thread-safe as per the POSIX standard. **Open to hear your ideas on how to reason about this if this is not exactly what you meant.**

For completeness, here is the C code and output. `mnt` is mounted vi Mountpoint.

Output

```
~/workspace/rewind » ./readdir_parallel ~/mnt
[139900944557824] .
[139900944557824] bench10GB.bin
[139900944557824] bench5MB.bin
[139900923578112] A
[139900923578112] new-file
[139900934067968] ..
[139900944557824] bench_dir_100
```

Code

```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <dirent.h>
#include <pthread.h>

#define NUM_THREADS 3

void *readdir_parallel(void *dir_ptr) {
    DIR *dir = (DIR *)dir_ptr;
    struct dirent *entry;

    // Read directory entries until end is reached
    while ((entry = readdir(dir)) != NULL) {
        printf("[%ld] %s\n", pthread_self(), entry->d_name);
    }

    pthread_exit(NULL);
}

int main(int argc, char *argv[]) {
    if (argc != 2) {
        printf("Usage: %s <directory>\n", argv[0]);
        return EXIT_FAILURE;
    }

    DIR *dir = opendir(argv[1]);
    if (dir == NULL) {
        perror("opendir");
        return EXIT_FAILURE;
    }

    // Create threads to read directory entries in parallel
    pthread_t threads[NUM_THREADS];
    for (int i = 0; i < NUM_THREADS; i++) {
        if (pthread_create(&threads[i], NULL, readdir_parallel, dir) != 0) {
            fprintf(stderr, "Error creating thread %d\n", i);
            break;
        }
    }

    // Wait for all threads to finish
    for (int i = 0; i < NUM_THREADS; i++) {
        pthread_join(threads[i], NULL);
    }

    closedir(dir);
    return EXIT_SUCCESS;
}
```

jamesbornholt (approved):

Issue #807, opened 2024-03-08
Support providing `AES256` SSE type with each PutObject request


vladem:
### Tell us more about this new feature.

[The current implementation](https://github.com/awslabs/mountpoint-s3/blob/74df3e2e00c4a7c567832a40287830ee81fe2d26/mountpoint-s3/src/cli.rs#L279) of SSE settings (which is currently hidden behind a compile-time feature flag) only allows specifying `aws:kms`, `aws:kms:dsse` as the possible options to pass to S3 servers. 

There are cases when it's desired to pass `AES256` sse-type instead, e.g. with the policy provided in the [comment](https://github.com/awslabs/mountpoint-s3/issues/534#issuecomment-1961818855):
```
{
    "Sid": "...",
    "Effect": "Deny",
    "Principal": "*",
    "Action": "s3:PutObject",
    "Resource": "...",
    "Condition": {
        "StringNotEquals": {
            "s3:x-amz-server-side-encryption": "AES256"
        }
    }
}
```

dannycjones:
This was released in v1.6.0! https://github.com/awslabs/mountpoint-s3/releases/tag/mountpoint-s3-1.6.0

Merged in #827.

PR #827, opened 2024-03-26
Add AES256 SSE type

vladem:
## Description of change

Add support for SSE-S3 (`sse_type` flag still follows the convention of `aws s3 cp` and `aws s3api putobject`) 

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/807

---

### DIFF
--- a/mountpoint-s3-client/tests/put_object.rs
+++ b/mountpoint-s3-client/tests/put_object.rs
@@ -404,6 +404,7 @@ async fn check_sse(
         .expect("head object should succeed");
     let (expected_sse, expected_sdk_sse) = match expected_sse {
         None => (Some("AES256"), aws_sdk_s3::types::ServerSideEncryption::Aes256),
+        Some("AES256") => (Some("AES256"), aws_sdk_s3::types::ServerSideEncryption::Aes256),
         Some("aws:kms") => (Some("aws:kms"), aws_sdk_s3::types::ServerSideEncryption::AwsKms),
         Some("aws:kms:dsse") => (
             Some("aws:kms:dsse"),
@@ -450,6 +451,7 @@ async fn check_sse(
 #[test_case(Some("aws:kms:dsse"), Some(get_test_kms_key_id()))]
 #[test_case(Some("aws:kms:dsse"), None)]
 #[test_case(None, None)]
+#[test_case(Some("AES256"), None)]
 #[tokio::test]
 #[cfg(not(feature = "s3express_tests"))]
 async fn test_put_object_sse(sse_type: Option<&str>, kms_key_id: Option<String>) {
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -608,7 +608,7 @@ where
     validate_mount_point(&args.mount_point)?;
     #[cfg(feature = "sse_kms")]
     {
-        validate_sse_args(&args.sse, &args.sse_kms_key_id)?;
+        validate_sse_args(args.sse.as_deref(), args.sse_kms_key_id.as_deref())?;
     }
 
     let (client, runtime, s3_personality) = client_builder(&args)?;
@@ -922,10 +922,13 @@ fn validate_mount_point(path: impl AsRef<Path>) -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Disallow specifying `--sse-kms-key-id` when `--sse=AES256` as this is not allowed by the S3 API.
+/// We are not able to perform this check via clap API (the closest it has is `conflicts_with` method),
+/// thus having a custom validation.
 #[cfg(feature = "sse_kms")]
-fn validate_sse_args(sse_type: &Option<String>, sse_kms_key_id: &Option<String>) -> anyhow::Result<()> {
-    if sse_kms_key_id.is_some() && *sse_type == Some("AES256".to_owned()) {
-        Err(anyhow!("--sse_kms_key_id can not be used with AES256 sse_type"))
+fn validate_sse_args(sse_type: Option<&str>, sse_kms_key_id: Option<&str>) -> anyhow::Result<()> {
+    if sse_kms_key_id.is_some() && sse_type == Some("AES256") {
+        Err(anyhow!("--sse-kms-key-id can not be used with --sse AES256"))
     } else {
         Ok(())
     }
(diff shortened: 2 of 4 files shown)


mountpoint-s3/src/cli.rs
```diff
@@ -606,6 +606,10 @@ where
     tracing::debug!("{:?}", args);
 
     validate_mount_point(&args.mount_point)?;
+    #[cfg(feature = "sse_kms")]
+    {
+        validate_sse_args(&args.sse, &args.sse_kms_key_id)?;
```
vladem, line 611: I was thinking of embedding this check in `CliArgs` with clap API, but seems that it is just not expressive enough for this use case.

The intention is to make `--sse_kms_key_id` to conflict with a specific value of `--sse`. The closest API clap has is [conflicts_with](https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with), which makes it possible to restrict specifying `--sse_kms_key_id` when `--sse` is present. If there was an API `conflicts_with_if_eq` (similar to an existing [required_if_eq](https://docs.rs/clap/latest/clap/struct.Arg.html#method.required_if_eq)), than the check would be super neat.

But for now [custom validation](https://docs.rs/clap/latest/clap/_derive/_tutorial/chapter_3/index.html#custom-validation) as a last resort seems right.

arsh: This is important context that should be in the code as comments.

arsh (commented):
Just a minor comment.

mountpoint-s3-client/tests/put_object.rs
```diff
@@ -450,6 +451,7 @@ async fn check_sse(
 #[test_case(Some("aws:kms:dsse"), Some(get_test_kms_key_id()))]
 #[test_case(Some("aws:kms:dsse"), None)]
 #[test_case(None, None)]
+#[test_case(Some("AES256"), None)]
```
passaro, line 454: Can you specify `kms_key_id` here? If not, how does it fail?

mountpoint-s3/src/cli.rs
```diff
@@ -918,6 +922,15 @@ fn validate_mount_point(path: impl AsRef<Path>) -> anyhow::Result<()> {
     Ok(())
 }
 
+#[cfg(feature = "sse_kms")]
+fn validate_sse_args(sse_type: &Option<String>, sse_kms_key_id: &Option<String>) -> anyhow::Result<()> {
```
passaro, line 926: nit: prefer `Option<&str>` for arguments

mountpoint-s3/src/cli.rs
```diff
@@ -918,6 +922,15 @@ fn validate_mount_point(path: impl AsRef<Path>) -> anyhow::Result<()> {
...
+    if sse_kms_key_id.is_some() && *sse_type == Some("AES256".to_owned()) {
```
passaro, line 927: Could you rewrite this check as a pattern match?

mountpoint-s3/src/cli.rs
```diff
@@ -918,6 +922,15 @@ fn validate_mount_point(path: impl AsRef<Path>) -> anyhow::Result<()> {
...
+        Err(anyhow!("--sse_kms_key_id can not be used with AES256 sse_type"))
```
passaro, line 928: nit: use the actual flag
```suggestion
        Err(anyhow!("--sse_kms_key_id can not be used with --sse AES256"))
```

mountpoint-s3-client/tests/put_object.rs
vladem: no, it's not allowed by s3 api, it fails on `PutObjectRequest:write`:

```
ClientError(ResponseError(MetaRequestResult { response_status: 400, crt_error: Error(14343, "aws-c-s3: AWS_ERROR_S3_INVALID_RESPONSE_STATUS, Invalid response status from request"), error_response_headers: Some(Headers { inner: 0x7f10c00175d0 }), error_response_body: Some("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>InvalidArgument</Code><Message>Server Side Encryption with AWS KMS managed key requires HTTP header x-amz-server-side-encryption : aws:kms</Message><ArgumentName>x-amz-server-side-encryption</ArgumentName><ArgumentValue>AES256</ArgumentValue>...</Error>") }))
```

does it make sense to put a client-side check for this? (e.g. [here](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-client/src/s3_crt_client/put_object.rs#L57))

mountpoint-s3/src/cli.rs
jamesbornholt: nit: it's actually `--sse-kms-key-id` (clap converts _ to - in arguments by default)

mountpoint-s3/src/cli.rs
vladem: the `if` is clearer for my eye, but may be I am just not used to this style:
```
match sse_kms_key_id {
    Some(_) => {
        match sse_type {
            Some("AES256") => Err(anyhow!("--sse-kms-key-id can not be used with --sse AES256")),
            _ => Ok(())
        }
    }
    _ => Ok(())
}
```

mountpoint-s3/src/cli.rs
vladem: makes sense, added

mountpoint-s3/src/cli.rs
vladem: nice catch, fixed the error message

mountpoint-s3/src/cli.rs
dannycjones: Forgive me if I'm wrong, I think Ale was thinking something more along the lines of...

```rust
match (sse_type, sse_kms_key_id) {
    (Some("AES256"), Some(_key_id)) => Err(...)
    ...
}
```

jamesbornholt (approved):

mountpoint-s3/src/cli.rs
jamesbornholt: Using my timezone advantage to merge this before you can bikeshed it :-)

passaro: For posterity, now that we can no longer bikeshed: my preference would have been for @dannycjones' suggestion, but after the change to `&str` (and removal of `.to_owned()`) it would have been a minor nit.

Issue #833, opened 2024-04-03
Full throughput limit is not being used 


karthik1900:
### Mountpoint for Amazon S3 version

mount-s3 1.5.0

### AWS Region

us-east-1

### Describe the running environment

Running on Amazon EC2 g5.12xlarge (24 cores)
operating system: Amazon linux2

### Mountpoint options

```shell
mount-s3 <BUCKET> <DIR>

also tried the following options
1. maximum-throughput-gbps: 50/100
2. max-threads: 16/24/32/48
3. part-size: 8MB / 16MB / 32MB / 80MB
```

### What happened?

The said machine g5.12xlarge has a bandwidth of 40Gbps. While reading a file of size 73.6Gb, the max throughput I observed is around 6Gbps.

PS: I didn't see any server side errors in syslogs

monthonk:
Hey @karthik1900, that throughput seems really low. Can you share more about how you read data from the mounted directory? Based on our benchmark results (https://awslabs.github.io/mountpoint-s3/dev/bench/), increasing number of concurrent reads and opening a file in `O_DIRECT` mode might help.

karthik1900:
Hey @monthonk, thanks for the quick response

I am using a simple python open file
```
with open("myfile", "rb") as f:
    byte = f.read()
```

Will try with the O_DIRECT mode and update here

karthik1900:
When I use fio to to read the file, I am getting about 22 Gb/s, I guess the issue is with my read
```
fio --name=read --rw=read --bs=4m --direct=1 --filename="myfile"
```

monthonk:
Thanks for the updates. For your workload, you can try spawning threads to read multiple files at the same time. That should get you close to the instance bandwidth. We will also update documentation on our side.

karthik1900:
Hey @monthonk 

Is there a way where we can set a configuration which will allow multiple threads to read one large file (as s3 allows reading in chunks)

dannycjones:
Hi! We merged that particular documentation change, so we now have a section in troubleshooting slower throughput than expected: https://github.com/awslabs/mountpoint-s3/blob/main/doc/TROUBLESHOOTING.md#slower-throughput-than-expected

If there's any more questions here, please let us know - or we can close the issue. Thanks!

PR #834, opened 2024-04-03
Add new troubleshooting section for 'slower throughput than expected'

dannycjones:
## Description of change

Adding a section that to help direct customers towards opening files concurrently over just reading files in sequence.

It also helps capture our expectations around single file throughput - i.e. that it depends on the instance type / CPU, and that we expect to achieve the best speeds by processing files concurrently.

Relevant issues:
- #833

Documentation change only.

---


jamesbornholt (commented):
Probably also worth a specific note about the common `cp -R` pattern not being parallel by default.

jamesbornholt (approved):

PR #844, opened 2024-04-10
Remove fuse_tests from asan

jamesbornholt:
## Description of change

They frequently trigger a deadlock inside ASan's allocator. We're really using ASan to test the CRT bindings anyway, so the S3 tests are really what matter most. I don't love this change because future features won't get added automatically, but I'm more worried right now about unblocking a flaky CI.

No.

---


passaro (approved):

jamesbornholt:
This is a workflow change to fix CI, so merging without it passing.

PR #849, opened 2024-04-12
 Add option to disable trailing checksums for uploads

### LINKED ISSUE #516: Amazon S3 on Outposts support for write workloads

jamesbornholt:
## Description of change

Some S3 implementations (notably S3 on Outposts) don't support the S3 additional checksums feature, which causes uploads to fail. This change adds a new command-line flag to disable additional checksums for uploads. We also automatically disable additional checksums when we detect (via endpoint resolution) that the target is an Outposts bucket.

Even when trailing checksums are disabled, we still validate the checksums of parts right before committing an upload.
This change adopts a new CRT ability to decouple upload review checksums from the actual S3 headers (https://github.com/awslabs/aws-c-s3/pull/421), so that we can still validate upload checksums locally without sending them to S3.

We don't have (and likely won't have) Outposts in CI, but I've tested manually like this:
```
[ec2-user@ip-172-31-44-14 ~]$ ./mount-s3 arn:aws:s3-outposts:us-east-1:691736938728:outpost/op-0f129d3877cfede5e/accesspoint/bornholt-test-bucket-ap ~/mnt
bucket arn:aws:s3-outposts:us-east-1:691736938728:outpost/op-0f129d3877cfede5e/accesspoint/bornholt-test-bucket-ap is mounted at /home/ec2-user/mnt
[ec2-user@ip-172-31-44-14 ~]$ echo "hello world!" > ~/mnt/test.txt
[ec2-user@ip-172-31-44-14 ~]$ cat ~/mnt/test.txt
hello world!
[ec2-user@ip-172-31-44-14 ~]$ aws s3api get-object --bucket arn:aws:s3-outposts:us-east-1:691736938728:outpost/op-0f129d3877cfede5e/accesspoint/bornholt-test-bucket-ap --key test.txt outfile
{
    "AcceptRanges": "bytes",
    "LastModified": "2024-04-12T02:20:29+00:00",
    "ContentLength": 13,
    "ETag": "\"8e8732f3d5618766a7ddfea2a6594f7b-1\"",
    "ContentType": "text/plain",
    "Metadata": {},
    "StorageClass": "OUTPOSTS"
}
[ec2-user@ip-172-31-44-14 ~]$ cat outfile
hello world!
```
(Outpost access point aliases also work the same way)

Relevant issues: fixes #516.

Not by default.

---

### DIFF
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -634,6 +634,8 @@ where
     let (client, runtime, s3_personality) = client_builder(&args)?;
 
     let bucket_description = args.bucket_description();
+    tracing::debug!("using S3 personality {s3_personality:?} for {bucket_description}");
+
     let fuse_config = args.fuse_session_config();
 
     let mut filesystem_config = S3FilesystemConfig::default();
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -290,10 +290,11 @@ pub struct CliArgs {
 
     #[clap(
         long,
-        help = "Disable S3 additional checksums for object uploads",
+        help = "Checksum algorithm to use for S3 uploads [default: crc32c]",
         help_heading = BUCKET_OPTIONS_HEADER,
+        value_name = "ALGORITHM",
     )]
-    pub disable_upload_checksums: bool,
+    pub upload_checksums: Option<UploadChecksums>,
 }
 
 #[derive(Debug, Clone)]
@@ -320,6 +321,25 @@ impl ValueEnum for BucketType {
         match self {
             Self::GeneralPurpose => Some(clap::builder::PossibleValue::new("general-purpose")),
             Self::Directory => Some(clap::builder::PossibleValue::new("directory")),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum UploadChecksums {
+    Crc32c,
+    Off,
+}
+
+impl ValueEnum for UploadChecksums {
+    fn value_variants<'a>() -> &'a [Self] {
+        &[Self::Crc32c, Self::Off]
+    }
+
+    fn to_possible_value(&self) -> Option<clap::builder::PossibleValue> {
+        match self {
+            Self::Crc32c => Some(clap::builder::PossibleValue::new("crc32c")),
+            Self::Off => Some(clap::builder::PossibleValue::new("off")),
         }
     }
 }
@@ -654,13 +674,18 @@ where
     filesystem_config.storage_class = args.storage_class;
     filesystem_config.allow_delete = args.allow_delete;
     filesystem_config.allow_overwrite = args.allow_overwrite;
-    filesystem_config.use_upload_checksums = !args.disable_upload_checksums;
-    if !s3_personality.supports_additional_checksums() {
+    filesystem_config.s3_personality = s3_personality;
+    filesystem_config.server_side_encryption = ServerSideEncryption::new(args.sse, args.sse_kms_key_id);
+
+    // Written in this awkward way to force us to update it if we add new checksum types
+    filesystem_config.use_upload_checksums = match args.upload_checksums {
+        Some(UploadChecksums::Crc32c) | None => true,
+        Some(UploadChecksums::Off) => false,
+    };
+    if !s3_personality.supports_additional_checksums() && args.upload_checksums.is_none() {
         tracing::info!("disabling upload checksums because target S3 personality does not support them");
         filesystem_config.use_upload_checksums = false;
     }
-    filesystem_config.s3_personality = s3_personality;
-    filesystem_config.server_side_encryption = ServerSideEncryption::new(args.sse, args.sse_kms_key_id);
 
     let prefetcher_config = Default::default();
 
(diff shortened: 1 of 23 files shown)


mountpoint-s3/src/cli.rs
```diff
@@ -632,10 +652,13 @@ where
     filesystem_config.storage_class = args.storage_class;
     filesystem_config.allow_delete = args.allow_delete;
     filesystem_config.allow_overwrite = args.allow_overwrite;
-    filesystem_config.s3_personality = s3_personality;
-    {
-        filesystem_config.server_side_encryption = ServerSideEncryption::new(args.sse, args.sse_kms_key_id);
+    filesystem_config.use_upload_checksums = !args.disable_upload_checksums;
+    if !s3_personality.supports_additional_checksums() {
+        tracing::info!("disabling upload checksums because target S3 personality does not support them");
+        filesystem_config.use_upload_checksums = false;
```
arsh, line 658: It would be nice if customers didn't have to upgrade Mountpoint when Outpost starts supporting it. Or you think we'd need more changes in MP to support it anyways?

mountpoint-s3/src/cli.rs
```diff
@@ -866,13 +889,13 @@ fn get_region(args_region: Option<String>, instance_info: &InstanceInfo) -> (Str
     (DEFAULT_REGION.to_owned(), false)
 }
 
-fn get_s3_personality(
-    args_personality: Option<S3PersonalityArg>,
+fn infer_s3_personality(
```
arsh, line 892: do we log the personality anywhere? if not, should we?

mountpoint-s3/src/upload.rs
```diff
@@ -193,17 +202,17 @@ impl<Client: ObjectClient> Debug for UploadRequest<Client> {
 fn verify_checksums(review: UploadReview, expected_size: u64, expected_checksum: Crc32c) -> bool {
     let mut uploaded_size = 0u64;
     let mut uploaded_checksum = Crc32c::new(0);
-    for part in review.parts {
+    for (i, part) in review.parts.iter().enumerate() {
         uploaded_size += part.size;
 
         let Some(checksum) = &part.checksum else {
-            error!("missing part checksum");
+            error!(part_number = i, "missing part checksum");
```
arsh, line 209: didn't i see a comment saying part numbers start at 1?

mountpoint-s3/src/cli.rs
jamesbornholt: I think we could probably infer it from other log messages, but easy enough to add anyway.

mountpoint-s3/src/cli.rs
jamesbornholt: I don't expect we'd need more changes, but I think it's better for us to do this detection now and remove it in a future upgrade when ready. Otherwise, Outposts customers would have to add the disable flag themselves today, and then remember to remove it in the future. This way, we can do that upgrade for them automatically when ready.

arsh (dismissed):

mountpoint-s3/src/cli.rs
```diff
@@ -286,20 +287,39 @@ pub struct CliArgs {
         value_parser = clap::builder::NonEmptyStringValueParser::new(),
     )]
     pub sse_kms_key_id: Option<String>,
+
+    #[clap(
+        long,
+        help = "Disable S3 additional checksums for object uploads",
+        help_heading = BUCKET_OPTIONS_HEADER,
+    )]
+    pub disable_upload_checksums: bool,
```
dannycjones, line 296: I'm a little uncertain about this one - do we want a boolean, or would we rather have some kind of mode parameter like `--trailing-checksums <none|crc32c>`? Today we use CRC-32C or turn them off, but we could allow customers to choose between additional checksums (SHA-1, SHA-256, CRC-32) in the future if we really wanted to.

Maybe it provides a path for Outpost customers to opt-in to checksums without being forced to upgrade to get checksums too if they became available, as discussed here: https://github.com/awslabs/mountpoint-s3/pull/849/files#r1565606507?

We could have `--disable-upload-checksums` easily alias to `--trailing-checksums none` in the future if we did make that change, so I don't have a strong preference here.

dannycjones (commented):
Generally looks good - I do have a query on the CLI argument.

I'd like us to get into the habit of updating the changelog as we add changes - do you mind to update for MP and MP Client in this PR?

mountpoint-s3/src/cli.rs
jamesbornholt: Yeah, I like this idea, done.

mountpoint-s3/src/cli.rs
dannycjones: We now have a way to let customers opt-in to checksums on older versions too now, nice. Resolving this one.

dannycjones (commented):
One minor comment change, otherwise LGTM!

dannycjones (approved):
@jamesbornholt updated the comment I mentioned, please check and merge if you're happy.

PR #852, opened 2024-04-14
Allow building on 32-bit platforms

vladem:
## Description of change

Enforcing a limit on `S3ClientConfig::part_size` to be `4GiB` on a 32-bit platform (this is the maximum value which fits in `usize` on a 32-bit platform). `4GiB` is the maximum possible RAM size. Making `part_size` greater than the available RAM won't ever be efficient.

Relevant issues: none

No. On 32-bit platforms we don't pass compilation now. On 64-bit platforms behaviour stays exactly the same.

---

### DIFF
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -160,7 +160,7 @@ pub struct CliArgs {
         long,
         help = "Part size for multi-part GET and PUT",
         default_value = "8388608",
-        value_parser = value_parser!(u64).range(1..),
+        value_parser = value_parser!(u64).range(1..usize::MAX as u64),
         help_heading = CLIENT_OPTIONS_HEADER
     )]
     pub part_size: u64,
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -78,6 +78,7 @@ macro_rules! event {
         }
     }
 }
+
 /// Configurations for the CRT-based S3 client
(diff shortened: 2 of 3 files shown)


mountpoint-s3/src/cli.rs
```diff
@@ -160,7 +160,7 @@ pub struct CliArgs {
         long,
         help = "Part size for multi-part GET and PUT",
         default_value = "8388608",
-        value_parser = value_parser!(u64).range(1..),
+        value_parser = value_parser!(u64).range(1..usize::MAX as u64),
```
jamesbornholt, line 163: I thought this upper bound was already implicit, does it actually change anything?

vladem: on 64-bit it doesn't change anything, the limit is already there:
```
$ mount-s3 --version
mount-s3 1.6.0
$ mount-s3 <bucket> <path> --debug --foreground --part-size=18446744073709551616
error: invalid value '18446744073709551616' for '--part-size <PART_SIZE>': number too large to fit in target type
```

but in the context of 32-bit system, where `usize::MAX` (type expected by the `S3CrtClient`) is less than `u64::MAX` there will be a change, since the cli argument type is `u64` (and there is no good way to specify `usize`, `value_parser!` macro does not compile with `usize`)

jamesbornholt: Ah, I missed that it was `usize::MAX` not `u64::MAX`!

jamesbornholt (approved):

PR #854, opened 2024-04-16
PoC to buffer write requests

passaro:
Buffer write requests into part-size buffers before sending them to the CRT client. 

**Draft**: it will need some tidy up and fixing test failures (cancellation/error reporting).

---

### DIFF
--- a/mountpoint-s3-client/Cargo.toml
+++ b/mountpoint-s3-client/Cargo.toml
@@ -11,6 +11,7 @@ description = "High-performance Amazon S3 client for Mountpoint for Amazon S3."
 mountpoint-s3-crt = { path = "../mountpoint-s3-crt", version = "0.7.0" }
 mountpoint-s3-crt-sys = { path = "../mountpoint-s3-crt-sys", version = "0.7.0" }
 
+async-channel = "2.1.1"
 async-trait = "0.1.57"
 auto_impl = "1.1.2"
 base64ct = { version = "1.6.0", features = ["std"] }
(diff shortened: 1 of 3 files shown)


mountpoint-s3-client/Cargo.toml
```diff
@@ -11,6 +11,7 @@ description = "High-performance Amazon S3 client for Mountpoint for Amazon S3."
 mountpoint-s3-crt = { path = "../mountpoint-s3-crt", version = "0.7.0" }
 mountpoint-s3-crt-sys = { path = "../mountpoint-s3-crt-sys", version = "0.7.0" }
 
+async-channel = "2.1.1"
```
dannycjones, line 14: Is the channel provided by tokio (which we already have a dep on) useful here?

https://docs.rs/tokio/latest/tokio/sync/mpsc/index.html

mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -172,12 +198,30 @@ impl PutObjectRequest for S3PutObjectRequest {
             create_mpu.await.unwrap()?;
         }
 
-        // Write will fail if the request has already finished (because of an error).
-        self.body
-            .meta_request
-            .write(slice, false)
-            .await
-            .map_err(S3RequestError::CrtError)?;
+        // Send data to the write queue.
```
vladem, line 201: the backpressure is not implemented in any form with this change, right? was it the same before switching to async write API?

By not having backpressure I mean that `write()` syscalls complete as soon as they've buffered data in RAM, user application has no way of finding out that S3 requests do not adjust to their writing speed and the buffer just keep growing. This is the behaviour before now and before the async write API?

passaro:
Dropping in favor of #874.

Issue #768, opened 2024-02-21
`--metadata-ttl` option requires `--cache` directory


mhnap:
Not sure whether it's indented, but `--metadata-ttl` option requires `--cache` directory while `--cache` directory is not needed and is not used for metadata caching. Also `--cache` help states that it would `Enable caching of object metadata and content to the given directory`, but if I'm not mistaken caching of object metadata is not related to `--cache` directory.

Issue #759, opened 2024-02-16
Clarify the expected behavior of the `--metadata-ttl` option set to zero


mhnap:
In an attempt to disable the metadata cache, I set the `--metadata-ttl=0`, referencing https://github.com/awslabs/mountpoint-s3/pull/616 as a potential guide. However, this resulted in a significant increase in `List` and `Head` requests, exceeding even the number observed with the cache completely disabled.

While I'm uncertain about the intended behavior of `--metadata-ttl=0`, I assume it should either disable the metadata cache (if feasible) or be prevented from being accepted as a CLI option.

jamesbornholt:
Yes, `--metadata-ttl=0` disables metadata caching entirely, which is almost certainly not a good thing, as filesystem performance with zero metadata caching is very bad. For Mountpoint, that means sending many more LIST/HEAD requests than necessary, as you're seeing.

When caching is disabled (i.e. no `--cache` or `--metadata-ttl` is set), the actual setting is roughly the same as `--metadata-ttl=0.1` (noting that you can't actually set that as a float), as this comment explains:

https://github.com/awslabs/mountpoint-s3/blob/022f915f76535e517175a8e11291eedaeec92932/mountpoint-s3/src/fs.rs#L303-L313

This `0.1` setting is also the reason our documentation has a [caveat around consistency](https://github.com/awslabs/mountpoint-s3/blob/main/doc/SEMANTICS.md#consistency-and-concurrency):

> However, Mountpoint may return stale metadata for an existing object within 1 second of the object being modified or deleted in your S3 bucket by another client. This occurs only if the object was accessed through Mountpoint immediately before being modified or deleted in your S3 bucket. The stale metadata will only be visible through metadata operations such as stat on individual files. Directory listings will never be stale and always reflect the current metadata. These cases do not apply to newly created objects, which are always immediately visible through Mountpoint. Stale metadata can be refreshed by either opening the file or listing its parent directory.

We probably should just forbid `--metadata-ttl=0`, or convert it to mean the same thing as the default value, as I can't imagine a setting where it would actually work well. Until then, I think `--metadata-ttl=1` is likely the effect you're looking for.

mhnap:
@jamesbornholt Thank you for such a detailed explanation. Now it's clear to me why I saw such behavior for the `--metadata-ttl=0` case. I just didn't realize that `--metadata-ttl=0` would result in different settings than disabled caching.

[...]

Issue #752, opened 2024-02-14
Disabling data cache isn't mentioned in the documentation


mhnap:
https://github.com/awslabs/mountpoint-s3/pull/616 introduced disabling of data cache by setting `--max_cache_size=0`, but this behavior is not mentioned in the [Caching configuration](https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#caching-configuration) section of the documentation. Would be great to include this information also there and maybe even add more details (is metadata cache stored in the same `--cache` dir, is metadata cache size can be configured, or it's limited to some value, etc).

jamesbornholt:
This is intentionally undocumented right now, I think, as we weren't sure how useful metadata caching would be on its own. If you have a use case for it we'd be super interested to hear more!

mhnap:
Aha, got it. My goal was to perform benchmarks with enabled cache for different `--max-cache-size` and `--metadata-ttl` values, and observe how they affect Get/List/Head request numbers and total test time. To isolate and understand better test results, I used to set `--max-cache-size` or `--metadata-ttl` to zero. However, I don't have any use case for metadata caching on its own other than benchmarks for now. Still, I think it would be great to see more details in the documentation if possible.

passaro:
Hi @mhnap. We just merged #855, which enables (and documents) metadata-only caching. Closing #752 and #768.

PR #855, opened 2024-04-17
Allow --metadata-ttl without --cache and set default with --cache to 60s

passaro:
## Description of change

Allow users to explicitly enable metadata caching, independently of object content caching, by setting the `--metadata-ttl` option. Before this change, users were required to also set `--cache <DIR>` in order to configure the metadata TTL. To enable just the metadata cache they had to resort to the workaround of restricting the cache size to zero (`--max-cache-size 0`).
We are also introducing the following special values for `--metadata-ttl`:

* `minimal`: describe the default setting (when `--cache` is not used) where Mountpoint caches metadata for up to 1s, refreshes it on `open`, and does not keep a negative cache.
* `indefinite`: new value recommended for users with static buckets, disables refreshing of metadata. (Implemented by setting the TTL to the maximum allowed value, ~100 years).

**When `--cache` is set, we are now also setting the default TTL for metadata to `60s` (`1min`) vs the previous `1s`.**

Finally, setting `--metadata-ttl 0` now resolves to `minimal` and logs a warning to notify users.

The caching options section looks like this:
```
Caching options:
      --cache <DIRECTORY>
          Enable caching of object content to the given directory and set metadata TTL to 60s
      --metadata-ttl <SECONDS|indefinite|minimal>
          Time-to-live (TTL) for cached metadata in seconds [default: minimal, or 60s if --cache is set]
      --max-cache-size <MiB>
          Maximum size of the cache directory in MiB [default: preserve 5% of available space]
```

Relevant issues: #752, #759, #768  

Two changes to existing behavior:
1. When setting `--cache` but not `--metadata-ttl`, the default TTL is `60s` vs the previous `1s`.
2. Setting `--metadata-ttl 0` resolves to `minimal` and logs a warning.

---

### DIFF
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -1,7 +1,11 @@
 ## Unreleased
 
+### New features
+* Metadata caching can now be configured independently of data caching. When passing the `--metadata-ttl <seconds>` argument without also specifying `--cache <directory>`, Mountpoint will cache file metadata in memory for up to the given TTL, but will not cache object data. The `--metadata-ttl` argument also accepts two special values: `minimal` to enable only the minimal necessary caching, and `indefinite` to cache indefinitely. These modes can help accelerate workloads that touch many files but do not need to cache object data for re-use (for example, listing a directory and then reading each file within it once). ([#855](https://github.com/awslabs/mountpoint-s3/pull/855))
+
 ### Breaking changes
-* When using the `--cache` flag, the default metadata time-to-live (TTL) is now set to 60 seconds instead of 1 second. The `--metadata-ttl` flag can still be set to further configure it, but can now also be used without the `--cache` flag in order to only enable caching of metadata. In addition to values in seconds, `--metadata-ttl` can now also be set to `minimal` (strict consistency) or `indefinite` (no updates). The `--metadata-ttl 0` setting is no longer supported, is now interpreted as `minimal`, and will be removed in a future release. ([#855](https://github.com/awslabs/mountpoint-s3/pull/855))
+* The `--metadata-ttl 0` setting is no longer supported and will be removed in a future release. The new `--metadata-ttl minimal` has a similar effect, but behaves better when latency for S3 requests is high. ([#855](https://github.com/awslabs/mountpoint-s3/pull/855))
+* When using the `--cache` flag, the default metadata TTL is now set to 60 seconds (`--metadata-ttl 60`) instead of 1 second. ([#855](https://github.com/awslabs/mountpoint-s3/pull/855))
 
 ### Other changes
 * The checksum algorithm to use for uploads to S3 can now be chosen with the `--upload-checksums <ALGORITHM>` command-line argument. The only supported values in this release are `crc32c` (the default, and the existing behavior) and `off`, which disables including checksums in uploads. The `off` value allows uploads to S3 implementations that do not support [additional checksums](https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/). This option defaults to `off` when the bucket name is an S3 on Outposts bucket access point (either an ARN or a bucket alias). ([#849](https://github.com/awslabs/mountpoint-s3/pull/849))
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -1,10 +1,10 @@
 ## Unreleased
 
 ### Breaking changes
-* No breaking changes.
+* When using the `--cache` flag, the default metadata time-to-live (TTL) is now set to 60 seconds instead of 1 second. The `--metadata-ttl` flag can still be set to further configure it, but can now also be used without the `--cache` flag in order to only enable caching of metadata. In addition to values in seconds, `--metadata-ttl` can now also be set to `minimal` (strict consistency) or `indefinite` (no updates). The `--metadata-ttl 0` setting is no longer supported, is now interpreted as `minimal`, and will be removed in a future release. ([#855](https://github.com/awslabs/mountpoint-s3/pull/855))
 
 ### Other changes
-* The checksum algorithm to use for uploads to S3 can now be chosen with the `--upload-checksums <ALGORITHM>` command-line argument. The only supported values in this release are `crc32c` (the default, and the existing behavior) and `off`, which disables including checksums in uploads. The `off` value allows uploads to S3 implementations that do not support [additional checksums](https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/). This option defaults to `off` when the bucket name is an S3 on Outposts bucket access point (either an ARN or a bucket alias). ([#849](https://github.com/awslabs/mountpoint-s3/pull/849)).
+* The checksum algorithm to use for uploads to S3 can now be chosen with the `--upload-checksums <ALGORITHM>` command-line argument. The only supported values in this release are `crc32c` (the default, and the existing behavior) and `off`, which disables including checksums in uploads. The `off` value allows uploads to S3 implementations that do not support [additional checksums](https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/). This option defaults to `off` when the bucket name is an S3 on Outposts bucket access point (either an ARN or a bucket alias). ([#849](https://github.com/awslabs/mountpoint-s3/pull/849))
 
 ## v1.6.0 (April 11, 2024)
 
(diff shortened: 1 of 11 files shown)


mountpoint-s3/tests/cli.rs
```diff
@@ -214,36 +215,31 @@ fn allow_other_conflict() -> Result<(), Box<dyn std::error::Error>> {
...
-    const INVALID_TTL_STRING: &str = "20000000000000000000";
     cmd.arg("test-bucket")
         .arg(dir.path())
-        .arg("--cache")
-        .arg(cache_dir.path())
         .arg("--metadata-ttl")
-        .arg(INVALID_TTL_STRING);
-    let error_message = "'--metadata-ttl <SECONDS>': number too large to fit in target type";
+        .arg(invalid_ttl);
+    let error_message = "'--metadata-ttl <SECONDS|indefinite|minimal>': TTL must be a valid number of seconds";
```
dannycjones, line 242: Probably just stick `", or"` on the end, since I was wondering why the error doesn't mention indefinite, etc..... when it does!

dannycjones (commented):
Looks good!

Before merging:
- Add changelog entry
- Update CONFIGURATION.md https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#caching-configuration

mountpoint-s3/src/cli.rs
```diff
@@ -237,21 +238,19 @@ pub struct CliArgs {
...
         value_name = "DIRECTORY",
     )]
     pub cache: Option<PathBuf>,
 
     #[clap(
         long,
-        help = "Time-to-live (TTL) for cached metadata in seconds [default: 1s]",
-        value_name = "SECONDS",
-        value_parser = parse_ttl_seconds,
+        help = "Time-to-live (TTL) for cached metadata in seconds [default: minimal, or 60s if --cache is set]",
```
jamesbornholt, line 249: nit to not make people think `60s` is a valid value to pass:
```suggestion
        help = "Time-to-live (TTL) for cached metadata in seconds [default: minimal, or 60 seconds if --cache is set]",
```

mountpoint-s3/src/cli.rs
```diff
@@ -639,15 +638,21 @@ where
...
+    let mut metadata_cache_ttl = args.metadata_ttl.unwrap_or_else(|| {
+        if args.cache.is_some() {
+            // When the data cache is enabled, use 1min as metadata-ttl.
+            TimeToLive::Custom(Duration::from_secs(60))
+        } else {
+            TimeToLive::Minimal
+        }
+    });
+    if matches!(metadata_cache_ttl, TimeToLive::Custom(Duration::ZERO)) {
+        tracing::warn!("'--metadata-ttl 0' is not supported, using 'minimal' instead");
```
jamesbornholt, line 650: We need to make clear that we're planning to remove this in a future release. So two things:
1. Wording needs to be more direct: something like `The '--metadata-ttl 0' setting is no longer supported, is now interpreted as 'minimal', and will be removed in a future release. Use '--metadata-ttl minimal' instead.`
2. We might want to send this message somewhere other than tracing -- like making sure it gets printed to stderr even when mounting in background mode.

mountpoint-s3/src/fs.rs
```diff
@@ -334,6 +335,26 @@ impl Default for CacheConfig {
...
+            TimeToLive::Custom(ttl) => Self {
+                serve_lookup_from_cache: true,
+                file_ttl: ttl,
+                dir_ttl: ttl,
+                ..Default::default()
+            },
+        }
+    }
+}
+
```
jamesbornholt, line 357: Even though API breakage isn't as much a concern in this crate, I probably wouldn't do this as a `From` impl, as we'll just end up having to undo it in the future if we ever add more cache configs.

mountpoint-s3/src/lib.rs
```diff
@@ -12,6 +12,7 @@ mod object;
 pub mod prefetch;
 pub mod prefix;
 mod sync;
+pub mod time_to_live;
```
jamesbornholt, line 15: nit: maybe put this under the `fs` module

mountpoint-s3/src/time_to_live.rs
```diff
@@ -0,0 +1,66 @@
+use std::{fmt::Display, num::ParseIntError, str::FromStr, time::Duration};
+
+use thiserror::Error;
+
+/// User-configurable time-to-live (TTL) for metadata caching.
+#[derive(Debug, Clone, Copy)]
+pub enum TimeToLive {
+    Minimal,
+    Indefinite,
+    Custom(Duration),
```
jamesbornholt, line 10: bikeshedding on naming: `Duration(Duration)`

mountpoint-s3/src/time_to_live.rs
```diff
@@ -0,0 +1,66 @@
...
+        TimeToLive::MAXIMUM_TTL_SECONDS,
+        TimeToLive::MAXIMUM_TTL_YEARS
+    )]
+    TooLarge,
+}
+
+impl TimeToLive {
+    const MINIMAL: &'static str = "minimal";
+    const INDEFINITE: &'static str = "indefinite";
+    const MAXIMUM_TTL_YEARS: u64 = 100;
```
jamesbornholt, line 28: Add a comment to document why we need a maximum that isn't just 2^64

jamesbornholt (commented):
Just some minor stuff

mountpoint-s3/src/cli.rs
passaro: 1. done
2. agree, but not trivial at first glance. I'll think about it a bit more.

mountpoint-s3/src/cli.rs
```diff
@@ -639,15 +636,26 @@ where
...
+        } else {
+            TimeToLive::Minimal
+        }
+    });
+    if matches!(metadata_cache_ttl, TimeToLive::Duration(Duration::ZERO)) {
+        const ZERO_TTL_WARNING: &str = "The '--metadata-ttl 0' setting is no longer supported, is now interpreted as 'minimal', and will be removed in a future release. Use '--metadata-ttl minimal' instead";
+        tracing::warn!("{}", ZERO_TTL_WARNING);
+        if !args.foreground {
+            // Ensure warning is visible even when not redirecting logs to stdout.
+            eprintln!("{}", ZERO_TTL_WARNING);
```
jamesbornholt, line 652: What do you think about adding some color here just to make it stand out a bit for interactive users? I did this before for Shuttle with the `owo_colors` crate: https://github.com/awslabs/shuttle/blob/623c3b5a309f8537990b26987fe3410b2c710f0e/src/sync/atomic/mod.rs#L88-L95

jamesbornholt (commented):
One minor suggestion, and then needs a rebase.

mountpoint-s3/CHANGELOG.md
```diff
@@ -1,10 +1,10 @@
 ## Unreleased
 
 ### Breaking changes
-* No breaking changes.
+* When using the `--cache` flag, the default metadata time-to-live (TTL) is now set to 60 seconds instead of 1 second. The `--metadata-ttl` flag can still be set to further configure it, but can now also be used without the `--cache` flag in order to only enable caching of metadata. In addition to values in seconds, `--metadata-ttl` can now also be set to `minimal` (strict consistency) or `indefinite` (no updates). The `--metadata-ttl 0` setting is no longer supported, is now interpreted as `minimal`, and will be removed in a future release. ([#855](https://github.com/awslabs/mountpoint-s3/pull/855))
```
jamesbornholt, line 4: Maybe split this up because only part of it is breaking

### New features
* Metadata caching can now be configured independently of data caching. When passing the `--metadata-ttl <seconds>` argument without also specifying `--cache <directory>`, Mountpoint will cache file metadata in memory for up to the given TTL, but will not cache object data. The `--metadata-ttl` argument also accepts two special values: `minimal` to enable only the minimal necessary caching, and `indefinite` to cache indefinitely. These modes can help accelerate workloads that touch many files but do not need to cache object data for re-use (for example, listing a directory and then reading each file within it once).

### Breaking changes
* The `--metadata-ttl 0` setting is no longer supported and will be removed in a future release. The new `--metadata-ttl minimal` has a similar effect, but behaves better when latency for S3 requests is high.

jamesbornholt (commented):
LGTM, just one changelog thing.

mountpoint-s3/CHANGELOG.md
passaro: I've also added a breaking change entry for the new default TTL with cache.

jamesbornholt (approved):

PR #860, opened 2024-04-23
Avoid creating a new rule engine for every endpoint resolution

monthonk:
## Description of change

Currently, we are creating a new rule engine for every bucket endpoint resolution. This operation is really expensive because we have to parse the JSON file for the rule engine and we're doing this whenever we create a new http request via [new_request_template](https://github.com/awslabs/mountpoint-s3/blob/268b672c81b3b8820af29051b599deaf04207b88/mountpoint-s3-client/src/s3_crt_client.rs#L345-L346).

With this change we will only create the rule engine once for an endpoint config which is reusable and it should improve performance for all http requests.

Flame graph of the latency before this change.

full_app_flamegraph

No

---

### DIFF
--- a/mountpoint-s3-client/src/endpoint_config.rs
+++ b/mountpoint-s3-client/src/endpoint_config.rs
@@ -55,6 +55,7 @@ pub struct EndpointConfig {
     use_dual_stack: bool,
     endpoint: Option<Uri>,
     addressing_style: AddressingStyle,
+    endpoint_rule_engine: RuleEngine,
 }
 
 impl EndpointConfig {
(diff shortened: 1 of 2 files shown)


arsh:
@monthonk can you attach the before-this-change flamegraphs to the description?

arsh (dismissed):

mountpoint-s3-client/src/endpoint_config.rs
```diff
@@ -67,6 +68,7 @@ impl EndpointConfig {
             use_dual_stack: false,
             endpoint: None,
             addressing_style: AddressingStyle::Automatic,
+            endpoint_rule_engine: RuleEngine::new(&Default::default()).unwrap(),
```
passaro, line 71: Have we considered making this static?

passaro (approved):
Looks good.

passaro (approved):

Issue #710, opened 2024-01-23
Implement statfs operation to report basic stats


lzw06061139:
### Mountpoint for Amazon S3 version

mount-s3 1.3.2（build from main branch）

### Describe the running environment

Running in a ubuntu server（Release：16.04， kernel：4.15.0-142-generic）

### Mountpoint options

```shell
mount-s3  --region us-east-1 --endpoint-url {my-endpoint} --debug --debug-crt -l ~/ {my-bucket} {my-mountpoint}
```

### What happened?

Successfully mounted {my-bucket} on {my-mountpoint}，I can see the mounted filesystem by running commnd:  `cat /proc/mounts` ，but running `df -h`  can't report the mounted filesystem(others can be reported)

monthonk:
Hi @lzw06061139, thanks for reporting the issue. This is expected behavior because Mountpoint does not report file system stats to the kernel, so `df -h` would filter it out. However, you should still be able to see the mount by running `df -a` command.

I willl change this to feature request to implement `statfs` operation (https://docs.rs/fuser/latest/fuser/trait.Filesystem.html#method.statfs) which should make `df -h` includes Mountpoint by default.

lzw06061139:
> Hi @lzw06061139, thanks for reporting the issue. This is expected behavior because Mountpoint does not report file system stats to the kernel, so `df -h` would filter it out. However, you should still be able to see the mount by running `df -a` command.
> 
> I willl change this to feature request to implement `statfs` operation to report basic stats (https://docs.rs/fuser/latest/fuser/trait.Filesystem.html#method.statfs) so `df -h` would include Mountpoint in its output by default.

 Thank you for your reply @monthonk . `df -a` can see the mount with zero disk space usage. By the way,when do you expect the new feature `statfs` will be implemented?

monthonk:
Sorry, we don't have any target date yet. But we will let you know in this thread once we have something to share.

dannycjones:
This was implemented in #1118 and released in [Mountpoint v1.12.0](https://github.com/awslabs/mountpoint-s3/releases/tag/mountpoint-s3-1.12.0)! 🚀

PR #871, opened 2024-05-08
Implement statfs with synthetic values

dannycjones:
## Description of change

Draft prototype change for now. May not be merged.

Some applications do not work as expected when a file system reports available space as zero. Some applications may check for available space, and fail if that's a small number (like zero). I believe even `df -h` will not report the file system in the list by default for some reason - and this may fix it (TBC).

Next steps:
- Are these the right fields to change? What use cases are we solving?
- Confirm that we do actually want to implement `statfs` by default, diverging from the values provided by Fuser.
- Converge on an agreed value for these fields.
  - u64 max may have some unexpected consequences - we have no idea. Half of u64 is probably large enough.
  - u32 max is too small - that would only be a bit over 4TiB using default block size of 1024.
  - In 2021, AWS shared that there are over 100 trillion objects in Amazon S3. Maybe we could take inspiration from there. https://aws.amazon.com/blogs/aws/amazon-s3s-15th-birthday-it-is-still-day-1-after-5475-days-100-trillion-objects/
- Merge.

Relevant issues: #710

Mountpoint will now report a non-zero number of available blocks, free blocks, and free files. Before, it reported zero for all three values.

---

### DIFF
--- a/mountpoint-s3/Cargo.toml
+++ b/mountpoint-s3/Cargo.toml
@@ -29,7 +29,7 @@ lazy_static = "1.4.0"
 libc = "0.2.126"
 linked-hash-map = "0.5.6"
 metrics = "0.22.1"
-nix = { version = "0.27.1", features = ["user"] }
+nix = { version = "0.27.1", features = ["user", "fs"] }
 regex = "1.7.1"
 serde = { version = "1.0.190", features = ["derive"] }
 serde_json = "1.0.95"
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -593,6 +593,43 @@ pub struct DirectoryEntry {
     lookup: LookedUp,
 }
 
+/// Reply to a `statfs` call
+#[derive(Debug)]
+pub struct StatFs {
+    /// Total number of blocks
+    pub total_blocks: u64,
+    /// Number of free blocks
+    pub free_blocks: u64,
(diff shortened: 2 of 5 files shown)


passaro:
How do we think this should show up in `df`? This is what I now get on AL2:
```
$ df -ha
Filesystem      Size  Used Avail Use% Mounted on
[...]
mountpoint-s3      0 -4.0Z  4.0Z    - /path/to/mnt
```

While before it was:
```
mountpoint-s3      0     0     0    - /path/to/mnt
```

Note that in both cases I needed the `--all` flag. Did you expect that to change?

dannycjones:
> How do we think this should show up in `df`? This is what I now get on AL2:
> 
> ```
> $ df -ha
> Filesystem      Size  Used Avail Use% Mounted on
> [...]
> mountpoint-s3      0 -4.0Z  4.0Z    - /path/to/mnt
> ```
> 
> While before it was:
> 
> ```
> mountpoint-s3      0     0     0    - /path/to/mnt
> ```
> 
> Note that in both cases I needed the `--all` flag. Did you expect that to change?

I was hoping it would change - based only on previous discussions. Looks like it doesn't.

What's also unexpected there is that we have a negative used value which I imagine we'd want to avoid.

dannycjones:
Closing this PR for now since I have no intention at the moment of working on it further. However, hopefully this is a good starting point if we do want to move forward with it.

PR #879, opened 2024-05-15
Minor fix update to INSTALL.md

thundergolfer:
## Description of change

Came across this discrepancy when building `main` from source.

No breaking change. Docs only.

---

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -194,7 +194,7 @@ You can build Mountpoint for Amazon S3 from source. However, binaries built in t
 4. Build Mountpoint for Amazon S3:
    ```
    cd mountpoint-s3
-   cargo build --release
+   cargo +nightly build --release
    ```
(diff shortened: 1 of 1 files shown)


doc/INSTALL.md
```diff
@@ -194,7 +194,7 @@ You can build Mountpoint for Amazon S3 from source. However, binaries built in t
 4. Build Mountpoint for Amazon S3:
    ```
    cd mountpoint-s3
-   cargo build --release
+   cargo +nightly build --release
```
passaro, line 197: There should be no need to use the nightly toolchain. We build mountpoint with stable (currently rustc 1.78.0).

passaro: Are you seeing any issues without it?

thundergolfer: Ahh maybe I was using an older stable which only had a feature in nightly. Perhaps this doc could then have a note/hint about not using nightly and upgrading to latest stable.

thundergolfer: Better yet, the project could use `rust-toolchain.toml`

Issue #861, opened 2024-04-24
Assuming role configured in AWS profile does not work outside of `aws` partition (China, US Gov Cloud, etc.)


csy97:
### Mountpoint for Amazon S3 version

mount-s3 1.6.0

### AWS Region

cn-northwest-1

### Describe the running environment

When I use the command AWS_DEFAULT_REGION=cn-northwest-1 AWS_PROFILE=bwmtest2 mount-s3 --region cn-northwest-1 --debug --log-directory . / testtmppp /mnt/ mount-s3 will fail when mounted

```
[root@VM-0-5-centos .aws]# AWS_DEFAULT_REGION=cn-northwest-1 AWS_PROFILE=bwmtest2 mount-s3 --region cn-northwest-1   --debug --log-directory ./   testtmppp /mnt/
Error: Failed to create S3 client

Caused by:
    0: initial ListObjectsV2 failed for bucket testtmppp in region cn-northwest-1
    1: Client error
    2: No signing credentials found
Error: Failed to create mount process
```

But the fact that I was able to get the IAM role correctly

```
AWS_DEFAULT_REGION=cn-northwest-1 AWS_PROFILE=bwmtest2 aws sts get-caller-identity
{
    "Account": "714736990101",
    "UserId": "AROA2M2NE66K5E6KBFALA:botocore-session-1713966126",
    "Arn": "arn:aws-cn:sts::714736990101:assumed-role/s3fullrole/botocore-session-1713966126"
}
```

### Mountpoint options

```shell
AWS_DEFAULT_REGION=cn-northwest-1 AWS_PROFILE=bwmtest2 mount-s3 --region cn-northwest-1   --debug --log-directory ./   testtmppp /mnt/
```

### What happened?

Note that this problem only occurs when I need to ASSUME, if I use ak/sk directly I don't have this problem!
This is my credentials file.

```
[default]
aws_access_key_id = ak
aws_secret_access_key = sk
[bwmtest2]
output=json
region=cn-northwest-1
role_arn=arn:aws-cn:iam::714736990101:role/s3fullrole
source_profile=default
```

I think this is a bug in China, the resources in China are segregated from global, when I refer to a source_profile mount-s3 in credentials it sends the request to global's sts endpoint instead of China's endpoint, thus causing the authentication to fail.

The steps to reproduce this are very simple in the China region, through the source_profile assume role can appear this

[...]

PR #880, opened 2024-05-16
Add assume role authorization tests

monthonk:
## Description of change

Currently, we don't have any tests that verify mountpoint can assume a role configured in the CLI config file. This change adds some tests to cover them.

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/861

No

---

### DIFF
--- a/mountpoint-s3-client/tests/auth.rs
+++ b/mountpoint-s3-client/tests/auth.rs
@@ -100,9 +100,9 @@ async fn test_static_provider() {
 /// test. So the [test_profile_provider] test below is forked into its own process, where it can set
 /// the environment variables it needs to point to an isolated CLI configuration file without
 /// affecting the rest of the real test runner.
-async fn test_profile_provider_async() {
+async fn test_profile_provider_static() {
     let sdk_client = get_test_sdk_client().await;
-    let (bucket, prefix) = get_test_bucket_and_prefix("test_profile_provider");
+    let (bucket, prefix) = get_test_bucket_and_prefix("test_profile_provider_static");
 
     let key = format!("{prefix}/hello");
     let body = b"hello world!";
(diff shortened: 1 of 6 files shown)


mountpoint-s3-client/tests/auth.rs
```diff
@@ -187,12 +187,72 @@ async fn test_profile_provider_async() {
...
-        runtime.block_on(test_profile_provider_async());
+        runtime.block_on(test_profile_provider_static());
+    }
+
+    #[test]
+    fn test_profile_provider_sts() {
+        // rusty_fork doesn't support async tests, so build an SDK-usable runtime manually
+        let runtime = tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap();
+        runtime.block_on(test_profile_provider_assume_role());
     }
```
dannycjones, line 256: Can we maintain the naming convention here with something like?:

```rust
#[test]
fn test_profile_provider_assume_role() {
  let runtime = todo!();
  runtime.block_on(test_profile_provider_assume_role_async());
}
```


same for all the `rusty_fork_test!` tests

mountpoint-s3/tests/common/s3.rs
```diff
@@ -34,6 +34,39 @@ pub fn get_test_region() -> String {
...
+
+    let test_region = get_test_region();
+    let mut rng = rand::thread_rng();
+    regions
+        .into_iter()
+        .filter(|region| region.to_string() != test_region)
+        .choose(&mut rng)
+        .unwrap()
+        .to_string()
+}
```
dannycjones, line 68: I'd rather we try to avoid randomness between tests. Instead, what if we just have something like...

```suggestion
// Get a region other than what configured in S3_REGION
pub fn get_non_test_region() -> String {
    match get_test_region() {
        "us-east-1" => String::new("us-west-2"),
        _ => String::new("us-east-1"),
    }
}
```

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -484,6 +486,160 @@ fn mount_with_sse(
...
+    let mut cmd = Command::cargo_bin("mount-s3")?;
+    let child = cmd
+        .arg(&bucket)
+        .arg(mount_point.path())
+        .arg(format!("--prefix={prefix}"))
+        .arg("--auto-unmount")
+        .env("AWS_CONFIG_FILE", config_file.path())
+        .env("AWS_PROFILE", profile_name)
+        .env_remove("AWS_REGION")
+        .env_remove("AWS_DEFAULT_REGION")
```
dannycjones, line 590: Can you also add a env var to disable IMDS, so this test would still work if CRT add it to their region chain?

```suggestion
        .env("AWS_CONFIG_FILE", config_file.path())
        .env("AWS_PROFILE", profile_name)
        .env("AWS_EC2_METADATA_DISABLED", "true")
        .env_remove("AWS_REGION")
        .env_remove("AWS_DEFAULT_REGION")
```

dannycjones (commented):
Just a few minor comments

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -705,3 +922,42 @@ fn unmount_and_check_log(mut process: Child, mount_path: &Path, expected_log_lin
...
+            )
+            .unwrap();
+            if let Some(session_token) = credentials.session_token() {
+                writeln!(credentials_file, "aws_session_token={session_token}").unwrap();
+            }
+            Ok(Some(credentials_file))
+        }
+        None => Ok(None),
+    }
+}
```
dannycjones, line 963: To keep things simple, I'd suggest we move to an approach where we always create the config file entry for the source profile (so its consistent between local and CI).

Would it be possible to also use the default chain like the `auth` tests. I know we'll need to block on the future / async fn which we didn't do in this module yet.

https://github.com/awslabs/mountpoint-s3/blob/8c3e18e80a2fa1ee53eaf5cb5c286d003609683c/mountpoint-s3-client/tests/auth.rs#L43-L51

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -484,6 +486,221 @@ fn mount_with_sse(
...
+    // Credentials are provided as environment variables when running on GitHub Actions,
+    // so we will have to build a credentials file from those variables if they exist, otherwise
+    // we are going to use default profile from the test instance.
+    let credentials_file = create_shared_credentials_file_if_possible(source_profile)?;
+
+    // First, make sure we can mount with the default profile
+    let mut cmd = Command::cargo_bin("mount-s3")?;
+    if let Some(file) = &credentials_file {
+        cmd.env("AWS_SHARED_CREDENTIALS_FILE", file.path());
+    }
```
dannycjones, line 653: Related to the other comment:

I'm hoping each test just becomes:
- Populate source profile
- Populate profile under test (which draws from source)
- Configure only `AWS_CONFIG_FILE`

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -484,6 +486,221 @@ fn mount_with_sse(
...
+    Ok(())
+}
+
+#[test]
+fn run_fail_when_assume_role_with_invalid_arn() -> Result<(), Box<dyn std::error::Error>> {
+    let (bucket, prefix) = get_test_bucket_and_prefix("run_fail_when_assume_role_with_invalid_arn");
+    let invalid_role = "arn:aws:iam::123456789123:role/invalid-role";
+    let region = get_test_region();
+    let mount_point = assert_fs::TempDir::new()?;
+    let profile_name = "fork_test";
```
dannycjones, line 639:
```suggestion
    let profile_with_bad_arn = "bad_profile";
```

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -484,6 +487,186 @@ fn mount_with_sse(
...
+    // verify mount status and mount entry
+    assert!(exit_status.success());
+    assert!(mount_exists("mountpoint-s3", mount_point.path().to_str().unwrap()));
+
+    test_read_files(&bucket, &prefix, &region, &mount_point.to_path_buf());
+
+    unmount(mount_point.path());
+
+    Ok(())
+}
```
dannycjones, line 608: We are still setting a region in the profile, right? I'm not sure this tests anything right now.

Do we instead want to make region optional when passing to `create_cli_config_file(...)`?

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -484,6 +487,186 @@ fn mount_with_sse(
...
+    let child = cmd
+        .arg(&bucket)
+        .arg(mount_point.path())
+        .arg(format!("--prefix={prefix}"))
+        .arg("--auto-unmount")
+        .env("AWS_CONFIG_FILE", config_file.path())
+        .env("AWS_PROFILE", profile_name)
+        .env_remove("AWS_ACCESS_KEY_ID")
+        .env_remove("AWS_SECRET_ACCESS_KEY")
+        .env_remove("AWS_SESSION_TOKEN")
```
dannycjones, line 551: Should we be removing `AWS_REGION`, `AWS_DEFAULT_REGION` here too?

mountpoint-s3-client/tests/auth.rs
```diff
@@ -187,12 +187,96 @@ async fn test_profile_provider_async() {
...
+        .unwrap();
+        if let Some(session_token) = credentials.session_token() {
+            writeln!(credentials_file, "aws_session_token={session_token}").unwrap();
+        }
+        std::env::remove_var("AWS_ACCESS_KEY_ID");
+        std::env::remove_var("AWS_SECRET_ACCESS_KEY");
+        std::env::remove_var("AWS_SESSION_TOKEN");
+        std::env::set_var("AWS_SHARED_CREDENTIALS_FILE", credentials_file.path().as_os_str());
+        credentials_file
+    });
```
dannycjones, line 236: Can we adopt the same logic as the fork tests, using the default creds provider to always create a config file?

mountpoint-s3/tests/fuse_tests/fork_test.rs
monthonk: yeah, this was a miss from refactoring. It should be optional.

dannycjones (approved):
LGTM!

PR #882, opened 2024-05-17
Provide `ErrorMetadata` to `crate::fuse` layer with errors from the client

vladem:
> [!NOTE]
> The majority of the changes (`646` lines) are for the updated `Cargo.lock` caused by the addition of `httpmock` dev dependency 

## Description of change
1. `ProvideErrorMetadata` trait aims to provide additional information with errors. After this change, code in `fuse.rs` can access `error_code`, `http_code`, `s3_error_code` and [some other fields](https://github.com/awslabs/mountpoint-s3/blob/d590b7da62f145589ea66af9d7fdc5e21f2f1502/mountpoint-s3-client/src/error_metadata.rs#L2) for errors occurring on `S3FuseFilesystem::lookup()` (for other fuse operations we currently return **no** metadata). **Currently, we don't use this information in any way apart from the tests.**
3. There [are tests added](https://github.com/awslabs/mountpoint-s3/blob/d590b7da62f145589ea66af9d7fdc5e21f2f1502/mountpoint-s3/tests/fs.rs#L1525) for common `lookup` errors: forbidden, throttled, not found, unhandled error from s3;

Relevant issues: None

 ## What is **not** in this change
1. Emission of serialised errors in a separate file (for reference see the expected approach here: https://github.com/awslabs/mountpoint-s3/pull/894);
2. Additional data with errors on **fuse operations other then lookup()** (e.g. throttling on `read()` will have no `error_code` or `http_code`);
3. S3-response data for `HeadObjectError::NotFound` and `ListObjectsError::NoSuchBucket`;
4. Returning mount errors in a way suitable for logging (e.g. forbidden on the initial ListObjectsV2), which will likely require replacing [anyhow::Result](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/cli.rs#L637) with a type implementing `ProvideErrorMetadata` trait.

This does not change any behaviour visible to customer.

---


mountpoint-s3/src/fs/error.rs
```diff
@@ -40,6 +41,7 @@ macro_rules! err {
             message: format!($message, $($args)*),
             source: $source,
             level: $level,
+            metadata: None,
```
passaro, line 44: Eventually, we'll want to propagate metadata here as well, right?

mountpoint-s3/src/inode.rs
```diff
@@ -531,7 +532,11 @@ impl Superblock {
                             error=?e,
                             "DeleteObject failed for unlink",
                         );
-                        Err(InodeError::ClientError(anyhow!(e).context("DeleteObject failed")))?;
+                        let metadata = e.meta().clone();
+                        Err(InodeError::ClientError {
+                            source: anyhow!(e).context("DeleteObject failed"),
+                            metadata,
+                        })?;
```
passaro, line 539: This calls for a "constructor" on `InodeError` that takes a `ProvideErrorMetadata` source and a context.

passaro (commented):
Looks good overall. Just a few suggestions.

I was unsure about returning the metadata by reference, by I see this the [approach in the Rust SDK](https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/aws-smithy-types/src/error/metadata.rs), so it may be worth to adopt as well.

Before finalizing, can you try and implement the propagation for another type of error, so we can get a better picture?

mountpoint-s3-client/src/error_metadata.rs
```diff
@@ -0,0 +1,15 @@
...
+
+pub trait ProvideErrorMetadata {
+    fn meta(&self) -> &ErrorMetadata;
+}
+
+/// Empty error metadata
+pub const EMPTY_ERROR_METADATA: ErrorMetadata = ErrorMetadata {
+    http_code: None,
+    s3_error_code: None,
+};
```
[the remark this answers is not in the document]
vladem: just in case, it can not be (which would be a bit simpler):
```
pub const EMPTY: Self = Default::default();
```
seemingly, `default()` is non-const:
```
cannot call non-const fn `<ErrorMetadata as std::default::Default>::default` in constants
```

passaro (commented):
I've done a first pass and added a few minor suggestions. But the main question is about the `ErrorMetadata` and whether it belongs to the client layer as it is.

mountpoint-s3-client/src/error_metadata.rs
```diff
@@ -0,0 +1,29 @@
+#[derive(Default, Debug, Clone, PartialEq)]
+pub struct ErrorMetadata {
+    pub http_code: Option<i32>,
+    pub s3_error_code: Option<String>,
+    pub s3_error_message: Option<String>,
+    pub error_code: Option<String>,
```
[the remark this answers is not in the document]
vladem: Valid observation, those fields are never set in the client.

One approach we could take is:
```
// mountpoint-s3-client
pub struct ClientErrorMetadata {
    pub http_code: Option<i32>,
    pub s3_error_code: Option<String>,
    pub s3_error_message: Option<String>,
}

// mountpoint-s3
pub struct ErrorMetadata {
    pub client_error_meta: ClientErrorMetadata,
    pub error_code: Option<String>,
    pub s3_bucket_name: Option<String>,
    pub s3_object_key: Option<String>,
}
``` 

I think it won't even require a second trait, since only client errors are generic parameters, `InodeError` / `fs::Error` are always used when the type is known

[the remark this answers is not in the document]
vladem: just as an alternative, this is how rust SKD achieves extendibility of the meta struct (all of their error's are "client" errors though):

```rust
pub struct ErrorMetadata {
    code: Option<String>,
    message: Option<String>,
    extras: Option<HashMap<&'static str, String>>,
}

const EXTENDED_REQUEST_ID: &str = "s3_extended_request_id";

impl RequestIdExt for ErrorMetadata {
    fn extended_request_id(&self) -> Option<&str> {
        self.extra(EXTENDED_REQUEST_ID)
    }
}
```

We could do the same and leave `ErrorMetadata::extras` empty in client, than fill them just before converting the error to `InodeError` and implement a trait/accessor method for each field:

```rust
impl InodeError {
    fn error_code(&self) -> Option<&str> {}
    fn s3_bucket_name(&self) -> Option<&str> {}
    fn s3_object_key(&self) -> Option<&str> {}
}
```

This looks to be boilerplate though as accessors should be implemented for each fs error (at least for `PrefetchReadError` / `UploadWriteError`).

mountpoint-s3/tests/common/mod.rs
```diff
@@ -101,6 +105,49 @@ pub fn tokio_block_on<F: Future>(future: F) -> F::Output {
...
+    let auth_config = CredentialsProviderStaticOptions {
+        access_key_id: credentials.access_key_id(),
+        secret_access_key: credentials.secret_access_key(),
+        session_token: credentials.session_token(),
+    };
+    let credentials_provider = CredentialsProvider::new_static(&Allocator::default(), auth_config).unwrap();
+    S3ClientAuthConfig::Provider(credentials_provider)
+}
+
+// This helper will be also used in `fuse_tests/fork_test.rs` to check that fuse operation errors are actually logged.
```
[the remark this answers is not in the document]
vladem: to this https://github.com/awslabs/mountpoint-s3/pull/894/files#diff-b6f1caa405e63c89fd632c2d50612b88726aa868accde4c8c860077ed054403aR860

but this comment for now

mountpoint-s3/tests/fs.rs
```diff
@@ -1465,6 +1480,242 @@ async fn test_readdir_rewind_with_local_files_only() {
...
+    </ListBucketResult>
+    "#;
+    template.replace("__PREFIX__", prefix).replace("__BUCKET__", bucket)
+}
+
+#[test_case(true, false; "head object")]
+#[test_case(false, true; "list object")]
+#[tokio::test]
+async fn test_lookup_throttled_mock(head_object_throttled: bool, list_object_throttled: bool) {
+    let _ = EnvVarGuard::new("AWS_MAX_ATTEMPTS", "2"); // retry S3 request 2 times instead of 10
```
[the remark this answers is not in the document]
vladem: just noticed an option to set this via client config

mountpoint-s3/src/inode.rs
```diff
@@ -1635,6 +1647,42 @@ pub enum InodeError {
     },
 }
 
+impl InodeError {
+    /// Constructs InodeError::ClientError enriching metadata with error_code, bucket and key.
```
passaro, line 1651: Is `error_code` going to be anything other than `MOUNTPOINT_ERROR_CLIENT`?

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -976,8 +1007,15 @@ fn try_parse_generic_error(request_result: &MetaRequestResult) -> Option<S3Reque
             let message = error_elem
                 .get_child("Message")
                 .and_then(|e| e.get_text())
-                .unwrap_or(error_code_str);
-            Some(S3RequestError::Forbidden(message.into_owned()))
+                .unwrap_or(error_code_str.clone());
+            Some(S3RequestError::Forbidden(
+                message.to_string(),
+                Box::new(ClientErrorMetadata {
```
passaro, line 1013: One aspect of this change that does not convince me completely is that most (all?) metadata information seems to be "intrinsic" to the type of error rather than additional. This here is the only case where we add the code/message, but it is largely duplicated.
Are there other cases to be implemented in the future that drive your approach?
Otherwise, do we really need to add the (boxed) metadata struct, or could we build it on the fly based on the enum case (giving up on the return reference)?

mountpoint-s3/src/inode.rs
vladem: I can not think of a case for that now, will hardcode the code in the function body

mountpoint-s3-client/src/error_metadata.rs
```diff
@@ -0,0 +1,25 @@
+/// Additional data fetched from S3 response, which caused an error
+#[derive(Default, Debug, Clone, PartialEq)]
+pub struct ClientErrorMetadata {
+    /// http code of the response, e.g. 403
+    pub http_code: Option<i32>,
+    /// error code from the response xml body, e.g. "AccessDenied"
+    pub s3_error_code: Option<String>,
```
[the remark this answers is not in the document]
vladem: makes sense, error data does not have to come from s3, I can imagine that CRT calls sts behind the hood:
https://docs.aws.amazon.com/STS/latest/APIReference/CommonErrors.html

i'll keep `error_message` though, as `message` outside of the "error context" does not make a lot of sense

mountpoint-s3-client/src/s3_crt_client.rs
vladem: in the [last commit](https://github.com/awslabs/mountpoint-s3/pull/882/commits/2ec7627756199a0a1426185fa497c013f51cb24c) I've changed this to be constructed in `ProvideErrorMetadata::meta()` for `ResponseError` and `Throttled`, which required changing signature to `fn meta(&self) -> ClientErrorMetadata` (no ref in returned value)

I like the fact that this logic is grouped in `ProvideErrorMetadata` trait implementation for `S3RequestError` and that now we can use `Default::default()` instead of `&ClientErrorMetadata::EMPTY`

ps. we make clients of `mountpoint-s3-client` to always copy metadata even if a reference could've suffice, but it does not change current usages and I think that is a reasonable trade-off to make to have a more clear code

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -1064,9 +1068,15 @@ fn try_parse_generic_error(request_result: &MetaRequestResult) -> Option<S3Reque
         // redirect
         400 => try_parse_forbidden(request_result).or_else(|| try_parse_redirect(request_result)),
         403 => try_parse_forbidden(request_result),
-        0 => try_parse_throttled(request_result).or_else(|| {
-            try_parse_canceled_request(request_result).or_else(|| try_parse_no_credentials(request_result))
-        }),
+        0 => {
+            if let Some(err) = try_parse_throttled(request_result) {
```
passaro, line 1072: nit: this is fine, but inconsistent with the `400` case. I'd leave it as a chain of `.or_else`.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -877,9 +878,10 @@ pub enum S3RequestError {
     #[error("Wrong region (expecting {0})")]
     IncorrectRegion(String),
 
-    /// Forbidden
+    /// Forbidden (metadata is boxed to avoid allocating unnecessary space in case of Result::Ok,
+    /// see "result_large_err" clippy warning for details)
     #[error("Forbidden: {0}")]
-    Forbidden(String),
+    Forbidden(String, Box<ClientErrorMetadata>),
```
passaro, line 884: minor: I'd just simplify this case as well and add optional code and status here

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -896,6 +902,42 @@ impl S3RequestError {
...
+                    None
+                };
+                ClientErrorMetadata {
+                    http_code,
+                    ..Default::default()
+                }
+            }
+            Self::Forbidden(_, metadata) => (**metadata).clone(),
+            Self::Throttled => {
+                // CRT does not set S3 response data for error codes other then AWS_ERROR_S3_INVALID_RESPONSE_STATUS
```
passaro, line 921: nit: not sure this comment is relevant anymore. Maybe some of it (the first two lines?) could make more sense in `try_parse_throttled`.

mountpoint-s3/src/inode.rs
```diff
@@ -1635,6 +1641,42 @@ pub enum InodeError {
...
+    /// Detailed information about an error is gathered in different frames of the call stack.
+    /// To make it manageable the idea is to enrich metadata with error_code, bucket and key
+    /// on the construction of mountpoint crate's errors, i.e. InodeError, PrefetchReadError
+    /// and UploadPutError.
+    fn client_error<E>(err: E, context: Option<&'static str>, bucket: &str, key: &str) -> Self
+    where
+        E: ProvideErrorMetadata + std::error::Error + Send + Sync + 'static,
+    {
+        let metadata = ErrorMetadata {
+            client_error_meta: err.meta().clone(),
```
passaro, line 1656: No need to clone, right?

mountpoint-s3/src/inode/readdir.rs
```diff
@@ -353,7 +353,7 @@ impl RemoteIter {
                     self.full_path.as_str(),
                 )
                 .await
-                .map_err(|e| InodeError::ClientError(anyhow::Error::new(e)))?;
+                .map_err(|e| InodeError::client_error(e, None, &self.bucket, &self.full_path))?;
```
passaro, line 356: tiny nit: should we just add a context string here as well? And simply make it non-optional?

passaro (commented):
Looking good. Just a few minor suggestions.

mountpoint-s3-client/src/s3_crt_client.rs
vladem: Hm, the problem is that for HeadObject we [provide](https://github.com/awslabs/mountpoint-s3/blob/2ec7627756199a0a1426185fa497c013f51cb24c/mountpoint-s3-client/src/s3_crt_client.rs#L1006) "\<no message\>" as the first tuple member, which is not the actual `error_code` from s3 response.

Reporting this in the event log may be misleading, I'd suggest that we keep it as it is now. Another option is to make it `Forbidden(Option<String>, Option<String>, Option<i32>),` and change our existing logging (remove "\<no message\>").

mountpoint-s3-client/src/s3_crt_client.rs
vladem: Removed it for now. Not sure if this comment explains anything that happens in `try_parse_throttled`. I can add a comment to the match block, where we invoke parsers:
```rust
403 => try_parse_forbidden(request_result),
// CRT does not set S3 response status for error codes other then AWS_ERROR_S3_INVALID_RESPONSE_STATUS
0 => {
```

mountpoint-s3-client/src/s3_crt_client.rs
vladem: do you mind adding a suggestion of how you'd expect it to look like? :)

(I though the current version is what you've mentioned previously by "flatten": https://github.com/awslabs/mountpoint-s3/pull/882#discussion_r1633067922)

vladem: will something like this be better?
```rust
        0 => try_parse_throttled(request_result)
            .or_else(|| try_parse_canceled_request(request_result))
            .or_else(|| Some(try_parse_no_credentials_or_generic(request_result))),
```

vladem: will keep this variant in new commit, looks neat

passaro: Yes, I meant the latter variant. "flatten" compared to nesting the second `or_else` inside the first. Sorry I wasn't clear :)

mountpoint-s3-client/src/s3_crt_client.rs
passaro: I see. I'd be fine with the all `Option` variant, but happy either way.

vladem: ok, all `Option` lgtm too, the only side effect of this change is logging, Forbidden on HeadObject will now be logged slightly differently:

> 2024-06-18T14:04:07.100018Z  WARN lookup{req=60 ino=132 name="list.txt"}: mountpoint_s3::fuse: lookup failed: inode error: error from ObjectClient: HeadObject failed: Client error: Forbidden: **None**

before it was:
> 2024-06-03T15:57:04.665373Z  WARN lookup{req=50 ino=5 name="list.txt"}: mountpoint_s3::fuse: lookup failed: inode error: error from ObjectClient: HeadObject failed: Client error: Forbidden: **\<no message\>**

vladem: logging in the case of an existing message will become slightly more verbose

> 2024-06-18T14:17:47.133309Z  WARN read{req=12 ino=2 fh=1 offset=0 size=4096 name="encrypted_with_kms"}: mountpoint_s3::fuse: read failed: get request failed: get object request failed: Client error: Forbidden: **Some("**User: <arn> is not authorized to perform: kms:Decrypt on resource: <arn> because no session policy allows the kms:Decrypt action**")**

the issue is that `Option<String>` does not implement `Display`, so to avoid having `Some` bit in logs, we can do something like this:
```rust
#[derive(Default, Debug)]
pub struct S3RequestErrorCode(Option<String>);

impl Display for S3RequestErrorCode {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        if let Some(error_code) = &self.0 {
            write!(f, "{}", error_code)
        } else {
            write!(f, "<no message>")
        }
    }
}
...
    /// Forbidden. Contains response's error_code, error_message and an http_status.
    #[error("Forbidden: {0}")]
    Forbidden(S3RequestErrorCode, Option<String>, Option<i32>),
...
```

But I don't like this overcomplication. Lets keep `Forbidden(String, ClientErrorMetadata)` (`Box` is not necessary anymore).

passaro (approved):
LGTM

PR #884, opened 2024-05-22
Improve cancellation test for PutObject write

passaro:
## Description of change

Remove reliance of timing when testing PutObject write cancellation. 

No

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client/put_object.rs
+++ b/mountpoint-s3-client/src/s3_crt_client/put_object.rs
@@ -252,3 +252,10 @@ impl PutObjectRequest for S3PutObjectRequest {
         })
     }
 }
+
+impl S3PutObjectRequest {
+    /// The number of bytes written to this request so far.
+    pub fn bytes_written(&self) -> u64 {
+        self.total_bytes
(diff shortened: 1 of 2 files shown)


mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -252,3 +252,10 @@ impl PutObjectRequest for S3PutObjectRequest {
         })
     }
 }
+
+impl S3PutObjectRequest {
+    /// The number of bytes written to this request so far.
+    pub fn bytes_written(&self) -> u64 {
```
monthonk, line 258: Do we plan to expose this function at `PutObjectRequest` interface so that other variants of `PutObject` can report this too? If that's the case then we probably want to add some TODO.

monthonk (approved):
LGTM!

mountpoint-s3-client/src/s3_crt_client/put_object.rs
passaro: Added TODO

monthonk (approved):

PR #888, opened 2024-05-24
Small refactors of inode code

jamesbornholt:
## Description of change

This is three commits, each a small refactoring to better separate `inode.rs` and `fs.rs`, which I did as part of a larger refactoring coming later:
1. Move the process ID out of `inode.rs` -- the WriteHandle was just storing the PID and never using it itself, since the check happens in `fs.rs`, so the PID can live on the `fs.rs` handle instead
2. Clean up the lifecycle of `inode.rs` WriteHandles by making them immediately call `start_writing`, and add a symmetric ReadHandle with the same lifecycle and purpose, in place of `start_reading` and `finish_reading`.
3. Move `inc_file_size` off the Inode and onto the WriteHandle since it only makes sense to use while writing.

No.

---


passaro (approved):
Looks good!

PR #896, opened 2024-05-31
Save associated binary for core dump

monthonk:
## Description of change

To be able to analyze the core dump we also need the binary it was generated from. This adds a new script for uploading the binary associated with the core dump when tests are failing in the CI.

Tested here https://github.com/monthonk/mountpoint-s3/actions/runs/9318610444/job/25651288788#step:8:324, verified that the core dump and the binary are both uploaded to the test bucket.

No

---

### DIFF
--- a/.github/actions/scripts/save-coredump.sh
+++ b/.github/actions/scripts/save-coredump.sh
@@ -3,7 +3,7 @@
 bucket_name=$S3_BUCKET_NAME
 bucket_prefix=$S3_BUCKET_TEST_PREFIX
 coredump_path=/var/lib/systemd/coredump/
-coredump_pattern=core.mountpoint_s3*
+coredump_pattern=core.*
 
 # upload core dump files to S3
 aws s3 cp ${coredump_path} s3://${bucket_name}/${bucket_prefix}coredump/ --recursive --exclude "*" --include "${coredump_pattern}"
--- a/.github/actions/scripts/save-coredump.sh
+++ b/.github/actions/scripts/save-coredump.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+bucket_name=$S3_BUCKET_NAME
+bucket_prefix=$S3_BUCKET_TEST_PREFIX
+coredump_path=/var/lib/systemd/coredump/
+coredump_pattern=core.mountpoint_s3*
+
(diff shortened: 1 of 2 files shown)


.github/actions/scripts/save-coredump.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+bucket_name=$S3_BUCKET_NAME
+bucket_prefix=$S3_BUCKET_TEST_PREFIX
```
vladem, line 4: it'll be possible to find that from the test logs, right?

.github/actions/scripts/save-coredump.sh
```diff
@@ -0,0 +1,21 @@
...
+coredump_path=/var/lib/systemd/coredump/
+coredump_pattern=core.mountpoint_s3*
```
vladem, line 6: what is the source for this `mountpoint_s3` suffix? it is not the binary name, right? (as I can see from `cargo test --features=s3_tests --no-run` some of our binaries don't include this suffix, e.g. `target/debug/deps/integration_tests-5661f8ddaa0dd083`, will dumps for those be uploaded?)

arsh (dismissed):
LGTM as a way to quickly enable it and hopefully gather core dumps. We can revisit if we want a more sophisticated thing long-term.

.github/actions/scripts/save-coredump.sh
monthonk: yes, you can see the prefix and file name from the logs.

vladem (approved):

dannycjones (commented):
~~We already merged, but for future changes of a similar nature~~

We already have the changes I was suggesting, my bad!

.github/actions/scripts/save-coredump.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# upload core dump files to S3
+aws s3 cp ${coredump_path} s3://${bucket_name}/${bucket_prefix}coredump/ --recursive --exclude "*" --include "${coredump_pattern}"
```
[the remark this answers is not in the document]
dannycjones: I see we already have that with the `bucket_prefix`! Nice.

PR #900, opened 2024-06-04
Initialize the CRT eagerly in tests

jamesbornholt:
## Description of change

We think the lazy initialization might be the cause of some of our issues, because it happens on an ephemeral thread. Let's try initializing it at load time.

No.

---


jamesbornholt:
The DCO app is currently broken (https://github.com/dcoapp/app/issues/211), so I'm overriding the merge requirements for that.

Issue #906, opened 2024-06-11
Specifying KMS Key ID (not ARN) or KMS Alias for upload causes Mountpoint to crash


monthonk:
### Mountpoint for Amazon S3 version

mount-s3 1.7.0

### AWS Region

ap-southeast-1

### Describe the running environment

Running in EC2 on Amazon Linux 2023 using instance profile credentials against an S3 bucket in the same account.

### Mountpoint options

```shell
mount-s3 --sse aws:kms --sse-kms-key-id '<kms-key-id>' doc-example-bucket mnt
```

### What happened?

When a user explicitly specifies a KMS key id for their uploads at mount time Mountpoint will verify that the returned key in the response matches with the key provided by user. https://github.com/awslabs/mountpoint-s3/blob/650269904f617f96879f7ee7d93f8bca4e5096e5/mountpoint-s3/src/fs.rs#L493-L498

As a result, Mountpoint is going to kill itself when the keys mismatch because it thinks that the SSE settings were corrupted in transit or on S3 side. But the problem is that S3 always return KMS arn in the response while the key that user provided could be a key arn, key id, or key alias. So, it will be crashing even if the provided value from the user points to the same KMS key.

### Relevant log output

```shell
2024-06-11T14:51:28.479165Z DEBUG fuser::request: FUSE( 90) ino 0x0000000000000001 LOOKUP name "new_file.txt"
2024-06-11T14:51:28.479390Z DEBUG lookup{req=90 ino=1 name="new_file.txt"}:head_object{id=24 bucket="doc-example-bucket" key="new_file.txt"}: mountpoint_s3_client::s3_crt_client::head_object: new request
2024-06-11T14:51:28.479623Z DEBUG lookup{req=90 ino=1 name="new_file.txt"}:list_objects{id=25 bucket="doc-example-bucket" continued=false delimiter="/" max_keys="1" prefix="new_file.txt/"}: mountpoint_s3_client::s3_crt_client::list_objects: new request
[...]

PR #908, opened 2024-06-12
Restrict `--sse-kms-key-id` to Key ARN only

vladem:
## Description of change

Mountpoint currently crashes on runtime if `--sse-kms-key-id` is in one of [other](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#key-id) forms. It is desirable to know about an error on mount time.

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/906

## Help updates
```
$ mount-s3 --help
<...>
      --sse-kms-key-id <AWS_KMS_KEY_ARN>
          AWS Key Management Service (KMS) key ARN to use with KMS server-side encryption when uploading new objects. Key ID, Alias and Alias ARN are all not supported.
<...>
```

This is a breaking change for workloads which pass `--sse-kms-key-id` in any form that is not ARN and do not perform writes.

---


mountpoint-s3/src/cli.rs
```diff
@@ -1004,4 +1016,21 @@ mod tests {
             parsed.expect_err("invalid bucket name");
         }
     }
+
+    // https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#key-id
+    #[test_case("arn:aws:kms:eu-west-1:151381207180:key/dabe1478-fe48-47ca-b6f8-ca044b643a82", true; "KMS Key ARN")]
+    #[test_case("arn:aws:kms:us-west-2:111122223333:key/mrk-1234abcd12ab34cd56ef1234567890ab", true; "Multi-region KMS Key ARN")]
```
vladem, line 1022: We allow [multi-region key arns](https://docs.aws.amazon.com/kms/latest/developerguide/multi-region-keys-overview.html), but I decided not to add a test for it at this point. 

I've verified locally that a write succeeds for a multi-region key replicated from `us-east-1` to `eu-west-1`.

vladem:
```
$ aws s3api head-object --bucket * --key aaaaaaa2.txt | grep SSEKMSKeyId
    "SSEKMSKeyId": "arn:aws:kms:eu-west-1:*:key/mrk-*",
```

monthonk (approved):
LGTM

PR #917, opened 2024-06-17
Fix an issue where mountpoint-s3-client could interpret a HTTP 206 Partial success response as an error

passaro:
## Description of change

We are removing a workaround in mountpoint-s3-client that reduced the number of requests to S3 and is no longer needed. When introduced in #285, the workaround used a default CRT meta-request instead of an auto-ranged-get for small requests, which avoided a redundant HeadObject request that the CRT performed on every auto-ranged-get. Since then, the CRT has been updated to avoid the extra requests when a range is specified, so we can always use auto-ranged-get.

No.

---


fuatbasik (approved):
LGTM!

Issue #721, opened 2024-01-30
Improvements to error reporting (RFC)


dannycjones:
Opening this issue to create a space to share feedback or comments - an informal Request For Comment (RFC).

- How could error reporting in Mountpoint be improved?
- What kind of requirements do customers have for how to learn from or monitor errors?
- What kind of systems need to process these errors, if any?

PR #919, opened 2024-06-20
Logging errors in a structured format

vladem:
## Description of change

Adds `--event-log-directory` CLI flag. If specified all of the logs originating from `mountpoint_s3::logging::event_log` will be written as `\n`-separated json's in this file. We do this logging in `fuse_error!` and `fuse_unsupported!` macros for failed fuse operations.  

Note that the application log will also contain an additional log line per failed fuse operation with TRACE level **even if the feature flag / cli option are not provided**.

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/721

If compiled without `event_log` feature flag there should be no changes to existing behaviour.

---

### DIFF
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -248,11 +248,20 @@ pub struct CliArgs {
     #[clap(long, help = "Enable debug logging for AWS Common Runtime", help_heading = LOGGING_OPTIONS_HEADER)]
     pub debug_crt: bool,
 
+    #[cfg(feature = "event_log")]
+    #[clap(
+        long,
+        help = "Write structured event log files to a directory (can be the same as --log-directory)",
+        help_heading = LOGGING_OPTIONS_HEADER,
+        value_name = "DIRECTORY",
+    )]
+    pub event_log_directory: Option<PathBuf>,
+
     #[clap(
         long,
         help = "Disable all logging. You will still see stdout messages.",
         help_heading = LOGGING_OPTIONS_HEADER,
-        conflicts_with_all(["log_directory", "debug", "debug_crt", "log_metrics"])
+        conflicts_with_all(NO_LOG_CONFLICTS_WITH),
     )]
     pub no_log: bool,
 
@@ -325,6 +334,17 @@ pub struct CliArgs {
     pub bind: Option<Vec<String>>,
 }
 
+#[cfg(feature = "event_log")]
+const NO_LOG_CONFLICTS_WITH: [&str; 5] = [
+    "log_directory",
+    "debug",
+    "debug_crt",
+    "log_metrics",
+    "event_log_directory",
+];
+#[cfg(not(feature = "event_log"))]
+const NO_LOG_CONFLICTS_WITH: [&str; 4] = ["log_directory", "debug", "debug_crt", "log_metrics"];
+
 #[derive(Debug, Clone)]
 pub enum BucketType {
     GeneralPurpose,
@@ -385,7 +405,7 @@ impl CliArgs {
         self.prefix.as_ref().cloned().unwrap_or_default()
     }
 
-    fn logging_config(&self) -> LoggingConfig {
+    fn logging_config(&self, _is_fork_parent: bool) -> LoggingConfig {
         let default_filter = if self.no_log {
             String::from("off")
         } else {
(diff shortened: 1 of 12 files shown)


mountpoint-s3/src/cli.rs
```diff
@@ -227,11 +227,20 @@ pub struct CliArgs {
     #[clap(long, help = "Enable debug logging for AWS Common Runtime", help_heading = LOGGING_OPTIONS_HEADER)]
     pub debug_crt: bool,
 
+    #[cfg(feature = "event_log")]
+    #[clap(
+        long,
+        help = "Write structured event log for failed fuse operations to a directory (allowed to be the same as --log-directory). Note that if `MOUNTPOINT_LOG=off` no events will be emitted.",
```
passaro, line 233: Help text could be simpler. E.g.:
```suggestion
        help = "Write structured event log files to a directory (can be the same as --log-directory)",
```
I'd leave the `MOUNTPOINT_LOG` mention for the docs.

mountpoint-s3/src/logging/event_log.rs
```diff
@@ -0,0 +1,35 @@
+use crate::fs::error_metadata::{MOUNTPOINT_ERROR_INTERNAL, MOUNTPOINT_ERROR_UNSUPPORTED};
+use crate::fs::Error;
+use tracing::event;
+
+pub fn log_fs_error_event(error: &Error, fuse_operation: &str, fuse_request_id: u64) {
```
passaro, line 5: nit: `log_fuse_error_event`?

mountpoint-s3/src/logging/event_log.rs
```diff
@@ -0,0 +1,35 @@
+use crate::fs::error_metadata::{MOUNTPOINT_ERROR_INTERNAL, MOUNTPOINT_ERROR_UNSUPPORTED};
```
passaro, line 1: Would it make sense to briefly describe the event log here? And also add some comments on the 2 functions?

mountpoint-s3/src/cli.rs
```diff
@@ -432,7 +460,7 @@ where
     );
 
     if args.foreground {
-        init_logging(args.logging_config()).context("failed to initialize logging")?;
+        init_logging(args.logging_config(/* is_fork_parent= */ false)).context("failed to initialize logging")?;
```
passaro, line 463: I am not a fan of this style to document a bool. But in this case maybe we can avoid the parameter entirely (see other comment).

passaro (commented):
Looks good overall. Left a couple of observations.

mountpoint-s3/src/cli.rs
```diff
@@ -372,10 +392,18 @@ impl CliArgs {
             filter
         };
 
+        let _event_log_directory: Option<PathBuf> = None;
+        #[cfg(feature = "event_log")]
+        let _event_log_directory = if _is_fork_parent {
```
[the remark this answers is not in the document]
vladem: Ideally parent needs it's own log file then (otherwise there is a possibility of mixed up log records). Today parent and child will log to the same file (but I think there is [a possibility](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/logging.rs#L97) of them writing to different files, so it's kind of an undefined behaviour). I don't want to create a separate file for parent now, neither we can allow a UB in case of the event log, so I decided to remove event logging from parent.

LMK if you have other thoughts though, otherwise I can just add a comment with reasoning.

NB. all of the mount errors [must be in child](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/cli.rs#L466), I don't expect any to come from the parent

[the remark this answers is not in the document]
passaro: I suspect there are failures in the parent (e.g. failed to instantiate/talk to child) that we may want to log eventually. And we may need to fix the UB.

But okay to leave it out of this PR. Just comment on it.

mountpoint-s3/src/cli.rs
```diff
@@ -372,10 +392,18 @@ impl CliArgs {
             filter
         };
 
+        let _event_log_directory: Option<PathBuf> = None;
```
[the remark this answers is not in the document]
vladem: made it to return `Option<PathBuf>` as supplying path to `LoggingConfig` is the only place where we use this method and `LoggingConfig` owns paths that it holds ( i.e. `log_directory` now)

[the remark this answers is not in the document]
passaro: minor: If you have to `.clone()` before returning from the method, rather prefer returning a reference and let the caller `clone`/`to_owned` it.

mountpoint-s3/src/cli.rs
vladem: hm, I'd say those 2 bits to be deleted from the help message are important (that we only log failed fuse operations and that `MOUNTPOINT_LOG=off` disables the event logging), do you thing it's worth to add an `event log` section in [CONFIGURATION.md](https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#logging) or since it's unreleased better keep it in comments in [mountpoint-s3/src/logging/event_log.rs](https://github.com/awslabs/mountpoint-s3/pull/919/files/cb98dfa433eafbe0c98e3fef2e7c58da4c380c89#diff-07e82ed97e76743ed5dd7709689574da3efb385fdb1b347941b6fc6289f1c8fd)?

vladem:
This will be addressed later.

mountpoint-s3/src/cli.rs
```diff
@@ -406,9 +426,24 @@ impl CliArgs {
             log_directory: self.log_directory.clone(),
             log_to_stdout: self.foreground,
             default_filter,
+            event_log_directory: self.get_event_log_directory(_is_fork_parent).map(Path::to_path_buf),
         }
     }
 
+    fn get_event_log_directory(&self, _is_fork_parent: bool) -> Option<&Path> {
```
ujinho, line 433: Shouldn't we lock the entire function by feature flag?

Issue #527, opened 2023-09-20
Compatibility for Alpine Linux


choeflake:
### Tell us more about this new feature.

Currently the Mountpoint release does not contain a build for Alpine Linux. Will it be possible to introduce an Alpine Linux version? I would very appreciate this since some systems could only run Alpine Linux (docker images).

dannycjones:
Thanks for the feature request, Cornelis!

I'm assuming you are not blocked to build mountpoint-s3 from source? You'd need to install relevant dependencies as part of a [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/), to ensure the image isn't too large.

Of course, I can see why having a compatible release would make it easier and skip out the build entirely.

gabrik:
Hi, I have the same issue, as I'm looking to use it in an Alpine AMI image.

I'm trying to build from source, but got some errors during linking with `libclang`

### libclang:
```
shared library at /usr/lib/llvm16/lib/libclang.so.16.0.6 could not be opened: Dynamic loading not supported
```

Here the `Dockerfile` I'm using:

```
FROM rust:alpine3.18 as build

RUN apk add pkgconf alpine-sdk clang-dev fuse3-dev git cmake libunwind-dev

RUN git clone https://github.com/awslabs/mountpoint-s3 -b v1.0.2
RUN cd mountpoint-s3 && git submodule update --init --recursive

RUN cd mountpoint-s3 && make


FROM alpine:latest as mount-s3-alpine
COPY --from=build /root/mountpoint-s3/target/release/mount-s3 /

ENTRYPOINT [ "/mount-s3", "-f" ]
```

Any suggestion?

choeflake:
> Thanks for the feature request, Cornelis!
> 
> I'm assuming you are not blocked to build mountpoint-s3 from source? You'd need to install relevant dependencies as part of a [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/), to ensure the image isn't too large.

Thanks for the reply. I did not manage it to successful build on an Alpine image. In my case I assumed that Alpine was required as I had to extend from an existing image. But I found that I also could go with an Ubuntu based image, so I continued with that.
I cannot remember why I could not successful build as it is a few days ago. The final image has to run on X64 and my laptop used for building is ARM based. That could also be the or an issue.

> Of course, I can see why having a compatible release would make it easier and skip out the build entirely.

[...]

PR #922, opened 2024-06-24
Use MaybeUninit when getting thread ID from CRT request metrics

dannycjones:
## Description of change

Alpine Linux is not officially supported by Mountpoint, however I saw in https://github.com/awslabs/mountpoint-s3/issues/527#issuecomment-2171576001 that building currently fails for what appears to be a trivial reason - we do not use a more generic way of initializing a null pointer.

It would fail with this error:

```
388.8    Compiling mountpoint-s3-crt v0.7.0 (/mountpoint-s3/mountpoint-s3-crt)
389.1 warning: `mountpoint-s3-crt-sys` (lib) generated 2 warnings
389.3 error[E0308]: mismatched types
389.3     --> mountpoint-s3-crt/src/s3/client.rs:1184:40
389.3      |
389.3 1184 |         let mut out: aws_thread_id_t = 0;
389.3      |                      ---------------   ^ expected `*mut c_void`, found `usize`
389.3      |                      |
389.3      |                      expected due to this
389.3      |
389.3      = note: expected raw pointer `*mut c_void`
389.3                        found type `usize`
389.3 help: if you meant to create a null pointer, use `std::ptr::null_mut()`
389.3      |
389.3 1184 |         let mut out: aws_thread_id_t = std::ptr::null_mut();
389.3      |                                        ~~~~~~~~~~~~~~~~~~~~
389.3 
389.5 For more information about this error, try `rustc --explain E0308`.
389.5 error: could not compile `mountpoint-s3-crt` (lib) due to 1 previous error
```

This change updates the code to use [`std::mem::MaybeUninit<aws_thread_id_t>`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#out-pointers) rather than `0long` for the thread ID. The type of the thread ID can vary depending on where Mountpoint is built, and this will allow Mountpoint to compile where the thread ID is not type `long`.

Note, we do not appear to be following the advised practice for formatting the thread ID. _"For portability, aws_thread_id_t must not be printed directly."_. Since we only use try to print the thread ID for logging, I think this is OK for now. Really, we should be using [`aws_thread_id_t_to_string`](https://docs.rs/mountpoint-s3-crt-sys/latest/mountpoint_s3_crt_sys/fn.aws_thread_id_t_to_string.html) however using this relies on the C macro `AWS_THREAD_ID_T_REPR_BUFSZ` which is not made available today by our bindings.

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/527#issuecomment-2171576001

### Testing

I've verified that this change compiles on Alpine 3.18. The change will go through standard integration tests as part of the pull request and merge.

For reference, I was using this Dockerfile:

```Dockerfile
FROM alpine:3.18

# Install build tools
RUN apk add gcc curl git zlib g++ cmake make pkgconfig fuse fuse-dev clang clang-dev

# Install rust
RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y && \
    source "$HOME/.cargo/env"

ENV RUSTFLAGS="-C target-feature=-crt-static"

# Build mountpoint-s3
RUN git clone \
        --recurse-submodules --shallow-submodules \
        https://github.com/awslabs/mountpoint-s3.git && \
    source "$HOME/.cargo/env" && \
    cd mountpoint-s3 && \
    cargo build --release

RUN echo "user_allow_other" >> /etc/fuse.conf

# Run in foreground mode so that the container can be detached without exiting Mountpoint
ENTRYPOINT [ "/mount-s3", "-f" ]

```

This does not change existing behavior. This allows Mountpoint to build on platforms where the thread ID type in the CRT is not `long`.

---

### DIFF
--- a/mountpoint-s3-crt/src/s3/client.rs
+++ b/mountpoint-s3-crt/src/s3/client.rs
@@ -1184,15 +1184,11 @@ impl RequestMetrics {
     /// Get the ID of the thread the request was made from
     pub fn thread_id(&self) -> Option<ThreadId> {
         let mut out: MaybeUninit<aws_thread_id_t> = MaybeUninit::uninit();
-        // SAFETY: `inner` is a valid aws_s3_request_metrics, `out` will be a valid pointer if no error was set
+        // SAFETY: `inner` is a valid aws_s3_request_metrics, `out` will point to initialized memory if no error was set
         let thread_id = unsafe {
             aws_s3_request_metrics_get_thread_id(self.inner.as_ptr(), out.as_mut_ptr())
                 .ok_or_last_error()
                 .ok()?;
-            assert!(
-                !out.as_ptr().is_null(),
-                "Thread ID should be available if call succeeded",
-            );
             out.assume_init()
         };
         Some(thread_id.into())
--- a/mountpoint-s3-crt/src/s3/client.rs
+++ b/mountpoint-s3-crt/src/s3/client.rs
@@ -13,9 +13,11 @@ use crate::s3::s3_library_init;
 use crate::{aws_byte_cursor_as_slice, CrtError, ResultExt, ToAwsByteCursor};
 use futures::Future;
 use mountpoint_s3_crt_sys::*;
+
 use std::ffi::{OsStr, OsString};
 use std::fmt::Debug;
 use std::marker::PhantomPinned;
+use std::mem::MaybeUninit;
 use std::os::unix::prelude::OsStrExt;
 use std::pin::Pin;
 use std::ptr::NonNull;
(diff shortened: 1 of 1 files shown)


mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -1181,14 +1183,19 @@ impl RequestMetrics {
...
+        // SAFETY: `inner` is a valid aws_s3_request_metrics, `out` will be a valid pointer if no error was set
+        let thread_id = unsafe {
+            aws_s3_request_metrics_get_thread_id(self.inner.as_ptr(), out.as_mut_ptr())
                 .ok_or_last_error()
-                .ok()?
+                .ok()?;
+            assert!(
+                !out.as_ptr().is_null(),
+                "Thread ID should be available if call succeeded",
+            );
```
jamesbornholt, line 1195: This assertion is always true, `out.as_ptr()` is a pointer _to the_ `aws_thread_id_t`. I think you can just drop it; the early return tells us that it succeeded.

dannycjones: Got it. In this case, we're giving the function a pointer to a memory location to populate and we then assume its initialized after the call completes successfully.

We'd want to check the pointer isn't null if we passed a _null pointer_ where the function should replace that with a valid pointer and wanted to check its been updated to a new location of memory.

jamesbornholt (approved):

PR #924, opened 2024-06-26
Include test files in cargo

ahmarsuhail:
## Description of change

Cargo publish was failing without these files. 

verified that it works after this change by running `cargo publish --dry-run -p mountpoint-s3-crt-sys` . Crate size is 9.9MB

no

---


monthonk (approved):
LGTM.

I think this is a gap we can improve in our CI. We probably should include dry-run for cargo publish in [crates workflow](https://github.com/awslabs/mountpoint-s3/blob/main/.github/workflows/crates.yml), but that will be in another PR.

PR #926, opened 2024-06-27
Re-implement the prefetcher using backpressure mechanism

monthonk:
## Description of change

The prefetcher now uses only one GetObject request to fetch data in advance. This request has a range of entire object but use backpressure mechanism to control how much data it wants to fetch into the part queue instead of spawning up to two requests in parallel.

This should make the throughput more stable because previously the two request tasks could compete with each other when fetching data from S3. Also, it will be easier to control how much data we want to store in the part queue.

No, the change only for internal interface.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client/get_object.rs
+++ b/mountpoint-s3-client/src/s3_crt_client/get_object.rs
@@ -46,13 +46,16 @@ impl S3CrtClient {
                 .map_err(S3RequestError::construction_failure)?;
         }
 
-        if let Some(range) = range {
+        let next_offset = if let Some(range) = range {
             // Range HTTP header is bounded below *inclusive*
             let range_value = format!("bytes={}-{}", range.start, range.end.saturating_sub(1));
             message
                 .set_header(&Header::new("Range", range_value))
                 .map_err(S3RequestError::construction_failure)?;
-        }
+            range.start
+        } else {
+            0
+        };
 
         let key = format!("/{key}");
         message
@@ -60,6 +63,7 @@ impl S3CrtClient {
             .map_err(S3RequestError::construction_failure)?;
 
         let (sender, receiver) = futures::channel::mpsc::unbounded();
+        let read_window_offset = next_offset + self.inner.initial_read_window_size as u64;
 
         let mut options = S3CrtClientInner::new_meta_request_options(message, S3Operation::GetObject);
         options.part_size(self.inner.read_part_size as u64);
@@ -84,6 +88,9 @@ impl S3CrtClient {
             request,
             finish_receiver: receiver,
             finished: false,
+            enable_backpressure: self.inner.enable_backpressure,
+            next_offset,
+            read_window_offset,
         })
     }
 }
@@ -101,13 +108,21 @@ pub struct S3GetObjectRequest {
     #[pin]
     finish_receiver: UnboundedReceiver<Result<GetBodyPart, Error>>,
     finished: bool,
+    enable_backpressure: bool,
+    next_offset: u64,
+    read_window_offset: u64,
 }
 
 impl GetObjectRequest for S3GetObjectRequest {
     type ClientError = S3RequestError;
 
     fn increment_read_window(mut self: Pin<&mut Self>, len: usize) {
+        self.read_window_offset += len as u64;
         self.request.meta_request.increment_read_window(len as u64);
+    }
+
+    fn read_window_offset(self: Pin<&Self>) -> u64 {
+        self.read_window_offset
     }
 }
 
@@ -122,7 +137,14 @@ impl Stream for S3GetObjectRequest {
         let this = self.project();
 
         if let Poll::Ready(Some(val)) = this.finish_receiver.poll_next(cx) {
-            return Poll::Ready(Some(val.map_err(|e| ObjectClientError::ClientError(e.into()))));
+            let result = match val {
+                Ok(item) => {
+                    *this.next_offset += item.1.len() as u64;
+                    Some(Ok(item))
+                }
+                Err(e) => Some(Err(ObjectClientError::ClientError(e.into()))),
+            };
+            return Poll::Ready(result);
         }
 
         match this.request.poll(cx) {
@@ -134,7 +156,16 @@ impl Stream for S3GetObjectRequest {
                 *this.finished = true;
                 Poll::Ready(Some(Err(e)))
             }
-            Poll::Pending => Poll::Pending,
+            Poll::Pending => {
+                // If the request is still not finished but the read window is not enough to poll
+                // the next chunk we might want to return error instead of keeping the request blocked.
+                if *this.enable_backpressure && this.read_window_offset <= this.next_offset {
+                    return Poll::Ready(Some(Err(ObjectClientError::ClientError(
+                        S3RequestError::EmptyReadWindow,
+                    ))));
+                }
+                Poll::Pending
+            }
         }
     }
 }
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -1,12 +1,37 @@
 //! This module implements a prefetcher for GetObject requests.
 //!
-//! It works by making increasingly larger GetObject requests to the CRT. We want the chunks to be
-//! large enough that they can make effective use of the CRT's fan-out parallelism across the S3
-//! frontend, but small enough that we don't accumulate a lot of unread object data in memory or
-//! wastefully download data we'll never read. As the reader continues to make sequential reads,
-//! we increase the size of the GetObject requests up to some maximum. If the reader ever makes a
-//! non-sequential read, we abandon the prefetching and start again with the minimum request size.
+//! It works by relying on the CRT's flow-control window feature. The prefetcher creates a single
+//! GetObject request with entire length of the object (starting from the first read offset) and
+//! makes increasingly larger read window. We want the chunks to be large enough that they can make
+//! effective use of the CRT's fan-out parallelism across the S3 frontend, but small enough that we
+//! don't accumulate a lot of unread object data in memory or wastefully download data we'll never
+//! read. As the reader continues to make sequential reads, we increase the size of the read window
+//! up to some maximum. If the reader ever makes a non-sequential read, we abandon the prefetching
+//! and start again with a new GetObject request with minimum read window size.
+//!
+//! In more technical details, the prefetcher creates a RequestTask when receiving the first read
+//! request from the file system or after it has just been reset. The RequestTask consists of two main
+//! components.
+//! 1.  An ObjectPartStream that has a role to continuously fetch data from the sources which can be
+//!     either S3 or the cache on disk. The ObjectPartStream is spawned and run in a separate thread
+//!     from the prefetcher.
+//! 2.  A PartQueue, where we store data received from the ObjectPartStream, waiting to be read from
+//!     the prefetcher via a RequestTask function.
+//!
+//! A backpressure mechanism is needed to control how much data we want to store in the part queue at
+//! a time as we don't want to download the entire object into memory. For the client part stream, we
+//! may be able to rely on the CRT flow-control flow window to block when we don't increase the read
+//! window size, but for the caching part stream we don't have the machinery to do that yet. That's why
+//! we introduce the BackpressureController and BackpressureLimiter to help solving this.
+//!
+//! Essentially, the BackpressureController and BackpressureLimiter is a pair of sender/receiver of a
+//! channel, created at RequestTask initialization. The sender is handed to the RequestTask. Its role
+//! is to communicate with its receiver to tell "when" it is ready to receive more data. The receiver
+//! is handed to the ObjectPartStream where the stream should call a provided function "before" fetching
+//! more data from the sources and put them into the part queue. The BackpressureLimiter should be used
+//! as a mean to block ObjectPartStream thread to fetch more data.
 
+mod backpressure_controller;
 mod caching_stream;
 mod part;
 mod part_queue;
@@ -14,7 +39,6 @@ mod part_stream;
 mod seek_window;
 mod task;
 
-use std::collections::VecDeque;
 use std::fmt::Debug;
 use std::time::Duration;
 
@@ -24,6 +48,8 @@ use metrics::{counter, histogram};
 use mountpoint_s3_client::error::{GetObjectError, ObjectClientError};
 use mountpoint_s3_client::types::ETag;
 use mountpoint_s3_client::ObjectClient;
+use part::PartOperationError;
+use part_stream::RequestTaskConfig;
 use thiserror::Error;
 use tracing::trace;
 
@@ -79,6 +105,15 @@ pub enum PrefetchReadError<E> {
 
     #[error("integrity check failed")]
     Integrity(#[from] IntegrityError),
+
+    #[error("part read failed")]
+    PartReadFailed(#[from] PartOperationError),
+
+    #[error("backpressure must be enabled with non-zero initial read window")]
+    BackpressurePreconditionFailed,
+
+    #[error("read window increment failed")]
+    ReadWindowIncrement,
 }
 
 pub type DefaultPrefetcher<Runtime> = Prefetcher<ClientPartStream<Runtime>>;
@@ -110,10 +145,8 @@ where
 
 #[derive(Debug, Clone, Copy)]
 pub struct PrefetcherConfig {
-    /// Size of the first request in a prefetch run
-    pub first_request_size: usize,
-    /// Maximum size of a single prefetch request
-    pub max_request_size: usize,
+    /// Maximum size of the read window
+    pub max_read_window_size: usize,
     /// Factor to increase the request size by whenever the reader continues making sequential reads
     pub sequential_prefetch_multiplier: usize,
     /// Timeout to wait for a part to become available
@@ -127,19 +160,10 @@ pub struct PrefetcherConfig {
 }
 
 impl Default for PrefetcherConfig {
+    #[allow(clippy::identity_op)]
     fn default() -> Self {
-        #[allow(clippy::identity_op)]
         Self {
-            // This is a weird looking number! We really want our first request size to be 1MiB,
-            // which is a common IO size. But Linux's readahead will try to read an extra 128k on on
-            // top of a 1MiB read, which we'd have to wait for a second request to service. Because
-            // FUSE doesn't know the difference between regular reads and readahead reads, it will
-            // send us a READ request for that 128k, so we'll have to block waiting for it even if
-            // the application doesn't want it. This is all in the noise for sequential IO, but
-            // waiting for the readahead hurts random IO. So we add 128k to the first request size
-            // to avoid the latency hit of the second request.
-            first_request_size: 1 * 1024 * 1024 + 128 * 1024,
-            max_request_size: 2 * 1024 * 1024 * 1024,
+            max_read_window_size: 2 * 1024 * 1024 * 1024,
             sequential_prefetch_multiplier: 8,
             read_timeout: Duration::from_secs(60),
             // We want these large enough to tolerate a single out-of-order Linux readahead, which
@@ -147,7 +171,7 @@ impl Default for PrefetcherConfig {
             // making a guess about where the optimal cut-off point is before it would be faster to
             // just start a new request instead.
             max_forward_seek_wait_distance: 16 * 1024 * 1024,
-            max_backward_seek_distance: 1 * 1024 * 1024,
+            max_backward_seek_distance: 1024 * 1024,
         }
     }
 }
@@ -206,11 +230,7 @@ pub struct PrefetchGetObject<Stream: ObjectPartStream, Client: ObjectClient> {
     client: Arc<Client>,
     part_stream: Arc<Stream>,
     config: PrefetcherConfig,
-    // Invariant: the offset of the first byte in this task's part queue is always
-    // self.next_sequential_read_offset.
-    current_task: Option<RequestTask<Client::ClientError>>,
-    // Currently we only every spawn at most one future task (see [spawn_next_request])
-    future_tasks: VecDeque<RequestTask<Client::ClientError>>,
+    backpressure_task: Option<RequestTask<Client::ClientError>>,
     // Invariant: the offset of the last byte in this window is always
     // self.next_sequential_read_offset - 1.
     backward_seek_window: SeekWindow,
@@ -221,7 +241,6 @@ pub struct PrefetchGetObject<Stream: ObjectPartStream, Client: ObjectClient> {
     /// Start offset for sequential read, used for calculating contiguous read metric
     sequential_read_start_offset: u64,
     next_sequential_read_offset: u64,
-    next_request_size: usize,
     next_request_offset: u64,
     size: u64,
 }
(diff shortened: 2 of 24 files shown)


mountpoint-s3/src/prefetch/part_stream.rs
```diff
@@ -223,12 +227,35 @@ where
...
                                 curr_offset += part.len() as u64;
-                                part_queue_producer.push(Ok(part));
+                                part_queue_producer.push(Ok(part), Some(next_read_window_offset));
+                            }
+                            debug_assert!(
+                                curr_offset <= last_offset,
+                                "current offset must not be larger than last offset"
+                            );
+
+                            if remaining_window_size > 0 {
```
passaro, line 237: It's not clear to me why we are calling `increment_read_window` here or why we need the sender/receiver pair. I'd argue this logic requires some comment or maybe a rethink.

mountpoint-s3/src/prefetch/task.rs
```diff
@@ -14,36 +15,62 @@ pub struct RequestTask<E: std::error::Error> {
...
             start_offset: offset,
             total_size: size,
             part_queue,
+            read_window_updater,
         }
     }
 
+    // Push a given list of parts in front of the part queue
+    pub async fn push_front(&mut self, mut parts: Vec<Part>) -> Result<(), PrefetchReadError<E>> {
+        // Merge all parts into one single part.
```
passaro, line 58: This could be expensive (especially because of checksumming). Couldn't we just keep a vector of parts?

mountpoint-s3/src/prefetch/part_stream.rs
```diff
@@ -168,49 +169,52 @@ where
...
 
-                pin_mut!(get_object_result);
+                pin_mut!(request);
                 loop {
-                    match get_object_result.next().await {
+                    match request.next().await {
                         Some(Ok((offset, body))) => {
                             trace!(offset, length = body.len(), "received GetObject part");
-                            metrics::counter!("s3.client.total_bytes", "type" => "read").increment(body.len() as u64);
+
```
passaro, line 211: Are we not reporting `total_bytes` anymore?

passaro (commented):
Not a full review, but I do have a few questions before continuing.

mountpoint-s3/src/prefetch.rs
```diff
@@ -110,35 +112,50 @@ where
...
-            // This is a weird looking number! We really want our first request size to be 1MiB,
-            // which is a common IO size. But Linux's readahead will try to read an extra 128k on on
-            // top of a 1MiB read, which we'd have to wait for a second request to service. Because
-            // FUSE doesn't know the difference between regular reads and readahead reads, it will
-            // send us a READ request for that 128k, so we'll have to block waiting for it even if
-            // the application doesn't want it. This is all in the noise for sequential IO, but
-            // waiting for the readahead hurts random IO. So we add 128k to the first request size
-            // to avoid the latency hit of the second request.
-            first_request_size: 1 * 1024 * 1024 + 128 * 1024,
+            initial_read_window_size,
```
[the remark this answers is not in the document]
monthonk: Yes, that's right. The CRT will return a part size anyway even if the read window is smaller, so the change here should reflect that. The latency could also be impacted as the first request size is larger.

mountpoint-s3/src/prefetch.rs
```diff
@@ -284,31 +323,47 @@ where
...
         while to_read > 0 {
-            let Some(current_task) = self.current_task.as_mut() else {
-                // If [prepare_requests] didn't spawn a request, we've reached the end of the object.
+            let Some(current_task) = self.backpressure_task.as_mut() else {
                 trace!(offset, length, "read beyond object size");
                 break;
             };
             debug_assert!(current_task.remaining() > 0);
 
+            if let Some(len) = self.next_read_window_size {
```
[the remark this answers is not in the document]
monthonk: I will try to consolidate it, probably easier if this logic is in `RequestTask`.

mountpoint-s3/src/prefetch.rs
```diff
@@ -284,31 +323,47 @@ where
...
+                trace!(
+                    current_window = self.current_read_window_size,
+                    next_window = len,
+                    "incrementing read window"
+                );
+                current_task.increment_read_window(len);
+                self.current_read_window_size = len;
```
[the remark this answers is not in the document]
monthonk: It actually remembers the last window size we have increased. Probably bad choice of naming here.

mountpoint-s3/src/prefetch.rs
```diff
@@ -458,61 +469,45 @@ where
...
+        if offset >= available_soon_offset {
             trace!(
                 requested_offset = offset,
                 available_offset = available_offset,
                 "seek failed: not enough data available"
             );
             return Ok(false);
         }
+
+        // Make sure there is enough data in the read window
```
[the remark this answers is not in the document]
monthonk: I think you're right, will update this.

mountpoint-s3/src/prefetch/task.rs
monthonk: My thinking is that it's fairly cheap compare to creating a new request and this way we don't have to change the part queue structure. I will see if we can improve this.

mountpoint-s3/src/prefetch.rs
```diff
@@ -206,22 +248,19 @@ pub struct PrefetchGetObject<Stream: ObjectPartStream, Client: ObjectClient> {
...
+    backpressure_task: Option<RequestTask<Client::ClientError>>,
     // Invariant: the offset of the last byte in this window is always
     // self.next_sequential_read_offset - 1.
     backward_seek_window: SeekWindow,
     bucket: String,
     object_id: ObjectId,
     // preferred part size in the prefetcher's part queue, not the object part
     preferred_part_size: usize,
+    current_read_window_size: usize,
+    next_read_window_size: Option<usize>,
```
vladem, line 260: May it make things simpler if we'd always reason in terms of absolute offsets, including this `next_read_window_size` variable?

I understand that now you use `next_read_window_size` as a relative value, which will be added to the current allowed offset. The problem is that since we mix absolute offsets and relative values, the code is hard to follow (e.g. you need to look around to understand what's `fn prepare_next_read_window` is doing): you always need to keep in mind what is relative and what is absolute. 

May be if we'd just record the last allowed offset that we've set for the request and the previous one, we'd still be able to calculate the value to use in `current_task.increment_read_window(..)`?

As aside -- didn't get why the calculation of `next_read_window_size` is separated from the call to `increment_read_window`. This creates the need for this `PrefetchGetObject::next_read_window_size` field, which could've been just a local variable.

mountpoint-s3/src/prefetch.rs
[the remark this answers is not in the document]
vladem: +1 for resetting if `next_read_window_offset < available_soon_offset`.

So `max_forward_seek_wait_distance` [today](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/prefetch.rs#L149) is `16MB` and we assume that this is more efficient for the existing request to download `16MB` instead of creating a new one. Monthon's code looks aligned with this assumption (after line [482](https://github.com/awslabs/mountpoint-s3/pull/926/files#diff-ff2b2b24fbf3d7ad046bb857a44d84da70bbdd714b0ea9460f5d1b8a08408effR482) we know that it's less than `16MB` to be downloaded left to serve the read).

I'd say the problem here is that we may increment the window exactly to the requested byte and no further. Any window incrementing rule which prevents this looks arbitrary or complicated, so I'd just reset too.

mountpoint-s3/src/prefetch.rs
```diff
@@ -284,31 +323,47 @@ where
...
+                trace!(
+                    current_window = self.current_read_window_size,
+                    next_window = len,
+                    "incrementing read window"
+                );
+                current_task.increment_read_window(len);
+                self.current_read_window_size = len;
+                self.next_read_window_size = None;
+            }
             let part = match current_task.read(to_read as usize).await {
```
vladem, line 348: not clear how do we make sure that this will never block (which would mean a deadlock as this function is the only one increasing the window), I believe you guarantee that `to_read` is always in the window, a comment may help

mountpoint-s3/src/prefetch.rs
```diff
@@ -354,54 +408,27 @@ where
...
-        let start = self.next_request_offset;
-        if start >= self.size {
-            return None;
-        }
+    /// Spawn a backpressure GetObject request which has a range from current offset to the end of the file.
+    /// We will be using flow-control window to control how much data we want to download into the prefetcher.
+    fn spawn_read_backpressure_request(&mut self) -> RequestTask<Client::ClientError> {
+        let start = self.next_sequential_read_offset;
+        let object_size = self.size as usize;
+        let range = RequestRange::new(object_size, start, object_size);
```
vladem, line 430: dumb question, but we do have a test for backpressure-enabled requests, that they handle ranges well, i.e when request reads from the middle of an object?

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -109,6 +137,9 @@ impl<E: std::error::Error + Send + Sync> PartQueueProducer<E> {
             trace!("closed channel");
         } else {
             self.bytes_sent.fetch_add(part_len, Ordering::SeqCst);
+            let _ = self
+                .next_read_window_offset
```
vladem, line 141: do we actually need to receive `next_read_window_offset` from the client? don't we have enough information in `Prefetcher` / `RequestTask` to compute that value? 

having to pass `next_read_window_offset` from `ObjectPartStream` to `Prefetcher` through multiple layers so it can just do:
```rust
let next_read_window_offset = current_task.next_read_window_offset();
self.prepare_next_read_window(remaining_window, next_read_window_offset);
```
looks overcomplicated, or am I missing something?

mountpoint-s3/src/prefetch.rs
monthonk: I think the tests that reset the prefetcher should cover this, like random read or forward seek, but I didn't add new tests specifically for it.

mountpoint-s3/src/prefetch/part_queue.rs
monthonk: I just updated the code to simplify the flow. It should not be too complicated now.

mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -99,14 +107,23 @@ pub struct S3GetObjectRequest {
...
 
 impl GetObjectRequest for S3GetObjectRequest {
     type ClientError = S3RequestError;
 
     fn increment_read_window(mut self: Pin<&mut Self>, len: usize) {
+        self.next_read_window_offset += len as u64;
         self.request.meta_request.increment_read_window(len as u64);
     }
+
+    fn next_read_window_offset(self: Pin<&Self>) -> u64 {
```
passaro, line 124: This may require some comment or a better name. Why `next`? Isn't this the upper bound of the current read window?

mountpoint-s3/src/prefetch/backpressure_controller.rs
```diff
@@ -0,0 +1,118 @@
...
+    next_read_window_offset: u64,
+    last_request_offset: u64,
+}
+
+#[derive(Debug)]
+pub struct BackpressureLimiter {
+    read_window_incrementing_queue: Receiver<usize>,
+}
+
+/// Creates a [BackpressureController] and its related [BackpressureLimiter].
```
passaro, line 19: Could we explain what's the function of these 2 types? Their methods and fields could also do with some comments (and/or clearer names).

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -207,29 +262,65 @@ where
...
+                    let next_offset = offset + body.len() as u64;
+                    debug_assert!(
+                        next_offset <= block_aligned_byte_range.end,
+                        "current offset must be withing the block range"
+                    );
 
-                    // Split the body into blocks.
                     let mut body: Bytes = body.into();
+
+                    // S3 doesn't provide checksum for us if the request range is not aligned to
```
passaro, line 273: It is not clear to me what this whole block is doing.

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -183,7 +218,27 @@ where
             }
         };
 
-        pin_mut!(get_object_result);
+        pin_mut!(get_object_request);
+
+        // Current read window might be larger or smaller by client's read window at this point so we need to match them
```
passaro, line 223: This block needs some clarification. Why do we need this check? How do we match them?

mountpoint-s3/src/prefetch.rs
```diff
@@ -347,7 +386,17 @@ where
...
 
             self.next_sequential_read_offset += part_bytes.len() as u64;
-            self.prepare_requests();
+
+            // Increase the read window size if available bytes in the queue are less than half of previous read window size
+            let remaining_in_queue = current_task
+                .available_offset()
+                .saturating_sub(self.next_sequential_read_offset) as usize;
+            if remaining_in_queue < self.preferred_read_window_size / 2 {
+                self.preferred_read_window_size = (self.preferred_read_window_size
```
passaro, line 395: Do we need `self.preferred_read_window_size` now? Maybe it should belong to `RequestTask`, which could handle the read window increase when triggered here.

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -93,15 +94,33 @@ impl<E: std::error::Error + Send + Sync> PartQueue<E> {
...
+
+        if let Some(current_part) = current_part.as_mut() {
+            part.extend(current_part).unwrap();
+            *current_part = part;
+        } else {
+            *current_part = Some(part);
+        }
+    }
+
+    pub fn bytes_received(&self) -> u64 {
```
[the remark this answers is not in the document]
monthonk: I think `bytes_received` could be the size of the object now where previously it should not be larger than 2GB per task.

mountpoint-s3-client/src/s3_crt_client/get_object.rs
monthonk: In my mind current offset is at the start of the read window (the lower bound), so the upper bound would be `next`. I will add the comment and also happy to change if it doesn't make sense.

mountpoint-s3/src/prefetch.rs
monthonk: I'm not sure where it should be either, but I believe the prefetcher should be in control of scaling up or down of the read window size. That's why I keep it here.

vladem: +1 to Alex, I'd say this logic belongs to `RequestTask` (or even to `BackpressureLimiter`?), in that way window-moving logic won't be scattered around the code base and there will be no necessity for `RequestTask` to look inside `RequestTask` for `available_offset`.

by the way,  wouldn't a simplified interface like this suffice for `BackpressureController`?
```rust
impl BackpressureController {
    // Used by `PrefetchGetObject` to report to `BackpressureController` the offset 
    // until which the data was read by the user application.
    pub fn set_next_sequential_read_offset(&mut self, next_sequential_read_offset: usize)
}
```

Ultimately `next_sequential_read_offset` seems to be the only bit of information, which `PrefetchGetObject` should report to `BackpressureLimiter` (via `BackpressureController`) so that the later can make a decision to move the window (apart from the information provided on `RequestTask` instantiation: `preferred_part_size`, `object_size`).

vladem: the whole thing becomes much clearer than:
1. we have a `RequestTask` pumping data from CRT to the queue;
2. we have a `PrefetchGetObject` reading that data from the queue when user requests it;
3. `RequestTask` and `PrefetchGetObject` are synchronised through an additional queue, which is used by `PrefetchGetObject` to report when the data is consumed;
4. `RequestTask` stops pumping next data parts from CRT (in `next_read_window_blocking` call) until `PrefetchGetObject` consumes something from the previous part (and reports `next_sequential_read_offset`)

monthonk: Just to clarify, the logic here is not for moving the read window but for deciding if we should use bigger window size. Agree that the window moving logic should be in the `RequestTask`.

vladem: Looks overcomplicated when we implement window calculation in `PrefetchGetObject`. As I mentioned it needs to fetch an atomic `available_offset` from the `RequestTask` and pass the new window size through the channel to `ClientPartStream`, while it could've passed more basic information in form of consumed offset.

But if you have a strong opinion of why window calculation logic should be in `PrefetchGetObject`, please, share your reasoning!

monthonk: Not a really strong opinion, if the logic is in `RequestTask` we still have to pass information about how to scale up the request (like `max_read_window_size` or `sequential_prefetch_multiplier`) to the task as they are currently in the `PrefetcherConfig`. I'm also thinking about if we have some kind of memory limiter in the future, we will also have to push the information/logic about how to scale down to the task.

vladem: Ack, some of the information originates from `PrefetcherConfig` and we want to keep it there.

mountpoint-s3/src/prefetch.rs
```diff
@@ -7,14 +7,14 @@
 //! we increase the size of the GetObject requests up to some maximum. If the reader ever makes a
 //! non-sequential read, we abandon the prefetching and start again with the minimum request size.
 
+mod backpressure_controller;
```
passaro, line 10: Can you update the comment above? It may be a good place to also describe the overall new design and how the various components in this module interact.

passaro (commented):
Had another go and left a few comments.

Overall, I'm not too convinced of where the backpressure mechanism and the prefetching logic have ended up, but I don't really have a better proposal. It may only be a matter of better documenting which choices are driven by the current CRT API (and may evolve later), but happy to discuss further.

mountpoint-s3/src/prefetch/backpressure_controller.rs
```diff
@@ -0,0 +1,151 @@
...
+/// `wait_for_read_window_increment` to wait for feedback from the consumer. This method
+/// could block when they know that the producer requires read window incrementing.
+///
+/// [BackpressureController] will be given to the consumer side of the object stream.
+/// It can be used anywhere to set preferred read window size for the stream and tell the
+/// producer when its read window should be increased.
+pub fn new_backpressure_controller(
+    preferred_read_window_size: usize,
+    max_read_window_size: usize,
+    read_window_size_multiplier: usize,
```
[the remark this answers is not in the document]
monthonk: Yes, it is now. We need to tell it to scale up, but the scaling strategy must be configured when we create `BackpressureController`.

mountpoint-s3/src/prefetch/part.rs
```diff
@@ -21,6 +22,51 @@ impl Part {
...
+    /// Aligning part with the right boundaries will reduce the compute time for validating the checksum at read.
+    pub fn split_to_parts(id: ObjectId, offset: u64, mut body: Bytes, alignment: usize) -> Vec<Self> {
+        let mut curr_offset = offset;
+        let mut parts = Vec::new();
+
+        // First, align the front to part boundaries
+        if offset % alignment as u64 != 0 {
+            let distance_to_align = alignment - (curr_offset % alignment as u64) as usize;
+            let chunk_size = distance_to_align.min(body.len());
+            if chunk_size > 0 {
```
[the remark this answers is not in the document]
monthonk: When body is empty, I will make the check explicit.

mountpoint-s3/src/prefetch/part_queue.rs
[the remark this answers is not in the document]
monthonk: I can change it back to `usize`.

mountpoint-s3/src/prefetch/task.rs
```diff
@@ -1,65 +1,98 @@
...
 
 use crate::prefetch::part::Part;
-use crate::prefetch::part_queue::{unbounded_part_queue, PartQueue};
+use crate::prefetch::part_queue::PartQueue;
 use crate::prefetch::PrefetchReadError;
 
+use super::backpressure_controller::BackpressureController;
+use super::part_stream::RequestRange;
+
+pub struct RequestTaskConfig {
```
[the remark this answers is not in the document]
monthonk: Will move it to where we define `ObjectPartStream`.

mountpoint-s3/src/prefetch/task.rs
```diff
@@ -1,65 +1,98 @@
...
-            start_offset: offset,
-            total_size: size,
-            part_queue,
+    // Push a given list of parts in front of the part queue
+    pub async fn push_front(&mut self, mut parts: Vec<Part>) -> Result<(), PrefetchReadError<E>> {
+        // Merge all parts into one single part.
+        // This could result in a really big part, but we normally use this only for backward seek
+        // so its size should not be bigger than the prefetcher's `max_backward_seek_distance`.
+        let part = parts.iter_mut().reduce(|acc, e| {
+            acc.extend(e).unwrap();
```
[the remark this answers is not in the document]
monthonk: You're right, this should be propagated.

mountpoint-s3/src/prefetch/task.rs
```diff
@@ -1,65 +1,98 @@
...
         self.remaining -= part.len();
+
+        // We read some data out of the part queue so the read window should be moved
+        let next_offset = part.offset() + part.len() as u64;
+        self.backpressure_controller.next_offset_hint(next_offset).await?;
+
+        let remaining_in_queue = self.available_offset().saturating_sub(next_offset) as usize;
+        // If the part queue is empty it means we are reading faster than the task could prefetch,
+        // so we should use larger window for the task.
+        if remaining_in_queue == 0 {
```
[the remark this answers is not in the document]
monthonk: Yes, now we have the `BackpressureController` to keep moving the read window with `preferred_read_window_size` which mean data the part queue should always be filled up to this number. if the part queue is not empty it means current `preferred_read_window_size` is large enough to serve the reader. We only want to scale up when the part queue is empty, indicating that the reader is reading faster than we could prefetch.

mountpoint-s3/src/prefetch/part_stream.rs
```diff
@@ -241,7 +216,113 @@ where
...
+    Client: ObjectClient + Send + Sync + 'static,
+{
+    async fn get_from_client(
+        &mut self,
+        initial_read_window_offset: u64,
+    ) -> Result<(), PrefetchReadError<Client::ClientError>> {
+        let bucket = &self.config.bucket;
+        let id = &self.config.object_id;
+
+        // Normally, initial read window size should be very small (~1MB) so that we can serve the first read request as soon as possible,
```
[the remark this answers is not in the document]
vladem: May be I am missing something here, though. Happy to discuss.

vladem (commented):
This POC shows great results, but some aspects of it bother me. Mainly:
1. memory consumption and behaviour change for cache;
2. huge size of the change;
3. amount of code duplication in `part_stream`/`caching_stream`.

I suggest to tackle 3/ first and see if it helps with the size of the change. We may consider enabling backpressure in 2 iterations (only for regular stream at first). Finally we can address issues with caching stream and enable backpressure for it too.

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -132,56 +153,96 @@ where
...
+            cache_key: cache_key.clone(),
+            preferred_part_size: self.config.preferred_part_size,
+            original_range: range,
+            block_index: block_range.start,
+            block_offset: block_range.start * block_size,
+            buffer: ChecksummedBytes::default(),
+            cache: self.cache.clone(),
+        };
+
+        // Start by issuing the first request that has a range up to initial read window offset.
```
vladem, line 210: nit: if we had a single function, encapsulating the logic of slitting a request into 2, we won't need to have this duplicated explanation (we could've just use this function in both `ClientPartStream` and `CachingStream`)

monthonk:
I'm going to split this into smaller PRs.

PR #933, opened 2024-07-04
Reset prefetcher on any error

passaro:
## Description of change

The prefetcher did not reset its internal state (and cancel current tasks) when encountering an error on a forward seek. As a result, successive reads could try and read from a part queue in an invalid state. Note that this behavior was inconsistent with sequential reads, which would instead trigger a reset on error.

This change ensures that the prefetcher is always reset when encountering an error, whether while reading or seeking. If an application retries a read operation after a failure, the prefetcher will always spawn a new  request task.

Relevant issues: #930

No

---

### DIFF
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -246,7 +246,52 @@ where
             next_seq_offset = self.next_sequential_read_offset,
             "read"
         );
+        let result = self.try_read(offset, length).await;
+        if result.is_err() {
+            self.reset_prefetch_to_offset(offset);
+        }
+        result
+    }
+}
 
+impl<Stream, Client> PrefetchGetObject<Stream, Client>
+where
+    Stream: ObjectPartStream,
+    Client: ObjectClient + Send + Sync + 'static,
+{
+    /// Create and spawn a new prefetching request for an object
+    fn new(
+        client: Arc<Client>,
+        part_stream: Arc<Stream>,
+        config: PrefetcherConfig,
+        bucket: &str,
+        key: &str,
+        size: u64,
+        etag: ETag,
+    ) -> Self {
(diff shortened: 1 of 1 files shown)


monthonk (dismissed):
Nice test! it works well with my changes to the prefetcher too.

mountpoint-s3/src/prefetch.rs
```diff
@@ -246,7 +246,52 @@ where
             next_seq_offset = self.next_sequential_read_offset,
             "read"
         );
+        let result = self.try_read(offset, length).await;
+        if result.is_err() {
+            self.reset_prefetch_to_offset(offset);
```
vladem, line 251: Just wondering if there are any other errors (apart from `PrefetchReadError` on `try_seek`) that will trigger reset of the prefetcher after this change. But even if we don't know this set exactly, I think, resetting is a safe choice to make. I can not imagine errors which would allow further reading from the task.

passaro: Good point! 

There is only one case: failing integrity check when extending the response buffer. We were previously canceling the tasks, but not fully resetting the prefetcher state. My latest commit fixes that.

mountpoint-s3/src/prefetch.rs
```diff
@@ -318,55 +357,12 @@ where
             }
 
             let part_len = part_bytes.len() as u64;
-            let result = response.extend(part_bytes);
-            match result {
-                Ok(()) => {}
-                Err(e @ IntegrityError::ChecksumMismatch(_, _)) => {
-                    // cancel inflight tasks
-                    self.current_task = None;
```
vladem, line 326: nice that we've cleaned this up, could've been misleading why we do the cleanup twice

vladem (approved):

PR #941, opened 2024-07-15
Re-open the logging file on SIGHUP

vladem:
## Description of change

Re-open the logging file [to support rotating](https://github.com/logrotate/logrotate/issues/338#issuecomment-652499270) with no data loss. We re-open the file in our `MakeWriter` implementation to avoid reopening the file in the middle of the log line ([context](https://github.com/tokio-rs/tracing/issues/1931#issuecomment-1042364210)).

This PR is WIP (no tests and probably we'll revisit the rotation mechanism completely).

Relevant issues: None

This is not a breaking change. Though we'll need to check for possible performance regressions as we now acquire `RwLock` on each `event!` invocation (i.e. on each logging site), but that is inline with [tracing/logging-appender](https://github.com/tokio-rs/tracing/blob/master/tracing-appender/src/rolling.rs#L88) implementation.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -999,12 +999,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
-name = "cfg_aliases"
-version = "0.1.1"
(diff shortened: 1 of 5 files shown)


vladem:
Will assess demand in https://github.com/awslabs/mountpoint-s3/issues/958 and may re-open accordingly.

Issue #815, opened 2024-03-13
Add --bind option to specify address when running on a host with multiple NICs


rocketbitz:
### Tell us more about this new feature.

AWS is providing more and more instances with multiple network interfaces (in some cases, each individually capable of 100Gbps) and it would be a valuable feature to be able to bind individual mountpoint instances to each interface and maximize throughput.

For reference, [rclone has a feature](https://rclone.org/docs/#bind-string) like this already.

arsh:
Thanks @rocketbitz for opening this issue. I'll do some research internally and post here next steps.

arsh:
We realized that this will require a new feature in the Common Runtime (CRT) S3 client that mountpoint uses.

https://github.com/awslabs/aws-c-s3/issues/415

Oblynx:
The upstream issue seems to be resolved. Is this supported now?

rocketbitz:
Technically yes, but the performance is very poor. You are not able to come even close to saturating line rates on the instances that have multiple NICs with this.

PR #943, opened 2024-07-16
Add support for distributing requests over multiple NW interfaces behind feature flag

dannycjones:
## Description of change

This change adds experimental support for distributing S3 requests over multiple network interfaces. It introduces the feature behind a compile-time feature flag.

We have requests to support distributing network requests over multiple network interfaces (#815). This is especially useful on larger EC2 instances where we have a very large amount of aggregate throughput but we only leverage one of the network interfaces today. This change is one step towards addressing this, by allowing the operator of Mountpoint to specify a list of network interfaces.

The AWS CRT added (experimental) support for multiple network interfaces in https://github.com/awslabs/aws-c-s3/pull/444.

This change is currently being tested on `dl1.24xlarge` to verify the expected behavior.

Relevant issues: #815

No change to existing behavior.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -93,6 +93,7 @@ pub struct S3ClientConfig {
     max_attempts: Option<NonZeroUsize>,
     read_backpressure: bool,
     initial_read_window: usize,
+    network_interface_names: Option<Vec<String>>,
 }
 
 impl Default for S3ClientConfig {
@@ -109,6 +110,7 @@ impl Default for S3ClientConfig {
             max_attempts: None,
             read_backpressure: false,
             initial_read_window: DEFAULT_PART_SIZE,
+            network_interface_names: None,
         }
     }
 }
@@ -186,6 +188,13 @@ impl S3ClientConfig {
     #[must_use = "S3ClientConfig follows a builder pattern"]
     pub fn initial_read_window(mut self, initial_read_window: usize) -> Self {
         self.initial_read_window = initial_read_window;
+        self
+    }
+
+    /// Set a list of network interfaces to distribute S3 requests over
+    #[must_use = "S3ClientConfig follows a builder pattern"]
+    pub fn network_interface_names(mut self, network_interface_names: Vec<String>) -> Self {
+        self.network_interface_names = Some(network_interface_names);
         self
     }
 }
@@ -359,6 +368,10 @@ impl S3CrtClientInner {
             )));
         }
         client_config.part_size(config.part_size);
+
+        if let Some(network_interface_names) = config.network_interface_names {
+            client_config.network_interface_names(network_interface_names);
+        }
 
         let user_agent = config.user_agent.unwrap_or_else(|| UserAgent::new(None));
         let user_agent_header = user_agent.build();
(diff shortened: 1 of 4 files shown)


mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -61,6 +61,55 @@ pub struct ClientConfig {
...
+    network_interface_names: Option<NetworkInterfaceNames>,
+}
+
+/// This struct bundles together the list of owned strings for the network interfaces, and the
+/// cursors pointing to them.
+///
+/// This is necessary because the CRT expects a pointer to the list of cursors,
+/// and the lifetime of the cursors must meet or exceed the lifetime of the [ClientConfig].
+/// By wrapping the array of cursors in this struct, we know the cursors will be valid
+/// and we can ensure the memory will be freed when dropped at the same time as [ClientConfig].
```
jamesbornholt, line 78: I'm pretty sure the CRT [copies these strings out](https://github.com/awslabs/aws-c-s3/blob/0ab4d58ef0bd97970d43828cb6b57a3de5747343/source/s3_client.c#L597-L598), so we don't need machinery to keep them around like this forever, only until `aws_s3_client_new` is finished. That means you can probably just create the `Vec<aws_byte_cursor>` inside the `Client::new` method and throw it away once you're done.

mountpoint-s3/src/cli.rs
```diff
@@ -293,6 +293,16 @@ pub struct CliArgs {
...
+
+    #[cfg(feature = "multiple-nw-iface")]
+    #[clap(
+        long,
+        help = "One or more comma-delimited network interfaces for Mountpoint to use when accessing S3. Requires Linux 5.7+ or running as root.",
+        help_heading = CLIENT_OPTIONS_HEADER,
+        value_delimiter = ',',
+        value_name = "INTERFACE_NAMES",
+    )]
+    pub network_interfaces: Option<Vec<String>>,
```
jamesbornholt, line 305: I don't love comma-delimited arguments. I kind of think we should do this Docker-style, with multiple uses of the argument (and a shorter name):
```
--bind eth0 --bind eth1 --bind eth2
```

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -93,6 +93,7 @@ pub struct S3ClientConfig {
     max_attempts: Option<NonZeroUsize>,
     read_backpressure: bool,
     initial_read_window: usize,
+    network_interface_names: Option<Vec<String>>,
```
jamesbornholt, line 96: Is there a difference between `None` and `Some(vec![])`? I think probably not, so this might not need to be Option.

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -61,6 +61,55 @@ pub struct ClientConfig {
...
+    /// Create a new [NetworkInterfaceNamesInner].
+    pub fn new(names: Vec<String>) -> Self {
+        // Turn this into a read-only representation of the vector of strings.
+        let names = names.into_boxed_slice();
+
+        let cursors = names
+            .iter()
+            .map(|name| {
+                // SAFETY: The names are stored alongside the cursors in NetworkInterfaceNamesInner.
+                //         The lifetime of NetworkInterfaceNamesInner will always meet the lifetime of the cursors.
```
jamesbornholt, line 100: The other thing that matters is that the `String`s will never be mutated, since that could reallocate the backing memory for the string and make these cursors dangle. But as above, I think you can just do this cursor array as a temporary at construction time and then it doesn't matter.

monthonk (approved):

PR #947, opened 2024-07-18
Add `--log-rotation-frequency` CLI flag

vladem:
## Description of change

This is an alternative approach for log rotation, proposed in the [PR](https://github.com/awslabs/mountpoint-s3/pull/941). With this approach MP does not rely on external tool (e.g. `logrotate`) to do the file copying, but does it itself.

Relevant issues: None

This is not a breaking change. Though we'll need to check for possible performance regressions as we now [acquire](https://github.com/tokio-rs/tracing/blob/master/tracing-appender/src/rolling.rs#L88) RwLock on each event! invocation (i.e. on each logging site).

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1212,6 +1212,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


vladem:
Will assess demand in https://github.com/awslabs/mountpoint-s3/issues/958 and may re-open accordingly.

PR #965, opened 2024-08-01
Allow partial repeat of `readdir` response

vladem:
## Description of change
When user application gets interrupted in a `readdir` syscall the underlying chain of `readdir` fuse requests gets reset to an offset which is considered stale by Mountpoint. In that case Mountpoint still completes the interrupted `readdir` request, but kernel partially discards the response. We already cache the last response, so we can use it to serve the request which follows the interrupt. 

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/955

This is not a breaking change. Previously an error was returned, now it'll be handled properly.

---

### DIFF
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Other changes
 
-* Fix an issue where an interrupt during `readdir` syscall lead to an error. ([#965](https://github.com/awslabs/mountpoint-s3/pull/965))
+* Fix an issue where an interrupt during `readdir` syscall leads to an error. ([#965](https://github.com/awslabs/mountpoint-s3/pull/965))
 
 ## v1.13.0 (December 2, 2024)
 


mountpoint-s3/src/fs.rs
```diff
@@ -1016,18 +1016,22 @@ where
 
         let readdir_handle = dir_handle.handle.lock().await;
 
-        if offset != dir_handle.offset() {
+        if offset != dir_handle.offset() && offset > -1 {
```
dannycjones, line 1019: nit: this is probably a bit clearer

```suggestion
        if offset != dir_handle.offset() && offset >= 0 {
```

mountpoint-s3/tests/fs.rs
```diff
@@ -1482,6 +1482,45 @@ async fn test_readdir_rewind_with_local_files_only() {
...
+    let first_response = ls(&fs, dir_handle, 0, max_entries).await;
+    assert!(first_response.len() == 25);
+
+    // FUSE( 11) INTERRUPT unique RequestId(10)
+    // don't really interrupt the application, but emulate the resulting fuse requests
+
+    // FUSE( 12) READDIRPLUS fh FileHandle(1), offset 1, size 4096
+    // first response was partially discarded because of an interrupt, but the returned entries should be the same for the second request
+    let second_response = ls(&fs, dir_handle, 1, max_entries).await;
+    assert_eq!(&first_response[1..], &second_response[..]);
```
dannycjones, line 1508: I'd prefer we generalize over this (`test_case(..)`) and provide some edge cases.

In particular, I'm looking for:
- `1` as you have
- `24`, i.e right at the end of the page so we know we're not doing anything wrong on the edges

dannycjones (changes requested):
Nice!

Thanks for adding the new test case. I think it can be easily adapted to cover some edge cases (see comment).

Do you think this deserves a new changelog entry?

mountpoint-s3/src/fs.rs
```diff
@@ -1016,18 +1016,22 @@ where
```
arsh, line 1016: on line 1010 we have a special case; can you check if removing it does not break the tests?

vladem: I've added a test for using cached response after a rewind. It works as expected, as the cached response is updated after the first request after rewind completes.

After commenting out the block on [line 1010](https://github.com/awslabs/mountpoint-s3/blob/e2680a231511832472c23c086ecb1493ad73201e/mountpoint-s3/src/fs.rs#L1010C12-L1010C35) only rewind tests are failing, which is expected: 
```
test test_readdir_rewind_with_new_files::default_default_expects ... FAILED
test test_readdir_rewind_ordered ... FAILED
test test_readdir_repeat_response_partial::first_in_the_end_second_in_full ... ok
test test_readdir_then_open_cached ... ok
test test_readdir_repeat_response_partial::first_in_the_beginning_second_in_full ... ok
test test_readdir_repeat_response_partial::first_in_full_second_in_the_end ... ok
test test_readdir_rewind_unordered ... FAILED
test test_readdir_rewind_with_local_files_only ... FAILED
test test_readdir_vs_readdirplus ... ok
test test_readdir_rewind_with_new_files::s3filesystemconfig_s3_personality_s3personality_expressonezone_default_default_expects ... FAILED
test test_readdir_repeat_response_partial::first_in_full_second_in_the_beginning ... ok
```

dannycjones (commented):
Sorry, forgot to post these comments. Just a concern on outdated comments

mountpoint-s3/src/fs.rs
vladem: Decided to explicitly destroy the cache on rewind.

(After another look at it I had a suspicion that when rewind happens while cached response starts at offset 0, this response will be repeated. In an unordered case it would have lead to wrong results. It proved to be wrong, as we don't try to use cache in this case as `offset == dir_handle.offset() == 0`, but destroying the cache, while being redundant, protects us from possible bugs in future)

mountpoint-s3/src/fs.rs
vladem: made it `offset > 0` since in case `offset == 0` we do the rewind and don't want to use cache (`offset != dir_handle.offset()` will also be `false`, but still)

mountpoint-s3/CHANGELOG.md
dannycjones, line 1: We have a conflict here, needs a rebase.

mountpoint-s3/tests/fs.rs
```diff
@@ -1485,6 +1485,80 @@ async fn test_readdir_rewind_with_local_files_only() {
     assert_eq!(new_entries.len(), 3); // 1 new local file + 2 dirs (. and ..) = 3 entries
 }
 
+// Check that request with an out-of-order offset which is in bounds of previously cached response is served well.
+// This is relevant for situation when user application is interrupted in a readdir system call, which makes kernel
+// to partially discard previous response and request some entires from it again:
```
dannycjones, line 1490: typo/nit

```suggestion
// Check that request with an out-of-order offset which is in bounds of previously cached response is served well.
// This is relevant for situation when user application is interrupted in a readdir system call, which makes the
// kernel partially discard previous response and request some entries from it again:
```

dannycjones (commented):
Just rebase and some nits (since we're already updating the branch).

dannycjones:
@vladem would you mind rebasing this one?

dannycjones (approved):
LGTM, one minor nit I missed before

dannycjones (approved):
LGTM!

PR #967, opened 2024-08-02
Consolidate test credential helpers into creds modules

dannycjones:
## Description of change

Across the code base, we have many ways to assume or fetch credentials. We had different needs at different times, and so this was done in each case where needed. This ended up being quite complex.

This change brings most of those functions under a `test::common::creds` module in both crates.

The goal here is to simplify, especially to help reduce the changes in PRs like #684.

Relevant issues: #684

No breaking change, tests only.

---

### DIFF
--- a/mountpoint-s3-client/tests/common/creds.rs
+++ b/mountpoint-s3-client/tests/common/creds.rs
@@ -2,12 +2,7 @@
 //!
 //! Some methods are duplicated in `mountpoint-s3::tests::common::creds`.
 
-use aws_config::sts::AssumeRoleProvider;
-use aws_config::Region;
-use aws_credential_types::provider::ProvideCredentials;
 use aws_credential_types::Credentials;
-
-use crate::common::get_test_region;
 
 /// Detect if running on GitHub Actions (GHA) and if so,
 /// emit masking string to avoid credentials accidentally being printed.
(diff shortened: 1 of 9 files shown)


mountpoint-s3-client/tests/common/creds.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// See the [session policies section of the AWS IAM User Guide][session_policies] for more detail.
+    ///
+    /// [session_policies]: https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies.html#policies_session
+    pub async fn get_scoped_down_credentials<Policy: AsRef<str>>(policy: Policy) -> Credentials {
+        let provider = AssumeRoleProvider::builder(get_subsession_iam_role())
+            .region(Region::new(get_test_region()))
+            .session_name("test_mountpoint-s3-client")
+            .policy(policy.as_ref())
+            .build()
+            .await;
```
dannycjones, line 68: I dropped a nonce from the session name to simplify. Please let me know if its worth keeping.

monthonk: I don't think we use them so should be fine. But is it still useful to keep the session name at all? they are going to have the same value now.

dannycjones: Session name is required (and does provide a little context in things like CloudTrail in the off chance we ever wanted that).

https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html#API_AssumeRole_RequestParameters

monthonk: I see, thought it was optional.

monthonk (approved):
LGTM

PR #979, opened 2024-08-09
Wrap data cache reader with BufReader

dannycjones:
## Description of change

Opening this mainly to observe if there's any improvement, but we should probably merge anyway.

We're currently making individual system calls for reading from the chunks on disk. We should be using buffered readers. Since we're only doing this a little at the start of each 1MiB file, it's probably not that much of an impact.

Relevant issues: N/A

It doesn't change behavior but may impact performance when using the data cache.

No changelog entry required, small change to how we read the data cache.

---

### DIFF
--- a/mountpoint-s3/src/data_cache/disk_data_cache.rs
+++ b/mountpoint-s3/src/data_cache/disk_data_cache.rs
@@ -1,7 +1,7 @@
 //! Module for the on-disk data cache implementation.
 
 use std::fs;
-use std::io::{ErrorKind, Read, Seek, Write};
+use std::io::{BufReader, ErrorKind, Read, Seek, Write};
 use std::os::unix::fs::{DirBuilderExt, OpenOptionsExt};
 use std::path::{Path, PathBuf};
 use std::time::Instant;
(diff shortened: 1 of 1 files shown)


dannycjones:
As expected, nothing special in benchmark results. That being said, we should be preferring buffered reads where possible.

mountpoint-s3/src/data_cache/disk_data_cache.rs
```diff
@@ -230,11 +230,12 @@ impl DiskDataCache {
         block_idx: BlockIndex,
         block_offset: u64,
     ) -> DataCacheResult<Option<ChecksummedBytes>> {
-        let mut file = match fs::File::open(path.as_ref()) {
+        let file = match fs::File::open(path.as_ref()) {
             Ok(file) => file,
             Err(err) if err.kind() == ErrorKind::NotFound => return Ok(None),
             Err(err) => return Err(err.into()),
         };
+        let mut file = BufReader::new(file);
```
unexge, line 238: Would increasing the default buffer size of 8 KiB help here? As we know our blocks are 1 MiB, higher buffer size might be better?

dannycjones: That's a really good question, I hadn't actually considered how it decides how much to buffer and how the size of that buffer can help. I'll need to think about it a bit more and give some different values a try.

dannycjones: Thinking about this a bit more critically, maybe we shouldn't use a buffered reader at all. What we really do here is read the whole chunk in no matter what, so we should just load the whole thing into memory allocating it up front (1MiB + a little bit for the header).

passaro: From a very quick look at the `BufReader` code: https://doc.rust-lang.org/src/std/io/buffered/bufreader.rs.html#281, the small (default) buffer may actually be good enough to read the header while the 1MiB block skips it.

But agree to measure performance before changing anything.

dannycjones:
Closing this for now.

PR #980, opened 2024-08-10
Re-implement the prefetcher using backpressure mechanism

monthonk:
## Description of change

The prefetcher now uses only one `GetObject` request to fetch data in advance. This request has a range of entire object but use backpressure mechanism to control how much data it wants to fetch into the part queue instead of spawning up to two requests in parallel.

This should make the throughput more stable because previously the two request tasks could compete with each other when fetching data from S3. Also, it will be easier to control how much data we want to store in the part queue.

No, the change only for internal interface.

No

---

### DIFF
--- a/mountpoint-s3/src/checksums.rs
+++ b/mountpoint-s3/src/checksums.rs
@@ -63,7 +63,7 @@ impl ChecksummedBytes {
     /// This operation just increases the reference count and sets a few indices,
     /// so there will be no validation and the checksum will not be recomputed.
     pub fn split_off(&mut self, at: usize) -> ChecksummedBytes {
-        assert!(at < self.len());
+        assert!(at <= self.len());
 
         let start = self.range.start;
         let prefix_range = start..(start + at);
@@ -73,6 +73,27 @@ impl ChecksummedBytes {
         Self {
             buffer: self.buffer.clone(),
             range: suffix_range,
+            checksum: self.checksum,
+        }
+    }
+
+    /// Splits the checksummed bytes into two at the given index.
+    ///
+    /// Afterwards self contains elements [at, len), and the returned Bytes contains elements [0, at).
+    ///
+    /// This operation just increases the reference count and sets a few indices,
+    /// so there will be no validation and the checksum will not be recomputed.
+    pub fn split_to(&mut self, at: usize) -> ChecksummedBytes {
+        assert!(at <= self.len());
+
+        let start = self.range.start;
+        let prefix_range = start..(start + at);
+        let suffix_range = (start + at)..self.range.end;
+
+        self.range = suffix_range;
+        Self {
+            buffer: self.buffer.clone(),
+            range: prefix_range,
             checksum: self.checksum,
         }
     }
@@ -317,6 +338,26 @@ mod tests {
         assert_eq!(expected, new_checksummed_bytes.buffer);
         assert_eq!(expected_part1, checksummed_bytes.buffer_slice());
         assert_eq!(expected_part2, new_checksummed_bytes.buffer_slice());
+        assert_eq!(expected_checksum, checksummed_bytes.checksum);
+        assert_eq!(expected_checksum, new_checksummed_bytes.checksum);
+    }
+
+    #[test]
+    fn test_split_to() {
+        let split_to_at = 4;
+        let bytes = Bytes::from_static(b"some bytes");
+        let expected = bytes.clone();
+        let expected_checksum = crc32c::checksum(&expected);
+        let mut checksummed_bytes = ChecksummedBytes::new(bytes);
+
+        let mut expected_part2 = expected.clone();
+        let expected_part1 = expected_part2.split_to(split_to_at);
+        let new_checksummed_bytes = checksummed_bytes.split_to(split_to_at);
+
+        assert_eq!(expected, checksummed_bytes.buffer);
+        assert_eq!(expected, new_checksummed_bytes.buffer);
+        assert_eq!(expected_part1, new_checksummed_bytes.buffer_slice());
+        assert_eq!(expected_part2, checksummed_bytes.buffer_slice());
         assert_eq!(expected_checksum, checksummed_bytes.checksum);
         assert_eq!(expected_checksum, new_checksummed_bytes.checksum);
     }
--- a/mountpoint-s3/src/checksums.rs
+++ b/mountpoint-s3/src/checksums.rs
@@ -63,7 +63,7 @@ impl ChecksummedBytes {
     /// This operation just increases the reference count and sets a few indices,
     /// so there will be no validation and the checksum will not be recomputed.
     pub fn split_off(&mut self, at: usize) -> ChecksummedBytes {
-        assert!(at <= self.len());
+        assert!(at < self.len());
 
         let start = self.range.start;
         let prefix_range = start..(start + at);
@@ -73,27 +73,6 @@ impl ChecksummedBytes {
         Self {
             buffer: self.buffer.clone(),
             range: suffix_range,
-            checksum: self.checksum,
-        }
-    }
-
-    /// Splits the checksummed bytes into two at the given index.
-    ///
-    /// Afterwards self contains elements [at, len), and the returned Bytes contains elements [0, at).
-    ///
-    /// This operation just increases the reference count and sets a few indices,
-    /// so there will be no validation and the checksum will not be recomputed.
-    pub fn split_to(&mut self, at: usize) -> ChecksummedBytes {
-        assert!(at <= self.len());
-
-        let start = self.range.start;
-        let prefix_range = start..(start + at);
-        let suffix_range = (start + at)..self.range.end;
-
-        self.range = suffix_range;
-        Self {
-            buffer: self.buffer.clone(),
-            range: prefix_range,
             checksum: self.checksum,
         }
     }
@@ -338,26 +317,6 @@ mod tests {
         assert_eq!(expected, new_checksummed_bytes.buffer);
         assert_eq!(expected_part1, checksummed_bytes.buffer_slice());
         assert_eq!(expected_part2, new_checksummed_bytes.buffer_slice());
-        assert_eq!(expected_checksum, checksummed_bytes.checksum);
-        assert_eq!(expected_checksum, new_checksummed_bytes.checksum);
-    }
-
-    #[test]
-    fn test_split_to() {
-        let split_to_at = 4;
-        let bytes = Bytes::from_static(b"some bytes");
-        let expected = bytes.clone();
-        let expected_checksum = crc32c::checksum(&expected);
-        let mut checksummed_bytes = ChecksummedBytes::new(bytes);
-
-        let mut expected_part2 = expected.clone();
-        let expected_part1 = expected_part2.split_to(split_to_at);
-        let new_checksummed_bytes = checksummed_bytes.split_to(split_to_at);
-
-        assert_eq!(expected, checksummed_bytes.buffer);
-        assert_eq!(expected, new_checksummed_bytes.buffer);
-        assert_eq!(expected_part1, new_checksummed_bytes.buffer_slice());
-        assert_eq!(expected_part2, checksummed_bytes.buffer_slice());
         assert_eq!(expected_checksum, checksummed_bytes.checksum);
         assert_eq!(expected_checksum, new_checksummed_bytes.checksum);
     }
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -806,15 +806,14 @@ mod tests {
         };
 
         let mut get_failures = HashMap::new();
-        // We only have one request with backpressure, so we are going to inject the failure at
-        // 2nd read from that request stream.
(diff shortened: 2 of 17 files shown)


mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -80,11 +86,25 @@ where
...
+        Self {
+            client,
+            cache,
+            backpressure_limiter,
+            config,
+        }
     }
 
-    async fn get_from_cache(self, range: RequestRange, part_queue_producer: PartQueueProducer<Client::ClientError>) {
+    async fn get_from_cache(
```
passaro, line 103: It seems that we are changing how often this method is being called: before every time the prefetcher asked for more data (new `RequestTask`), we would first check the cache and fall back to the client at the first cache miss. Now new `RequestTask`s are only spawned on out-of-order reads, while sequential data is requested via backpressure. This means that a fully sequential read will switch entirely to the client after a single cache miss.

We should discuss whether this is acceptable and in any case document it.

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -229,11 +283,39 @@ where
...
+                //
+                // This is because the read window might be enough to fetch "some data" from S3 but not the entire block.
+                // For example, consider that we got a file system read request with range 2MB to 4MB and we have to start
+                // reading from block_offset=0 and block_size=5MB. The first read window might have a range up to 4MB which
+                // is enough to serve the read request but if the prefetcher is not able to read anything it cannot tell
+                // the stream to move its read window.
+                //
+                // A side effect from this is delay on the cache updating which is actually good for performance but it makes
+                // testing a bit more complicated because the cache might not be updated immediately.
+                let part_range = self
```
passaro, line 302: Why the duplication with `make_part`? I'm struggling to see what is different and why. Could we review this logic and maybe extract or document it?

monthonk: `make_part` only trim data to the block boundaries but here we want to push some data (i.e. `chunk`) to the part queue even before we have a full block. The reasons are in above comments. I will extract it to a new function.

mountpoint-s3/src/prefetch/part_stream.rs
```diff
@@ -180,36 +186,93 @@ where
...
+        let object_id = &self.config.object_id;
+        let first_read_window_end_offset = self.config.range.start() + self.config.initial_read_window_size as u64;
+
+        let part_composer = ClientPartComposer {
+            part_queue_producer,
+            object_id: object_id.clone(),
+            preferred_part_size: self.config.preferred_part_size,
+        };
+
+        // Normally, initial read window size should be very small (~1MB) so that we can serve the first read request as soon as possible,
```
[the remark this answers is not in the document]
monthonk: Will have a look at this.

mountpoint-s3/src/prefetch/task.rs
```diff
@@ -1,65 +1,82 @@
...
+
+        // We read some data out of the part queue so the read window should be moved
+        self.backpressure_controller.send_feedback(DataRead(part.len())).await?;
+
+        let next_offset = part.offset() + part.len() as u64;
+        let remaining_in_queue = self.available_offset().saturating_sub(next_offset) as usize;
+        // If the part queue is empty it means we are reading faster than the task could prefetch,
+        // so we should use larger window for the task.
+        if remaining_in_queue == 0 {
+            self.backpressure_controller.send_feedback(PartQueueStall).await?;
```
vladem, line 64: Sending `PartQueueStall` here is not as precise as we may be. Reading from the part queue on line 52 may get blocked, but we would not signal about this event to the controller.

I'd expose `is_empty` method in the PartQueue:
```
    pub fn is_empty(&self) -> bool {
        self.receiver.is_empty()
    }
```

Then call it before a read attempt and send `PartQueueStall` if it returns `true`. Moreover I would immediatelly signal about that to the limiter `self.read_window_updater.send(..)`.  This will make the system more robust as it guarantees that when there is no data in the `PartQueue`, limiter is **always** notified about that.

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -48,30 +49,35 @@ where
     where
         Client: ObjectClient + Clone + Send + Sync + 'static,
     {
-        let range = config.range.align(self.cache.block_size(), false);
```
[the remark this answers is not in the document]
monthonk: Yeah, I don't think we need to align the read window, but also don't want to remove the `RequestRange::align` right now as we might want to use it in the future.

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -278,16 +361,14 @@ where
                 &self.cache_key,
             );
             self.part_queue_producer.push(Ok(make_part(
```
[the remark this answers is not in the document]
passaro: My understanding is that `flush` is required to update the cache for the last (smaller) block. But the change you made to `compose_parts` makes sure that the data is already pushed to the queue (line 314). Am I missing something?

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -229,11 +283,39 @@ where
...
+            let mut offset = offset;
             while !body.is_empty() {
                 let remaining = (block_size as usize).saturating_sub(self.buffer.len()).min(body.len());
-                let chunk = body.split_to(remaining);
+                let chunk: ChecksummedBytes = body.split_to(remaining).into();
+
+                // We need to return some bytes to the part queue even before we can fill an entire caching block because
+                // we want to start the feedback loop for the flow-control window.
+                //
+                // This is because the read window might be enough to fetch "some data" from S3 but not the entire block.
```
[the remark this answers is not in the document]
passaro: Not sure I follow. Are we aligning to block boundaries or not?
Performance-wise, it would be a different trade-off, happy to discuss it. But we need to make sure we do not try to cache partial blocks.

mountpoint-s3/src/prefetch/caching_stream.rs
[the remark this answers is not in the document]
monthonk: Ah, you're right. We should not push to the part queue again here. Sorry, I got confused.

mountpoint-s3/src/prefetch/caching_stream.rs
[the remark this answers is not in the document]
monthonk: Sorry, I mean the request ranges are aligned to block boundaries, but the read window ranges may not. Agree, that we must not try to cache partial blocks.

mountpoint-s3/src/checksums.rs
```diff
@@ -77,6 +77,27 @@ impl ChecksummedBytes {
         }
     }
 
+    /// Splits the checksummed bytes into two at the given index.
```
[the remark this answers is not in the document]
monthonk: Actually, we don't really need it.

mountpoint-s3/src/prefetch.rs
```diff
@@ -365,74 +378,45 @@ where
...
-            if let Some(task) = self.spawn_next_request() {
-                self.future_tasks.push_back(task);
+    /// Spawn a backpressure GetObject request which has a range from current offset to the end of the file.
+    /// We will be using flow-control window to control how much data we want to download into the prefetcher.
+    fn spawn_read_backpressure_request(
+        &mut self,
+    ) -> Result<RequestTask<Client::ClientError>, PrefetchReadError<Client::ClientError>> {
+        let start = self.next_sequential_read_offset;
+        let object_size = self.size as usize;
+        let range = RequestRange::new(object_size, start, object_size);
```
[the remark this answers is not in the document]
monthonk: Not sure if that's always the case, probably will leave it as is for now.

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -129,19 +170,23 @@ where
...
         let cache_key = &self.config.object_id;
+        let first_read_window_end_offset = self.config.range.start() + self.config.initial_read_window_size as u64;
         let block_size = self.cache.block_size();
         assert!(block_size > 0);
 
         // Always request a range aligned with block boundaries (or to the end of the object).
         let block_aligned_byte_range =
             (block_range.start * block_size)..(block_range.end * block_size).min(range.object_size() as u64);
+        let request_len = (block_aligned_byte_range.end - block_aligned_byte_range.start) as usize;
+        let block_aligned_byte_range =
```
passaro, line 188: nit: do we still need the previous `block_aligned_byte_range`? could we instead just calculate the range here, or the start and len separately?

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -247,23 +314,24 @@ where
...
                 self.block_offset += block_size;
                 self.buffer = ChecksummedBytes::default();
             }
         }
+        Ok(())
+    }
 
+    /// Flush remaining data in the buffer to the cache. This can be called to write the last
+    /// block for the object.
+    fn flush(self) {
```
passaro, line 327: `flush` should not be needed anymore as a separate method. We should be able to move this logic at the end of `compose_parts`, right?

mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -309,6 +367,23 @@ fn update_cache<Cache: DataCache + Send + Sync>(
...
+    Some(Part::new(
+        object_id.clone(),
+        part_range.start(),
+        bytes.slice(trim_start..trim_end),
+    ))
+}
+
 /// Creates a Part that can be streamed to the prefetcher from the given cache block.
 /// If required, trims the block bytes to the request range.
 fn make_part(
```
passaro, line 389: minor: could we just inline this in `get_from_cache` now? in that context the assertion may become trivial and the reasoning behind `unwrap` would be clearer.

mountpoint-s3/src/prefetch/backpressure_controller.rs
```diff
@@ -0,0 +1,192 @@
...
+                self.read_window_end_offset += len as u64;
+                Some(self.read_window_end_offset)
+            } else {
+                None
+            };
+            return Ok(new_read_window_offset);
+        }
+
+        // Reaching here means there is not enough read window, so we block until it is large enough
+        while self.read_window_end_offset <= offset && self.read_window_end_offset < self.request_end_offset {
```
vladem, line 178: is `self.read_window_end_offset < self.request_end_offset` a precaution or before the last commit we actually could've get blocked, when `offset` was above the object length?

mountpoint-s3/src/prefetch/caching_stream.rs
monthonk: Agreed, moved to inline statement.

mountpoint-s3/src/prefetch/caching_stream.rs
monthonk: Yes, we can move it into `compose_parts` now.

mountpoint-s3/src/prefetch/backpressure_controller.rs
monthonk: It could be blocked after the request is completed, this check prevents that.

vladem: Ah, I see, for `read_from_request` the provided `offset` is always `<= object_size`. When `offset == object_size` we now that it won't be read cause EOF will be reached, but in other cases we want window to be `self.read_window_end_offset > offset`.

[For the caching use case](https://github.com/awslabs/mountpoint-s3/blob/ade08b63e18f21759620fc8dc75cfafde702e9cf/mountpoint-s3/src/prefetch/caching_stream.rs#L140), where we wait for increment before loading data from disk, may next `block_offset` be `>= object_size`?

monthonk: Yeah, previously we could have a problem after serving the last cache block. Now we will block until `self.read_window_end_offset` has been incremented to the end of the request then we bail out of the loop with second condition (even if the first condition is still true because `block_offset` is larger than `self.read_window_end_offset`).

monthonk: It gets blocked and that was bad, but we don't see it because the reader receives all the data they requested so they close the file and everything is dropped.

mountpoint-s3/src/prefetch/backpressure_controller.rs
```diff
@@ -122,12 +122,13 @@ impl BackpressureController {
     }
 
     // Send an increment read window request to the stream producer
-    async fn increment_read_window<E>(&self, len: usize) -> Result<(), PrefetchReadError<E>> {
+    async fn increment_read_window(&self, len: usize) {
         // This should not block since the channel is unbounded
-        self.read_window_updater
+        let _ = self
```
passaro, line 127: nit: since we are not returning the error, I'd rather use an explicit `if let Err..` to trace instead of `inspect_err`. Not blocking on this though.

passaro (approved):

vladem (approved):

PR #982, opened 2024-08-14
Make setattr silently fail on remote inodes

pawalt:
This commit changes the setattr behavior so that if an inode is remote, the operation is a no-op. This means that programs like `wget` that used to fail on our s3 mounts will now succeed, but some of their operations will not do anything.


pawalt:
Sorry folks I meant to PR this into our repo for testing :)

PR #983, opened 2024-08-19
Store bench results in S3

vladem:
## Description of change

This PR configures CI to run benchmark on schedule every 8 hours. Results are stored in the `S3_BENCH_BUCKET_NAME` bucket for further processing. 

Relevant issues: none

## Does this change impact existing behavior?

No.

## Does this change need a changelog entry in any of the crates?

No.
---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

PR #985, opened 2024-08-20
Use separate bucket for bench results, run on PRs

vladem:
## Description of change

In the previous [PR](https://github.com/awslabs/mountpoint-s3/pull/983) we configured results to be stored under a prefix of a bucket which CI does not have access to. Since a change is required a more stable approach would be to store results in a separate bucket.

This change also configures CI to store results to S3 on benches run from a PR.

Relevant issues: none

No.

No.

---


dannycjones (approved):
Nice adding on PRs too.

PR #986, opened 2024-08-21
Add runner for performing benchmarks with different configurations

dannycjones:
## Description of change

This is a runner I've written up for allowing me to experiment with different Mountpoint parameters and understand what sequential read performance can be achieved with each.

Publishing here to see if its useful to add to the main repository.

There's still a few limitations:
- It focuses only on sequential read. It would be better if it could run against all the available benchmarks we use today (at the expense of a longer runtime).
- It runs each experiment exactly once. It should be run more than once to get an accurate picture of the performance with each experiment.
- It currently requires many 100GiB objects. It doesn't need anything close to that much for the benchmark.
- It has configurations for multiple network cards, but that's currently still behind a feature flag. (Checkout `main` and build with feature flag, that works fine.)

No, this is a new script to support the team (and anyone) in performing benchmarking using FIO.

No, there's no change to behavior.

---


dannycjones:
There's quite a bit about the approach that I would like to change, plus I'd like to try and unify this and the existing regression benchmarks in some way.

dannycjones:
I'm going to keep this going in a branch alongside `main`. When investigations are complete, I'll work on introducing this to `main`.

PR #989, opened 2024-08-22
Update mountpoint-s3-crt-sys crate excludes to reduce package size

dannycjones:
## Description of change

With the latest CRT update, we would exceed the 10MiB limit imposed by crates.io.

This change takes inspiration from the [CRT Python bindings exclude list](https://github.com/awslabs/aws-crt-python/blob/main/MANIFEST.in) to remove files that are not useful for standard CRT usage. There's a few others that we identified that we can cut also.

No change.

No change, it does not change compilation output from mountpoint-s3-crt-sys.

---


mountpoint-s3-crt-sys/Cargo.toml
```diff
@@ -8,38 +8,56 @@ repository = "https://github.com/awslabs/mountpoint-s3"
...
     "crt/aws-lc/crypto/kyber/kat/*",
-    "crt/aws-lc/crypto/x509/*_test.cc",
     "crt/aws-lc/crypto/x509/test/*",
+    "crt/aws-lc/crypto/ocsp/test/*",
     "crt/aws-lc/fuzz/*",
     "crt/aws-lc/generated-src/crypto_test_data.cc",
-    "crt/aws-lc/ssl/ssl_test.cc",
-    "crt/aws-lc/ssl/test/*",
-    "crt/aws-lc/third_party/googletest/test/*",
+    "crt/aws-lc/ssl",
```
unexge, line 36: Is this okay because we're using s2n-tls?

dannycjones: I'm not 100% sure, so I've reverted this line in particular. On Linux, I believe so, yes - but it's unclear to me what would happen on other OS.

dannycjones: When excluding `crt/aws-lc/ssl/`:

```
Run cargo package -p mountpoint-s3-crt-sys --no-verify
    Updating crates.io index
   Packaging mountpoint-s3-crt-sys v0.8.0 (/home/runner/work/mountpoint-s3/mountpoint-s3/mountpoint-s3-crt-sys)
    Packaged 2139 files, 44.4MiB (8.3MiB compressed)
```

When not excluding `crt/aws-lc/ssl/`:

```
❯ cargo package -p mountpoint-s3-crt-sys --no-verify --allow-dirty
   Packaging mountpoint-s3-crt-sys v0.8.0 (/Users/djonesoa/devel/mountpoint-s3/mountpoint-s3-crt-sys)
    Packaged 2185 files, 45.7MiB (8.6MiB compressed)
```

unexge: Okay, maybe we can revisit this in future when we need to reduce the size again 😄

PR #990, opened 2024-08-22
Add `UNSTABLE_CACHE_KEY` environment variable

muddyfish:
## Description of change

Add `UNSTABLE_CACHE_KEY` environment variable.
Using `UNSTABLE_CACHE_KEY` allows users to specify a cache path disambiguator. This key can be used to ensure that cache directories are unique, even if the cache directory specified is shared.

Relevant issues: N/A

No breaking changes. This will change behaviour when the `UNSTABLE_CACHE_KEY` variable is passed.

No, this is being added as an unstable feature without documentation or promise for continued support.

---

### DIFF
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -789,8 +789,11 @@ where
         };
 
         if let Some(cache_config) = cache_config {
-            let managed_cache_dir =
-                ManagedCacheDir::new_from_parent(path).context("failed to create cache directory")?;
+            let managed_cache_dir = match env_unstable_cache_key() {
+                None => ManagedCacheDir::new_from_parent(path),
+                Some(cache_key) => ManagedCacheDir::new_from_parent_with_cache_key(path, cache_key)
+            }.context("failed to create cache directory")?;
+
             let cache = DiskDataCache::new(managed_cache_dir.as_path_buf(), cache_config);
             let prefetcher = caching_prefetch(cache, runtime, prefetcher_config);
             let mut fuse_session = create_filesystem(
@@ -956,6 +959,11 @@ fn parse_kms_key_arn(kms_key_arn: &str) -> anyhow::Result<String> {
 fn env_region() -> Option<String> {
     env::var_os("AWS_REGION").map(|val| val.to_string_lossy().into())
 }
+
+fn env_unstable_cache_key() -> Option<String> {
+    env::var_os("UNSTABLE_CACHE_KEY").map(|val| val.to_string_lossy().into())
+}
+
 
 /// Determine the region using the following sources (in order):
 ///  * `--region` flag (user-provided),
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -791,8 +791,9 @@ where
         if let Some(cache_config) = cache_config {
             let managed_cache_dir = match env_unstable_cache_key() {
                 None => ManagedCacheDir::new_from_parent(path),
-                Some(cache_key) => ManagedCacheDir::new_from_parent_with_cache_key(path, cache_key)
-            }.context("failed to create cache directory")?;
+                Some(cache_key) => ManagedCacheDir::new_from_parent_with_cache_key(path, cache_key),
+            }
+            .context("failed to create cache directory")?;
 
             let cache = DiskDataCache::new(managed_cache_dir.as_path_buf(), cache_config);
             let prefetcher = caching_prefetch(cache, runtime, prefetcher_config);
@@ -963,7 +964,6 @@ fn env_region() -> Option<String> {
 fn env_unstable_cache_key() -> Option<String> {
     env::var_os("UNSTABLE_CACHE_KEY").map(|val| val.to_string_lossy().into())
 }
-
 
 /// Determine the region using the following sources (in order):
 ///  * `--region` flag (user-provided),
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -790,8 +790,9 @@ where
         };
 
         if let Some(cache_config) = cache_config {
-            let cache_key = env_unstable_cache_key().unwrap_or(OsString::new());
-            let managed_cache_dir = ManagedCacheDir::new_from_parent_with_cache_key(path, cache_key).context("failed to create cache directory")?;
+            let cache_key = env_unstable_cache_key();
+            let managed_cache_dir = ManagedCacheDir::new_from_parent_with_cache_key(path, cache_key)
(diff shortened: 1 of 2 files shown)


mountpoint-s3/src/cli.rs
```diff
@@ -789,8 +789,12 @@ where
         };
 
         if let Some(cache_config) = cache_config {
-            let managed_cache_dir =
-                ManagedCacheDir::new_from_parent(path).context("failed to create cache directory")?;
+            let managed_cache_dir = match env_unstable_cache_key() {
```
passaro, line 792: This change muddies the role of `ManagedCacheDir`. I suggest we either:

1. Clean up and document `ManagedCacheDir` to support the added complexity,
2. Keep it as it is and use 2 instances here.

Since this is an `UNSTABLE` feature, I'm leaning towards (2).

mountpoint-s3/src/data_cache/cache_directory.rs
```diff
@@ -16,6 +16,7 @@ use thiserror::Error;
 /// Cache directory that has been created and emptied, and will be emptied when dropped.
 #[derive(Debug)]
 pub struct ManagedCacheDir {
+    outer_path: PathBuf,
```
passaro, line 19: If we want to add the extra complexity to this type, we should document it and make clear what are the 2 paths here, possibly with better names.

mountpoint-s3/src/data_cache/cache_directory.rs
```diff
@@ -30,65 +31,83 @@ pub enum ManagedCacheDirError {
...
     /// If the directory already exists, it will be deleted before being recreated.
-    pub fn new_from_parent<P: AsRef<Path>>(parent_path: P) -> Result<Self, ManagedCacheDirError> {
+    pub fn new_from_parent_with_cache_key<P: AsRef<Path>, P2: AsRef<Path>>(
+        parent_path: P,
+        cache_key: P2,
+    ) -> Result<Self, ManagedCacheDirError> {
+        let wrapping_path = parent_path.as_ref().join("mountpoint-cache");
         let managed_cache_dir = Self {
-            managed_path: parent_path.as_ref().join("mountpoint-cache"),
+            path_with_cache_key: wrapping_path.join(cache_key),
```
passaro, line 40: `join` can append nested directories or even replace with an absolute path. Should we check for it?

passaro: Or is there a more specific method?

mountpoint-s3/src/data_cache/cache_directory.rs
```diff
@@ -30,65 +31,83 @@ pub enum ManagedCacheDirError {
...
+        Self::create_dir(&self.wrapping_path)?;
+        Self::create_dir(self.as_path())
+    }
+
+    /// Create a directory, assuming the parent path exists.
+    fn create_dir(path: &Path) -> Result<(), ManagedCacheDirError> {
+        let mkdir_result = fs::DirBuilder::new().mode(0o700).create(path);
+        if let Err(mkdir_err) = mkdir_result {
+            match mkdir_err.kind() {
+                io::ErrorKind::AlreadyExists => tracing::warn!(
```
passaro, line 73: Will this always be triggered when there is no `cache_key`?

mountpoint-s3/src/data_cache/cache_directory.rs
passaro: Or rather use `Path` methods like `is_relative` and `components`

mountpoint-s3/src/data_cache/cache_directory.rs
```diff
@@ -30,65 +35,95 @@ pub enum ManagedCacheDirError {
...
     /// If the directory already exists, it will be deleted before being recreated.
-    pub fn new_from_parent<P: AsRef<Path>>(parent_path: P) -> Result<Self, ManagedCacheDirError> {
+    pub fn new_from_parent_with_cache_key(
+        parent_path: impl AsRef<Path>,
+        cache_key: Option<OsString>,
+    ) -> Result<Self, ManagedCacheDirError> {
+        let wrapping_path = parent_path.as_ref().join("mountpoint-cache");
         let managed_cache_dir = Self {
-            managed_path: parent_path.as_ref().join("mountpoint-cache"),
+            wrapping_path: wrapping_path.clone(),
```
passaro, line 44: nit: I'd suggest computing the inner path separately, e.g.

```rust
let path_with_cache_hash = match cache_key {
    None => wrapping_path.clone(),
    Some(cache_key) => wrapping_path.join(hash_cache_key(cache_key.as_bytes()))
};
let managed_cache_dir = Self {
    wrapping_path,
    path_with_cache_hash,
};
```

mountpoint-s3/src/data_cache/cache_directory.rs
```diff
@@ -29,66 +36,102 @@ pub enum ManagedCacheDirError {
...
                 _kind => return Err(ManagedCacheDirError::CleanupFailure(remove_dir_err)),
             }
         }
-        tracing::trace!(cache_subdirectory = ?self.managed_path, "cache sub-directory removal complete");
+        tracing::trace!(cache_subdirectory = ?self.mountpoint_cache_path, "cache sub-directory removal complete");
+        Ok(())
+    }
+
+    /// Create the cache sub-directory, assumes the parent path exists.
+    fn create_cache_dir(&self) -> Result<(), ManagedCacheDirError> {
```
passaro, line 75: nit: I'd rather inline this code in `new_from_parent_with_cache_key`, so all the logic to deal with the 2 levels is in a single place.

passaro (approved):
LGTM. Remaining suggestions are optional.

mountpoint-s3/src/data_cache/cache_directory.rs
```diff
@@ -46,15 +46,18 @@ impl ManagedCacheDir {
         let mountpoint_cache_path = parent_path.as_ref().join("mountpoint-cache");
         let managed_cache_path = match cache_key {
             None => mountpoint_cache_path.clone(),
-            Some(cache_key) => mountpoint_cache_path.join(hash_cache_key(cache_key.as_bytes())),
+            Some(ref cache_key) => mountpoint_cache_path.join(hash_cache_key(cache_key.as_bytes())),
```
passaro, line 49: FYI: this is fine, but we usually prefer using a reference in the match, like this:
```rust
        let managed_cache_path = match &cache_key {
            None => mountpoint_cache_path.clone(),
            Some(cache_key) => mountpoint_cache_path.join(hash_cache_key(cache_key.as_bytes())),
        };
```

passaro (approved):
LGTM

PR #991, opened 2024-08-22
Adjust read window based on used memory

vladem:
## Description of change

This is a draft PR at this point, mainly to run benchmarks.

Relevant issues: N/A

Yes, reads are impacted.

Yes, likely an entry in `mountpoint-s3` will be added.

---


mountpoint-s3/src/prefetch/part_stream.rs
```diff
@@ -24,11 +26,12 @@ pub trait ObjectPartStream {
     /// size for the parts, but implementations are allowed to ignore it.
     fn spawn_get_object_request<Client>(
         &self,
-        client: &Client,
+        client: Arc<Client>,
```
vladem, line 29: this change in the interface may not be required, but I'd like to make sure we agree on the memory limiting strategy first

Issue #987, opened 2024-08-21
Automatically adjust prefetching based on available memory


dannycjones:
### Tell us more about this new feature.

This issue tracks adding an automatic adjustment to the amount of prefetching based on available memory. Today, prefetching will fetch up to 3GiB per file handle which is fine for most cases but can use a lot of memory when many files are read concurrently.

unexge:
[Mountpoint v1.10.0](https://github.com/awslabs/mountpoint-s3/releases/tag/mountpoint-s3-1.10.0) has been released with some improvements on the prefetcher.

renanmagagnin:
[Mountpoint for Amazon S3 v1.20.0](https://github.com/awslabs/mountpoint-s3/releases/tag/mountpoint-s3-1.20.0) is now available with significant reduction in overall memory usage!

PR #999, opened 2024-09-05
Backwards seek window does not affect the read window

vladem:
## Description of change

This change fixes the situation when the `read_window_end_offset` set on the client would be more than the configured read window above the last read offset by client application. This was happening because we were accounting some bytes from the client twice -- first time it was read from the `receiver` (client) and the second time when it was read from `current_part` (which contains bytes from backwards seek window, when backwards seek occurs).

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/987

This is a part of memory control change, which affects existing behaviour but not in a breaking way.

This is a part of memory control change, which should have a changelog entry.

---

### DIFF
--- a/mountpoint-s3/src/prefetch/backpressure_controller.rs
+++ b/mountpoint-s3/src/prefetch/backpressure_controller.rs
@@ -92,9 +92,9 @@ impl BackpressureController {
             // Note, that this may come from a backwards seek, so offsets observed by this method are not necessarily ascending
             BackpressureFeedbackEvent::DataRead { offset, length } => {
                 let next_read_offset = offset + length as u64;
-                let remaining_window = self.read_window_end_offset.saturating_sub(next_read_offset);
+                let remaining_window = self.read_window_end_offset.saturating_sub(next_read_offset) as usize;
                 // Increment the read window only if the remaining window reaches some threshold i.e. half of it left.
-                if remaining_window < (self.preferred_read_window_size / 2) as u64
+                if remaining_window < (self.preferred_read_window_size / 2)
                     && self.read_window_end_offset < self.request_end_offset
                 {
                     let new_read_window_end_offset = next_read_offset
--- a/mountpoint-s3/src/prefetch/backpressure_controller.rs
+++ b/mountpoint-s3/src/prefetch/backpressure_controller.rs
@@ -8,7 +8,7 @@ use super::PrefetchReadError;
 #[derive(Debug)]
 pub enum BackpressureFeedbackEvent {
     /// An event where data with a certain length has been read out of the stream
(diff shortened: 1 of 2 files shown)


mountpoint-s3/src/prefetch/backpressure_controller.rs
```diff
@@ -8,7 +8,7 @@ use super::PrefetchReadError;
 #[derive(Debug)]
 pub enum BackpressureFeedbackEvent {
     /// An event where data with a certain length has been read out of the stream
-    DataRead(usize),
+    DataRead(u64, usize),
```
passaro, line 11: nit: let's be explicit with names and/or document
```suggestion
    DataRead { offset: u64, length: usize },
```

mountpoint-s3/src/prefetch/backpressure_controller.rs
```diff
@@ -93,21 +89,22 @@ impl BackpressureController {
     /// will ensure that the read window size is enough to read this offset and that it is always close to `preferred_read_window_size`.
     pub async fn send_feedback<E>(&mut self, event: BackpressureFeedbackEvent) -> Result<(), PrefetchReadError<E>> {
         match event {
-            BackpressureFeedbackEvent::DataRead(length) => {
-                self.next_read_offset += length as u64;
+            // Note, that this may come from a backwards seek, so offsets observed by this method are not necessarily ascending
+            BackpressureFeedbackEvent::DataRead(offset, length) => {
+                let next_read_offset = offset + length as u64;
+                let remaining_window = self.read_window_end_offset.saturating_sub(next_read_offset);
```
passaro, line 95: tiny nit: cast this to `usize` (and avoid `as u64` on the next line)

vladem: this will be an incorrect cast on 32-bit, so I'd leave it as is, if you don't mind

(not super important, as I don't think we have tests for 32-bit or if anyone using that)

vladem: From my perspective -- we shouldn't be using `usize`. It's safe if we always remember that length-s are `usize`, but event in `fuse.rs` interface [it's not](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/fuse.rs#L138). 

If you insist I can add a commit with a cast to `usize` though.

vladem: I made it uniform, I guess it a drop in the ocean to use u64 here only. And may be you're right that this is wasteful, idk.

passaro (approved):

PR #1005, opened 2024-09-09
Avoid extending part on backward seek

monthonk:
## Description of change

Currently, we combine parts from the seek window to `current_part` in the part queue whenever we seek backward which mean we also have to re-compute checksums for this combined part. It particularly affect read throughput in some use cases where backward seek rate is high. This change should improve the throughput for those use cases.

No

No, this  is a performance improvement change

---

### DIFF
--- a/mountpoint-s3/src/prefetch/part_queue.rs
+++ b/mountpoint-s3/src/prefetch/part_queue.rs
@@ -1,3 +1,4 @@
+use std::collections::VecDeque;
 use std::time::Instant;
 
 use tracing::trace;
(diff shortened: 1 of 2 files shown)


mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -62,12 +68,18 @@ impl<E: std::error::Error + Send + Sync> PartQueue<E> {
...
             Ok(current_part)
         } else {
-            // Do `try_recv` first so we can track whether the read is starved or not
-            if let Ok(part) = self.receiver.try_recv() {
+            // Read from the auxiliary queue first if it's not empty
+            if !self.front_queue.is_empty() {
+                Ok(self
+                    .front_queue
+                    .pop_front()
+                    .expect("checked above that the queue is not empty"))
```
unexge, line 76:
```suggestion
            if let Some(part) = self.front_queue.pop_front() {
                Ok(part)
```

would get rid of `.expect` call I think

mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -13,7 +14,11 @@ use crate::sync::{Arc, AsyncMutex};
 #[derive(Debug)]
 pub struct PartQueue<E: std::error::Error> {
     current_part: AsyncMutex<Option<Part>>,
-    receiver: Receiver<Result<Part, PrefetchReadError<E>>>,
+    /// The auxiliary queue that supports pushing parts to the front of the part queue in order to
+    /// allow backward seek.
+    front_queue: VecDeque<Part>,
```
passaro, line 19: Shouldn't `front_queue` actually be just a stack (`Vec`)?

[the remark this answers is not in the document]
monthonk: Agree, seems like we don't really need mutex or atomic here. I will check again and try to make `current_part` a part of `front_queue`.

passaro (approved):
Looks good!

PR #1009, opened 2024-09-12
Add tests for specifying multiple network interfaces in the S3 client

dannycjones:
## Description of change

We currently can't do binary tests where they depend on feature flags. (https://docs.rs/assert_cmd/2.0.6/assert_cmd/cargo/index.html#limitations)

Instead, I'll focus on testing how interface names are specified to the S3 client and when that returns errors.

Currently, we'll fail S3 requests and not client creation. The tests document that, it is certainly an area that desires improvement.

I was tempted to mark these 'Linux-only' using a config flag but I'm able to run them on macOS, just need to install the package for `ip` commands.

Relevant issues: #815

No, just new tests.

No.

---

### DIFF
--- a/mountpoint-s3-client/tests/network_interface_config.rs
+++ b/mountpoint-s3-client/tests/network_interface_config.rs
@@ -1,0 +1,125 @@
+#![cfg(feature = "s3_tests")]
+
+pub mod common;
+
+use test_case::test_case;
+
+use common::*;
+use mountpoint_s3_client::config::{EndpointConfig, S3ClientConfig};
+use mountpoint_s3_client::error::{HeadObjectError, ObjectClientError::ClientError, ObjectClientError::ServiceError};
+use mountpoint_s3_client::{ObjectClient, S3CrtClient};
+
+#[tokio::test]
+async fn test_empty_list() {
+    let (bucket, prefix) = get_test_bucket_and_prefix("test_empty_list");
+    let key = format!("{prefix}/no-such-key");
+
+    let interface_names = Vec::new();
+    let config = S3ClientConfig::new()
+        .endpoint_config(EndpointConfig::new(&get_test_region()))
+        .network_interface_names(interface_names);
+    let client = S3CrtClient::new(config).expect("client should create OK");
+
+    let err = client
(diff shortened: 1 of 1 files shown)


mountpoint-s3-client/tests/network_interface_config.rs
```diff
@@ -0,0 +1,125 @@
...
+    assert!(
+        matches!(err, ServiceError(HeadObjectError::NotFound)),
+        "HeadObject should 404 successfully"
+    );
+}
+
+/// This test demonstrates how the S3 client will fail today when configured with bad network interfaces.
+/// The behavior isn't great, but this documents what happens.
+///
+/// TODO: How can we get to a point where S3 client creation fails when invalid interface names are provided?
```
vladem, line 58: it would be possible if we could affect somehow the interface which is used for the request; then we could've verify validity of all provided interfaces at least on the initial `ListObjectsV2`, may be we can leverage that it's round-robin and do the initial `ListObjectsV2` multiple times defined by NiCs number? 

how does the error provided to `mount-s3` output / logs looks like, when there is a typo in interface's name?

dannycjones: >how does the error provided to mount-s3 output / logs looks like, when there is a typo in interface's name?

```

Caused by:
    0: initial ListObjectsV2 failed for bucket djonesoa-dub in region us-east-1
    1: Client error
    2: Unknown CRT error
    3: CRT error 1053: aws-c-io: AWS_IO_SOCKET_INVALID_OPTIONS, Invalid socket options.
```

You get something similar when it fails at file operation time (such as where one of the interfaces did exist) along with EIO.

dannycjones: >may be we can leverage that it's round-robin and do the initial ListObjectsV2 multiple times defined by NiCs number?

I think that could work. I was hoping not to mess too much with that right now since I'd prefer to implement this at the CRT level.

My thinking is that CRT could eagerly do the interface name to ID lookup, and propagate the bad network names immediately at client initialization.

vladem (approved):

PR #1013, opened 2024-09-13
Adjust read window based on used memory

monthonk:
## Description of change

Add a mechanism to track memory usage of the prefetcher and adjust its read window size based on available memory.

Relevant issues: https://github.com/awslabs/mountpoint-s3/issues/987

Yes, reads are impacted.

Yes, mountpoint-s3 changelog will be updated along with documentation in the following PR.

---


mountpoint-s3/src/prefetch/part_queue.rs
```diff
@@ -20,6 +22,7 @@ pub struct PartQueue<E: std::error::Error> {
     failed: bool,
     /// The total number of bytes sent to the underlying queue of `self.receiver`
     bytes_received: Arc<AtomicUsize>,
+    mem_limiter: Arc<MemoryLimiter<Client>>,
```
vladem, line 25: won't be needed if we don't account memory used by backward seeks? (my reasoning here is that this should be negligible)

mountpoint-s3/src/prefetch/backpressure_controller.rs
```diff
@@ -91,58 +115,125 @@ impl BackpressureController {
...
     }
 
-    // Try scaling up preferred read window size with a multiplier configured at initialization.
-    fn try_scaling_up(&mut self) {
+    // Scale up preferred read window size with a multiplier configured at initialization.
+    // Scaling up fails silently if there is no enough free memory to perform it.
+    fn scale_up(&mut self) {
         if self.preferred_read_window_size < self.max_read_window_size {
+            let new_read_window_size = self.preferred_read_window_size * self.read_window_size_multiplier;
+            // Also align the new read window size to the client part size
```
vladem, line 183: it took me a while to understand, why we're doing that; it's because CRT will round up the `read_window_end` to this value, consuming more memory, right? can we document this?

mountpoint-s3/src/prefetch/backpressure_controller.rs
```diff
@@ -91,58 +115,125 @@ impl BackpressureController {
...
+                .record((self.preferred_read_window_size / 1024 / 1024) as f64);
         }
     }
 }
 
+impl<Client: ObjectClient> Drop for BackpressureController<Client> {
+    fn drop(&mut self) {
+        // Free up memory we have reserved for the read window.
+        debug_assert!(self.request_end_offset >= self.next_read_offset);
+        let remaining_window = self.read_window_end_offset.saturating_sub(self.next_read_offset);
```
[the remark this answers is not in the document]
monthonk: `push_front` and read would happen in the same fuse request because that's what trigger backwards seek, so that should not happen. But if we're worried then we can exclude the used mem in the backwards window.

mountpoint-s3/src/prefetch/backpressure_controller.rs
monthonk: Actually, I would just drop this because we now use allocated memory stat from the CRT (which is more than the actual mem used) and it doesn't really matter if we round up our window to the part size.

mountpoint-s3/src/prefetch/backpressure_controller.rs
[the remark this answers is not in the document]
vladem: Sure, I see that it works today, but looks a bit fragile and I'd propose to have a mitigation here. 

As a matter of fact, if we [had pushed to the queue](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/prefetch.rs#L525) before deciding if the seek was successful, we'd be reserving without ever releasing.

The mitigation could be as I proposed (not accounting data from `PartQueue::front_queue`) or something else if it's more convenient.

mountpoint-s3/src/prefetch/backpressure_controller.rs
vladem: But the excess will end up in our `PartQueue` and will not be reported by CRT though after some point, right? Probably until `part_size` is not huge this overhead should be negligible. I am wondering how bad it'll get if `part_size` is, say, 512mb, and unaccounted data stays up in the queue.

May be we can go on with dropping this alignment for now, but add a metric in `read_from_request` counting the excess (`backpressure_limiter.read_window_end_offset - request.next().await.offset`)?

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -1208,6 +1208,24 @@ impl ObjectClient for S3CrtClient {
         }
     }
 
+    fn mem_usage_stats(&self) -> Option<MemoryUsageStats> {
+        let crt_buffer_pool_stats = self.inner.s3_client.poll_buffer_pool_usage_stats();
+        let mem_allocated = crt_buffer_pool_stats
```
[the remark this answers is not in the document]
vladem: I can see that there are implementation details of CRT which may be not interesting to all client users. My main suggestion here is to add a comment explaining what CRT metrics we're using and why. E.g. shouldn't we use `reserved` metrics?

mountpoint-s3/src/prefetch/backpressure_controller.rs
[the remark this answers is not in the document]
monthonk: The reason I wanted to push back is that we then have to decide whether the read is backwards seek at `BackpressureFeedbackEvent::DataRead` and if we don't be careful we could face the other side of the problem i.e. releasing memory more than we should and causing overflow which is also fragile in my opinion.

mountpoint-s3/src/prefetch/backpressure_controller.rs
monthonk: That's true if the part size were large then we would end up consuming much more memory than reserved.

vladem (approved):
A couple of minor comments (we may address in a separate PR).

monthonk:
@vladem thanks for reviewing! I just rebased and addressed your comments.

mountpoint-s3/src/prefetch/part_stream.rs
```diff
@@ -24,11 +26,12 @@ pub trait ObjectPartStream {
     /// size for the parts, but implementations are allowed to ignore it.
     fn spawn_get_object_request<Client>(
         &self,
-        client: &Client,
+        client: Arc<Client>,
```
vladem, line 29: nit: revert 

(@passaro's [change](https://github.com/awslabs/mountpoint-s3/pull/1032/files#diff-ff2b2b24fbf3d7ad046bb857a44d84da70bbdd714b0ea9460f5d1b8a08408effR307) will make Client cloneable, but probably may merge as is and address separately)

vladem (approved):

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -767,7 +767,9 @@ impl S3CrtClientInner {
             .set(metrics.num_requests_streaming_response as f64);
 
         // Buffer pool metrics
+        let start = Instant::now();
         let buffer_pool_stats = s3_client.poll_buffer_pool_usage_stats();
+        metrics::histogram!("s3.client.buffer_pool.get_usage_latecy_us").record(start.elapsed().as_micros() as f64);
```
dannycjones, line 772: typo:

```suggestion
        metrics::histogram!("s3.client.buffer_pool.get_usage_latency_us").record(start.elapsed().as_micros() as f64);
```

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -1208,6 +1210,13 @@ impl ObjectClient for S3CrtClient {
         }
     }
 
+    fn mem_usage_stats(&self) -> Option<BufferPoolUsageStats> {
+        let start = Instant::now();
+        let crt_buffer_pool_stats = self.inner.s3_client.poll_buffer_pool_usage_stats();
+        metrics::histogram!("s3.client.buffer_pool.get_usage_latecy_us").record(start.elapsed().as_micros() as f64);
```
dannycjones, line 1216:
```suggestion
        metrics::histogram!("s3.client.buffer_pool.get_usage_latency_us").record(start.elapsed().as_micros() as f64);
```

mountpoint-s3/src/prefetch/part_stream.rs
passaro: Is there any reason for the change? I can't see it.

monthonk: We want to use the client in `MemoryLimiter` and wrapping it in `Arc` makes it easier to do that. We were passing the client differently in a few places, for example, we use `Arc<Client>` in the [uploader](https://github.com/awslabs/mountpoint-s3/blob/6cda3049e6b0d627748c16977c97c6e4f6241645/mountpoint-s3/src/upload.rs#L27) but pass it as reference in the part stream. I just have less conflicts in my code if I pass it around as `Arc<Client>`.

passaro: My conflicting change is now merged. I solved the same problem by requiring `Client: Clone`, which is fine for `S3CrtClient` (while the mock, etc. need to be wrapped in `Arc`).

monthonk: Got it, will rebase and see if it solves the conflict.

vladem (approved):
🚀

PR #1018, opened 2024-09-17
Update mock-mount-s3 to require '--max-throughput-gbps' argument

dannycjones:
## Description of change

It was unclear to me when first using `mock-mount-s3` that it will limit throughput based on the value for `--max-throughput-gbps`, and that this will default to 10Gbps as real Mountpoint does if it can't detect the correct value.

This change requires the argument to be specified, so that the user can be forced to acknowledge this default or choose a better value for the use case.

Relevant issues: N/A

It does, but only for the mock binary. You now must specify `--max-throughput-gbps <Gbps>`.

No.

---

### DIFF
--- a/mountpoint-s3/src/bin/mock-mount-s3.rs
+++ b/mountpoint-s3/src/bin/mock-mount-s3.rs
@@ -10,7 +10,9 @@
 //!
 //! This binary is intended only for use in testing and development of Mountpoint.
 
+use anyhow::anyhow;
 use futures::executor::ThreadPool;
+
 use mountpoint_s3::cli::CliArgs;
 use mountpoint_s3::s3::S3Personality;
 use mountpoint_s3_client::mock_client::throughput_client::ThroughputMockClient;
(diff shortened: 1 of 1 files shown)


mountpoint-s3/src/bin/mock-mount-s3.rs
```diff
@@ -32,7 +34,11 @@ fn create_mock_client(args: &CliArgs) -> anyhow::Result<(ThroughputMockClient, T
 
     tracing::warn!("using mock client");
 
-    let max_throughput_gbps = args.maximum_throughput_gbps.unwrap_or(10) as f64;
+    let Some(max_throughput_gbps) = args.maximum_throughput_gbps else {
+        return Err(anyhow!(
+            "must set --max-throughput-gbps when using mock-mount-s3 binary"
```
monthonk, line 39: This doesn't match the real argument.
```suggestion
            "must set --maximum-throughput-gbps when using mock-mount-s3 binary"
```

monthonk (changes requested):
Does this binary really compile? I have to add `features = ["thread-pool"]` for the futures crate to make it work. Maybe we should also fix that.

dannycjones:
>Does this binary really compile? I have to add features = ["thread-pool"] for the futures crate to make it work. Maybe we should also fix that.

Yeah, I wasn't sure on the best practice here. I'll submit a possible fix in another PR, I think it will be an improvement.

monthonk (approved):
Sounds good! thanks Danny.

PR #1020, opened 2024-09-17
Fix bind command-line argument being ignored

dannycjones:
## Description of change

The `--bind` argument was not actually configuring the network interfaces in standard release builds of Mountpoint, as it still had the feature flag hiding the configuration.

This change fixes the issue for now. In the meantime, I'll take some time to understand how this type of issue can be avoided in future.

Relevant issues: #815

Fixes issue where `--bind` argument wasn't respected.

Added fix note in change log.

---

### DIFF
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -642,7 +642,6 @@ pub fn create_s3_client(args: &CliArgs) -> anyhow::Result<(S3CrtClient, EventLoo
     if let Some(ttl) = args.metadata_ttl {
         user_agent.key_value("mp-cache-ttl", &ttl.to_string());
     }
-    #[cfg(feature = "multiple-nw-iface")]
     if let Some(interfaces) = &args.bind {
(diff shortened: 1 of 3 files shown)


mountpoint-s3/src/cli.rs
```diff
@@ -667,7 +666,6 @@ pub fn create_s3_client(args: &CliArgs) -> anyhow::Result<(S3CrtClient, EventLoo
         .read_backpressure(true)
         .initial_read_window(initial_read_window_size)
         .user_agent(user_agent);
-    #[cfg(feature = "multiple-nw-iface")]
```
vladem, line 670: I've completely overlooked that on previous review. Is it possible to verify locally that it has the desired effect though? (e.g. compiling without any feature flags and specifying `eth0` and an inexistent interface as in your auto tests)

(also there's a conflict in changelog)

dannycjones: Yep, I verified locally (and double checked just now after merging from main).

passaro (approved):
LGTM

Issue #630, opened 2023-11-24
ls: cannot open directory '...': Transport endpoint is not connected


tchaton:
### Mountpoint for Amazon S3 version

1.1.1

### AWS Region

us-east-1

### Describe the running environment

Running on Amazon EC2

### What happened?

This is happening quite frequentally 

ls: cannot open directory `....`: Transport endpoint is not connected

### Relevant log output

```shell
The only log line I can see is the following.


2023-11-24T12:53:20.711406Z  WARN readdirplus{req=10524 ino=1 fh=1193 offset=1}: mountpoint_s3::fuse: readdirplus failed: out-of-order readdir, expected=10, actual=1
fusermount: /... not mounted
```

tchaton:
Addtitionally, we are observing a CPU spike every minute with --enable-metadata-caching --metadata-cache-ttl 60. I was hoping the listing would be lazy e.g if the users don't list or interact with the mount, no listing is done.

passaro:
About the CPU spikes: Mountpoint does not proactively refresh metadata when it expires. So it should behave just as you were expecting. I suspect that the activity you are observing is due to applications accessing the filesystem and the kernel in turn requesting updated metadata from Mountpoint.

tchaton:
Hey @passaro Let me update and give you more feedbacks.

tchaton:
But if you want to see some failures, you can do something like this.

Create 1 bucket with 1M files with size ranges from 100kb to 1GB.

And copy all the files to another bucket while maximazing the CPU usage to 100% on the machine.

```
docker run --rm -v ~/.aws:/root/.aws -v /{mount_to_bucket_1}/:/data/ peakcom/s5cmd --numworkers 64 cp /data/ s3://bucket_2
```

passaro:
@tchaton, unfortunately, I was not able to reproduce the issue with the command you suggested. It may depend on specific factors like the content of your bucket or the load on your instance.

However, my (unconfirmed) suspicion is that you are seeing the result of an out of memory issue, similar to that reported in #502. 
Would you be able to verify if your syslog contains lines similar to these (once you reproduce the `Transport endpoint is not connected` error):

```
kernel: Out of memory: Killed process 2684 (mount-s3)
systemd[1]: session-1.scope: A process of this unit has been killed by the OOM killer. 
systemd[1]: session-1.scope: Killing process 3172 (docker) with signal SIGKILL.
```

tchaton:
[...]

Issue #566, opened 2023-10-20
mountpoint-s3 has a high memory usage, resulting in an Out of memory: Kill process error in the analysis software diamond


sunl:
### Mountpoint for Amazon S3 version

mount-s3 v1.0.2

### AWS Region

cn-northwest-1

### Describe the running environment

Running in EC2(c6i.4xlarge) on Centos7 against an S3 Bucket in the same account.

### What happened?

My testing scenario is using diamond blast for genomics analysis, the command used are as follows:
```
docker run -v /data:/data 
-v /database:/data/db/RD/Metagenome/database \
--privileged --log-opt max-size=10m --log-opt max-file=3 \
dockerimage diamond blastp --threads 16 -b 3 -q /data/gene_catalogue.faa.${index} \
-d /data/db/RD/Metagenome/database/nr.dmnd -o /data/Unigenes_vs_nr_blt.txt.${index} \
--max-target-seqs 10 --evalue 0.0001 --outfmt 6 qseqid qlen sseqid stitle slen \
pident length mismatch gapopen qstart qend sstart send evalue bitscore
```
Among them, nr.dmnd is the reference database that the analysis relies on, stored in the S3 bucket. The S3 bucket is mounted to the /database directory on EC2 with mount-s3. 

One EC2 runs one task to process one sample, each task needs to run for approximately 6 hours. I simultaneously started 20 EC2 runs 20 tasks (index from 1 to 20). Among the 20 tasks, 2-3 out of 20 tasks often fail after running for over about 5 hours, and the task index that fail each time are not fixed. The following error message is observed:
```
Oct 11 15:28:26 ip-10-0-0-13 kernel: Out of memory: Kill process 3484 (diamond) score 750 or sacrifice child
Oct 11 15:28:27 ip-10-0-0-13 kernel: Out of memory: Kill process 3722 (diamond) score 750 or sacrifice child
```

Then I replaced mount-s3 with goofys, and all 20 tasks were successful every time, but the analysis time will be about 8% -10% longer than using mount-s3.

I suspect that this issue is caused by the high memory usage of mount s3. I am not sure if you have any suggestions or if you can optimize the memory usage of mount s3 in the future?

sauraank:
Hi! Can you please attach log outputs with adding the Logging options `--debug` and `--log-metrics` with mount-s3. This will help identifying the issue, if there is any, which is causing high memory usage?

sunl:
[...]

Issue #502, opened 2023-09-06
Memory usage keeps increasing on md5sum compute


shashi-banger:
### Mountpoint for Amazon S3 version

v1.0.1

### AWS Region

us-east-1

### Describe the running environment

Running on a local PC docker container. ALso experienced OOMKilled when running as a pod on AWS EKS

### What happened?

- Docker started in privileged mode
- mount of s3 bucket was successful
- When trying to execute md5sum on a 10GB file and monitoring `docker stats` memory usage keeps increasing steadily to 2GB and above
- Even tried to chunk by chunk processing using the following python code snippet

```python
def generate_file_md5(filepath):
    bufsize = 2**20 * 6
    buf = bytearray(bufsize)
    bufview = memoryview(buf)
    md5_hash = hashlib.md5()
    with open(filepath, 'rb', buffering=0) as f:
        while True:
            nread = f.readinto(bufview)
            if not nread:
                break
            md5_hash.update(bufview[:nread])
    return md5_hash.hexdigest()
```

same behaviour with above python code execution also

### Relevant log output

```shell
CONTAINER ID   NAME        CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O    PIDS
5363c739f1c3   keen_pike   20.20%    1.071GiB / 7.772GiB   13.78%    13.8GB / 108MB   0B / 270kB   16
```

shashi-banger:
Attaching debug logs:

[mountpoint-s3-2023-09-06T04-03-12Z.log](https://github.com/awslabs/mountpoint-s3/files/12532434/mountpoint-s3-2023-09-06T04-03-12Z.log)

shashi-banger:
Attaching debug logs for reference

[mountpoint-s3-2023-09-06T04-03-12Z.log](https://github.com/awslabs/mountpoint-s3/files/12532532/mountpoint-s3-2023-09-06T04-03-12Z.log)

shashi-banger:
Thank you for the response. Please check if it is worth considering a command line option or some configuration to limit the maximum prefetch size. This would allow users to set the memory limits for a container more reliably.

May be #510 fix will help. Will retry once it is fixed.

CrawX:
I second that it would be very convenient to have an option to limit the memory usage for scenarios where memory availability is limited. I understand that this will likely impact performance, but that's still better than getting OOM killed.

[...]

PR #1021, opened 2024-09-18
Add temporary way to configure amount of data prefetched per file handle

dannycjones:
## Description of change

In some use cases (such as those in multi-tenant settings), a very large number of file handles may be opened concurrently. While the default prefetching works fine for environments with fewer open file handles, we see that Mountpoint can use far too much memory otherwise.

This change introduces a temporary way to scale down the maximum amount that Mountpoint will prefetch per file handle. Today, the default is 2GiB and the new parameter can be used to reduce that.

The end goal is to make this parameter obsolete through the work in #987 to make prefetching scale down when available memory is low.

Relevant issues: #502, #566, #630, #987

No existing behavior is impacted. This allows a temporary way to override the prefetch window size.

When it is removed, it will not trigger any issue (so long as the root cause is solved on memory usage). We use an environment variable such that this can be removed without breaking deployments.

This change does not add a changelog entry as we do not recommend using this override.

---

### DIFF
--- a/mountpoint-s3/src/prefetch.rs
+++ b/mountpoint-s3/src/prefetch.rs
@@ -184,10 +184,10 @@ impl Default for PrefetcherConfig {
 /// This unstable override is expected to be removed once adaptive prefetching based on available memory is available:
 /// https://github.com/awslabs/mountpoint-s3/issues/987
 fn determine_max_read_size() -> usize {
-    let env_var_key = "UNSTABLE_MOUNTPOINT_MAX_PREFETCH_WINDOW_SIZE";
+    const ENV_VAR_KEY: &str = "UNSTABLE_MOUNTPOINT_MAX_PREFETCH_WINDOW_SIZE";
     const DEFAULT_READ_WINDOW_SIZE: usize = 2 * 1024 * 1024 * 1024;
 
-    match std::env::var_os(env_var_key) {
+    match std::env::var_os(ENV_VAR_KEY) {
         Some(val) => match val.to_string_lossy().parse() {
             Ok(val) => {
                 tracing::warn!(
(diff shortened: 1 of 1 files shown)


mountpoint-s3/src/prefetch.rs
```diff
@@ -176,6 +176,38 @@ impl Default for PrefetcherConfig {
...
 
+/// Provide the maximum read size for the prefetcher, for which there is one prefetcher per file handle.
+///
+/// This allows a way to override the prefetch window rather than using the hardcoded default within Mountpoint.
+/// We do not recommend using the override, and it may be removed at any time.
+///
+/// This unstable override is expected to be removed once adaptive prefetching based on available memory is available:
+/// https://github.com/awslabs/mountpoint-s3/issues/987
+fn determine_max_read_size() -> usize {
+    let env_var_key = "UNSTABLE_MOUNTPOINT_MAX_PREFETCH_WINDOW_SIZE";
```
passaro, line 187: nit: const?

mountpoint-s3/src/prefetch.rs
```diff
@@ -176,6 +176,38 @@ impl Default for PrefetcherConfig {
...
+    const DEFAULT_READ_WINDOW_SIZE: usize = 2 * 1024 * 1024 * 1024;
+
+    match std::env::var_os(env_var_key) {
+        Some(val) => match val.to_string_lossy().parse() {
+            Ok(val) => {
```
passaro, line 192: Do we need to set a lower bound? Have you checked what happens with 0? Or with a value smaller than part size or request size?

dannycjones: Checking with 0, it does appear to continue progressing. As discussed, it's likely that the read window begins at 1MiB+128KiB and then increments by this amount because any scale up would always be bigger than 0.

passaro (approved):
LGTM

PR #1022, opened 2024-09-18
Add support for concurrent downloads to prefetch_benchmark example

passaro:
## Description of change

Add a new optional `--downloads <N>` flag to the `client_benchmark` example binary in `mountpoint-s3-client`, which controls the number of concurrent downloads to perform. If omitted, it defaults to `N = 1` which reproduces the previous behavior.

Additional changes:
* Performs an initial HEAD request to determine the object size (and remove the `--size` flag)
* Introduce a `--read-size` flag to configure the read buffer size

No.

No. The change only affects one of the example binaries.
---

### DIFF
--- a/mountpoint-s3/examples/prefetch_benchmark.rs
+++ b/mountpoint-s3/examples/prefetch_benchmark.rs
@@ -1,3 +1,4 @@
+use std::str::FromStr;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::Instant;
@@ -7,7 +8,7 @@ use futures::executor::{block_on, ThreadPool};
 use mountpoint_s3::prefetch::{default_prefetch, Prefetch, PrefetchResult};
 use mountpoint_s3_client::config::{EndpointConfig, S3ClientConfig};
 use mountpoint_s3_client::types::ETag;
-use mountpoint_s3_client::S3CrtClient;
+use mountpoint_s3_client::{ObjectClient, S3CrtClient};
 use mountpoint_s3_crt::common::rust_log_adapter::RustLogAdapter;
 use tracing_subscriber::fmt::Subscriber;
 use tracing_subscriber::util::SubscriberInitExt;
@@ -32,7 +33,6 @@ fn main() {
         .about("Download a single key from S3 and ignore its contents")
         .arg(Arg::new("bucket").required(true))
         .arg(Arg::new("key").required(true))
-        .arg(Arg::new("size").required(true))
         .arg(
             Arg::new("throughput-target-gbps")
                 .long("throughput-target-gbps")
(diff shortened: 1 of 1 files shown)


mountpoint-s3/examples/prefetch_benchmark.rs
```diff
@@ -78,28 +92,38 @@ fn main() {
...
     for i in 0..iterations.unwrap_or(1) {
-        let runtime = ThreadPool::builder().pool_size(1).create().unwrap();
+        let runtime = client.event_loop_group();
         let manager = default_prefetch(runtime, Default::default());
-        let received_size = Arc::new(AtomicU64::new(0));
+        let received_bytes = Arc::new(AtomicU64::new(0));
 
         let start = Instant::now();
 
-        let mut request = manager.prefetch(client.clone(), bucket, key, size, ETag::for_tests());
```
dannycjones, line 88: um, why was `ETag::for_tests()` working in the first place? 😄 (not blocking)

mountpoint-s3/examples/prefetch_benchmark.rs
```diff
@@ -78,28 +92,38 @@ fn main() {
     }
     let client = Arc::new(S3CrtClient::new(config).expect("couldn't create client"));
 
+    let head_object_result = block_on(client.head_object(bucket, key)).expect("HeadObject failed");
+    let size = head_object_result.object.size;
+    let etag = ETag::from_str(&head_object_result.object.etag).unwrap();
+
     for i in 0..iterations.unwrap_or(1) {
-        let runtime = ThreadPool::builder().pool_size(1).create().unwrap();
+        let runtime = client.event_loop_group();
```
dannycjones, line 100: Why do we move this from a separate thread pool onto the CRT's event loop? This doesn't match what we do with Mountpoint.

dannycjones (commented):
One question on moving from a separate thread pool onto CRT's event loop. Otherwise, LGTM.

mountpoint-s3/examples/prefetch_benchmark.rs
passaro: It wasn't.

dannycjones: Well, that solves that one :)

mountpoint-s3/examples/prefetch_benchmark.rs
dannycjones: My mistake, I was thinking we were using it for the prefetch reads but we're passing this into the prefetcher _which does match what we're doing with MP_.

dannycjones (approved):
Ship it!

mountpoint-s3/examples/prefetch_benchmark.rs
passaro: In fact, this is what we do in Mountpoint: we get the [event loop from the client](https://github.com/awslabs/mountpoint-s3/blob/ed4735d71af0432491a361912da747ccaf39a21b/mountpoint-s3/src/cli.rs#L693) and use it to [initialize the prefetcher](https://github.com/awslabs/mountpoint-s3/blob/ed4735d71af0432491a361912da747ccaf39a21b/mountpoint-s3/src/cli.rs#L812).

PR #1023, opened 2024-09-18
Build dockerfile.source from source

muddyfish:
## Description of change

- Dockerfile.source now builds based on `centos:7`
- Add `/` to path so you can run `mount-s3` without a /
- Adds `build-image` and `push-image` targets to makefile

Result from `Dockerfile.source` now builds from Centos rather than AL2023, so has the same support as our release builds.

No. `Dockerfile.source` is used as an example rather than an end-user tool

---


dannycjones:
I think we decided to keep this targeting Amazon Linux 2023, and where we need to perform builds with wider compatibility we'll continue to use the scripts in `packaging/`.

PR #1029, opened 2024-09-20
Make cache writes async

passaro:
## Description of change

Spawn separate tasks to update cache blocks, rather than writing synchronously and blocking the read path.

No

No

---

### DIFF
--- a/mountpoint-s3/src/data_cache/in_memory_data_cache.rs
+++ b/mountpoint-s3/src/data_cache/in_memory_data_cache.rs
@@ -7,11 +7,11 @@ use async_trait::async_trait;
 
 use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheError, DataCacheResult};
 use crate::object::ObjectId;
-use crate::sync::AsyncRwLock;
+use crate::sync::RwLock;
 
 /// Simple in-memory (RAM) implementation of [DataCache]. Recommended for use in testing only.
 pub struct InMemoryDataCache {
-    data: AsyncRwLock<HashMap<ObjectId, HashMap<BlockIndex, ChecksummedBytes>>>,
+    data: RwLock<HashMap<ObjectId, HashMap<BlockIndex, ChecksummedBytes>>>,
     block_size: u64,
 }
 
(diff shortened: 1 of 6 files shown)


mountpoint-s3/src/prefetch/caching_stream.rs
```diff
@@ -297,64 +309,51 @@ where
...
-        Err(error) => {
-            warn!(key=?object_id, block_index, ?error, "failed to update cache");
-        }
-    };
-    metrics::histogram!("prefetch.cache_update_duration_us").record(start.elapsed().as_micros() as f64);
+    fn update_cache(&self, block: ChecksummedBytes, block_index: u64, block_offset: u64, object_id: &ObjectId) {
+        let object_id = object_id.clone();
+        let cache = self.cache.clone();
+        self.runtime
+            .spawn(async move {
```
vladem, line 345: Cool, so now we don't block reads on cache updates. I think we may increase cache misses for some workloads, in particular those which try to read a block twice through the same file handle in a short period of time. Is it worth an entry in changelog?

mountpoint-s3/src/data_cache/in_memory_data_cache.rs
```diff
@@ -3,13 +3,15 @@
...
+
 use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheError, DataCacheResult};
 use crate::object::ObjectId;
-use crate::sync::RwLock;
+use crate::sync::AsyncRwLock;
 
 /// Simple in-memory (RAM) implementation of [DataCache]. Recommended for use in testing only.
 pub struct InMemoryDataCache {
-    data: RwLock<HashMap<ObjectId, HashMap<BlockIndex, ChecksummedBytes>>>,
+    data: AsyncRwLock<HashMap<ObjectId, HashMap<BlockIndex, ChecksummedBytes>>>,
```
vladem, line 14: not sure if this change adds any value [if we don't hold the lock across await points ](https://docs.rs/async-lock/latest/async_lock/#relationship-with-stdsync)

passaro: You are correct! Reverted.

vladem (approved):

PR #1033, opened 2024-09-25
Fix resource utilization check in the CI

monthonk:
## Description of change

Our CI workflows work by checking out the code from branch `gh-pages` to retrieve benchmark results from previous commits and compare them to values in the current run. However, the resource utilization check was done after the benchmark result check which already has pulled in the branch `gh-pages` resulting in errors because the branch already exists. This change fixes that.

No

No, only CI changes

---


vladem (approved):
it's so weird that fetch fails, as the proceeding push originates from the same local repo; but let's try this out!

PR #1034, opened 2024-09-25
Allow PR checks to be run against any base branch

dannycjones:
## Description of change

I've been working in a fork and found that CI wasn't running in the PRs against my feature branch. This meant the code wasn't actually buildable since I'd missed including a file into the PR.

This change will cause all PRs to run integration tests and benchmarks.

We may want to allow a way to disable running actions in the future, but for now we just expand this to cover all base branches.

I'm also leaving `push` events out of scope, although we could consider a similar change. This may have more of an impact on community forks though so I'll leave it as is for now.

It impacts CI only. Pull requests not against `main` as the base branch will now run benchmarks and integration tests.

No.

---

### DIFF
--- a/.github/workflows/bench_pr.yml
+++ b/.github/workflows/bench_pr.yml
@@ -2,7 +2,6 @@ name: Benchmarks (PR)
 
 on:
   pull_request_target:
-    branches: [ "main" ]
     types: [ labeled, opened, reopened, synchronize ]
 
 permissions:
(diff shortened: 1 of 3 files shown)


passaro:
Thoughts on only extending to feature branches? e.g. `feature/**`

dannycjones:
> Thoughts on only extending to feature branches? e.g. `feature/**`

I don't really have an opinion here, although I'd be inclined to cover additional branch prefixes like `private-feature/**`.

passaro (approved):

PR #1039, opened 2024-09-30
Add option to pass memory limit to CRT.

adpeace:
The CRT memory limit can be performance-impacting and this allows it to be specified on the command-line (default is based on hardware and the target throughput specified, otherwise).

**Note**: This PR is intended for perf-benchmark-working-branch.

## Description of change

Add crt-mem-limit option to the mount-s3 CLI.  Pass this through to the C CRT client.

Use a PR to the branch as a mechanism to get another pair of eyes/knowledge share, but there is at least one TODO before this could be merged to mainline (called out in a comment), and we should make sure directionally we want to expose this parameter directly in this way.

No, the default behaviour should be unchanged.  Verified that not passing the argument gives a default assignment of the memory limit.

Not until merged to main, then, unsure.

---

### DIFF
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -325,6 +325,14 @@ pub struct CliArgs {
         value_name = "NETWORK_INTERFACE",
     )]
     pub bind: Option<Vec<String>>,
+
+    #[clap(
+        long,
+        help = "Set memory limit for CRT bufferpool",
+        help_heading = CLIENT_OPTIONS_HEADER,
+        value_name = "MiB",
+    )]
+    pub crt_mem_limit_mib: Option<u64>,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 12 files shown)


mountpoint-s3/src/cli.rs
```diff
@@ -325,6 +325,15 @@ pub struct CliArgs {
...
     pub bind: Option<Vec<String>>,
+
+    #[clap(
+        long,
+        help = "Set memory limit for CRT bufferpool, or 0 to use CRT default",
+        help_heading = CLIENT_OPTIONS_HEADER,
+        value_name = "MiB",
+        default_value = "0",
+    )]
+    pub crt_mem_limit: u64,
```
oleg-lvovitch-aws, line 336: Two thoughts:
- Can we call this `crt_mem_limt_mb` so the code is easier to read (I had to think about `1024*1024` below)
- Should this be an Option? I don't know if making "0" to mean the default is a great idea

oleg-lvovitch-aws (commented):
This looks generally sensible mechanically, but I worry that this is an additional publicly facing knob for MP that controls something very internal. I can't think of a legitimate customer case where we want to control just the CRT memory. 

I understand why we need this internally - feels like we want to compile this under a flag, probably. Making it a publicly facing knob would likely be just confusing to the customers, particularly given that  we already expose our own memory management mechanism

mountpoint-s3/src/cli.rs
adpeace: Yes, reasonable points, I will update the PR on both counts, thanks.

PR #1041, opened 2024-09-30
Add random 6-character suffix to log file names

dannycjones:
## Description of change

Before this change, Mountpoint can write to the same log file if multiple Mountpoint processes spawn within the same second. This can cause log entries to be interleaved and reduces the effectiveness of debug logging since its impossible to tell which log entries relate to which Mountpoint process.

This change updates the log file creation logic to now include a six-character suffix between the timestamp and the `.log` suffix. An example of a new log file name: `mountpoint-s3-2024-10-02T10-26-57Z.961grx.log`.

We update the log config logic so that it can be defined before the fork, and then persists after fork. We know this is safe as it contains no state unsafe to share over forks.

This change also updates the logging documentation to indicate that log file names could always change in the future and shouldn't be considered stable. It also clarifies this behavior, since customers may only infrequently run into this scenario.

Relevant issues: N/A

### Alternatives considered

- Include PID in log entries.
  - This would add additional bytes/noise to all logs to cover a case which for most customers will not occur.
  - In background mode, what is 'the PID of Mountpoint'? It should probably be the child process ID.
- Include subseconds in the file name.
  - This one doesn't really have much advantage over always include the PID, and still leaves us open to the (hopefully far less likely) possibility of two Mountpoints spawning in the same subsecond.

Yes. We now include a random suffix in the name of log files.

Before: `mountpoint-s3-2024-10-02T10-26-57Z.log`
After: `mountpoint-s3-2024-10-02T10-26-57Z.961grx.log`

Yes, the main Mountpoint change log has been updated.

---


dannycjones:
I was just testing this, it doesn't cover running in the background yet.

mountpoint-s3/src/cli.rs
```diff
@@ -501,6 +477,36 @@ impl CliArgs {
...
+
+    let log_file = cli_args.log_directory.as_ref().map(|dir| prepare_log_file_name(dir));
+
+    LoggingConfig {
+        log_file,
+        log_to_stdout: cli_args.foreground,
+        default_filter,
+    }
+}
+
```
dannycjones, line 509: I've moved this out of the `impl CliArgs` because it produces a different return value each time (due to randomness). I hope moving it out makes this clearer (along with RustDoc).

dannycjones: Sounds good, that's much nicer. Updated.

passaro (approved):

PR #1046, opened 2024-10-01
Add support for single PutObject in mountpoint-s3-client

passaro:
## Description of change

Re-introduce support for a single PutObject request in `mountpoint-s3-client`. The request is exposed with the name `put_object_single`, while the existing `put_object` method performs a multi-part upload. 

No, it adds a new method.

Yes, it will require an entry for `mountpoint-s3-client` and possibly for `mountpoint-s3-crt`.

---

### DIFF
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -344,12 +344,12 @@ impl PutObjectParams {
     }
 }
 
-/// How CRC32c checksums are used for parts of a multi-part PutObject request
+/// How CRC32c checksums are used for single PutObject or parts of a multi-part PutObject request
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
 pub enum PutObjectTrailingChecksums {
-    /// Checksums are computed, passed to upload review, and also sent to S3
+    /// Checksums are computed and sent to S3. Also passed to upload review for multi-part PutObject
     Enabled,
-    /// Checksums are computed, passed to upload review, but not sent to S3
+    /// Only for multi-part PutObject: checksums are computed and passed to upload review, but not sent to S3. Equivalent to [Disabled] for single PutObject
     ReviewOnly,
     /// Checksums are not computed on the client side
     #[default]
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -134,6 +134,15 @@ pub trait ObjectClient {
         params: &PutObjectParams,
     ) -> ObjectClientResult<Self::PutObjectRequest, PutObjectError, Self::ClientError>;
 
+    /// Put an object into the object store.
+    async fn put_object_single<'a>(
+        &self,
+        bucket: &str,
+        key: &str,
+        params: &PutObjectParams,
+        contents: impl AsRef<[u8]> + Send + 'a,
+    ) -> ObjectClientResult<PutObjectResult, PutObjectError, Self::ClientError>;
+
     /// Retrieves all the metadata from an object without returning the object contents.
     async fn get_object_attributes(
         &self,
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -139,7 +139,7 @@ pub trait ObjectClient {
         &self,
         bucket: &str,
         key: &str,
-        params: &PutObjectParams,
+        params: &PutObjectSingleParams,
         contents: impl AsRef<[u8]> + Send + 'a,
     ) -> ObjectClientResult<PutObjectResult, PutObjectError, Self::ClientError>;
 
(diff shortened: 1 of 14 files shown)


mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -109,6 +82,87 @@ impl S3CrtClient {
...
+    pub(super) async fn put_object_single<'a>(
+        &self,
+        bucket: &str,
+        key: &str,
+        params: &PutObjectParams,
+        contents: impl AsRef<[u8]> + Send + 'a,
+    ) -> ObjectClientResult<PutObjectResult, PutObjectError, S3RequestError> {
+        let span = request_span!(self.inner, "put_object_single", bucket, key);
+        let start_time = Instant::now();
+        let response_headers: Arc<Mutex<Option<Headers>>> = Default::default();
```
monthonk, line 95: It was a bit hard to understand why we need a lock here. Seems like we need the CRT to write response headers to this variable so we can extract the values later on a different thread. Probably worth adding a comment.

mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -109,6 +82,87 @@ impl S3CrtClient {
...
+            .map_err(S3RequestError::construction_failure)?;
+
+        let key = format!("/{key}");
+        message
+            .set_request_path(&key)
+            .map_err(S3RequestError::construction_failure)?;
+
+        let checksum_config = match params.trailing_checksums {
+            PutObjectTrailingChecksums::Enabled => Some(ChecksumConfig::trailing_crc32c()),
+            PutObjectTrailingChecksums::ReviewOnly => Some(ChecksumConfig::upload_review_crc32c()),
```
monthonk, line 142: How does `ReviewOnly` work with put_object single? Is it just no-op?

mountpoint-s3-crt/src/io/stream.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::common::allocator::Allocator;
+
+    #[test]
+    fn test_slice_cursor() {
```
monthonk, line 168: Can we have another test that does partial read from the stream and verify it doesn't end prematurely?

monthonk (commented):
Looks good overall, just have a few minor comments.

mountpoint-s3-client/src/s3_crt_client/put_object.rs
passaro: Yes. Not great, but the only alternative I can think of is using a different params type. Not sure if worth it.

mountpoint-s3-crt/src/io/stream.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// [InputStream] cannot outlive the slice (enforced by a lifetime restriction on the [InputStream].
+    pub fn new_from_slice(allocator: &Allocator, buffer: &'a [u8]) -> Result<Self, Error> {
+        let cursor = aws_byte_cursor {
+            len: buffer.len(),
+            ptr: buffer.as_ptr() as *mut u8,
+        };
+
+        // SAFETY: allocator is a valid aws_allocator. `Self` has a lifetime of 'a, so Rust
+        // will ensure that the return value from this function doesn't out live the buffer.
+        // We need to make sure in things that consume InputStream that we don't accidentally
```
vladem, line 85: could you please describe a wrong usage of `InputStream`? is it possible to make an attempt to read from a destroyed `buffer` from rust code using the interface defined below?

(no questions regarding `put_object_single` impl, as I believe we rely on the fact that CRT won't access the stream after invoking `finish_callback`, nor will it access the `aws_http_message` holding the pointer to it)

vladem (commented):
LGTM, a couple of questions.

mountpoint-s3-client/src/object_client.rs
```diff
@@ -134,6 +134,15 @@ pub trait ObjectClient {
         params: &PutObjectParams,
     ) -> ObjectClientResult<Self::PutObjectRequest, PutObjectError, Self::ClientError>;
 
+    /// Put an object into the object store.
+    async fn put_object_single<'a>(
+        &self,
+        bucket: &str,
+        key: &str,
+        params: &PutObjectParams,
```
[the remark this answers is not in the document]
muddyfish: I was planning on adding these to `PutObjectParams` and making it work for both

mountpoint-s3-client/src/s3_crt_client/put_object.rs
monthonk: I think it's fine to use the same type, but we should update the rustdoc on `PutObjectTrailingChecksums` enum.

monthonk (approved):
LGTM

mountpoint-s3-client/src/object_client.rs
```diff
@@ -356,6 +365,68 @@ pub type UploadReviewPart = mountpoint_s3_crt::s3::client::UploadReviewPart;
...
+        self.ssekms_key_id = value;
+        self
+    }
+}
+
+/// A checksum used by the object client for integrity checks on uploads.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub enum UploadChecksum {
+    Crc32c(crate::checksums::Crc32c),
```
vladem, line 418: a bit verbose, `Crc32c(checksums::Crc32c)` could be better

vladem (approved):
Looks good!

PR #1057, opened 2024-10-14
Return the new object ETag in PutObjectResult

passaro:
## Description of change

On success, both `put_object` (MPU) and `put_object_single` return the ETag of the newly uploaded object. 

No.

Yes. Changelogs will be updated in a separate change (including changes from #1046).

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client/put_object.rs
+++ b/mountpoint-s3-client/src/s3_crt_client/put_object.rs
@@ -1,3 +1,4 @@
+use std::ffi::OsString;
 use std::sync::{Arc, Mutex};
 use std::time::Instant;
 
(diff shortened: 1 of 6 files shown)


mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -254,16 +247,46 @@ fn try_get_header_value(headers: &Headers, key: &str) -> Option<String> {
...
         sse_type: try_get_header_value(&response_headers, SSE_TYPE_HEADER_NAME),
         sse_kms_key_id: try_get_header_value(&response_headers, SSE_KEY_ID_HEADER_NAME),
-    }
+    })
+}
+
+/// Creates `on_headers` callback that will send the response headers to the matching `Receiver`.
+fn response_headers_handler() -> (impl FnMut(&Headers, i32), Receiver<Headers>) {
+    let (response_headers_sender, response_headers) = oneshot::channel();
+    let mut response_headers_sender = Some(response_headers_sender);
```
monthonk, line 283: It took me a while to understand why do we need this. Can we add some comments?

As far as I know, `on_headers` is FnMut because meta request can call `on_headers` callback multiple times depending on request type, but for `put_object` and `put_object_single` we know that it only get called at completion. Is that correct?

mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -233,7 +226,7 @@ pub struct S3PutObjectRequest {
     start_time: Instant,
     total_bytes: u64,
     /// Headers of the CompleteMultipartUpload response, available after the request was finished
-    response_headers: Arc<Mutex<Option<Headers>>>,
+    response_headers: Receiver<Headers>,
```
monthonk, line 229: I would prefer `response_headers_receiver` and update the comment.

mountpoint-s3-client/src/s3_crt_client/put_object.rs
passaro: That's correct. Will add comment.

mountpoint-s3-client/src/s3_crt_client/put_object.rs
passaro: Pushing back a little on this. A `Receiver` is a future and that's how it is used here. But happy to clarify in the comment.

monthonk: It's just not clear from the naming that we have to call `.await` to really get the headers, but not blocking on it.

monthonk (approved):
LGTM! The failed test doesn't seem to relate to the change.

PR #1058, opened 2024-10-14
Remove use of ObjectInfo in S3 client HeadObject response

dannycjones:
## Description of change

Before this change, `head_object` and `list_objects` shared the same struct `ObjectInfo`. The data we returned from these two operations was common, and so it worked well. Now we want to add information that is specific to the two operations.

This change leaves `ObjectInfo` in place for `list_objects`, aligning with the [Object](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html) structure in the Rest XML API.

This change is in preparation to surfacing checksum algorithm from ListObjectsV2 requests as well as surfacing checksum values in HeadObject requests.

Relevant issues: N/A

There is no breaking changes to Mountpoint's file system semantics.

This is a breaking change to direct consumers of Mountpoint's S3 client. The `head_object` return type `HeadObjectResult` will no longer have an `object` field of type `ObjectInfo`. Accesses to those fields should now be on the `HeadObjectResult` struct itself.

Yes. A change log entry is added to the `mountpoint-s3-client` crate.

---

### DIFF
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -219,7 +219,7 @@ pub struct HeadObjectResult {
     pub last_modified: OffsetDateTime,
 
     /// Entity tag of this object.
-    pub etag: String,
+    pub etag: ETag,
 
     /// Storage class for this object.
     ///
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -210,11 +210,28 @@ pub enum ListObjectsError {
 #[derive(Debug)]
 #[non_exhaustive]
 pub struct HeadObjectResult {
-    /// The name of the bcuket
-    pub bucket: String,
+    /// Size of the object in bytes.
+    ///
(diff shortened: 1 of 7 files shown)


mountpoint-s3-client/src/object_client.rs
```diff
@@ -210,11 +210,28 @@ pub enum ListObjectsError {
...
+    /// Size of the object in bytes.
+    ///
+    /// Refers to the `Content-Length` HTTP header for HeadObject.
+    pub size: u64,
+
+    /// The time this object was last modified.
+    pub last_modified: OffsetDateTime,
+
+    /// Entity tag of this object.
+    pub etag: String,
```
passaro, line 222: Should we use `ETag` here?

dannycjones: We could. I didn't want to diverge from what we have in `ObjectInfo`, but maybe we should go ahead and update both.

I'll update just `HeadObjectResult` for now, and can follow up with a PR for `ObjectInfo` (for `list_objects`).

passaro (approved):
LGTM

PR #1064, opened 2024-10-15
Multilevel cache

vladem:
## Description of change

Allow using both caches when `--cache-express <bucket> --cache <directory>` options are specified, local cache is queried first.

Relevant issues: No

No.

Yes, will add in one of the future PRs.

---

### DIFF
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -58,7 +58,7 @@ lazy_static! {
     static ref RAMP_BYTES: Vec<u8> = ramp_bytes(0, RAMP_BUFFER_SIZE + RAMP_MODULUS);
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
 pub struct MockClientConfig {
     /// The bucket name this client will connect to
     pub bucket: String,
@@ -74,7 +74,7 @@ pub struct MockClientConfig {
 
 /// A mock implementation of an object client that we can manually add objects to, and then query
 /// via the [ObjectClient] APIs.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct MockClient {
     config: MockClientConfig,
     objects: Arc<RwLock<BTreeMap<String, MockObject>>>,
@@ -105,6 +105,16 @@ impl MockClient {
     /// Remove object for the mock client's bucket
     pub fn remove_object(&self, key: &str) {
         self.objects.write().unwrap().remove(key);
+    }
+
+    /// Remove all objects for the mock client's bucket
+    pub fn remove_all_objects(&self) {
+        self.objects.write().unwrap().clear();
+    }
+
+    /// Number of objects in the mock client's bucket
+    pub fn object_count(&self) -> usize {
+        self.objects.write().unwrap().len()
     }
 
     /// Returns `true` if this mock client's bucket contains the specified key
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -5,10 +5,11 @@ use std::io::{Read, Write};
 use std::num::NonZeroUsize;
 use std::os::fd::AsRawFd;
 use std::path::{Path, PathBuf};
+use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::{anyhow, Context as _};
-use clap::{value_parser, Parser, ValueEnum};
+use clap::{value_parser, ArgGroup, Parser, ValueEnum};
 use fuser::{MountOption, Session};
 use futures::task::Spawn;
 use mountpoint_s3_client::config::{AddressingStyle, EndpointConfig, S3ClientAuthConfig, S3ClientConfig};
@@ -26,7 +27,9 @@ use nix::unistd::ForkResult;
 use regex::Regex;
 use sysinfo::{RefreshKind, System};
 
-use crate::data_cache::{CacheLimit, DiskDataCache, DiskDataCacheConfig, ExpressDataCache, ManagedCacheDir};
+use crate::data_cache::{
+    CacheLimit, DiskDataCache, DiskDataCacheConfig, ExpressDataCache, ManagedCacheDir, MultilevelDataCache,
+};
 use crate::fs::{CacheConfig, ServerSideEncryption, TimeToLive};
 use crate::fuse::session::FuseSession;
 use crate::fuse::S3FuseFilesystem;
@@ -46,7 +49,15 @@ const CACHING_OPTIONS_HEADER: &str = "Caching options";
 const ADVANCED_OPTIONS_HEADER: &str = "Advanced options";
 
 #[derive(Parser, Debug)]
-#[clap(name = "mount-s3", about = "Mountpoint for Amazon S3", version = build_info::FULL_VERSION)]
+#[clap(
+    name = "mount-s3",
+    about = "Mountpoint for Amazon S3",
+    version = build_info::FULL_VERSION,
+    group(
+        ArgGroup::new("cache_group")
+            .multiple(true),
+    ),
+)]
 pub struct CliArgs {
     #[clap(help = "Name of bucket to mount", value_parser = parse_bucket_name)]
     pub bucket_name: String,
@@ -298,10 +309,10 @@ pub struct CliArgs {
     #[cfg(feature = "block_size")]
     #[clap(
         long,
-        help = "Size of a cache block in KiB [Default: 1024 (1 MiB) for disk cache, 512 (512 KiB) for S3 Express cache]",
+        help = "Size of a cache block in KiB [Default: 1024 (1 MiB) for disk cache and for S3 Express cache]",
         help_heading = CACHING_OPTIONS_HEADER,
         value_name = "KiB",
-        requires = "cache_group"
+        requires = "cache_group",
     )]
     pub cache_block_size: Option<u64>,
 
@@ -423,10 +434,7 @@ impl CliArgs {
         if let Some(kib) = self.cache_block_size {
             return kib * 1024;
         }
-        if self.cache_express_bucket_name().is_some() {
-            return 512 * 1024; // 512 KiB block size - default for express cache
-        }
-        1024 * 1024 // 1 MiB block size - default for disk cache
+        1024 * 1024 // 1 MiB block size - default for disk cache and for express cache
     }
 
     fn cache_express_bucket_name(&self) -> Option<&str> {
@@ -435,6 +443,27 @@ impl CliArgs {
             return Some(bucket_name);
         }
         None
+    }
+
+    fn disk_data_cache_config(&self) -> Option<(&Path, DiskDataCacheConfig)> {
+        match self.cache.as_ref() {
+            Some(path) => {
+                let cache_limit = match self.max_cache_size {
+                    // Fallback to no data cache.
+                    Some(0) => return None,
+                    Some(max_size_in_mib) => CacheLimit::TotalSize {
+                        max_size: (max_size_in_mib * 1024 * 1024) as usize,
+                    },
+                    None => CacheLimit::default(),
+                };
+                let cache_config = DiskDataCacheConfig {
+                    block_size: self.cache_block_size_in_bytes(),
+                    limit: cache_limit,
+                };
+                Some((path.as_path(), cache_config))
+            }
+            None => None,
+        }
     }
 
     /// Generates a logging configuration based on the CLI arguments.
@@ -756,6 +785,17 @@ pub fn create_s3_client(args: &CliArgs) -> anyhow::Result<(S3CrtClient, EventLoo
     Ok((client, runtime, s3_personality))
 }
 
+fn create_disk_cache(
+    cache_dir_path: &Path,
+    cache_config: DiskDataCacheConfig,
+) -> anyhow::Result<(ManagedCacheDir, DiskDataCache)> {
+    let cache_key = env_unstable_cache_key();
+    let managed_cache_dir = ManagedCacheDir::new_from_parent_with_cache_key(cache_dir_path, cache_key)
+        .context("failed to create cache directory")?;
+    let cache_dir_path = managed_cache_dir.as_path_buf();
+    Ok((managed_cache_dir, DiskDataCache::new(cache_dir_path, cache_config)))
+}
+
 fn mount<ClientBuilder, Client, Runtime>(args: CliArgs, client_builder: ClientBuilder) -> anyhow::Result<FuseSession>
 where
     ClientBuilder: FnOnce(&CliArgs) -> anyhow::Result<(Client, Runtime, S3Personality)>,
@@ -840,25 +880,61 @@ where
     tracing::trace!("using metadata TTL setting {metadata_cache_ttl:?}");
     filesystem_config.cache_config = CacheConfig::new(metadata_cache_ttl);
 
-    if let Some(path) = &args.cache {
-        let cache_limit = match args.max_cache_size {
-            // Fallback to no data cache.
-            Some(0) => None,
-            Some(max_size_in_mib) => Some(CacheLimit::TotalSize {
-                max_size: (max_size_in_mib * 1024 * 1024) as usize,
-            }),
-            None => Some(CacheLimit::default()),
-        };
-        if let Some(cache_limit) = cache_limit {
-            let cache_config = DiskDataCacheConfig {
-                block_size: args.cache_block_size_in_bytes(),
-                limit: cache_limit,
-            };
-            let cache_key = env_unstable_cache_key();
-            let managed_cache_dir = ManagedCacheDir::new_from_parent_with_cache_key(path, cache_key)
-                .context("failed to create cache directory")?;
+    match (args.disk_data_cache_config(), args.cache_express_bucket_name()) {
+        (None, Some(express_bucket_name)) => {
+            tracing::trace!("using S3 Express One Zone bucket as a cache for object content");
+            let express_cache = ExpressDataCache::new(
+                express_bucket_name,
+                client.clone(),
+                &args.bucket_name,
+                args.cache_block_size_in_bytes(),
+            );
 
-            let cache = DiskDataCache::new(managed_cache_dir.as_path_buf(), cache_config);
+            let prefetcher = caching_prefetch(express_cache, runtime, prefetcher_config);
+            let fuse_session = create_filesystem(
+                client,
+                prefetcher,
+                &args.bucket_name,
+                &args.prefix.unwrap_or_default(),
+                filesystem_config,
+                fuse_config,
+                &bucket_description,
+            )?;
+
+            Ok(fuse_session)
+        }
+        (Some((cache_dir_path, disk_data_cache_config)), None) => {
+            tracing::trace!("using local disk as a cache for object content");
+            let (managed_cache_dir, disk_cache) = create_disk_cache(cache_dir_path, disk_data_cache_config)?;
+
+            let prefetcher = caching_prefetch(disk_cache, runtime, prefetcher_config);
+            let mut fuse_session = create_filesystem(
+                client,
+                prefetcher,
+                &args.bucket_name,
+                &args.prefix.unwrap_or_default(),
+                filesystem_config,
+                fuse_config,
+                &bucket_description,
+            )?;
+
+            fuse_session.run_on_close(Box::new(move || {
+                drop(managed_cache_dir);
+            }));
+
+            Ok(fuse_session)
+        }
+        (Some((cache_dir_path, disk_data_cache_config)), Some(express_bucket_name)) => {
+            tracing::trace!("using both local disk and S3 Express One Zone bucket as a cache for object content");
+            let (managed_cache_dir, disk_cache) = create_disk_cache(cache_dir_path, disk_data_cache_config)?;
+            let express_cache = ExpressDataCache::new(
+                express_bucket_name,
+                client.clone(),
+                &args.bucket_name,
+                args.cache_block_size_in_bytes(),
+            );
+            let cache = MultilevelDataCache::new(Arc::new(disk_cache), express_cache, runtime.clone());
+
             let prefetcher = caching_prefetch(cache, runtime, prefetcher_config);
             let mut fuse_session = create_filesystem(
                 client,
@@ -874,43 +950,22 @@ where
                 drop(managed_cache_dir);
             }));
 
-            return Ok(fuse_session);
+            Ok(fuse_session)
+        }
+        _ => {
+            tracing::trace!("using no cache");
+            let prefetcher = default_prefetch(runtime, prefetcher_config);
+            create_filesystem(
+                client,
+                prefetcher,
+                &args.bucket_name,
+                &args.prefix.unwrap_or_default(),
+                filesystem_config,
+                fuse_config,
+                &bucket_description,
+            )
         }
     }
-
-    if let Some(express_bucket_name) = args.cache_express_bucket_name() {
-        // The cache can be shared across instances mounting the same bucket (including with different prefixes)
-        let source_description = &args.bucket_name;
-        let cache = ExpressDataCache::new(
-            express_bucket_name,
-            client.clone(),
-            source_description,
-            args.cache_block_size_in_bytes(),
-        );
-        let prefetcher = caching_prefetch(cache, runtime, prefetcher_config);
-        let fuse_session = create_filesystem(
-            client,
-            prefetcher,
-            &args.bucket_name,
-            &args.prefix.unwrap_or_default(),
-            filesystem_config,
-            fuse_config,
-            &bucket_description,
-        )?;
-
-        return Ok(fuse_session);
-    };
-
-    let prefetcher = default_prefetch(runtime, prefetcher_config);
-    create_filesystem(
-        client,
-        prefetcher,
-        &args.bucket_name,
-        &args.prefix.unwrap_or_default(),
-        filesystem_config,
-        fuse_config,
-        &bucket_description,
-    )
 }
 
 fn create_filesystem<Client, Prefetcher>(
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -309,7 +309,7 @@ pub struct CliArgs {
     #[cfg(feature = "block_size")]
     #[clap(
         long,
-        help = "Size of a cache block in KiB [Default: 1024 (1 MiB) for disk cache and for S3 Express cache]",
+        help = "Size of a cache block in KiB [Default: 1024 (1 MiB)]",
         help_heading = CACHING_OPTIONS_HEADER,
         value_name = "KiB",
         requires = "cache_group",
--- a/mountpoint-s3/src/data_cache/express_data_cache.rs
+++ b/mountpoint-s3/src/data_cache/express_data_cache.rs
@@ -82,15 +82,18 @@ where
         // TODO: optimize for the common case of a single chunk.
         let mut buffer = BytesMut::default();
         while let Some(chunk) = result.next().await {
-            let (offset, body) = chunk?;
-            if offset != buffer.len() as u64 {
-                return Err(DataCacheError::InvalidBlockOffset);
-            }
-            buffer.extend_from_slice(&body);
+            match chunk {
+                Ok((offset, body)) => {
+                    if offset != buffer.len() as u64 {
+                        return Err(DataCacheError::InvalidBlockOffset);
+                    }
+                    buffer.extend_from_slice(&body);
 
-            // Ensure the flow-control window is large enough.
-            // TODO: review if/when we add a header to the block.
-            result.as_mut().increment_read_window(self.block_size as usize);
+                    result.as_mut().increment_read_window(self.block_size as usize);
+                }
+                Err(ObjectClientError::ServiceError(GetObjectError::NoSuchKey)) => return Ok(None),
+                Err(e) => return Err(e.into()),
+            }
         }
         let buffer = buffer.freeze();
         DataCacheResult::Ok(Some(buffer.into()))
--- a/mountpoint-s3/src/data_cache/express_data_cache.rs
+++ b/mountpoint-s3/src/data_cache/express_data_cache.rs
@@ -78,6 +78,8 @@ where
         };
 
         pin_mut!(result);
+        // Guarantee that the request will start even in case of `initial_read_window == 0`.
+        result.as_mut().increment_read_window(self.block_size as usize);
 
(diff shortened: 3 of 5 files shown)


mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -102,7 +120,12 @@ where
         block_idx: BlockIndex,
         block_offset: u64,
         bytes: ChecksummedBytes,
+        object_size: usize,
     ) -> DataCacheResult<()> {
+        if object_size > self.max_object_size {
```
vladem, line 125: We may also want to skip the upload, if a block is already in the shared cache. This could be a more common pattern than with the local cache. Does this optimisation worth an extra `HeadObject` request with each `put_object`?

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -104,6 +104,16 @@ impl MockClient {
         self.objects.write().unwrap().remove(key);
     }
 
+    /// Remove all objects for the mock client's bucket
+    pub fn remove_all_objects(&self) {
+        self.objects.write().unwrap().clear();
+    }
+
+    /// Number of objects in the mock client's bucket
+    pub fn objects_number(&self) -> usize {
```
muddyfish, line 113: Isn't this normally called something like `object_count`?

mountpoint-s3/src/cli.rs
```diff
@@ -812,54 +823,90 @@ where
     tracing::trace!("using metadata TTL setting {metadata_cache_ttl:?}");
     filesystem_config.cache_config = CacheConfig::new(metadata_cache_ttl);
 
-    if let Some(path) = &args.cache {
+    let create_disk_cache = |args: &CliArgs| -> anyhow::Result<_> {
+        assert!(args.max_cache_size.is_none() || args.max_cache_size.unwrap() > 0);
+        let path = args.cache.as_ref().expect("cache path must be some");
```
muddyfish, line 828: Is it possible to change the function signature such that it can never be called with invalid args in the case that we check previously (which it seems we do)

mountpoint-s3/src/cli.rs
```diff
@@ -289,7 +300,7 @@ pub struct CliArgs {
         help = "Size of a cache block in KiB [Default: 1024 (1 MiB) for disk cache, 512 (512 KiB) for S3 Express cache]",
```
muddyfish, line 300: Should this change?

mountpoint-s3/src/cli.rs
```diff
@@ -812,54 +823,90 @@ where
...
 
-    if let Some(express_bucket_name) = args.cache_express_bucket_name() {
+        let cache_dir_path = managed_cache_dir.as_path_buf();
+        Ok((managed_cache_dir, DiskDataCache::new(cache_dir_path, cache_config)))
+    };
+
+    let create_express_cache = |args: &CliArgs, client| {
+        let express_bucket_name = args.cache_express_bucket_name().expect("express bucket must be set");
+        // bypass express cache for objects larger than 1 MiB
+        const EXPRESS_CACHE_MAX_OBJECT_SIZE: usize = 1024 * 1024;
```
muddyfish, line 850: Should we make this a static constant?

mountpoint-s3/src/cli.rs
```diff
@@ -812,54 +823,90 @@ where
...
-            client.clone(),
+            client,
             source_description,
             args.cache_block_size_in_bytes(),
-        );
+            EXPRESS_CACHE_MAX_OBJECT_SIZE,
+        )
+    };
+
+    let cache_size_positive = !matches!(args.max_cache_size, Some(0));
```
muddyfish, line 862: Does this catch negative numbers?

mountpoint-s3/src/cli.rs
```diff
@@ -812,54 +823,90 @@ where
...
+    if args.cache.is_some() && cache_size_positive && args.cache_express_bucket_name().is_some() {
```
muddyfish, line 863: Does a `match` block work here? It might be more readable

mountpoint-s3/src/data_cache/express_data_cache.rs
muddyfish: Might also be worth checking if we can do `put_if_none_match` headers and when they validate. Might be able to get away with not too many re-uploads

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -149,6 +172,8 @@ mod tests {
     use mountpoint_s3_client::mock_client::{MockClient, MockClientConfig};
     use mountpoint_s3_client::types::ETag;
 
+    const EXPRESS_CACHE_MAX_OBJECT_SIZE: usize = 1024 * 1024;
```
muddyfish, line 175: This should definitely be a constant somewhere as it's reused across files

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,323 @@
...
+            let cache_key = cache_key.clone();
+            let cache = self.disk_cache.clone();
+            let data_cloned = data.clone();
+            self.runtime
+                .spawn(async move {
+                    if let Err(error) = cache
+                        .put_block(cache_key.clone(), block_idx, block_offset, data_cloned, object_size)
+                        .await
+                    {
+                        warn!(cache_key=?cache_key, block_idx, ?error, "failed to update cache");
```
muddyfish, line 64: Should say failed to update local cache

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    async fn put_block(
+        &self,
+        cache_key: ObjectId,
+        block_idx: BlockIndex,
+        block_offset: u64,
+        bytes: ChecksummedBytes,
+        object_size: usize,
+    ) -> DataCacheResult<()> {
+        self.disk_cache
```
muddyfish, line 81: Could these be done in parallel? Should they be?

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,323 @@
...
+        let cache_key = ObjectId::new("a".into(), ETag::for_tests());
+
+        // put in both caches
+        cache
+            .put_block(cache_key.clone(), 0, 0, data.clone(), object_size)
+            .await
+            .expect("put should succeed");
+
+        // check it was put to express
+        assert_eq!(client.objects_number(), 1);
```
muddyfish, line 160: No check that the cache entry should match original bytes here

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,323 @@
...
+            data, entry,
+            "cache entry returned should match original bytes after put"
+        );
+
+        // delete entry from express
+        client.remove_all_objects();
+
+        // get entry from the local cache (with retries as it is async)
+        let mut retries = 10;
+        let entry = loop {
```
muddyfish, line 207: Why a loop here? Are we expecting it to fail sometimes?

mountpoint-s3/src/cli.rs
passaro: Agree with the match. Maybe we should have a method n `CliArgs` that returns an enum with the various cases. Something like:
```
NoCache,
DiskCache(DiskDataCacheConfig),
ExpressCache(ExpressCacheConfig),
MultiLevelCache(DiskDataCacheConfig, ExpressCacheConfig),
```

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -38,7 +39,13 @@ where
     /// Create a new instance.
     ///
     /// TODO: consider adding some validation of the bucket.
-    pub fn new(bucket_name: &str, client: Client, source_description: &str, block_size: u64) -> Self {
+    pub fn new(
+        bucket_name: &str,
+        client: Client,
+        source_description: &str,
+        block_size: u64,
+        max_object_size: usize,
```
passaro, line 47: This is for limiting the cache only for object up to a max size, right? Could it be a separate change?

mountpoint-s3/src/data_cache/express_data_cache.rs
passaro: But would `put_if_none_match` actually save anything?

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,323 @@
...
+        Self {
+            disk_cache,
+            express_cache,
+            runtime,
+        }
+    }
+}
+
+#[async_trait]
+impl<DiskCache: DataCache + Sync + Send + 'static, ExpressCache: DataCache + Sync, Runtime: Spawn + Sync> DataCache
```
passaro, line 31: nit: use `where`

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,323 @@
...
+        &self,
+        cache_key: &ObjectId,
+        block_idx: BlockIndex,
+        block_offset: u64,
+        object_size: usize,
+    ) -> DataCacheResult<Option<ChecksummedBytes>> {
+        if let Some(data) = self
+            .disk_cache
+            .get_block(cache_key, block_idx, block_offset, object_size)
+            .await?
```
passaro, line 44: Error handling requires a bit more thought: if there is an error here, we probably want to try the shared cache rather than bailing out.

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,323 @@
...
+        &self,
+        cache_key: ObjectId,
+        block_idx: BlockIndex,
+        block_offset: u64,
+        bytes: ChecksummedBytes,
+        object_size: usize,
+    ) -> DataCacheResult<()> {
+        self.disk_cache
+            .put_block(cache_key.clone(), block_idx, block_offset, bytes.clone(), object_size)
+            .await?;
```
passaro, line 83: Same issue as in `get_block`: an error on writing to disk should not prevent trying to write to the shared cache.

mountpoint-s3/src/data_cache/multilevel_cache.rs
vladem: it's definitely an option with something like
```rust
    async fn put_block(
        &self,
        cache_key: ObjectId,
        block_idx: BlockIndex,
        block_offset: u64,
        bytes: ChecksummedBytes,
        object_size: usize,
    ) -> DataCacheResult<()> {
        let fut1= self.disk_cache
            .put_block(cache_key.clone(), block_idx, block_offset, bytes.clone(), object_size);
        let fut2= self.express_cache
            .put_block(cache_key, block_idx, block_offset, bytes, object_size);
        let (res1, res2) = join!(fut1, fut2);
        res1?;
        res2?;
        Ok(())
    }
```

it may result in express cache being populated quicker, making data available to other instances, but the difference will not be too big

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,323 @@
...
+            .expect("cache entry should be returned");
+        assert_eq!(
+            data, entry,
+            "cache entry returned should match original bytes after put"
+        );
+
+        // delete entry from express
+        client.remove_all_objects();
+
+        // get entry from the local cache (with retries as it is async)
```
passaro, line 205: We control the runtime, though. Could we do better?

mountpoint-s3/src/data_cache/multilevel_cache.rs
vladem: yep, cache population is async so we don't really know how along after `get_block` completed the data'll be in cache

mountpoint-s3/src/data_cache/express_data_cache.rs
vladem: Good call, but I don't expect [If-None-Match](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestSyntax) header to result in an optimisation here. I'd expect the check of object existence to happen only after we've uploaded the body of the http request. [Documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html) doesn't say anything regarding that.

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -71,7 +71,7 @@ pub struct MockClientConfig {
 
 /// A mock implementation of an object client that we can manually add objects to, and then query
 /// via the [ObjectClient] APIs.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
vladem: Something like this?

```rust
#[derive(Debug, Clone)]
pub struct MockClient {
    inner: Arc<MockClientInner>,
}

#[derive(Debug)]
pub struct MockClientInner {
    config: MockClientConfig,
    objects: Arc<RwLock<BTreeMap<String, MockObject>>>,
    in_progress_uploads: Arc<RwLock<BTreeSet<String>>>,
    operation_counts: Arc<RwLock<HashMap<Operation, u64>>>,
}
```

This will require a bunch of `self.<field>` to `self.inner.<field>` changes in the file, but looks ok. Cloning the struct field-by-fields just requires less changes, but is slightly less memory efficient, yes.

mountpoint-s3/src/cli.rs
vladem: I can make `create_disk_cache` to receive unwrapped values instead of `&CliArgs`. The caller will need to unwrap them in some way, e.g:
```rust
create_disk_cache = |path: Path, max_cache_size: usize, cache_block_in_bytes: usize| -> anyhow::Result<_>
```

mountpoint-s3/src/cli.rs
vladem: Seems like something like this will work:
```rust
let cache_size_positive = !matches!(args.max_cache_size, Some(0));
match (&args.cache, args.cache_express_bucket_name(), cache_size_positive) {
    (None, Some(express_bucket_name), true) | (None, Some(express_bucket_name), false) | (Some(_), Some(express_bucket_name), false) => todo!(), // express
    (Some(path), None, true) => todo!(), // disk
    (Some(path), Some(express_bucket_name), true) => todo!(), // multilevel
    _ => (), // no cache
}
```

I don't think enum would make it much simpler / understandable, but let me know if you disagree.

passaro: This is really not great: `let cache_size_positive = !matches!(args.max_cache_size, Some(0));`

An intermediate option could be to have a `CliArgs::disk_data_cache_config(&self) -> Option<DiskDataCacheConfig>` method and then use a match like you suggest.

mountpoint-s3-client/src/mock_client.rs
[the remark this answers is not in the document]
passaro: We could likely review/remove the `Arc` in the fields though. No need to change now, I was only asking where you needed it `Clone`.

mountpoint-s3/src/data_cache/multilevel_cache.rs
muddyfish: Could we extend this so `get_block` or whatever returns a token that can be awaited/used to know when the cache has been written to?

mountpoint-s3/src/cli.rs
vladem: factored out, but it's `u64`

mountpoint-s3/src/cli.rs
vladem: made block size for both to be `1MiB`

mountpoint-s3/src/cli.rs
```diff
@@ -812,54 +823,90 @@ where
     tracing::trace!("using metadata TTL setting {metadata_cache_ttl:?}");
     filesystem_config.cache_config = CacheConfig::new(metadata_cache_ttl);
 
-    if let Some(path) = &args.cache {
+    let create_disk_cache = |args: &CliArgs| -> anyhow::Result<_> {
```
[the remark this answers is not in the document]
vladem: was not sure how to name the submodule, so kept new functions in the same module as the caller

mountpoint-s3/src/data_cache/express_data_cache.rs
vladem: factored out to the next PR

mountpoint-s3/src/data_cache/multilevel_cache.rs
vladem: Decided to execute them sequentially. The reasoning is:
1. we want the local cache for this instance to be available as soon as possible as it provides the best performance benefit on the repeated access to a block;
2. we are ok with having a delay when populating the shared cache, as it'll be used by other instances of MP and they may have their local cache;
3. starting the S3 request concurrently with the disk write may result in a longer delay for the disk operation.

mountpoint-s3/src/cli.rs
vladem: factored this out to the next PR

mountpoint-s3/src/cli.rs
vladem: Added a `disk_data_cache_config` and a match block as described in the comment.

mountpoint-s3-client/src/mock_client.rs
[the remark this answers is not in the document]
vladem: Ah, ok! The clone is needed in [multilevel cache unit-tests](https://github.com/awslabs/mountpoint-s3/pull/1064/files/8d9efd66ec6264356979f290b8450c1d8860168c#diff-9106dd30662c843c8114dbe0453395ba43b722d8777fda494f020e820e8b1644R129-R131).

mountpoint-s3/src/data_cache/multilevel_cache.rs
vladem: You mean something like this?
```rust
    async fn get_block(
        &self,
        cache_key: &ObjectId,
        block_idx: BlockIndex,
        block_offset: u64,
    ) -> DataCacheResult<(Option<ChecksummedBytes>, Option<RemoteHandle<..>>)>;
```

This doesn't look like a good fit for this API. Consider that there are multiple implementations of `get_block` and the second returned value will only make sense in the context of multilevel cache.

NB: there is a way to [spawn a task with handle](https://docs.rs/futures/latest/futures/task/trait.SpawnExt.html#method.spawn_with_handle), but the other tricky bit is that the task is canceled when the handle is dropped, so we'll need to hold this handle in the production code just for the sake of testing

mountpoint-s3/src/data_cache/multilevel_cache.rs
vladem: We can keep a reference to [ThreadPool](https://docs.rs/futures/latest/futures/executor/struct.ThreadPool.html), but I am not sure how may it help. We can not drop the thread pool or the multilevel cache, as later we need to make a `get_block` call which requires both. 

As @muddyfish suggested [below](https://github.com/awslabs/mountpoint-s3/pull/1064/files/8d9efd66ec6264356979f290b8450c1d8860168c#r1808709156) a way to make it more robust here would be to return a handle to the task from the `get_block`, but I am not sure if it is a better way compared to retries.

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,314 @@
...
+    disk_cache: Arc<DiskCache>,
+    express_cache: ExpressCache,
+    runtime: Runtime,
+}
+
+impl<DiskCache: DataCache, ExpressCache: DataCache, Runtime: Spawn>
+    MultilevelDataCache<DiskCache, ExpressCache, Runtime>
+{
+    pub fn new(disk_cache: Arc<DiskCache>, express_cache: ExpressCache, runtime: Runtime) -> Self {
+        assert_eq!(disk_cache.block_size(), express_cache.block_size());
```
muddyfish, line 21: Why does this matter?

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,314 @@
...
+                return DataCacheResult::Ok(Some(data));
+            }
+            Ok(None) => (),
+            Err(err) => warn!(cache_key=?cache_key, block_idx=block_idx, ?err, "error reading block from disk cache"),
+        }
+
+        if let Some(data) = self.express_cache.get_block(cache_key, block_idx, block_offset).await? {
+            trace!(cache_key=?cache_key, block_idx=block_idx, "block served from the express cache");
+            let cache_key = cache_key.clone();
+            let cache = self.disk_cache.clone();
```
muddyfish, line 55: Nit: can we call this `disk_cache`?

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,314 @@
...
+            let data_cloned = data.clone();
```
muddyfish, line 56: In the future, do you think this would be worth wrapping in an Arc to avoid a clone?

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,314 @@
...
+        block_offset: u64,
+        bytes: ChecksummedBytes,
+    ) -> DataCacheResult<()> {
+        match self
+            .disk_cache
+            .put_block(cache_key.clone(), block_idx, block_offset, bytes.clone())
+            .await
+        {
+            Ok(_) => (),
+            Err(error) => warn!(cache_key=?cache_key, block_idx, ?error, "failed to update the local cache"),
```
muddyfish, line 86: Stylistically, do we prefer this or using something like `.inspect_err` or `if let Err(...) = ...` for stuff like this?

mountpoint-s3/src/data_cache/multilevel_cache.rs
vladem: There is a method [block_size](https://github.com/awslabs/mountpoint-s3/pull/1064/files/6a8e202299cab80ee35bb6dac238dc1d2d8f6cba#diff-9106dd30662c843c8114dbe0453395ba43b722d8777fda494f020e820e8b1644R94) below, which's implemented based on this assumption. Later this method is used in CachingPartStream on [reads](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/prefetch/caching_stream.rs#L135) from cache and on [writes](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/prefetch/caching_stream.rs#L196) to it.

mountpoint-s3/src/data_cache/multilevel_cache.rs
vladem: `ChecksummedBytes` is cheaply cloneable  (it's size is about 7 `usize` fields), so I think it's ok to pass the value around.

mountpoint-s3/src/data_cache/multilevel_cache.rs
vladem: `if let Err(...) = ...` looks better, I'll use it here

mountpoint-s3/src/data_cache/multilevel_cache.rs
passaro: No, we would need a different runtime where we could control/block on task execution, maybe built on top of a [LocalPool](https://docs.rs/futures/latest/futures/executor/struct.LocalPool.html). May not be worth just to avoid the loop here, though.

mountpoint-s3/src/cli.rs
```diff
@@ -841,26 +880,34 @@ where
...
-            let cache_config = DiskDataCacheConfig {
-                block_size: args.cache_block_size_in_bytes(),
-                limit: cache_limit,
-            };
-            let cache_key = env_unstable_cache_key();
-            let managed_cache_dir = ManagedCacheDir::new_from_parent_with_cache_key(path, cache_key)
-                .context("failed to create cache directory")?;
+    match (args.disk_data_cache_config(), args.cache_express_bucket_name()) {
+        (None, Some(express_bucket_name)) => {
+            tracing::debug!("using express cache");
```
passaro, line 885: Either drop to `trace` level, or review the user messages (here and in the other cases) to match the args description/docs. Things to consider:
"cache" -> "object content cache" (vs metadata)
"express" -> "S3 Express One Zone"
"multilevel" -> not sure we have settled on a user-facing name

mountpoint-s3/src/cli.rs
```diff
@@ -438,6 +446,26 @@ impl CliArgs {
         None
     }
 
+    fn disk_data_cache_config(&self) -> Option<(&Path, DiskDataCacheConfig)> {
+        match self.cache.as_ref() {
+            Some(path) => {
+                let cache_limit = match self.max_cache_size {
+                    Some(0) => return None,
```
passaro, line 453: Add a comment here.

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -82,15 +82,20 @@ where
...
-                return Err(DataCacheError::InvalidBlockOffset);
+            match chunk {
+                Ok((offset, body)) => {
+                    if offset != buffer.len() as u64 {
+                        return Err(DataCacheError::InvalidBlockOffset);
+                    }
+                    buffer.extend_from_slice(&body);
+
+                    // Ensure the flow-control window is large enough.
+                    // TODO: review if/when we add a header to the block.
```
passaro, line 93: Outdated TODO?

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,321 @@
...
+
+use async_trait::async_trait;
+use futures::task::{Spawn, SpawnExt};
+use tracing::{trace, warn};
+
+use crate::object::ObjectId;
+
+use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheResult};
+
+pub struct MultilevelDataCache<DiskCache, ExpressCache, Runtime> {
```
passaro, line 11: Rustdocs

passaro (commented):
Just a couple of comments.

mountpoint-s3/src/data_cache/express_data_cache.rs
vladem: yeah, since we don't store header in the object. removed

mountpoint-s3/src/cli.rs
vladem: reduced to `trace` level, tried to make messages more descriptive

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,325 @@
...
+
+use async_trait::async_trait;
+use futures::task::{Spawn, SpawnExt};
+use tracing::{trace, warn};
+
+use crate::object::ObjectId;
+
+use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheResult};
+
+/// A data cache which uses both the local disk and S3 Express One Zone bucket as a storage.
```
passaro, line 11: I would simplify and just stick to only the first sentence.

If you want to provide more detail, you need to go all the way in and also mention how `get_block` can populate the local cache for example. But my preference would be to document the specific methods.

In any case, I would focus on the local vs shared distinction rather than assumptions on relative speed.

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -0,0 +1,325 @@
...
+    disk_cache: Arc<DiskCache>,
+    express_cache: ExpressCache,
+    runtime: Runtime,
+}
+
+impl<DiskCache: DataCache, ExpressCache: DataCache, Runtime: Spawn>
+    MultilevelDataCache<DiskCache, ExpressCache, Runtime>
+{
+    pub fn new(disk_cache: Arc<DiskCache>, express_cache: ExpressCache, runtime: Runtime) -> Self {
+        // Method `MultilevelDataCache::block_size` relies on block sizes of both caches to be equal.
```
passaro, line 25: A bit convoluted. We use the same blocks at both levels, so they need to have the same size. I'd mention it as a requirement in this method rustdoc.

mountpoint-s3/src/cli.rs
```diff
@@ -298,10 +309,10 @@ pub struct CliArgs {
     #[cfg(feature = "block_size")]
     #[clap(
         long,
-        help = "Size of a cache block in KiB [Default: 1024 (1 MiB) for disk cache, 512 (512 KiB) for S3 Express cache]",
+        help = "Size of a cache block in KiB [Default: 1024 (1 MiB) for disk cache and for S3 Express cache]",
```
passaro, line 312:
```suggestion
        help = "Size of a cache block in KiB [Default: 1024 (1 MiB)]",
```

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -82,15 +82,18 @@ where
...
-            if offset != buffer.len() as u64 {
-                return Err(DataCacheError::InvalidBlockOffset);
+            match chunk {
+                Ok((offset, body)) => {
+                    if offset != buffer.len() as u64 {
+                        return Err(DataCacheError::InvalidBlockOffset);
+                    }
+                    buffer.extend_from_slice(&body);
+
+                    result.as_mut().increment_read_window(self.block_size as usize);
```
passaro, line 92: Needs a comment. I already cannot tell anymore whether it is okay to increment **after** calling `next()` for example.

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -89,6 +91,7 @@ where
                     }
                     buffer.extend_from_slice(&body);
 
+                    // Ensure the flow-control window is large enough.
                     result.as_mut().increment_read_window(self.block_size as usize);
```
passaro, line 95: This seems unnecessary now, doesn't it?

passaro: Happy to keep it for now and review when we optimize for the single chunk case (see TODO above).

passaro (approved):
LGTM

mountpoint-s3/src/data_cache/express_data_cache.rs
vladem: We'll need to account for a case when the block object is larger than `block_size` for some reason. If we've just removed this line the `read` may freeze. If we've kept it as it is now MP may attempt to read an unbounded amount of data to RAM. 

This requires a bit more thinking, so I agree that it's better to address in the following PR.

PR #1065, opened 2024-10-16
Add GetObject support for object metadata

muddyfish:
## Description of change

Adds support for fetching user defined object metadata in GetObject calls.

Relevant issues: N/A

No

Yes, for mountpoint-s3-client.

---

### DIFF
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -525,7 +525,7 @@ impl UploadChecksum {
 /// This struct implements [`futures::Stream`], which you can use to read the body of the object.
 /// Each item of the stream is a part of the object body together with the part's offset within the
 /// object.
-#[cfg_attr(not(docsrs), async_trait)]
+#[async_trait]
 pub trait GetObjectRequest:
     Stream<Item = ObjectClientResult<GetBodyPart, GetObjectError, Self::ClientError>> + Send + Sync
 {
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -529,6 +529,11 @@ pub trait GetObjectRequest:
 {
     type ClientError: std::error::Error + Send + Sync + 'static;
 
+    /// Get the object's user defined metadata.
+    /// If the metadata has already been read, return immediately. Otherwise, resolve the future
+    /// when they're read.
+    async fn get_object_metadata(&mut self) -> Result<HashMap<String, String>, Self::ClientError>;
+
     /// Increment the flow-control window, so that response data continues downloading.
     ///
     /// If the client was created with `enable_read_backpressure` set true,
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -525,7 +525,7 @@ impl UploadChecksum {
 /// This struct implements [`futures::Stream`], which you can use to read the body of the object.
 /// Each item of the stream is a part of the object body together with the part's offset within the
 /// object.
-#[async_trait]
+#[cfg_attr(not(docsrs), async_trait)]
 pub trait GetObjectRequest:
     Stream<Item = ObjectClientResult<GetBodyPart, GetObjectError, Self::ClientError>> + Send + Sync
 {
@@ -534,7 +534,7 @@ pub trait GetObjectRequest:
     /// Get the object's user defined metadata.
     /// If the metadata has already been read, return immediately. Otherwise, resolve the future
     /// when they're read.
-    async fn get_object_metadata(self: Pin<&Self>) -> Result<ObjectMetadata, Self::ClientError>;
+    async fn get_object_metadata(&self) -> Result<ObjectMetadata, Self::ClientError>;
 
     /// Increment the flow-control window, so that response data continues downloading.
     ///
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -143,7 +143,7 @@ pub trait ObjectClient {
 ///
 /// [`ServiceError`]: ObjectClientError::ServiceError
 /// [`ClientError`]: ObjectClientError::ClientError
-#[derive(Debug, Error)]
+#[derive(Debug, Error, PartialEq)]
 pub enum ObjectClientError<S, C> {
     /// An error returned by the service itself
     #[error("Service error")]
@@ -534,7 +534,7 @@ pub trait GetObjectRequest:
     /// Get the object's user defined metadata.
     /// If the metadata has already been read, return immediately. Otherwise, resolve the future
     /// when they're read.
-    async fn get_object_metadata(&self) -> Result<ObjectMetadata, Self::ClientError>;
+    async fn get_object_metadata(&self) -> ObjectClientResult<ObjectMetadata, GetObjectError, Self::ClientError>;
 
     /// Increment the flow-control window, so that response data continues downloading.
     ///
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -357,6 +357,8 @@ pub enum GetObjectAttributesError {
     NoSuchKey,
 }
 
+pub type ObjectMetadata = HashMap<String, String>;
+
 /// Parameters to a [`put_object`](ObjectClient::put_object) request
 #[derive(Debug, Default, Clone)]
 #[non_exhaustive]
@@ -373,7 +375,7 @@ pub struct PutObjectParams {
     /// Custom headers to add to the request
     pub custom_headers: Vec<(String, String)>,
     /// User-defined object metadata
-    pub object_metadata: HashMap<String, String>,
+    pub object_metadata: ObjectMetadata,
 }
 
 impl PutObjectParams {
@@ -413,7 +415,7 @@ impl PutObjectParams {
     }
 
     /// Set user defined object metadata.
-    pub fn object_metadata(mut self, value: HashMap<String, String>) -> Self {
+    pub fn object_metadata(mut self, value: ObjectMetadata) -> Self {
         self.object_metadata = value;
         self
     }
@@ -456,7 +458,7 @@ pub struct PutObjectSingleParams {
     /// Custom headers to add to the request
     pub custom_headers: Vec<(String, String)>,
     /// User-defined object metadata
-    pub object_metadata: HashMap<String, String>,
+    pub object_metadata: ObjectMetadata,
 }
 
 impl PutObjectSingleParams {
@@ -496,7 +498,7 @@ impl PutObjectSingleParams {
     }
 
     /// Set user defined object metadata.
-    pub fn object_metadata(mut self, value: HashMap<String, String>) -> Self {
+    pub fn object_metadata(mut self, value: ObjectMetadata) -> Self {
         self.object_metadata = value;
         self
     }
@@ -525,14 +527,14 @@ impl UploadChecksum {
 /// object.
 #[cfg_attr(not(docsrs), async_trait)]
 pub trait GetObjectRequest:
-    Stream<Item = ObjectClientResult<GetBodyPart, GetObjectError, Self::ClientError>> + Send
+    Stream<Item = ObjectClientResult<GetBodyPart, GetObjectError, Self::ClientError>> + Send + Sync
 {
     type ClientError: std::error::Error + Send + Sync + 'static;
 
     /// Get the object's user defined metadata.
     /// If the metadata has already been read, return immediately. Otherwise, resolve the future
     /// when they're read.
-    async fn get_object_metadata(&mut self) -> Result<HashMap<String, String>, Self::ClientError>;
+    async fn get_object_metadata(self: Pin<&Self>) -> Result<ObjectMetadata, Self::ClientError>;
 
     /// Increment the flow-control window, so that response data continues downloading.
     ///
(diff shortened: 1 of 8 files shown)


mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -109,16 +139,26 @@ pub struct S3GetObjectRequest {
...
     /// Upper bound of the current read window. When backpressure is enabled, [S3GetObjectRequest]
     /// can return data up to this offset *exclusively*.
     read_window_end_offset: u64,
 }
 
+#[cfg_attr(not(docsrs), async_trait)]
 impl GetObjectRequest for S3GetObjectRequest {
     type ClientError = S3RequestError;
 
+    async fn get_object_metadata(&mut self) -> Result<HashMap<String, String>, Self::ClientError> {
```
vladem, line 154: does it have to be `mut`?

https://docs.rs/async-once-cell/latest/src/async_once_cell/lib.rs.html#1277

mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -109,16 +139,26 @@ pub struct S3GetObjectRequest {
...
     enable_backpressure: bool,
+    object_metadata_cb: Lazy<Result<ObjectMetadata, Canceled>, Receiver<ObjectMetadata>>,
     /// Next offset of the data to be polled from [poll_next]
     next_offset: u64,
     /// Upper bound of the current read window. When backpressure is enabled, [S3GetObjectRequest]
     /// can return data up to this offset *exclusively*.
     read_window_end_offset: u64,
 }
 
+#[cfg_attr(not(docsrs), async_trait)]
```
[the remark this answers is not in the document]
muddyfish: I'm not sure, but it was present in the original `object_client` codebase and I needed it. I can try seeing how it impacts the generated docs

mountpoint-s3-client/src/s3_crt_client/get_object.rs
muddyfish: Yes - otherwise the failure client doesn't compile as it then requires the `FailureGetRequest` to be `Sync`, which it isn't. Adding `mut` to the signature relaxes this requirement and I wasn't sure how to resolve the failure client issue, and I don't think reducing from`mut` to non-mutable later is a breaking change if we ever need to reduce it

mountpoint-s3-client/src/object_client.rs
```diff
@@ -428,6 +428,11 @@ pub trait GetObjectRequest:
 {
     type ClientError: std::error::Error + Send + Sync + 'static;
 
+    /// Get the object's user defined metadata.
+    /// If the metadata has already been read, return immediately. Otherwise, resolve the future
+    /// when they're read.
+    async fn get_object_metadata(&mut self) -> Result<HashMap<String, String>, Self::ClientError>;
```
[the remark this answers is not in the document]
vladem: Let's make this callable after the whole request was read, which is when the headers are certainly available and fits our use-case. This would mean the following signature:
```rust
async fn get_object_metadata(self: Pin<&mut Self>) -> Result<HashMap<String, String>, Self::ClientError>;
```

```rust
// failure_client.rs -- body changes
async fn get_object_metadata(self: Pin<&mut Self>) -> Result<HashMap<String, String>, Self::ClientError> {
    let this = self.project();
    this.request.get_object_metadata().await
}
```

```rust 
// s3_crt_client.rs -- no changes in body
async fn get_object_metadata(self: Pin<&mut Self>) -> Result<HashMap<String, String>, Self::ClientError> {
    self.object_metadata_cb
        .get_unpin()
        .await
        .clone()
        .map_err(|_| S3RequestError::RequestCanceled)
}
```

Also let's add a test for this scenario:
```rust 
let mut result = client
    .get_object(&bucket, &key, None, None)
    .await
    .expect("get_object should succeed");

let expected = &body;
pin_mut!(result);
while let Some(_) = result.next().await {
}
let actual_metadata = result.get_object_metadata().await.expect("Should return metadata");
assert_eq!(actual_metadata, metadata);
```

vladem (commented):
almost LGTM, let's make the `get_object_metadata` method callable after the request was consumed and process non UTF-8 headers in a safe way (+ resolve conflicts in Cargo.toml)

mountpoint-s3-client/src/object_client.rs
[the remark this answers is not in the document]
muddyfish: I decided to migrate to the pin approach anyway, but I don't want to make it require reading from the stream.

mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -109,16 +134,32 @@ pub struct S3GetObjectRequest {
...
 }
 
+#[async_trait]
 impl GetObjectRequest for S3GetObjectRequest {
     type ClientError = S3RequestError;
 
+    async fn get_object_metadata(self: Pin<&Self>) -> Result<ObjectMetadata, Self::ClientError> {
+        let empty_read_window = self.read_window_end_offset <= self.next_offset;
+        let request_required = self.object_metadata_cb.try_get().is_none();
+        if self.enable_backpressure && empty_read_window && request_required {
```
vladem, line 152: I propose to limit this to the `initial_read_window == 0` case, so the check will be as follows:
```rust
let request_required = self.object_metadata_cb.try_get().is_none();
if request_required && self.enable_backpressure && self.initial_read_window == 0 {
    return Err(S3RequestError::EmptyReadWindow);
}
```

(we'll need to store `initial_read_window ` in the request [on creation](https://github.com/awslabs/mountpoint-s3/blob/56dd238c43f8219ede7f449f92bd2632798a87aa/mountpoint-s3-client/src/s3_crt_client/get_object.rs#L71) of it)

mountpoint-s3-client/tests/get_object.rs
```diff
@@ -338,3 +339,79 @@ async fn test_get_object_cancel(read: bool) {
...
+        .await
+        .unwrap();
+
+    let client: S3CrtClient = get_test_client();
+
+    let result = client
+        .get_object(&bucket, &key, None, None)
+        .await
+        .expect("get_object should succeed");
+    pin_mut!(result);
```
passaro, line 369: Should we test that we can also call `get_object_metadata()` **after** streaming the body?

vladem: I'd also like to see a test case for a scenario when the meta request fails with an error, for a non-existing object for example. Will a `S3RequestError::RequestCanceled` be returned for `get_object_metadata`?

mountpoint-s3-client/src/object_client.rs
```diff
@@ -523,12 +525,17 @@ impl UploadChecksum {
 /// This struct implements [`futures::Stream`], which you can use to read the body of the object.
 /// Each item of the stream is a part of the object body together with the part's offset within the
 /// object.
-#[cfg_attr(not(docsrs), async_trait)]
+#[async_trait]
```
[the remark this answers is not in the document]
muddyfish: I removed it because it didn't appear to make any difference in how it showed up in the docs, and it doesn't appear to be necessary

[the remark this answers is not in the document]
passaro: Do you see `async fn get_object_metadata(...)` or does it get expanded? See here for example: https://docs.rs/mountpoint-s3-client/latest/mountpoint_s3_client/trait.ObjectClient.html

mountpoint-s3-client/src/object_client.rs
```diff
@@ -523,12 +525,17 @@ impl UploadChecksum {
...
 pub trait GetObjectRequest:
-    Stream<Item = ObjectClientResult<GetBodyPart, GetObjectError, Self::ClientError>> + Send
+    Stream<Item = ObjectClientResult<GetBodyPart, GetObjectError, Self::ClientError>> + Send + Sync
```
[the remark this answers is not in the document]
muddyfish: In `FailureGetRequest`, `get_object_metadata` calls `self.project_ref().request.get_object_metadata().await`. This usage of `request` across a future means the generic `GetObjectRequest` needs to be `Sync`, and it's not unless the `Stream` is also `Sync`.

mountpoint-s3-client/src/object_client.rs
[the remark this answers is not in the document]
muddyfish: Added it back again and verified it looks fine in doc preview

mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -109,16 +145,32 @@ pub struct S3GetObjectRequest {
...
     /// can return data up to this offset *exclusively*.
     read_window_end_offset: u64,
 }
 
+#[cfg_attr(not(docsrs), async_trait)]
 impl GetObjectRequest for S3GetObjectRequest {
     type ClientError = S3RequestError;
 
+    async fn get_object_metadata(&self) -> Result<ObjectMetadata, Self::ClientError> {
+        let request_required = self.lazy_object_metadata.try_get().is_none();
```
passaro, line 162: May just return it if already there...

mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -109,16 +145,32 @@ pub struct S3GetObjectRequest {
     finish_receiver: UnboundedReceiver<Result<GetBodyPart, Error>>,
     finished: bool,
     enable_backpressure: bool,
+    lazy_object_metadata: Lazy<Result<ObjectMetadata, Canceled>, Receiver<ObjectMetadata>>,
+    initial_read_window_empty: bool,
```
passaro, line 149: Couldn't we use `next_offset == read_window_end_offset` instead?

mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -67,15 +73,43 @@ impl S3CrtClient {
...
             span,
             |_| (),
-            |_, _| (),
+            move |headers, status| {
+                // Headers can be returned multiple times, but the object metadata doesn't change.
+                // Explicitly ignore the case where we've already set object metadata.
+                if let Some(object_metadata_sender) = object_metadata_sender_on_headers.lock().unwrap().take() {
+                    // On a 4xx / 5xx, don't send anything and cancel the future.
+                    // TODO - What happens on a 3xx status code?
+                    if status / 100 == 2 {
```
passaro, line 93: Check with a range here, e.g. `200..299`

mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -67,15 +73,43 @@ impl S3CrtClient {
...
             options,
             span,
             |_| (),
-            |_, _| (),
+            move |headers, status| {
+                // Headers can be returned multiple times, but the object metadata doesn't change.
+                // Explicitly ignore the case where we've already set object metadata.
+                if let Some(object_metadata_sender) = object_metadata_sender_on_headers.lock().unwrap().take() {
+                    // On a 4xx / 5xx, don't send anything and cancel the future.
+                    // TODO - What happens on a 3xx status code?
```
passaro, line 92: I'd move the status check before the `take()`. We only send the metadata on a successful request and rely on the `on_finish` callback to clean up if none succeed.

passaro (commented):
We are almost there. A couple of more suggestions.

mountpoint-s3-client/src/s3_crt_client/get_object.rs
vladem: No. We'll risk to return an error when headers are available. We know that for any successful request with non empty  `initial_read_window` headers will eventually be available. 

The proposed `next_offset == read_window_end_offset` may not work when the `initial_read_window` is not empty and the request was read till the end. In this scenario the specified condition will evaluate to true. If `body_callback` is called before the `headers_callback` with a slight delay and a call to `get_object_metadata` happening between them an error will be returned.

mountpoint-s3-client/src/object_client.rs
```diff
@@ -143,7 +143,7 @@ pub trait ObjectClient {
 ///
 /// [`ServiceError`]: ObjectClientError::ServiceError
 /// [`ClientError`]: ObjectClientError::ClientError
-#[derive(Debug, Error)]
+#[derive(Debug, Error, PartialEq)]
```
passaro, line 146: Is this required?

muddyfish: Yes, otherwise `assert_eq!(get_request.get_object_metadata().await, Ok(object_metadata));` fails in the tests. I could add a `.unwrap` but that doesn't feel right.

mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -87,10 +90,12 @@ impl S3CrtClient {
...
                 // Explicitly ignore the case where we've already set object metadata.
-                if let Some(object_metadata_sender) = object_metadata_sender_on_headers.lock().unwrap().take() {
-                    // On a 4xx / 5xx, don't send anything and cancel the future.
-                    // TODO - What happens on a 3xx status code?
-                    if status / 100 == 2 {
+
+                // Only set metadata if we have a 2xx status code. If we only get other status
+                // codes, then on_finish cancels.
+                if (200..300).contains(&status) {
+                    // This isn't to do with safety, only minor performance gains.
```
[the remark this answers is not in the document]
muddyfish: Do you mind if I do this in a followup PR? I've already gotten it changed as part of the CRC32 changes

vladem (approved):
LGTM

PR #1071, opened 2024-10-17
Add an integration test for the shared cache

vladem:
## Description of change

Add an integration test for the shared cache. It uses `S3_EXPRESS_ONE_ZONE_BUCKET_NAME` as a cache bucket and `S3_BUCKET_NAME` as a regular bucket.

Relevant issues: No

No.

No.

---

### DIFF
--- a/mountpoint-s3/tests/common/cache.rs
+++ b/mountpoint-s3/tests/common/cache.rs
@@ -77,14 +77,16 @@ impl<Cache: DataCache + Send + Sync + 'static> DataCache for CacheTestWrapper<Ca
         block_offset: u64,
         object_size: usize,
     ) -> DataCacheResult<Option<ChecksummedBytes>> {
-        let result = self
+        let result: Option<ChecksummedBytes> = self
             .inner
             .cache
             .get_block(cache_key, block_idx, block_offset, object_size)
-            .await?
-            .inspect(|_| {
-                self.inner.get_block_hit_count.fetch_add(1, Ordering::SeqCst);
-            });
+            .await?;
+
+        // The cache hit happens only if the `get_block` was successful and returned data
+        if result.is_some() {
+            self.inner.get_block_hit_count.fetch_add(1, Ordering::SeqCst);
+        }
 
         Ok(result)
     }
--- a/mountpoint-s3/tests/common/cache.rs
+++ b/mountpoint-s3/tests/common/cache.rs
@@ -1,0 +1,118 @@
+use std::{
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+
+use async_trait::async_trait;
+use mountpoint_s3::{
+    data_cache::{BlockIndex, ChecksummedBytes, DataCache, DataCacheResult},
+    object::ObjectId,
+};
+
+/// A wrapper around any type implementing [DataCache], which counts operations
+pub struct CacheTestWrapper<Cache> {
+    cache: Arc<Cache>,
+    get_block_ok_count: Arc<AtomicU64>,
+    get_block_hit_count: Arc<AtomicU64>,
+    get_block_failed_count: Arc<AtomicU64>,
+    put_block_ok_count: Arc<AtomicU64>,
+    put_block_failed_count: Arc<AtomicU64>,
+}
+
+impl<Cache> Clone for CacheTestWrapper<Cache> {
+    fn clone(&self) -> Self {
+        Self {
+            cache: self.cache.clone(),
+            get_block_ok_count: self.get_block_ok_count.clone(),
+            get_block_hit_count: self.get_block_hit_count.clone(),
+            get_block_failed_count: self.get_block_failed_count.clone(),
+            put_block_ok_count: self.put_block_ok_count.clone(),
+            put_block_failed_count: self.put_block_failed_count.clone(),
+        }
+    }
+}
+
+impl<Cache> CacheTestWrapper<Cache> {
+    pub fn new(cache: Arc<Cache>) -> Self {
+        CacheTestWrapper {
+            cache,
+            get_block_ok_count: Arc::new(AtomicU64::new(0)),
+            get_block_hit_count: Arc::new(AtomicU64::new(0)),
+            get_block_failed_count: Arc::new(AtomicU64::new(0)),
+            put_block_ok_count: Arc::new(AtomicU64::new(0)),
+            put_block_failed_count: Arc::new(AtomicU64::new(0)),
+        }
+    }
+
+    pub fn wait_for_put(&self, max_wait_duration: Duration) {
+        let st = std::time::Instant::now();
+        loop {
+            if st.elapsed() > max_wait_duration {
+                panic!("timeout on waiting for a write to the cache to happen")
+            }
+            if self.put_block_failed_count.load(Ordering::SeqCst) > 0
+                || self.put_block_ok_count.load(Ordering::SeqCst) > 0
+            {
+                break;
+            }
+            std::thread::sleep(Duration::from_millis(100));
+        }
+    }
+
+    pub fn get_block_hit_count(&self) -> u64 {
+        self.get_block_hit_count.load(Ordering::SeqCst)
+    }
+}
+
+#[async_trait]
+impl<Cache: DataCache + Send + Sync + 'static> DataCache for CacheTestWrapper<Cache> {
+    async fn get_block(
+        &self,
+        cache_key: &ObjectId,
+        block_idx: BlockIndex,
+        block_offset: u64,
+        object_size: usize,
+    ) -> DataCacheResult<Option<ChecksummedBytes>> {
+        let result = self
+            .cache
+            .get_block(cache_key, block_idx, block_offset, object_size)
+            .await
+            .inspect(|_| {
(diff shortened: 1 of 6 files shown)


mountpoint-s3/tests/fuse_tests/cache_test.rs
```diff
@@ -0,0 +1,121 @@
...
+    let key = get_object_key(&prefix, key_suffix, key_size);
+    let path = mount_point.path().join(&key);
+    let written = random_binary_data(object_size);
+    fs::write(&path, &written).expect("write should succeed");
+
+    // first read should be from the source bucket and be cached
+    let read = fs::read(&path).expect("read should succeed");
+    assert_eq!(read, written);
+
+    // cache population is async, 1 second should be enough for it to finish
```
muddyfish, line 74: Is there a better way for us to tell if it's done?

muddyfish: I'm thinking that avoiding flakey tests is something we should do here. Is it possible to make things public for tests only? Otherwise, I wonder if it's possible for us to return some awaitable on cache write complete. 

Can this impact regular on-disk cache?

vladem: another improvement which I can propose is to retry actual reads of the file until the object is served from cache (then now sleep after the first read will be required):
```rust
    // second read should be from the cache
    const MAX_WAIT_DURATION: std::time::Duration = std::time::Duration::from_secs(10);
    let st = std::time::Instant::now();
    loop {
        if st.elapsed() > MAX_WAIT_DURATION {
            panic!("timeout on waiting for data being served from the cache")
        }
        if let Ok(read) = fs::read(&path) {
            assert_eq!(read, written);
            break;
        }
        sleep(Duration::from_millis(100));
    }
```

Lmk if this looks better.

vladem: Replaced a single `sleep` with retrying read logic

mountpoint-s3/tests/common/fuse.rs
```diff
@@ -379,6 +379,47 @@ pub mod s3_session {
         }
     }
 
+    /// Create a FUSE mount backed by a real S3 client with a cache.
+    /// Note, that the mount uses S3 Standard as a source bucket.
+    pub fn new_with_cache_factory<Cache, CacheFactory>(
```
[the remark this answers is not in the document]
vladem: That is true, I'll need to make some functions of this module public to move it: `create_fuse_session`, `create_test_client`.

mountpoint-s3/tests/fuse_tests/cache_test.rs
vladem: Reverted back to `sleep` after @passaro [comment](https://github.com/awslabs/mountpoint-s3/pull/1071#discussion_r1831432734)

mountpoint-s3/tests/common/cache.rs
```diff
@@ -0,0 +1,118 @@
...
+
+/// A wrapper around any type implementing [DataCache], which counts operations
+pub struct CacheTestWrapper<Cache> {
+    cache: Arc<Cache>,
+    get_block_ok_count: Arc<AtomicU64>,
+    get_block_hit_count: Arc<AtomicU64>,
+    get_block_failed_count: Arc<AtomicU64>,
+    put_block_ok_count: Arc<AtomicU64>,
+    put_block_failed_count: Arc<AtomicU64>,
+}
```
passaro, line 23:
```suggestion
    inner: Arc<CacheTestWrapperInner<Cache>>,
}

struct CacheTestWrapperInner<Cache> {
    cache: Cache,
    get_block_ok_count: AtomicU64,
    get_block_hit_count: AtomicU64,
    get_block_failed_count: AtomicU64,
    put_block_ok_count: AtomicU64,
    put_block_failed_count: AtomicU64,
}
```

Or is there any reason to have all the separate `Arc`s?

passaro: Also please add some comments on the counters, especially to explain e.g. ok vs hit.

mountpoint-s3/tests/common/cache.rs
```diff
@@ -0,0 +1,118 @@
...
+            cache,
+            get_block_ok_count: Arc::new(AtomicU64::new(0)),
+            get_block_hit_count: Arc::new(AtomicU64::new(0)),
+            get_block_failed_count: Arc::new(AtomicU64::new(0)),
+            put_block_ok_count: Arc::new(AtomicU64::new(0)),
+            put_block_failed_count: Arc::new(AtomicU64::new(0)),
+        }
+    }
+
+    pub fn wait_for_put(&self, max_wait_duration: Duration) {
```
passaro, line 50: What is this doing? Waiting for the very first put call to return (almost)? How is it useful?

mountpoint-s3/tests/common/cache.rs
```diff
@@ -0,0 +1,118 @@
...
+        cache_key: &ObjectId,
+        block_idx: BlockIndex,
+        block_offset: u64,
+        object_size: usize,
+    ) -> DataCacheResult<Option<ChecksummedBytes>> {
+        let result = self
+            .cache
+            .get_block(cache_key, block_idx, block_offset, object_size)
+            .await
+            .inspect(|_| {
```
passaro, line 83: Consider replacing with a single `match`.

mountpoint-s3/tests/fuse_tests/cache_test.rs
```diff
@@ -0,0 +1,129 @@
...
+    let _session = create_fuse_session(
+        client,
+        prefetcher,
+        &bucket,
+        &prefix,
+        mount_point.path(),
+        Default::default(),
+    );
+
+    // Write an object, no caching happens yet
```
passaro, line 88: no caching -> you could assert on it now

mountpoint-s3/tests/fuse_tests/cache_test.rs
```diff
@@ -0,0 +1,129 @@
...
+    let key = get_object_key(&prefix, key_suffix, key_size);
+    let path = mount_point.path().join(&key);
+    let written = random_binary_data(object_size);
+    fs::write(&path, &written).expect("write should succeed");
+
+    // First read should be from the source bucket and be cached
+    let read = fs::read(&path).expect("read should succeed");
+    assert_eq!(read, written);
+
+    // Cache population is async, wait for it to happen
```
passaro, line 98: See comment above: you are not waiting for the cache to be populated, just for `put_block` to return.

vladem: After the `put_block` returned we [know](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/data_cache/express_data_cache.rs#L160) that the data is in the cache. 

Are you suggesting to wait for the `put_block` counter to change the value? e.g like this:
```rust

    pub fn wait_for_put(&self, max_wait_duration: Duration, previous_failed: u64, previous_ok: u64) {
        let st = std::time::Instant::now();
        loop {
            if st.elapsed() > max_wait_duration {
                panic!("timeout on waiting for a write to the cache to happen")
            }
            if self.put_block_failed_count.load(Ordering::SeqCst) > previous_failed
                || self.put_block_ok_count.load(Ordering::SeqCst) > previous_ok
            {
                break;
            }
            std::thread::sleep(Duration::from_millis(100));
        }
    }
```

passaro: But yes, you should compare with the previous count in some way.

passaro: If we are not actually using the failure counter, you could just simplify and just count all puts. We can always add counters later.

mountpoint-s3/tests/common/cache.rs
vladem: only one `inspect` left, so I think it makes sense to leave it as-is

mountpoint-s3/tests/fuse_tests/cache_test.rs
vladem: kept 2 counters only, made it to compare with the previous valie

mountpoint-s3/tests/common/cache.rs
passaro: nit: it still took me a while to figure out what is happening here, i.e. `.inspect` is called on the `Option` you get after `?`, so only when it is `some(..)`. Consider making it more explicit or adding a comment.

passaro (commented):
Looks good, only one minor suggestion.

mountpoint-s3/tests/common/cache.rs
```diff
@@ -0,0 +1,112 @@
...
+
+struct CacheTestWrapperInner<Cache> {
+    cache: Cache,
+    /// Number of times the `get_block` succeded and returned data
+    get_block_hit_count: AtomicU64,
+    /// Number of times the `put_block` was completed
+    put_block_count: AtomicU64,
+}
+
+impl<Cache> Clone for CacheTestWrapper<Cache> {
```
passaro, line 28: Why not just derive Clone?

passaro: `Cache` is not clone, derive will not work.

passaro (approved):
Merge it!

PR #1073, opened 2024-10-21
Pin Rust to 1.81.0

unexge:
- Create `rust-toolchain.toml`
- Use `actions-rust-lang/setup-rust-toolchain` GitHub action as it supports `rust-toolchain.toml`
---

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,3 @@
 [toolchain]
 channel = "1.81.0"
+components = ["rust-src"]


passaro:
@unexge, is the `rust-toolchain` action actually required? My understanding is that cargo will respect `rust-toolchain.toml` automatically.

unexge:
@passaro, not sure. It's [a popular request](https://github.com/dtolnay/rust-toolchain/issues?q=rust-toolchain+file) for [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain) to support rust-toolchain file, so I assumed it won't work. But let me give it a try.

.github/workflows/bench.yml
```diff
@@ -51,8 +52,10 @@ jobs:
         fuseVersion: 2
         libunwind: true
         fio: true
-    - name: Set up stable Rust
-      uses: dtolnay/rust-toolchain@stable
+    - name: Set up Rust toolchain
+      uses: dtolnay/rust-toolchain@master
```
passaro: My suggestion was to remove this step entirely, not to revert to `stable`

passaro: Alternative: we could look at some other actions, e.g. https://github.com/marketplace/actions/rust-toolchain-toml-install

unexge: Agh, I misunderstood. I think `dtolnay/rust-toolchain` installs `rustup`/`cargo`. We might look installing it in `.github/actions/install-dependencies` step, not sure if there are any downsides.

dannycjones: I was wondering about jumping to a new action, I saw this one that looks well maintained: https://github.com/actions-rust-lang/setup-rust-toolchain

unexge: Updated to use `actions-rust-lang/setup-rust-toolchain`. I had to override `rustflags` option though, it's `-D warnings` by default which turns all warnings into compile errors, and `vendor/fuser` has lots of warnings. It'd be nice to get rid of those warnings eventually, but not turning this on should be fine for now.

This new action also surfaces warnings on PRs, e.g., scroll to the bottom on https://github.com/awslabs/mountpoint-s3/pull/1073/files. We can turn this of it we find it noisy, but eventually I'd like to address these warnings as well.

passaro (approved):

PR #1074, opened 2024-10-22
Add workaround to random write errors in troubleshooting guide

muddyfish:
## Description of change

Adds a potential workaround for customers who need to do random writes in their workload. Mountpoint doesn't support this, but files can be written to on a non-MP volume and then moved into Mountpoint using standard tooling

Relevant issues: https://github.com/awslabs/mountpoint-s3-csi-driver/issues/278#issuecomment-2425204439 

No

No - this is just a docs change

---


dannycjones (approved):
LGTM, thanks!

Issue #1079, opened 2024-10-23
Allow custom telemetry handlers


aws-hans-pistor:
### Background
Different users of the mountpoint client are likely to be interested in different metrics that are relevant to their usecase. In order to support all of them, it would be nice if they could define a custom telemetry handler for their client & then whenever the on_telemetry callback is run, that handler is run as well.

For context, our client downloads different parts of the same object in S3 at high concurrency. We think there is some sort of throttling/slowdown going on somwhere along the request path, but in today's mountpoint we have no way of injecting metrics/logging for requests where the bandwidth is < X MB/s or the duration > Yms. It doesn't really make sense for mountpoint itself to support this usecase, but it also makes sense that users should have the capability to do this themselves

I'll be submitting a PR for this issue, but creating the issue as well if discussion is necessary.

vladem:
Hey @aws-hans-pistor, thanks for opening the issue and proposing the code change. We'd like to understand the use-case a bit better before proceeding.

I can see that the client already [logs](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-client/src/s3_crt_client.rs#L582) some information from telemetry with each HTTP request after it finishes (note that a single call to the client's method, e.g `get_object`,  may result in multiple underlying HTTP requests, e.g. `HeadObject` + multiple `GetObject`-s). You need `DEBUG` level logging to get those logs.

Even more information is logged when the `TRACE` level is enabled ([1](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-client/src/s3_crt_client.rs#L583), [2](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-crt/src/s3/client.rs#L1316)).

Ain't this logging sufficient for your use-case? If some information is not extracted from the CRT (e.g. extended request ids) we can consider fixing this for all users of the client. 

From the description it looks that you are also looking for a better way of controlling the verbosity of the logs:
> way of injecting metrics/logging for requests where the bandwidth is < X MB/s or the duration > Yms

[...]

PR #1080, opened 2024-10-23
Allow clients to define custom callbacks to handle telemetry

aws-hans-pistor:
## Description of change
Different users of mountpoint will care about different metrics returned for each requests, so allow them to define their own custom handlers for the on_telemetry callback in addition to the default metrics that mountpoint emits.

This allows users to do things like: 
- emit extended request ids ("x-amz-id-2")
- When some criteria is met, log out additional information

Relevant issues: #1079 

No there should be no breaking changes, the only visible change is that there's a new field to the S3ClientConfig which defines the custom telemetry handler

Just a note in mountpoint-s3-client letting users know this feature now exists

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -102,6 +102,7 @@ pub struct S3ClientConfig {
     read_backpressure: bool,
     initial_read_window: usize,
     network_interface_names: Vec<String>,
+    telemetry_callback: Option<Arc<dyn OnTelemetry>>,
 }
 
 impl Default for S3ClientConfig {
@@ -120,6 +121,7 @@ impl Default for S3ClientConfig {
             read_backpressure: false,
             initial_read_window: DEFAULT_PART_SIZE,
             network_interface_names: vec![],
+            telemetry_callback: None,
         }
     }
 }
@@ -221,6 +223,13 @@ impl S3ClientConfig {
         self.network_interface_names = network_interface_names;
         self
     }
+
+    /// Set a custom telemetry callback handler
+    #[must_use = "S3ClientConfig follows a builder pattern"]
+    pub fn telemetry_callback(mut self, telemetry_callback: Arc<dyn OnTelemetry>) -> Self {
+        self.telemetry_callback = Some(telemetry_callback);
+        self
+    }
 }
 
 /// Authentication configuration for the CRT-based S3 client
@@ -288,6 +297,7 @@ struct S3CrtClientInner {
     bucket_owner: Option<String>,
     credentials_provider: Option<CredentialsProvider>,
     host_resolver: HostResolver,
+    telemetry_callback: Option<Arc<dyn OnTelemetry>>,
 }
 
 impl S3CrtClientInner {
@@ -422,6 +432,7 @@ impl S3CrtClientInner {
             bucket_owner: config.bucket_owner,
             credentials_provider: Some(credentials_provider),
             host_resolver,
+            telemetry_callback: config.telemetry_callback,
         })
     }
 
@@ -551,6 +562,7 @@ impl S3CrtClientInner {
         let endpoint = options.get_endpoint().expect("S3Message always has an endpoint");
         let hostname = endpoint.host_name().to_str().unwrap().to_owned();
         let host_resolver = self.host_resolver.clone();
+        let telemetry_callback = self.telemetry_callback.clone();
 
         let start_time = Instant::now();
         let first_body_part = Arc::new(AtomicBool::new(true));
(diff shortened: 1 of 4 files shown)


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -1341,6 +1358,10 @@ impl ObjectClient for S3CrtClient {
     }
 }
 
+pub trait OnTelemetry: std::fmt::Debug + Send + Sync {
```
aws-hans-pistor, line 1361: I've defined a trait since I find it a little more readable than raw `Fn` types, let me know if you prefer differently

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -100,6 +100,7 @@ pub struct S3ClientConfig {
     read_backpressure: bool,
     initial_read_window: usize,
     network_interface_names: Vec<String>,
+    telemetry_callback: Option<Arc<dyn OnTelemetry>>,
```
aws-hans-pistor, line 103: I've used dynamic dispatch here since I don't think you want this type to leak into all the other types (i.e. S3CrtClient, S3CrtClientInner, S3ClientConfig, etc)

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -1341,6 +1358,10 @@ impl ObjectClient for S3CrtClient {
     }
 }
 
+pub trait OnTelemetry: std::fmt::Debug + Send + Sync {
+    fn on_telemetry(&self, request_metrics: &RequestMetrics, span: &Span);
```
jamesbornholt, line 1362: We will always have already entered the span before making this call, so is there any reason to also pass it explicitly?

aws-hans-pistor: I kept it because it looks like some metadata is stored within the span

```
let op = span_telemetry.metadata().map(|m| m.name()).unwrap_or("unknown");
```

Not sure if it's useful though

mountpoint-s3-client/src/s3_crt_client.rs
aws-hans-pistor: That's fair, I can swap to a type alias

mountpoint-s3-client/src/s3_crt_client.rs
jamesbornholt: Ah right. I think we should probably just pass that `op` as an argument to the call; I think it's the only important one. The callback can always retrieve the span with [`Span::current`](https://docs.rs/tracing/0.1.40/tracing/struct.Span.html#method.current) if they need it.

mountpoint-s3-client/src/s3_crt_client.rs
aws-hans-pistor: Actually this is not possible

`S3ClientConfig` requires that everything implements std::fmt::Debug, and type aliases can only implement auto-traits.

mountpoint-s3-client/src/s3_crt_client.rs
aws-hans-pistor: Cool I've updated the function signature to the operation as a &str. You're right, if callers really care then can grab it themselves.

aws-hans-pistor:
Resolved conflicts in latest revision

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -1342,6 +1359,10 @@ impl ObjectClient for S3CrtClient {
     }
 }
 
+pub trait OnTelemetry: std::fmt::Debug + Send + Sync {
+    fn on_telemetry(&self, request_metrics: &RequestMetrics, operation: &str);
```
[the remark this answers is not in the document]
aws-hans-pistor: I agree, i left `operation` because the existing telemetry code using it extensively

```
                    metrics::counter!("s3.requests.failures", "op" => op, "type" => request_type, "status" => http_status.unwrap_or(-1).to_string()).increment(1);
```


If you think it's unnecessary, will remove it.

aws-hans-pistor:
Another merge conflict with mainline

@vladem Can you restart the test runs again?

aws-hans-pistor:
Interface has changed, I will have to resolve this locally.

aws-hans-pistor:
I've rebased back onto upstream's mainline

mountpoint-s3-client/tests/common/mod.rs
```diff
@@ -87,6 +88,13 @@ pub fn get_test_backpressure_client(initial_read_window: usize, part_size: Optio
     S3CrtClient::new(config).expect("could not create test client")
 }
 
+pub fn get_test_client_with_custom_telemetry(telemetry_callback: Arc<dyn OnTelemetry>) -> S3CrtClient {
+    let config = S3ClientConfig::new()
+        .endpoint_config(EndpointConfig::new(&get_test_region()))
```
vladem, line 93:
```suggestion
        .endpoint_config(get_test_endpoint_config())
```

vladem (commented):
LGTM, one minor suggestion. Launched the tests, ready to approve after that.

vladem (approved):

muddyfish (approved):

aws-hans-pistor:
The failing test seems unrelated
https://github.com/awslabs/mountpoint-s3/blob/33c20c8ba9a0dd3a024915583ad3fc7b15e3e524/mountpoint-s3/src/upload/incremental.rs#L234

```
 ---- upload::incremental::tests::test_append_failure_on_object_replaced stdout ----
thread 'upload::incremental::tests::test_append_failure_on_object_replaced' panicked at mountpoint-s3/src/upload/incremental.rs:234:75:
```

PR #1082, opened 2024-10-24
Add additional checksum algorithms in mountpoint-s3-crt crate

dannycjones:
## Description of change

We want to provide bindings to some of the checksum algorithms provided by the AWS CRT. In particular, it may be interesting to see how these checksum algorithms compare for things like the internal checksumming we perform on the read path of Mountpoint file system. The performance of CRC64 versus our current CRC32C looks promising!

This change provides bindings for all algorithms exposed by the AWS CRT today.

Relevant issues: N/A

It removes the implementation of `std::hash::Hasher` on the checksum/digest algorithms. We do not depend on this in Mountpoint itself, but this will be called out in the change log for any consumers of the crates.

This adds new bindings. This has been noted in the `mountpoint-s3-crt` changelog.

---

### DIFF
--- a/mountpoint-s3-crt/src/checksums/sha1.rs
+++ b/mountpoint-s3-crt/src/checksums/sha1.rs
@@ -1,0 +1,115 @@
+use std::{mem::MaybeUninit, ptr::NonNull};
+
+use mountpoint_s3_crt_sys::{
+    aws_byte_buf, aws_byte_buf_init, aws_hash, aws_hash_destroy, aws_hash_finalize, aws_hash_update, aws_sha1_new,
+    AWS_SHA1_LEN,
+};
+
+use crate::common::allocator::Allocator;
+use crate::common::error::Error;
+use crate::{CrtError as _, ToAwsByteCursor};
+
+/// SHA1 checksum
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Sha1([u8; Self::LENGTH]);
+
+impl Sha1 {
+    /// Length of a SHA1 checksum.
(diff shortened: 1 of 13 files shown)


mountpoint-s3-crt/src/checksums/sha1.rs
```diff
@@ -0,0 +1,115 @@
...
+        // SAFETY: `self.inner` is a valid `aws_hash` and `buf` will outlive the call to `aws_hash_update`.
+        unsafe { aws_hash_update(self.inner.as_ptr(), &buf.as_aws_byte_cursor()).ok_or_last_error() }
+    }
+
+    /// Finalize the hash state and return the computed SHA1 checksum value.
+    pub fn finalize(self, allocator: &Allocator) -> Result<Sha1, Error> {
+        // SAFETY: allocator is a valid aws_allocator, and `aws_byte_buf_init` initializes the buffer on success.
+        let mut buffer = unsafe {
+            let mut buffer: MaybeUninit<aws_byte_buf> = MaybeUninit::uninit();
+            aws_byte_buf_init(buffer.as_mut_ptr(), allocator.inner.as_ptr(), Sha1::LENGTH).ok_or_last_error()?;
```
passaro, line 66: Ooops, we need to balance this with `aws_byte_buf_clean_up`. Probably worth creating a small wrapping type, assuming we don't have one already!

dannycjones: Yeah, glad that ASAN caught it too!

passaro (approved):
LGTM!

PR #1083, opened 2024-10-24
Add ability to request checksum in an S3 HeadObject request

dannycjones:
## Description of change

For a consumer of the S3 client, we need to return the checksum algorithm used with the object. This change adds the request option to retrieve checksum algorithm as part of a HeadObject request. The result now include the field, which will only be populated when requested and supported by the S3 implementation.

Relevant issues: N/A

We now require a `HeadObjectParams` struct when invoking `head_object` with the S3 client. This changes the signature, and so a change log entry will be added for the client crate. There is no behavior change.

Yes, a note is added for `mountpoint-s3-client` crate.

---

### DIFF
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -97,6 +97,7 @@ pub trait ObjectClient {
         &self,
         bucket: &str,
         key: &str,
+        params: &HeadObjectParams,
     ) -> ObjectClientResult<HeadObjectResult, HeadObjectError, Self::ClientError>;
 
     /// Put an object into the object store. Returns a [PutObjectRequest] for callers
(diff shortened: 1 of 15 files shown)


mountpoint-s3-client/src/object_client.rs
```diff
@@ -647,7 +682,7 @@ impl fmt::Display for ObjectAttribute {
 ///
 /// See [Checksum](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Checksum.html) in the *Amazon
 /// S3 API Reference* for more details.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
passaro, line 685: Thoughts on having the checksums typed once #1082 goes in?

dannycjones: I don't have a strong opinion. I'd be inclined to stick close to the REST API but I see we already diverged a bit by creating a `Checksum` struct in the first place.

dannycjones: Thinking about this a bit more:

- Probably HeadObject shouldn't be returning this struct, but instead each checksum on the `HeadObjectResult`. (This struct is mirroring how GetObjectAttributes returns the values.)
- We could have `Option<Crc32>`, etc. here. That sounds fine to me.

passaro: If we are talking about whether to wrap the `Option`al checksums in a `Checksum` struct, my slight preference is in favour - I don't think its really diverging from the API in meaningful ways. But I don't think it's really too interesting either way.

I care a bit more about the "typed" checksums. As I sees it we should either:
1. keep everything as `String`, neutral and close to the API. We'll offer separate facilities to parse checksums (and encode as base64).
2. consistently use our checksum types.

My preference is for (2), which seems much nicer for users.

passaro: None of this is blocking for this PR, just bringing it up because we may want to go through all related API changes in one release cycle.

passaro (approved):
Looks good. See comments on API changes, but happy to review in a separate PR.

PR #1085, opened 2024-10-28
Fix issue with fstat after upload

passaro:
## Description of change

As covered in #1044, Mountpoint would previously invalidate the inode for a file open for writing once it completed the upload, which resulted in calls to `fstat` to return `ESTALE`.

This change addresses the issue by updating the etag in the existing inode with the one returned from a successful upload. 

Fixes the issue in #1044.

Added an entry in `mountpoint-s3` changelog.

---

### DIFF
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -1,5 +1,13 @@
 ## Unreleased
 
+### New features
+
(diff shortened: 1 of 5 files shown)


muddyfish (approved):

dannycjones (approved):
nice

PR #1086, opened 2024-10-28
Add checksum algorithm to ListObjectsV2 response

dannycjones:
## Description of change

For a consumer of the S3 client, we need to return the checksum algorithm used with objects.

This change exposes the checksum algorithm that is sometimes returned as part of a ListObjectsV2 request. This functionality is not opt-in, and will now simply be exposed to be used by consumers of Mountpoint S3 client where supported.

This change also updates `ChecksumAlgorithm` to have a new `Unknown` variant. This is a common pattern in AWS SDKs. This will allow clients to recognize where an unknown algorithm is returned (should the S3 service start supporting additional algorithms), and avoid either returning `None` (if optional) or panicking when the error may be recoverable.

Relevant issues: N/A

Yes, it changes the `ObjectInfo` and `ChecksumAlgorithm` structs. Specifically, they are now marked `non_exhaustive` meaning that new fields could be added in the future (as has been done in this PR).

Yes, relevant change logs for breaking change (`non_exhaustive`) as well as new feature (exposing checksum algorithm) have been added to `mountpoint-s3-client`'s changelog.

---

### DIFF
--- a/mountpoint-s3-client/CHANGELOG.md
+++ b/mountpoint-s3-client/CHANGELOG.md
@@ -20,6 +20,7 @@
   The structure itself is not required to specify anything to achieve the existing behavior.
   ([#1083](https://github.com/awslabs/mountpoint-s3/pull/1083))
 * Both `ObjectInfo` and `ChecksumAlgorithm` structs are now marked `non_exhaustive`, to indicate that new fields may be added in the future.
+  `ChecksumAlgorithm` no longer implements `Copy`.
   ([#1086](https://github.com/awslabs/mountpoint-s3/pull/1086))
 
 ## v0.11.0 (October 17, 2024)
--- a/mountpoint-s3-client/CHANGELOG.md
+++ b/mountpoint-s3-client/CHANGELOG.md
@@ -7,6 +7,9 @@
   ([#1083](https://github.com/awslabs/mountpoint-s3/pull/1083))
 * Expose checksum algorithm in `ListObjectsResult`'s `ObjectInfo` struct.
   ([#1086](https://github.com/awslabs/mountpoint-s3/pull/1086))
+* `ChecksumAlgorithm` has a new variant `Unknown(String)`,
+  to accomodate algorithms not recognized by the client should they be added in future.
+  ([#1086](https://github.com/awslabs/mountpoint-s3/pull/1086))
 
 ### Breaking changes
 
--- a/mountpoint-s3-client/CHANGELOG.md
+++ b/mountpoint-s3-client/CHANGELOG.md
@@ -5,6 +5,8 @@
 * Add parameter to request checksum information as part of a `HeadObject` request.
   If specified, the result should contain the checksum for the object if available in the S3 response.
   ([#1083](https://github.com/awslabs/mountpoint-s3/pull/1083))
+* Expose checksum algorithm in `ListObjectsResult`'s `ObjectInfo` struct.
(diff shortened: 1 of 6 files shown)


mountpoint-s3-client/src/mock_client.rs
```diff
@@ -472,6 +474,24 @@ impl MockObject {
...
+            checksum_sha256,
+        } = &self.checksum;
+
+        match (checksum_crc32, checksum_crc32c, checksum_sha1, checksum_sha256) {
+            (Some(_), None, None, None) => Some(ChecksumAlgorithm::Crc32),
+            (None, Some(_), None, None) => Some(ChecksumAlgorithm::Crc32c),
+            (None, None, Some(_), None) => Some(ChecksumAlgorithm::Sha1),
+            (None, None, None, Some(_)) => Some(ChecksumAlgorithm::Sha256),
+            (None, None, None, None) => None,
+            _ => unreachable!("should not have more than one checksum"),
```
passaro, line 492: Is `unreachable` appropriate here? It seems inconsistent with the comment on the field:
```
    /// Typically, at most one of the checksums should be set.
```

dannycjones: Yeah, it is leaving the door open in case the assumption of only one checksum should change but it also means we'd crash here.

I can update the comment to assume there will _never_ be more than one checksum. Alternatively, we can just return the first one set (by replacing `None` with `_` in the patterns).

I'd inclined to do the latter, what do you think?

passaro: Let's go with the latter.

mountpoint-s3-client/src/object_client.rs
```diff
@@ -707,6 +711,28 @@ impl Checksum {
             checksum_sha256: None,
         }
     }
+
+    /// Provide [ChecksumAlgorithm] for the [Checksum], if set and recognized.
+    ///
+    /// This method assumes that at most one checksum will be set and will return the first matched.
+    /// This method cannot account for any new checksum algorithms that may be introduced in the future,
+    /// returning [None] when no algorithm is recognized.
```
passaro, line 719: The last sentence does not seem relevant: it's `Checksum` itself that does not account for new algorithms, rather than this method in particular.

passaro (commented):
Looks good, just a couple of minor suggestions.

mountpoint-s3-client/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 * Add parameter to request checksum information as part of a `HeadObject` request.
   If specified, the result should contain the checksum for the object if available in the S3 response.
   ([#1083](https://github.com/awslabs/mountpoint-s3/pull/1083))
+* Expose checksum algorithm in `ListObjectsResult`'s `ObjectInfo` struct.
+  ([#1086](https://github.com/awslabs/mountpoint-s3/pull/1086))
+* `ChecksumAlgorithm` has a new variant `Unknown(String)`,
```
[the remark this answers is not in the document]
dannycjones: I have a separate note in the breaking changes section for that.

mountpoint-s3-client/src/object_client.rs
dannycjones: Fair enough, will remove. We already have "if set and recognized" in the first line.

passaro (approved):

PR #1093, opened 2024-10-31
Update ChecksumAlgorithm field of client's ObjectInfo struct to be Vec over single element

dannycjones:
## Description of change

On reviewing the S3 API documentation, the checksum algorithm field is a list of algorithms. Additionally, when reviewing other SDKs such as the [Rust SDK, we see that they are presenting this field as `Option<Vec<String>>`](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/types/struct.Object.html) rather than a single optional element. (Note, we do drop the `Option` still.

We'd prefer to align with the SDK interface. Our tenet here is to ensure our S3 client is consistent with the official SDKs where there's no significant effort required. This is making a breaking change while we're already planning to make a number of breaking changes to the client.

Relevant issues:

- Follow up on #1086, which added checksum algorithms to the list objects response.

Yes, it changes the S3 client behavior to return a different type. We are however merging this before a new crate release, so this will not be an additional breaking change.

There's no behavior change to Mountpoint file system.

There is already an existing entry in `mountpoint-s3-client`'s changelog. This PR has been added to the list of PRs for that entry.

---

### DIFF
--- a/mountpoint-s3-client/CHANGELOG.md
+++ b/mountpoint-s3-client/CHANGELOG.md
@@ -6,7 +6,8 @@
   If specified, the result should contain the checksum for the object if available in the S3 response.
   ([#1083](https://github.com/awslabs/mountpoint-s3/pull/1083))
 * Expose checksum algorithm in `ListObjectsResult`'s `ObjectInfo` struct.
-  ([#1086](https://github.com/awslabs/mountpoint-s3/pull/1086))
+  ([#1086](https://github.com/awslabs/mountpoint-s3/pull/1086)),
+  ([#1093](https://github.com/awslabs/mountpoint-s3/pull/1093))
 * `ChecksumAlgorithm` has a new variant `Unknown(String)`,
   to accomodate algorithms not recognized by the client should they be added in future.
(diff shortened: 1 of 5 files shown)


mountpoint-s3-client/CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
   If specified, the result should contain the checksum for the object if available in the S3 response.
   ([#1083](https://github.com/awslabs/mountpoint-s3/pull/1083))
 * Expose checksum algorithm in `ListObjectsResult`'s `ObjectInfo` struct.
-  ([#1086](https://github.com/awslabs/mountpoint-s3/pull/1086))
+  ([#1086](https://github.com/awslabs/mountpoint-s3/pull/1086)),
+  ([#1093](https://github.com/awslabs/mountpoint-s3/pull/1093))
```
passaro, line 10: nit:
```suggestion
  ([#1086](https://github.com/awslabs/mountpoint-s3/pull/1086),
  [#1093](https://github.com/awslabs/mountpoint-s3/pull/1093))
```

passaro (approved):
LGTM!

PR #1096, opened 2024-11-04
Encapsulate test resources in `TestSession` struct

unexge:
## Description of change

Currently we return tuples from session constructor functions. This makes adding new resources to testing session challenging. Wrapping all testing resources in `TestSession` will help us to add new resources easily and it will also simply the type signatures of the functions using the test session.

No, just tests.

No

---

### DIFF
--- a/mountpoint-s3/tests/common/fuse.rs
+++ b/mountpoint-s3/tests/common/fuse.rs
@@ -75,6 +75,20 @@ impl TestSessionConfig {
     }
 }
 
+// Holds resources for the testing session and cleans them on drop.
+pub struct TestSession {
+    pub mount_dir: TempDir,
+    pub session: BackgroundSession,
+    pub test_client: TestClientBox,
+}
+
+pub trait TestSessionCreator: FnOnce(&str, TestSessionConfig) -> TestSession {}
+
+// Since trait aliases are not stable yet, we can't just `type TestSessionCreator = FnOnce(...)`.
+// As a workaround we can subtrait `FnOnce(...)` and have this blank impl to allow
+// `FnOnce(...)` in place of `impl TestSessionCreator`.
+impl<T> TestSessionCreator for T where T: FnOnce(&str, TestSessionConfig) -> TestSession {}
+
 fn create_fuse_session<Client, Prefetcher>(
     client: Client,
     prefetcher: Prefetcher,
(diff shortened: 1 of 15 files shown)


mountpoint-s3/tests/fuse_tests/write_test.rs
```diff
@@ -336,7 +327,9 @@ where
         filesystem_config,
         ..Default::default()
     };
-    let (mount_point, _session, _test_client) = creator_fn("fstat_after_writing", session_config);
+    let test_session = creator_fn("fstat_after_writing", session_config);
+    let mount_point = test_session.mount_dir;
+    let _test_client = test_session.test_client;
```
dannycjones, line 332: I believe we can drop this.

```suggestion
```

If so, can we apply it everywhere for unused `_test_client`?

mountpoint-s3/tests/common/fuse.rs
```diff
@@ -75,6 +75,20 @@ impl TestSessionConfig {
     }
 }
 
+// Holds resources for the testing session and cleans them on drop.
+pub struct TestSession {
+    pub mount_dir: TempDir,
+    pub session: BackgroundSession,
+    pub test_client: TestClientBox,
+}
```
dannycjones, line 83: Not blocking: we could make these private if we really wanted to, with `mount_dir(&self) -> &Path` and `test_client(&mut self) -> &mut TestClientBox`.

I don't believe the session is even used by anyone (but needs to stick around).

dannycjones (changes requested):
Thanks for giving some love to the tests!

I have just a minor change to remove cases where we have `let _test_client = test_session.client`, since its unnecessary.

mountpoint-s3/tests/fuse_tests/write_test.rs
unexge: Good catch! Fixed with https://github.com/awslabs/mountpoint-s3/pull/1096/commits/8506e33761d9cd56d982401c4e0ae6326e9574b1

mountpoint-s3/tests/common/fuse.rs
unexge: It's a good point but since we try to borrow `TestSession` as both immutable (via `mount_dir(&self)`) and mutable (via `test_client(&mut self)`) it causes `cannot borrow test_session as mutable because it is also borrowed as immutable`. I think it requires more refactoring to properly seperate them.

dannycjones (approved):
LGTM!

mountpoint-s3/tests/common/fuse.rs
dannycjones: Ah, makes sense. I was thinking we probably want to add interior mutability to the mock client but all good for now.

PR #1101, opened 2024-11-05
Remove 16MiB part size override in FIO read benchmarks

dannycjones:
## Description of change

The read benchmarks currently override the part size. It's unclear why this was added (#779), but we think its best to remove the override such that the benchmarks represent the default configuration.

Relevant issues: N/A

It changes benchmark configuration. We expect the effect to be minimal.

No impact to Mountpoint itself.

No.

---


adpeace:
I am working on a refactoring in this bit of code: one thing to watch out for is that the file layout will fail for the 100GB files if the part size isn't large enough.

PR #1102, opened 2024-11-05
Implement checksums for MockClient upload path

dannycjones:
## Description of change

This change updates the mock client's write path to compute and store checksums, matching behavior approximate to S3. We want this so that we can use the mock client for both uploads and downloads and verify checksum behavior for the client.

The change stores the checksums as strings, as this is the observed behavior of S3. We can always update if this turns out to be a bad assumption.

Relevant issues: N/A

This change updates the `mountpoint-s3-client` mock client to add checksum persistence to the write path. This was previously a gap in the mock client.

I do not think this needs a changelog entry, as it does not change the behavior of the S3 client itself. Happy to discuss.

---

### DIFF
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -1734,8 +1734,7 @@ mod tests {
         let s3_key = "key1";
         let content = vec![42u8; 512];
         let content_checksum = crc32c::checksum(&content);
-        let put_object_params =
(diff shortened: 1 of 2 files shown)


passaro (dismissed):

dannycjones:
Thanks @passaro, I also hooked up GetObjectAttributes while we're here: https://github.com/awslabs/mountpoint-s3/pull/1102/commits/051ced8d7f78605ba440d8d0aeecd7bb86d873db

passaro (approved):
LGTM!

PR #1104, opened 2024-11-06
Don't run benchmarks with debug, plus minor refactoring

adpeace:
## Description of change

Remove the `--debug` flag when mounting S3 for the benchmarks.

This is now available via an S3_DEBUG environment variable which, when set, will add `--debug` back again.

Also, within `fs_bench.sh`, merge the read and write benchmark methods into a single one, which is paramterized, since they were almost identical.  This avoids having to make the change described above in two places and simplifies the code going forwards.

This changes the benchmarks to run without --debug to the mount command, which creates a discontinuity in benchmark results, and may improve them (though there's no actual performance improvement here).

No.

---

### DIFF
--- a/mountpoint-s3/scripts/fs_bench.sh
+++ b/mountpoint-s3/scripts/fs_bench.sh
@@ -104,8 +104,20 @@ should_run_job() {
     fi
 }
 
-read_benchmark () {
-  jobs_dir=mountpoint-s3/scripts/fio/read
+# Run all benchmarks within a category.  Fio job definitions should exist under a directory
+# with the category name, passed as first argument, inside mountpoint-s3/scripts/fio.
+#
+# Params:
+# $1: benchmark category.
+run_benchmarks() {
+  category=$1
+  jobs_dir=mountpoint-s3/scripts/fio/$category
+
+  if [ $category == "read" ]; then
+    part_size="--part-size=16777216"
+  else
+    part_size=""
+  fi
 
   for job_file in "${jobs_dir}"/*.fio; do
 
@@ -122,7 +134,7 @@ read_benchmark () {
 
     # cleanup mount directory and log directory
     cleanup() {
-      echo "read_benchmark:cleanup"
+      echo "${category}_benchmark:cleanup"
       # unmount file system only if it is mounted
       ! mountpoint -q ${mount_dir} || sudo umount ${mount_dir}
       rm -rf ${mount_dir}
@@ -143,7 +155,7 @@ read_benchmark () {
       --allow-delete \
       --log-directory=${log_dir} \
       --prefix=${S3_BUCKET_TEST_PREFIX} \
-      --part-size=16777216 \
+      $part_size \
       ${optional_args}
     mount_status=$?
     set -e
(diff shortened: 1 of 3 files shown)


mountpoint-s3/scripts/fs_bench.sh
```diff
@@ -36,6 +36,10 @@ if [[ -n "${S3_ENDPOINT_URL}" ]]; then
   optional_args+="--endpoint-url=${S3_ENDPOINT_URL}"
 fi
 
+if [[ -n "${S3_DEBUG}" ]]; then
+  optional_args+= " --debug"
+fi
```
dannycjones, line 41: nit (and in the other files)

```suggestion
if [[ -n "${S3_DEBUG}" ]]; then
  optional_args+=" --debug"
fi
```

mountpoint-s3/scripts/fs_bench.sh
```diff
@@ -104,8 +108,20 @@ should_run_job() {
...
 }
 
-read_benchmark () {
-  jobs_dir=mountpoint-s3/scripts/fio/read
+# Run all benchmarks within a category.  Fio job definitions should exist under a directory
+# with the cateogry name, passed as first argument, inside mountpoint-s3/scripts/fio.
+#
+# Params:
+# $1: benchmark category.
+run_benchmarks() {
```
dannycjones, line 116:
```suggestion
# Run all benchmarks within a category.  Fio job definitions should exist under a directory
# with the category name, passed as first argument, inside mountpoint-s3/scripts/fio.
#
# Params:
# $1: benchmark category.
run_benchmarks() {
```

mountpoint-s3/scripts/fs_bench.sh
```diff
@@ -104,8 +108,20 @@ should_run_job() {
...
+  category=$1
+  jobs_dir=mountpoint-s3/scripts/fio/$category
+
+  if [ $category == "read" ]; then
+    part_size="--part-size=16777216"
+  else
+    part_size=""
+  fi
```
dannycjones, line 124: I want to suggest that we use `--write-part-size`, to only impact the files uploaded.

This way, the read pattern matches that of Mountpoint's defaults (even though the objects in S3 may be different from those uploaded by default).

dannycjones (commented):
Just putting early review

mountpoint-s3/scripts/fs_bench.sh
adpeace: thanks, will fix.

mountpoint-s3/scripts/fs_bench.sh
adpeace: thanks, will fix.

mountpoint-s3/scripts/fs_bench.sh
adpeace: I intentionally didn't do that because I am trying to preserve the existing behvaiour with this change, as it's intended to only be a refactor.  Therefore, I'd prefer to leave as-is, but let's discuss if you don't agree.

dannycjones: Fair enough, happy to discuss in a later change

dannycjones (approved):
LGTM!

PR #1108, opened 2024-11-06
Lay out files using fio in the benchmark.

adpeace:
## Description of change

This makes the benchmarks more self-contained, creating the state that they need to run rather than relying on pre-created state (potentially with different/unknown mount options and/or object properties).

It does change behaviour a little: previously the multi-thread tests would use the same object whereas now each thread uses its own object -- arguably this is a more useful test but results in higher S3 usage.

Also note that the cache benchmark is unchanged in this commit, because it makes assumptions about the filename used by the tests.

Removing the assumption that each test will operate on a single file prepares us for future mixed read/write tests, and allows different fio jobs to be run in parallel safely.

Yes, see above: previously the multi-thread tests would use the same object whereas now each thread uses its own object -- arguably this is a more useful test but results in higher S3 usage.

No.

---

### DIFF
--- a/doc/BENCHMARKING.md
+++ b/doc/BENCHMARKING.md
@@ -42,6 +42,7 @@ You can use the following steps.
         export S3_BUCKET_NAME=bucket_name
         export S3_BUCKET_TEST_PREFIX=prefix_path/
         export S3_ENDPOINT_URL=endpoint_url # optional
+        # these filenames only needed by cache benchmark
         export S3_BUCKET_BENCH_FILE=bench_file_name
         export S3_BUCKET_SMALL_BENCH_FILE=small_bench_file_name
         # to filter by job name; e.g. to only run small jobs
--- a/mountpoint-s3/scripts/fio/read_latency/ttfb.fio
+++ b/mountpoint-s3/scripts/fio/read_latency/ttfb.fio
@@ -1,9 +1,11 @@
 [global]
 name=fs_bench
-bs=1B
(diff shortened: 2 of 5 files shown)


adpeace:
Here's the benchmark failure.  Not sure why these benchmarks are regressing, am investigating now.

Possible performance regression was detected for benchmark.
Benchmark result of this commit is worse than the previous benchmark result exceeding threshold `2`.

| Benchmark suite | Current: ac81bb4f34199700d5b015401778875ec3d6536c | Previous: 50433e6a7b37b692abd5075aff355976f233006b | Ratio |
|-|-|-|-|
| `random_read_four_threads_direct_io` | `4.12275390625` MiB/s | `20.4248046875` MiB/s | `4.95` |
| `random_read_four_threads` | `4.3408203125` MiB/s | `19.6359375` MiB/s | `4.52` |
| `random_read_direct_io` | `1.01923828125` MiB/s | `3.701171875` MiB/s | `3.63` |
| `random_read` | `1.19501953125` MiB/s | `4.25361328125` MiB/s | `3.56` |

This comment was automatically generated by [workflow](https://github.com/awslabs/mountpoint-s3/actions?query=workflow%3ABenchmarks%20(PR)) using [github-action-benchmark](https://github.com/marketplace/actions/continuous-benchmark).

adpeace:
Requesting 2 reviews due to the above issue.

passaro (dismissed):

dannycjones (dismissed):

dannycjones (approved):

PR #1111, opened 2024-11-07
Add debug print for inode number in mknod

c-hagem:
## Description of change

`mknod` now prints the newly created inode number. This can help with tracing the lifetime of an inode between FUSE requests.

No, small debug log addition only.

No.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -406,6 +406,10 @@ where
             .superblock
             .create(&self.client, parent, name, InodeKind::File)
             .await?;
+        // Log newly created Inode number
+        let inode_nr = lookup.inode.ino();
+        debug!("mknod: created inode number {inode_nr}");
+
         let attr = self.make_attr(&lookup);
         Ok(Entry {
             ttl: lookup.validity(),
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -406,10 +406,7 @@ where
             .superblock
             .create(&self.client, parent, name, InodeKind::File)
             .await?;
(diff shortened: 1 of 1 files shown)


mountpoint-s3/src/fs.rs
```diff
@@ -406,6 +406,10 @@ where
             .superblock
             .create(&self.client, parent, name, InodeKind::File)
             .await?;
+        // Log newly created Inode number
+        let inode_nr = lookup.inode.ino();
+        debug!("mknod: created inode number {inode_nr}");
```
dannycjones, line 411: I'd take some inspiration from this line:

https://github.com/awslabs/mountpoint-s3/blob/534918e96337dee222b158df9d4bd92a05d791b9/mountpoint-s3/src/fs.rs#L342

I'd probably write something like this, such that it matches the format of other logs. I'd also drop the comment since its implied.

```suggestion
        debug!(ino = lookup.inode.ino(), "new inode created");
```

This should format the field nicely, and `tracing` library should ensure context of `mknod` is already captured.

Hopefully it'll look something like this:

```
2024-11-07T15:02:24.409045Z DEBUG mknod{req=12 parent=1 name="hello.md"}: mountpoint_s3_client::fs: new inode created ino=5
```

dannycjones (commented):
Nice, I have some styling comments but this is the right spot and content.

dannycjones:
@c-hagem we use the PR description in the commits to `main` now. Can you update the description to be as you would like when it is merged?

No changelog entry is needed!

dannycjones (approved):
LGTM!

I updated the PR description as this will be included in the commit message when merged.

Congratulations on the first PR merged! :)

PR #1117, opened 2024-11-07
Bypass the shared cache for large objects

vladem:
## Description of change

This change makes `get_block` and `put_block` for objects larger than `1MiB` be a no-op in the shared cache. 

Relevant issues: N/A

No, it is under the feature flag.

Yes, in the following PRs.

---

### DIFF
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -28,8 +28,8 @@ use regex::Regex;
 use sysinfo::{RefreshKind, System};
 
 use crate::data_cache::{
-    CacheLimit, DiskDataCache, DiskDataCacheConfig, ExpressDataCache, ManagedCacheDir, MultilevelDataCache,
-    EXPRESS_CACHE_MAX_OBJECT_SIZE,
+    CacheLimit, DiskDataCache, DiskDataCacheConfig, ExpressDataCache, ExpressDataCacheConfig, ManagedCacheDir,
+    MultilevelDataCache, EXPRESS_CACHE_MAX_OBJECT_SIZE,
 };
 use crate::fs::{CacheConfig, ServerSideEncryption, TimeToLive};
 use crate::fuse::session::FuseSession;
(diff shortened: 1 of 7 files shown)


mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -247,4 +275,37 @@ mod tests {
...
+        // PUT and GET for a large object should be no-op
+        cache
+            .put_block(cache_key_1.clone(), 0, 0, data_1.clone(), data_1.len())
+            .await
+            .expect("cache should be accessible");
+        let get_result = cache
+            .get_block(&cache_key_1, 0, 0, data_1.len())
+            .await
+            .expect("cache should be accessible");
+        assert!(get_result.is_none());
```
passaro, line 309: Should we check the client as well here?

mountpoint-s3/src/data_cache/multilevel_cache.rs
```diff
@@ -143,7 +155,13 @@ mod tests {
         let client = MockClient::new(config);
         (
             client.clone(),
-            ExpressDataCache::new(bucket, client, "unique source description", BLOCK_SIZE),
+            ExpressDataCache::new(
```
passaro, line 158: We should also add a test for the multilevel cache.

mountpoint-s3/src/cli.rs
```diff
@@ -888,6 +889,7 @@ where
                 client.clone(),
                 &args.bucket_name,
                 args.cache_block_size_in_bytes(),
+                EXPRESS_CACHE_MAX_OBJECT_SIZE,
```
passaro, line 892: Time to add a `Config` type for the express cache? We could set a default value for max_object_size there.

passaro (commented):
A couple of suggestions, none blocking.

mountpoint-s3/src/data_cache/express_data_cache.rs
vladem: added `assert_eq!(client.object_count(), 0, "cache must be empty");`

mountpoint-s3/src/cli.rs
vladem: I can add the following struct to group configuration arguments of `ExpressDataCache::new`:
```rust
/// Configuration for a [ExpressDataCache].
#[derive(Debug)]
pub struct ExpressDataCacheConfig {
    /// Name of a S3 Express bucket to store the blocks.
    bucket_name: String,
    /// Name of the mounted bucket.
    source_bucket_name: String,
    /// Size of data blocks.
    block_size: u64,
    /// The maximum size of an object to be cached.
    max_object_size: usize,
}
```

I wouldn't implement a `Default` for it though, because only `max_object_size` has a sensible default value. I'd just leave `EXPRESS_CACHE_MAX_OBJECT_SIZE` constant for that purpose.

muddyfish (approved):

passaro (approved):
LGTM!

PR #1123, opened 2024-11-11
Add support for getting object checksums in GetObject

muddyfish:
## Description of change

- Adds new method `get_object_checksum` to `get_object` to retrieve the whole object checksum.
- Adds new parameter `checksum_mode` to `GetObjectParams` to configure if we want to request checksums from S3.
  -  If checksums aren't requested, `get_object_checksum` returns an error.
- Refactor `S3CrtClient` to store a cell of the object headers rather than metadata.

Relevant issues: 

No

Yes

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client/get_object.rs
+++ b/mountpoint-s3-client/src/s3_crt_client/get_object.rs
@@ -10,23 +10,26 @@ use std::task::{Context, Poll};
 use futures::channel::mpsc::UnboundedReceiver;
 use futures::Stream;
 use mountpoint_s3_crt::common::error::Error;
-use mountpoint_s3_crt::http::request_response::Header;
+use mountpoint_s3_crt::http::request_response::{Header, Headers};
 use mountpoint_s3_crt::s3::client::MetaRequestResult;
 use pin_project::pin_project;
 use thiserror::Error;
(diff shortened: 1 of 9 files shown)


mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -95,27 +106,16 @@ impl S3CrtClient {
                 // Only set metadata if we have a 2xx status code. If we only get other status
```
passaro, line 106: nit: set headers

mountpoint-s3-client/tests/get_object.rs
```diff
@@ -465,3 +466,88 @@ async fn test_get_object_user_metadata_after_stream(size: usize, metadata: HashM
         .expect("should return metadata");
     assert_eq!(actual_metadata, metadata);
 }
+
+#[test_case(ChecksumAlgorithm::Crc32; "Checksum CRC32")]
```
passaro, line 470: nit: do you need the explicit label? isn't the enum enough to build a meaningful test case name?

passaro (commented):
Looks good. Only a couple of minor suggestions.

passaro (approved):

PR #1125, opened 2024-11-11
Verify object metadata in Express cache

muddyfish:
## Description of change

- Verify S3 Express cache objects have valid object metadata which matches the keys
- Verifies the CRC32C of the object content post-download from S3
  - If checksum is missing, return `BlockChecksumMissing`.

Relevant issues: N/A

Changes S3 Express cache to require object metadata and CRC32C to be present. Old caches will not be used. 

No

---

### DIFF
--- a/mountpoint-s3/src/data_cache/express_data_cache.rs
+++ b/mountpoint-s3/src/data_cache/express_data_cache.rs
@@ -5,7 +5,7 @@ use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheError, DataCacheRe
 
 use async_trait::async_trait;
 use base64ct::{Base64, Encoding};
-use bytes::{Bytes, BytesMut};
+use bytes::BytesMut;
 use futures::{pin_mut, StreamExt};
 use mountpoint_s3_client::error::{GetObjectError, ObjectClientError};
 use mountpoint_s3_client::types::{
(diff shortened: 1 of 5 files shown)


mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -1,18 +1,135 @@
...
+    block_offset: u64,
+    etag: String,
+    source_key: String,
+    source_bucket_name: String,
+    header_checksum: u32,
+}
+
+impl BlockHeader {
+    pub fn new(block_idx: BlockIndex, block_offset: u64, cache_key: &ObjectId, source_bucket_name: &str) -> Self {
+        let header_checksum = Self::get_header_checksum(block_idx, block_offset, cache_key, source_bucket_name).value();
```
passaro, line 42: The "header_checksum" should depend on the data_checksum (checksum of the data in the block) as well.

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -246,9 +376,41 @@ mod tests {
...
             data_1, entry,
             "cache entry returned should match original bytes after put"
         );
     }
+
+    proptest! {
+        #[test]
+        fn proptest_block_header_creates_small_s3_keys(block_header: BlockHeader) {
+            // Ensure we can always serialise to S3 keys smaller than 1kb
+            prop_assert!(block_header.to_s3_key("prefix").len() <= 1024);
```
passaro, line 390: This needs to also vary the prefix.

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -117,42 +253,33 @@ where
...
-    );
-    format!("{}/{}/{:010}", prefix, hashed_cache_key, block_idx)
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::checksums::ChecksummedBytes;
     use crate::sync::Arc;
+    use proptest::{prop_assert, proptest};
```
passaro, line 282: We should also test validation failures

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -1,18 +1,135 @@
...
+                .chain_update(&self.etag)
+                .finalize(),
+        );
+        format!("{}/{}/{:010}", prefix, hashed_cache_key, self.block_idx)
+    }
+
+    /// Convert to object metadata that is HTTP header safe (ASCII only)
+    pub fn to_headers(&self) -> HashMap<String, String> {
+        let source_key_encoded = Base64::encode_string(self.source_key.as_bytes());
+        HashMap::from([
```
vladem, line 68: Is it a reasonable risk that headers may start going over `2KiB`? E.g. if S3 allows bigger keys in future.

We could check that the total size of the headers (keys + "x-amz-meta" + values) is less than 2KIB and do not do the request if it is not.

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -94,17 +119,36 @@ where
...
+            .map_err(|err| DataCacheError::IoFailure(err.into()))?;
+
+        let checksum = result
+            .get_object_checksum()
+            .await
+            .map_err(|err| DataCacheError::IoFailure(err.into()))?;
+        let crc32c = crc32c_from_base64(&checksum.checksum_crc32c.ok_or(DataCacheError::BlockChecksumMissing)?)
+            .map_err(|err| DataCacheError::IoFailure(err.into()))?;
+
+        let block_metadata = BlockMetadata::new(block_idx, block_offset, cache_key, &self.bucket_name, crc32c);
```
vladem, line 149: This is a bit non-symmetric, that all of the arguments to `BlockMetadata::new` are from the "expected" parameters, passed to `get_block`, with the exception of  `crc32c` which is read from the actual block. But I think this is ok, since having block's crc32 in metadata's checksum mitigates another risk.

muddyfish: I could migrate the `crc32c` field to be arguments to both `to_put_object_params` and `validate_object_metadata`, but I feel that would probably be worse

vladem (approved):

mountpoint-s3/src/data_cache/express_data_cache.rs
vladem: I think it's ok, the main part is that we actually compare other params to the expected values.

PR #1128, opened 2024-11-13
Ensure bucket passed as S3 express cache is an express bucket

muddyfish:
## Description of change

Ensures that the S3 Express cache bucket passed is an S3 Express bucket

Relevant issues: N/A

Yes. Previously general purpose buckets could be passed in, but would not be useful

No

---


muddyfish:
Closing this out. We'll implement this differently by trying to write to the cache and validating that the storage class is correct

PR #1130, opened 2024-11-13
Add mixed read/write tests to the benchmark.

adpeace:
## Description of change

As we make changes that might impact mixed read/write workloads it is useful to have some examples of these in our benchmarks.  This change adds a 20/80, 50/50, and 80/20 read/write workload to the benchmarks.

The results reporting is updated to support multiple job types in a single benchmark.  This works by continuing to average over iterations as before, but averages each job separately then sums the averages to produce the final throughput number for that benchmark (i.e. (avg(read throughput) + avg(write throughput)) for the mixed benchmarks).

No.

No.

---


passaro (approved):
LGTM!

PR #1134, opened 2024-11-14
Move PR desc instructions from template to CONTRIBUTING.md

dannycjones:
## Description of change

Until this change, we were using HTML comments in the PR template to provide instructions to contributors so they know what to include in a PR title and description. Since changing the default on GitHub to use the PR description as the squash commit message, we now see the HTML comments in comment messages which is not desired at all.

This change replaces HTML comments with non-comment TODOs which should be addressed and removed. These are visible to reviewers, who should prompt the author to address them before merging.

We move some of the more detailed instructions into `CONTRIBUTING.md` which is where we describe the contribution process more broadly.

There's some minor simplification to the template given we can no longer provide clear instructions via HTML comment.

This changes the default description for code contributions to the repository only.

No change to the file system or S3 client crates.

No.

---


passaro (approved):
LGTM!

PR #1137, opened 2024-11-14
Set max_background FUSE config to 64 by default.

adpeace:
This improves sequential read performance on instances with multiple 100Gbps network interfaces.  It controls the number of requests that are allowed in the pending queue that are classified as background, which includes at least some read requests.  It also indirectly controls the "congestion threshold", which is set by default to 75% of the max background value.  When the congestion threshold is reached, FUSE will stop sending the asynchronous part of readaheads from paged IO to the filesystem.

Testing on 2 NIC instances shows up to approximately 50% speed-up on a sequential read workload with 32 open files, from 65 to 99Gbps, for paged IO.  Although we don't have enough instrumentation to fully understand the change in queueing behaviour in FUSE, we think it is likely because we're able to serve sufficient readahead requests for the object before hitting the congestion threshold when the limit is higher, thus allowing mountpoint to start prefetching later parts of the object sooner.

The value of 64 was picked by experimentation with values between 16 (the default) and 256, as well as specifically setting the congestion threshold.  Increasing the value generally led to better performance up to 64, after which performance doesn't improve further (at least not significantly).  We wanted to choose the lowest value that seemed reasonable for the desired performance improvement, to reduce the chance of affecting a workload that wasn't being tested.

As well as the standard regression tests, the change was tested on trn1 instances with a 256KB sequential read workload reading 32 files in parallel over 1, 2, and 4 network interfaces.  It does not regress our standard benchmarks nor performance on this test with 1 NIC in use.

This change also temporarily introduces two environment variables to tune the behaviour, so we can isolate this change if a particular workload is found to regress.

This improves performance on large instance types.  There's a risk of regression for workloads we don't test.

Yes, will submit a separate PR.

---


adpeace:
Note that i updated the PR comment reducing the quoted improvement because I had originally misquoted the 1 NIC test result as the baseline, instead of the baseline 2 NIC test result.

mountpoint-s3/src/fs.rs
```diff
@@ -155,6 +155,34 @@ where
...
+                .expect("invalid env var value for UNSTABLE_MAX_BACKGROUND");
+            let old = config
+                .set_max_background(max_background)
+                .expect("unable to set max background");
+            tracing::warn!("set max background to {} from {}", max_background, old)
+        } else {
+            let _ = config
+                .set_max_background(64)
+                .expect("unable to set max background to default value");
+        }
```
dannycjones, line 173: Since we're putting this in the `main` branch, I'd rather us handle parsing and setting errors properly and gracefully fail.

As an example, this is a change where we introduced an override for setting the prefetch read window size: https://github.com/awslabs/mountpoint-s3/commit/de6d145c2343b3bce433d1368c71f7eaf2f2d4d7

dannycjones (changes requested):
Overall happy with the change, feedback is on how to handle config name and what happens when configuration goes wrong.

mountpoint-s3/src/fs.rs
```diff
@@ -155,6 +155,34 @@ where
 
     pub async fn init(&self, config: &mut KernelConfig) -> Result<(), libc::c_int> {
         let _ = config.add_capabilities(fuser::consts::FUSE_DO_READDIRPLUS);
+        // Set max_background FUSE parameter to 64 by default, or override with environment variable:
+        // NOTE: Support for this environment variable may be removed in future without notice.
+        if let Some(user_max_background) = std::env::var_os("UNSTABLE_MAX_BACKGROUND") {
```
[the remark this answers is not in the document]
adpeace: Ack, will change, thanks for pointing this out.

mountpoint-s3/src/fs.rs
dannycjones: Discussed a bit offline, this is where we're landing.

- If customer is intending to modify an unstable configuration but we don't understand it or can't use it, we want to be loud and tell them its not working. (Crash preferable?)
- If customer sets some unstable configuration, we want to confirm it so they're reassured its applied and can be reminded that an unstable configuration is in use.
- If we remove it in the future, we want to be able to silently take it away without breaking the existing use case (as is the 'contract' with most of these unstable features). In this case, we should silently fail (maybe with a warning for a bit if we feel it is necessary enough).


Given this, keeping the crash here is the right choice (in lieu of plumbing everything through to the CLI error reporting which I don't want to recommend right now).

adpeace: Ack, thanks.  I have still improved the error messages. One thing I hadn't realized initially is that panics here don't actually propagate to the CLI (and this pattern already exists elsewhere), in that the mount will fail but the CLI prints the "success" error message and the metrics thread will continue to run (even with `-f`).  I haven't debugged fully yet though think this could be related to the context in which `init` is called, probably from a fuse thread.

I think we should merge as-is anyway, and then later refactor all similar panics to address this issue, since the unstable flag is experimental, and a failure wouldn't be encountered during normal code paths (setting `max_background` only fails if `0` is passed, which it isn't in the default/safe code path).

mountpoint-s3/src/fs.rs
```diff
@@ -153,8 +153,53 @@ where
...
+                .unwrap_or_else(|_| panic!("Unable to set FUSE max_background configuration to {}", max_background));
+            tracing::warn!(
+                "Successfully overridden FUSE max_background configuration to {} (was {}) from unstable environment variable.",
+                max_background,
+                old
+            );
+        } else {
+            let _ = config
+                .set_max_background(64)
+                .expect("unable to set FUSE max_background to default value");
```
dannycjones, line 185: Since the customer isn't overriding anything, I'd prefer we just warn here.

```suggestion
            const DEFAULT_MAX_BACKGROUND: u16 = 64;
            let max_background_result = config
                .set_max_background(DEFAULT_MAX_BACKGROUND);
            if max_background_result.is_err() {
                tracing::warn!("failed to set FUSE max_background to {}, using Kernel default", DEFAULT_MAX_BACKGROUND);
            }
```

Especially since panics here aren't propagating to a mount failure right now.

mountpoint-s3/src/fs.rs
```diff
@@ -153,8 +153,53 @@ where
...
+            let congestion_threshold =
+                Self::parse_env_var_to_u16(ENV_VAR_KEY_CONGESTION_THRESHOLD, user_congestion_threshold);
+            let old = config
+                .set_congestion_threshold(congestion_threshold)
+                .unwrap_or_else(|_| panic!("unable to set FUSE congestion_threshold to {}", congestion_threshold));
+            tracing::warn!(
+                "Successfully overridden FUSE congestion_threshold configuration to {} (was {}) from unstable environment variable.",
+                congestion_threshold, // Note: fixed a bug here, was using max_background
+                old
+            );
```
dannycjones, line 200: I don't think the comment makes sense here, it was only relevant during code review?

```suggestion
            tracing::warn!(
                "Successfully overridden FUSE congestion_threshold configuration to {} (was {}) from unstable environment variable.",
                congestion_threshold,
                old
            );
```

dannycjones (commented):
When the non-override path is updated to avoid crashing, happy for this to go in. (We covered performance in previous iterations, happy to skip the perf steps)

I was playing a little with the error path here, I think ultimately we return 'success' too early when running as a daemon... However, we're also not propagating the error when using `--foreground`.

Let's get this change in now.

mountpoint-s3/src/fs.rs
adpeace: I personally marginally prefer the original, because this should never happen (setting the config should only fail if we don't give it a positive integer).  However, I think it's fine either way so will update.

dannycjones (approved):
LGTM!

PR #1140, opened 2024-11-15
Fix flaky write_with_sse_kms_key_id_ok test

c-hagem:
Fixes the flakiness of write_with_sse_kms_key_id_ok test, which was caused by not properly unmounting and dropping child. 

This is not a breaking change; no changelog entry required (as this just fixes a test).

Before this change, this test fails in ~10  out of 100 runs, after this change it fails 0 times out of 100 runs.

---

### DIFF
--- a/mountpoint-s3/tests/fuse_tests/fork_test.rs
+++ b/mountpoint-s3/tests/fuse_tests/fork_test.rs
@@ -828,10 +828,11 @@ fn write_with_sse_kms_key_id_ok() {
     assert!(kms_key_arn.starts_with("arn:") && kms_key_arn.contains(":key"));
     let (bucket, prefix) = get_test_bucket_and_prefix("write_with_sse_kms_key_id_ok");
     let mount_point = assert_fs::TempDir::new().expect("can not create a mount dir");
-    let _ = mount_with_sse(&bucket, mount_point.path(), &prefix, &kms_key_arn, None);
+    let child = mount_with_sse(&bucket, mount_point.path(), &prefix, &kms_key_arn, None);
     let f_name = "f.txt";
     write_to_file(mount_point.path(), f_name).expect("should be able to write to the file");
     unmount(&mount_point);
+    wait_for_exit(child);
 
     let sdk_client = tokio_block_on(get_test_sdk_client(get_test_region().as_str()));
     let key = format!("{}{}", prefix, f_name);


muddyfish (approved):

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -828,10 +828,11 @@ fn write_with_sse_kms_key_id_ok() {
     assert!(kms_key_arn.starts_with("arn:") && kms_key_arn.contains(":key"));
     let (bucket, prefix) = get_test_bucket_and_prefix("write_with_sse_kms_key_id_ok");
     let mount_point = assert_fs::TempDir::new().expect("can not create a mount dir");
-    let _ = mount_with_sse(&bucket, mount_point.path(), &prefix, &kms_key_arn, None);
+    let child = mount_with_sse(&bucket, mount_point.path(), &prefix, &kms_key_arn, None);
     let f_name = "f.txt";
     write_to_file(mount_point.path(), f_name).expect("should be able to write to the file");
     unmount(&mount_point);
+    wait_for_exit(child);
```
vladem, line 835: Is [this](https://github.com/awslabs/mountpoint-s3/actions/runs/11861156791/job/33057921585) the right example of an error which you're intending to fix?

In which case, I am not sure if I understood the fix. The test fails before `wait_for_exit`: in `write_to_file`, called on line [833](https://github.com/awslabs/mountpoint-s3/pull/1140/files#diff-b6f1caa405e63c89fd632c2d50612b88726aa868accde4c8c860077ed054403aR833). Do you mind explaining the fix in more detail? (it's a good practice to add a comment in code with such explanation)

dannycjones: I'm not sure if we should add a comment since it looks like we're doing the right thing everywhere else. However, I agree that the fix here is currently unclear and we should make sure that the PR description (and later this will be the commit) includes why this was a problem.

vladem: Agree, no comments needed and if it helps to resolve issues locally, it would be great to merge this and see if it helps in the CI. Just curious about the idea behind the fix.

dannycjones: I've set this to merge for now since its bringing this test consistent with the other tests. I'm still surprised that this would fix the issue.

PR #1141, opened 2024-11-15
Validate that shared cache bucket is usable

muddyfish:
## Description of change

- Validates the shared cache bucket is write-able
- Validates the shared cache bucket supports the `EXPRESS_ONEZONE` storage class

Relevant issues: N/A

Yes, the shared cache bucket is now validated that it supports the `EXPRESS_ONEZONE` storage class

No

---

### DIFF
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -896,7 +896,8 @@ where
         (None, Some((config, bucket_name, cache_bucket_name))) => {
             tracing::trace!("using S3 Express One Zone bucket as a cache for object content");
             let express_cache = ExpressDataCache::new(client.clone(), config, bucket_name, cache_bucket_name);
-            block_on(express_cache.verify_cache_valid())?;
+            block_on(express_cache.verify_cache_valid())
+                .with_context(|| format!("initial PutObject failed for shared cache bucket {cache_bucket_name}"))?;
 
             let prefetcher = caching_prefetch(express_cache, runtime, prefetcher_config);
             let fuse_session = create_filesystem(
@@ -936,7 +937,8 @@ where
             tracing::trace!("using both local disk and S3 Express One Zone bucket as a cache for object content");
             let (managed_cache_dir, disk_cache) = create_disk_cache(cache_dir_path, disk_data_cache_config)?;
             let express_cache = ExpressDataCache::new(client.clone(), config, bucket_name, cache_bucket_name);
-            block_on(express_cache.verify_cache_valid())?;
+            block_on(express_cache.verify_cache_valid())
+                .with_context(|| format!("initial PutObject failed for shared cache bucket {cache_bucket_name}"))?;
             let cache = MultilevelDataCache::new(Arc::new(disk_cache), express_cache, runtime.clone());
 
             let prefetcher = caching_prefetch(cache, runtime, prefetcher_config);
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -11,6 +11,7 @@ use std::time::Duration;
 use anyhow::{anyhow, Context as _};
 use clap::{value_parser, ArgGroup, Parser, ValueEnum};
 use fuser::{MountOption, Session};
+use futures::executor::block_on;
 use futures::task::Spawn;
 use mountpoint_s3_client::config::{AddressingStyle, EndpointConfig, S3ClientAuthConfig, S3ClientConfig};
 use mountpoint_s3_client::error::ObjectClientError;
@@ -895,6 +896,7 @@ where
         (None, Some((config, bucket_name, cache_bucket_name))) => {
             tracing::trace!("using S3 Express One Zone bucket as a cache for object content");
             let express_cache = ExpressDataCache::new(client.clone(), config, bucket_name, cache_bucket_name);
+            block_on(express_cache.verify_cache_valid())?;
 
             let prefetcher = caching_prefetch(express_cache, runtime, prefetcher_config);
             let fuse_session = create_filesystem(
@@ -934,6 +936,7 @@ where
             tracing::trace!("using both local disk and S3 Express One Zone bucket as a cache for object content");
             let (managed_cache_dir, disk_cache) = create_disk_cache(cache_dir_path, disk_data_cache_config)?;
             let express_cache = ExpressDataCache::new(client.clone(), config, bucket_name, cache_bucket_name);
+            block_on(express_cache.verify_cache_valid())?;
             let cache = MultilevelDataCache::new(Arc::new(disk_cache), express_cache, runtime.clone());
 
             let prefetcher = caching_prefetch(cache, runtime, prefetcher_config);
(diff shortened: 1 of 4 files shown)


mountpoint-s3/tests/fuse_tests/cache_test.rs
```diff
@@ -143,6 +149,62 @@ fn disk_cache_write_read(key_suffix: &str, key_size: usize, object_size: usize)
...
+        panic!("wrong error type");
+    }
+}
+
+#[tokio::test]
+async fn express_cache_verify_fail_forbidden() {
+    let bucket_name = get_standard_bucket();
+    let cache_bucket_name = get_express_bucket();
+
+    let policy = r#"{"Statement": [
```
vladem, line 177: A comment that `s3express:CreateSession` is missing so a failure is expected?

mountpoint-s3/tests/fuse_tests/cache_test.rs
```diff
@@ -143,6 +149,62 @@ fn disk_cache_write_read(key_suffix: &str, key_size: usize, object_size: usize)
...
+    let cache = ExpressDataCache::new(client.clone(), Default::default(), &bucket_name, &cache_bucket_name);
+    let err = cache
+        .verify_cache_valid()
+        .await
+        .expect_err("cannot use standard bucket as shared cache");
+
+    if let ObjectClientError::ClientError(ResponseError(request_result)) = err {
+        let body = request_result.error_response_body.as_ref().expect("should have body");
+        let body = body.clone().into_string().unwrap();
+        assert!(body.contains("<Code>InvalidStorageClass</Code>"));
```
vladem, line 166: how do we make S3 to return an error in this case?

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -72,6 +72,30 @@ where
             bucket_name: bucket_name.to_owned(),
         }
     }
+
+    pub async fn verify_cache_valid(&self) -> ObjectClientResult<(), PutObjectError, Client::ClientError> {
```
vladem, line 76: A docstring explaining what check is done in this method and why is that sufficient? (e.g. what is `storage_class("EXPRESS_ONEZONE".to_string())` and why we don't do a `GetObject` check)

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -72,6 +72,30 @@ where
...
+        let object_key = format!("{}/_mountpoint_cache_metadata", &self.prefix);
+        // This data is human-readable, and not expected to be read by Mountpoint.
+        // The file format used here is NOT stable.
+        // For now, let's just include the data that's guaranteed to be correct as it's what
+        // calculates the prefix.
+        let data = format!(
+            "source_bucket={}\nblock_size={}",
```
vladem, line 83: We don't actually put a restriction that the given directory bucket must be used to cache data for a single source bucket. Nothing critical will happen, if customer tries to do that (we'll just overwrite the object here), but it may get misleading (this object tells that a bucket is used to cache data for bucket A, while it also holds bucket B's data ).

Maybe we can try creating an empty object instead? Would it be sufficient as a check?

mountpoint-s3/src/cli.rs
```diff
@@ -895,6 +896,7 @@ where
         (None, Some((config, bucket_name, cache_bucket_name))) => {
             tracing::trace!("using S3 Express One Zone bucket as a cache for object content");
             let express_cache = ExpressDataCache::new(client.clone(), config, bucket_name, cache_bucket_name);
+            block_on(express_cache.verify_cache_valid())?;
```
vladem, line 899: how does the error looks in the log? would it be useful to add some context about the put operation? (similar to [ListObjectsV2](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/cli.rs#L1061))
```rust
.with_context(|| format!("initial PutObject failed for bucket {cache_bucket_name}"))
```

vladem (commented):
a couple of suggestions, mostly LGTM

mountpoint-s3/tests/fuse_tests/cache_test.rs
muddyfish: We're sending expected storage class is express onezone, which standard doesn't support, and thus fails with an invalid storage class error

mountpoint-s3/src/cli.rs
muddyfish: With the additional context, it looks like this:
```
2024-11-18T14:17:14.010171Z  WARN put_object_single{id=1 bucket="..." key="5f74d6e14553da6f2273143c53378cc2b8942ef97ff06bbc01756abc0c4bd6c9/_mountpoint_cache_metadata"}: mountpoint_s3_client::s3_crt_client: duration=4.885103ms request_id=RN7K3ZYAH0Z72YS5 request_result=MetaRequestResult { response_status: 400, crt_error: Error(14343, "aws-c-s3: AWS_ERROR_S3_INVALID_RESPONSE_STATUS, Invalid response status from request"), error_response_headers: Some(Headers { inner: 0x7f80c00172c0 }), error_response_body: Some("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>InvalidStorageClass</Code><Message>The storage class you specified is not valid</Message><StorageClassRequested>EXPRESS_ONEZONE</StorageClassRequested><RequestId>RN7K3ZYAH0Z72YS5</RequestId><HostId>3TlUgk4EOMSHtVRavhSrsoorc/bi6mpLOU0yuw96IK7cRzVFOxL5KbsfyTQUHIzReLqDH5tonuk=</HostId></Error>") } meta request failed
Error: initial PutObject failed for shared cache bucket ...

Caused by:
    0: Client error
    1: Unknown response error: MetaRequestResult { response_status: 400, crt_error: Error(14343, "aws-c-s3: AWS_ERROR_S3_INVALID_RESPONSE_STATUS, Invalid response status from request"), error_response_headers: Some(Headers { inner: 0x7f80c00172c0 }), error_response_body: Some("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>InvalidStorageClass</Code><Message>The storage class you specified is not valid</Message><StorageClassRequested>EXPRESS_ONEZONE</StorageClassRequested><RequestId>RN7K3ZYAH0Z72YS5</RequestId><HostId>3TlUgk4EOMSHtVRavhSrsoorc/bi6mpLOU0yuw96IK7cRzVFOxL5KbsfyTQUHIzReLqDH5tonuk=</HostId></Error>") }
```

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -72,6 +72,33 @@ where
             bucket_name: bucket_name.to_owned(),
         }
     }
+
+    pub async fn verify_cache_valid(&self) -> ObjectClientResult<(), PutObjectError, Client::ClientError> {
```
vladem, line 76: nit: we can now remove the [todo](https://github.com/awslabs/mountpoint-s3/blob/7a265f5cfd20e6ce2ba76b18785d3b7a75e339c8/mountpoint-s3/src/data_cache/express_data_cache.rs#L66)

vladem (approved):
LGTM

PR #1146, opened 2024-11-19
Add metrics to express data cache

muddyfish:
## Description of change

Adds metrics to express data cache
Fixes a bug where getting a cache miss would be reported as an error rather than a cache miss

Relevant issues: N/A

Adds metrics, no user facing functionality changes.

No

---

### DIFF
--- a/mountpoint-s3/src/data_cache/express_data_cache.rs
+++ b/mountpoint-s3/src/data_cache/express_data_cache.rs
@@ -1,7 +1,7 @@
+use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheError, DataCacheResult};
 use crate::object::ObjectId;
 use std::collections::HashMap;
-
-use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheError, DataCacheResult};
+use std::time::Instant;
 
 use async_trait::async_trait;
 use base64ct::{Base64, Encoding};
@@ -111,11 +111,16 @@ where
         block_offset: u64,
         object_size: usize,
     ) -> DataCacheResult<Option<ChecksummedBytes>> {
+        let start = Instant::now();
+
         if object_size > self.config.max_object_size {
+            metrics::counter!("express_data_cache.block_hit").increment(0);
+            metrics::counter!("express_data_cache.over_max_object_size", "type" => "read").increment(1);
             return Ok(None);
         }
 
         if block_offset != block_idx * self.config.block_size {
+            emit_failure_metric_read("invalid_block_offset");
             return Err(DataCacheError::InvalidBlockOffset);
         }
 
@@ -130,8 +135,14 @@ where
             .await
         {
             Ok(result) => result,
-            Err(ObjectClientError::ServiceError(GetObjectError::NoSuchKey)) => return Ok(None),
-            Err(e) => return Err(e.into()),
+            Err(ObjectClientError::ServiceError(GetObjectError::NoSuchKey)) => {
+                metrics::counter!("express_data_cache.block_hit").increment(0);
+                return Ok(None);
+            }
+            Err(e) => {
+                emit_failure_metric_read("get_object_failure");
+                return Err(DataCacheError::IoFailure(e.into()));
+            }
         };
 
         pin_mut!(result);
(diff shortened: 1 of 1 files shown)


mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -114,11 +116,14 @@ impl S3CrtClient {
                 let _ = sender.unbounded_send(Ok((offset, data.into())));
             },
             move |result| {
-                // FIXME - Ideally we'd include a reason why we failed here.
-                object_headers_setter_on_finish.or_set(Err(ObjectHeadersError::UnknownError));
                 if result.is_err() {
-                    Err(parse_get_object_error(result).map(ObjectClientError::ServiceError))
+                    let err = parse_get_object_error(result);
```
passaro, line 120: This is problematic: the result has not been processed yet - see [these](https://github.com/awslabs/mountpoint-s3/blob/b61f4b966f616ed3d231857403946149520aad2a/mountpoint-s3-client/src/s3_crt_client.rs#L514) [comments](https://github.com/awslabs/mountpoint-s3/blob/b61f4b966f616ed3d231857403946149520aad2a/mountpoint-s3-client/src/s3_crt_client.rs#L648).

I think this is the same mistake that causes issues for PUT: https://github.com/awslabs/mountpoint-s3/issues/1007.

Unless we review the callbacks, the correct thing to do would be to wait for the request to finish and return the fully parsed error.

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -99,6 +99,30 @@ where
...
+                Ok(None)
+            }
+            e => {
+                self.emit_failure_metric("get_object_failure", MetricsType::Read);
+                Err(DataCacheError::IoFailure(e.into()))
+            }
+        }
+    }
+
+    fn emit_failure_metric(&self, reason: &'static str, type_: MetricsType) {
```
passaro, line 120: I'd suggest inlining the metrics calls. This function makes it less clear what you are tracking.

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -113,11 +137,15 @@ where
         block_offset: u64,
         object_size: usize,
     ) -> DataCacheResult<Option<ChecksummedBytes>> {
```
passaro, line 139: Too many points of failure.
Shall we factor out the logic into an inner method and only handle the metrics here?

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -132,24 +160,31 @@ where
             .await
         {
             Ok(result) => result,
-            Err(ObjectClientError::ServiceError(GetObjectError::NoSuchKey)) => return Ok(None),
-            Err(e) => return Err(e.into()),
+            Err(err) => return self.handle_get_object_err(err),
```
vladem, line 163: it looks that `ObjectClientError::ServiceError(GetObjectError::NoSuchKey)` is not even possible here, but I'd check what this [method](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-client/src/s3_crt_client/get_object.rs#L125) returns

mountpoint-s3/src/data_cache/express_data_cache.rs
muddyfish: I'm not planning on refactoring into an inner method, as that makes the reasons harder to propagate

mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -214,6 +256,17 @@ where
...
 }
 
+#[inline]
+fn emit_failure_metric_read(reason: &'static str) {
+    metrics::counter!("express_data_cache.block_hit").increment(0);
+    metrics::counter!("express_data_cache.block_err", "reason" => reason, "type" => "read").increment(1);
+}
+
+#[inline]
+fn emit_failure_metric_write(reason: &'static str) {
```
passaro, line 266: nit, non-blocking: we could just inline

mountpoint-s3/src/data_cache/express_data_cache.rs
passaro: An alternative could be to still split into an inner function which records the failure (and reason), but move the hit/miss metric to the wrapper.

Happy to review the approach later, though.

passaro (approved):
Left a few comments, but completely optional.

PR #1156, opened 2024-11-21
Fix compilation error on macOS/arm

passaro:
Add a cast in the new `statfs` test: `libc::fsfilcnt_t` is not `u64` on all platforms.

No.

No.

---


c-hagem (approved):
Looks good to me. Thanks for catching that.

PR #1163, opened 2024-11-22
Optimise shared cache for single read use cases

muddyfish:
Use an optional BytesMut to avoid copying data on the first write. 

No

No

---


mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -149,16 +149,19 @@ where
...
+                    if offset != buffer.as_ref().map_or(0, |buffer| buffer.len()) as u64 {
                         emit_failure_metric_read("invalid_block_offset");
                         return Err(DataCacheError::InvalidBlockOffset);
                     }
-                    buffer.extend_from_slice(&body);
+                    if let Some(ref mut buffer) = buffer {
+                        buffer.extend_from_slice(&body);
+                    } else {
+                        buffer.replace(BytesMut::from(Bytes::from(body)));
+                    }
```
passaro, line 164: I'd suggest to have `buffer: Bytes` and then just do:
```suggestion
                    buffer = if buffer.is_empty() {
                        Bytes::from(body)
                    } else {
                        // Unlikely: we expect `get_object` to return a single chunk.
                        let mut buffer = BytesMut::from(buffer);
                        buffer.extend_from_slice(&body);
                        buffer.freeze()
                    };
```

It would also get rid of the `map`s elsewhere.

PR #1170, opened 2024-11-26
Fix flaky out_of_order_write test

passaro:
The `out_of_order_write` tests failed to account that in incremental upload mode the previously written content of a file could be already uploaded when a subsequent `write` fails (as expected in the test). In this case, the upload would occasionally be triggered by a `flush` call as a consequence of the test runner process being forked.

No.

No.

---


dannycjones (approved):
LGTM

PR #1176, opened 2024-11-29
Update documentation for supported bucket types

vladem:
Update the documentation to enumerate supported bucket types.

No.

No.

---

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -105,7 +105,7 @@ Mountpoint also respects access control lists (ACLs) applied to objects in your 
 
 By default, Mountpoint will automatically mount your S3 bucket given only the bucket name, and will automatically select the appropriate S3 HTTPS endpoint. However, you can override this automation if you need finer control over how Mountpoint connects to your bucket.
 
-### S3 Bucket Types
+### S3 bucket types
 
 Mountpoint for Amazon S3 supports the following S3 bucket types: general purpose and directory. For more information on general purpose bucket naming rules see [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html), and for directory buckets see [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/directory-buckets-overview.html).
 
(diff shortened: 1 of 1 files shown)


muddyfish (dismissed):

doc/CONFIGURATION.md
```diff
@@ -105,6 +105,10 @@ Mountpoint also respects access control lists (ACLs) applied to objects in your
 
 By default, Mountpoint will automatically mount your S3 bucket given only the bucket name, and will automatically select the appropriate S3 HTTPS endpoint. However, you can override this automation if you need finer control over how Mountpoint connects to your bucket.
 
+### S3 Bucket Types
```
passaro, line 108: nit: lower case for consistency with other sections

doc/CONFIGURATION.md
```diff
@@ -105,6 +105,10 @@ Mountpoint also respects access control lists (ACLs) applied to objects in your
...
+
+Mountpoint for Amazon S3 supports the following S3 bucket types: general purpose and directory. For more information on general purpose bucket naming rules see [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html), and for directory buckets see [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/directory-bucket-naming-rules.html).
```
passaro, line 110: Before pointing to the guide, should we provide examples? e.g. `amzn-s3-demo-bucket` vs `amzn-s3-demo-bucket--az_id--x-s3`

It would make it clear that users have to provide the "full" bucket name, including the suffix for directory buckets.

dannycjones, line 110: What's the purpose of this documentation? If its just to state that we support them, I'd recommend we just link out to overview of that bucket type like this one: https://docs.aws.amazon.com/AmazonS3/latest/userguide/directory-buckets-overview.html

dannycjones: A little related to my comment on the purpose of these docs (https://github.com/awslabs/mountpoint-s3/pull/1176/files#r1863577958), I think we should be explicit with the intention. If we want to help customers understand how to use these buckets, then giving examples would be good.

For instance, we should recommend to customers to provide the bucket name including any suffix.

Again if this is the purpose, we should also mention that S3 Object Lambda can be used (and maybe Multi-Region Access Points) and link to their configuration sections.

passaro (approved):

passaro (approved):

PR #1186, opened 2024-12-05
Tidy up build info using unwrap

passaro:
With Rust 1.83, we can use `Option::unwrap` in a `const` context and remove the untidy hack we had to resort to in order to get the commit hash.

No.

No.

---


passaro:
Ha. We cannot in fact use this:
```
error[E0080]: evaluation of constant value failed
  --> mountpoint-s3/src/build_info.rs:31:35
   |
31 |     const COMMIT_HASH_STR: &str = built::GIT_COMMIT_HASH_SHORT.unwrap();
   |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'called `Option::unwrap()` on a `None` value', mountpoint-s3/src/build_info.rs:31:64
```

Let's keep the original hack.

PR #1189, opened 2024-12-06
Add errno to fuse_error macro

renanmagagnin:
Added errno to fuse_error macro

---

### DIFF
--- a/mountpoint-s3/tests/fuse_tests/fork_test.rs
+++ b/mountpoint-s3/tests/fuse_tests/fork_test.rs
@@ -810,7 +810,7 @@ fn read_with_no_permissions_for_a_key_sse() {
         read_result.expect("should be able to read a default-encrypted file after the first read failure");
     }
 
-    let log_line_pattern = format!("^.*WARN.*{encrypted_object}.*read failed: get request failed: get object request failed: Client error: Forbidden: User: .* is not authorized to perform: kms:Decrypt on resource: {key_id} because no session policy allows the kms:Decrypt action.*$");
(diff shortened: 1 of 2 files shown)


muddyfish (approved):

mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -810,7 +810,7 @@ fn read_with_no_permissions_for_a_key_sse() {
         read_result.expect("should be able to read a default-encrypted file after the first read failure");
     }
 
-    let log_line_pattern = format!("^.*WARN.*{encrypted_object}.*read failed: get request failed: get object request failed: Client error: Forbidden: User: .* is not authorized to perform: kms:Decrypt on resource: {key_id} because no session policy allows the kms:Decrypt action.*$");
+    let log_line_pattern = format!("^.*WARN.*{encrypted_object}.*read 5 failed: get request failed: get object request failed: Client error: Forbidden: User: .* is not authorized to perform: kms:Decrypt on resource: {key_id} because no session policy allows the kms:Decrypt action.*$");
```
passaro, line 813: The error message does not look right. Why are we adding the error number here?

dannycjones: We had a chat, we're probably looking for something like "read failed returning errno 5"

dannycjones: It may be worth taking a look at this libc(?) function that can return a string form of the error: https://man7.org/linux/man-pages/man3/errno.3.html

The idea is if we can instead get to something like:

>"read failed returning EPERM"

dannycjones: Oops, see we're already working in this PR: https://github.com/awslabs/mountpoint-s3/pull/1197

PR #1190, opened 2024-12-09
Create new latency benchmark for writing of one-byte files

renanmagagnin:
Creates a new latency benchmark for writing of one-byte files. Also creates a folder for writing latency benchmarks and extends the `fs_latency_bench.sh` script to handle multiple folders.

---


mountpoint-s3/scripts/fio/write/seq_write_small.fio
```diff
@@ -0,0 +1,15 @@
+[global]
+name=fs_bench
+bs=256k
+runtime=30s
+time_based
+group_reporting
+
+[sequential_write_small_file]
+size=5m
```
muddyfish, line 9: We should have a quick discussion on if 5MB is the right size for this

muddyfish (commented):
~~Can you include what results format we get as a comment?~~

Nevermind, I see we're running a benchmark action which should include it

mountpoint-s3/scripts/fio/write_latency/seq_write_small.fio
```diff
@@ -0,0 +1,14 @@
+[global]
+name=fs_bench
+bs=1B
+
+[sequential_write_small_file]
```
monthonk, line 5: It's not really clear from the name what number we want to measure here. Could we have better naming for it?

mountpoint-s3/scripts/fs_latency_bench.sh
```diff
@@ -127,8 +127,9 @@ do
 done
 
 
-# start time to first byte benchmark
-jobs_dir=mountpoint-s3/scripts/fio/read_latency
+run_start_time_to_first_byte_benchmarks() {
```
monthonk, line 130: We probably should change this function name too. It's not time to first byte for the write benchmark.

mountpoint-s3/scripts/fs_latency_bench.sh
```diff
@@ -126,9 +126,9 @@ do
...
 
-
-# start time to first byte benchmark
-jobs_dir=mountpoint-s3/scripts/fio/read_latency
+run_file_benchmarks() {
+  category=$1
+  jobs_dir=mountpoint-s3/scripts/fio/${category}_latency
 for job_file in "${jobs_dir}"/*.fio; do
   mount_dir=$(mktemp -d /tmp/fio-XXXXXXXXXXXX)
   job_name=$(basename "${job_file}")
```
monthonk, line 134: We should update indentation for this block. Sorry, I missed this part in the last review.

monthonk (changes requested):

PR #1192, opened 2024-12-09
Wait for CreateMPU before returning from put_object

passaro:
`S3CrtClient::put_object` was originally implemented so that it would complete immediately and return a `PutObjectRequest` implementation. Any error from the S3 request would only be returned on calling `write` or `complete` on the `PutObjectRequest`. With this change, we modify `put_object` to await for the initial `CreateMultipartUpload` request to complete and only then either return a `PutObjectRequest` or propagate the error from the request. This is analogous to what done for `get_object` in #1171 and addresses an issue where errors were not propagated correctly (#1007).

At the file handle level, however, we still want the `open` operation to complete quickly, without waiting for `CreateMultipartUpload` to complete. In order to preserve the previous behavior, `upload::atomic` was adapted to spawn a concurrent task in the background when calling `put_object`. 

Yes.

Yes, for `mountpoint-s3-client`. No user-visible changes in `mountpoint-s3`.

---


monthonk (dismissed):
lgtm, nice that we can maintain the same behavior for `open`.

monthonk (approved):

PR #1197, opened 2024-12-10
Add label to errno in error message for clarity

renanmagagnin:
This change improves the clarity of [fuse_error with errono](https://github.com/awslabs/mountpoint-s3/pull/1189).

---

### DIFF
--- a/mountpoint-s3/src/fuse.rs
+++ b/mountpoint-s3/src/fuse.rs
@@ -38,7 +38,7 @@ macro_rules! event {
 macro_rules! fuse_error {
     ($name:literal, $reply:expr, $err:expr) => {{
         let err = $err;
-        event!(err.level, "{} {} failed: {:#}", $name, err.to_errno(), err);
+        event!(err.level, "{} Errno: {} failed: {:#}", $name, err.to_errno(), err);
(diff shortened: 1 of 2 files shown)


mountpoint-s3/src/fuse.rs
```diff
@@ -38,7 +38,7 @@ macro_rules! event {
 macro_rules! fuse_error {
     ($name:literal, $reply:expr, $err:expr) => {{
         let err = $err;
-        event!(err.level, "{} {} failed: {:#}", $name, err.to_errno(), err);
+        event!(err.level, "{} Errno: {} failed: {:#}", $name, err.to_errno(), err);
```
passaro, line 41: I would prefer at least something like this:
```suggestion
        event!(err.level, "{} failed with errno {}: {:#}", $name, err.to_errno(), err);
```

passaro: But also happy to consider alternatives (here or on a separate PR), e.g.:
* as a named attribute in `event!` (assuming it's allowed), maybe together with the operation name,
* or as part of the error `Debug` implementation.

renanmagagnin: I think the suggested change is great. Let me know if you think it's worth looking more deeply into the alternatives :)

PR #1200, opened 2024-12-12
Improve get_object interface for backpressure

monthonk:
Currently, we support flow-control window for GetObject requests by allowing applications to call `GetObjectResponse::increment_read_window` but it is tricky to use because we need to hold onto the stream itself in order to control the feedback loop while also consuming the data.

This change introduces a new trait `ClientBackpressureHandle` for controlling the read window so that the stream and the flow-control paths are decoupled.

Applications can now call `GetObjectResponse::take_backpressure_handle` to get a backpressure handle from the response and use this handle to extend the read window.

Yes, there is a breaking change for `mountpoint-s3-client`.

Yes, for `mountpoint-s3-client`.

---

### DIFF
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -26,12 +26,8 @@ pub use etag::ETag;
 #[cfg_attr(not(docsrs), async_trait)]
 #[auto_impl(Arc)]
 pub trait ObjectClient {
-    type GetObjectResponse: GetObjectResponse<
-        ClientError = Self::ClientError,
-        BackpressureHandle = Self::BackpressureHandle,
-    >;
+    type GetObjectResponse: GetObjectResponse<ClientError = Self::ClientError>;
     type PutObjectRequest: PutObjectRequest<ClientError = Self::ClientError>;
-    type BackpressureHandle: ClientBackpressureHandle + Send + Sync;
     type ClientError: std::error::Error + ProvideErrorMetadata + Send + Sync + 'static;
 
     /// Query the part size this client uses for GET operations to the object store. This
@@ -604,7 +600,10 @@ pub trait ClientBackpressureHandle {
     /// Increment the flow-control read window, so that response data continues downloading.
     fn increment_read_window(&mut self, len: usize);
 
-    /// Get the upper bound of the current read window. When backpressure is enabled, [GetObjectRequest] can
+    /// Move the upper bound of the read window to the given offset if it's not already there.
+    fn ensure_read_window(&mut self, desired_end_offset: u64);
+
+    /// Get the upper bound of the read window. When backpressure is enabled, [GetObjectRequest] can
     /// return data up to this offset *exclusively*.
     fn read_window_end_offset(&self) -> u64;
 }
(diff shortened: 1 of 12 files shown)


mountpoint-s3-client/src/object_client.rs
```diff
@@ -595,33 +618,20 @@ impl UploadChecksum {
...
     Stream<Item = ObjectClientResult<GetBodyPart, GetObjectError, Self::ClientError>> + Send + Sync
 {
+    type BackpressureHandle: ClientBackpressureHandle;
     type ClientError: std::error::Error + Send + Sync + 'static;
 
+    /// Take the backpressure handle from the response.
+    ///
+    /// If `enable_read_backpressure` is false this call will return `None`,
+    /// no backpressure is being applied and data is being downloaded as fast as possible.
+    fn take_backpressure_handle(&mut self) -> Option<Self::BackpressureHandle>;
```
passaro, line 628: Why `take_`? We should document that it can only be called once, or rather return a reference and make it `Clone`.

mountpoint-s3-client/src/object_client.rs
```diff
@@ -27,8 +26,12 @@ pub use etag::ETag;
 #[cfg_attr(not(docsrs), async_trait)]
 #[auto_impl(Arc)]
 pub trait ObjectClient {
-    type GetObjectResponse: GetObjectResponse<ClientError = Self::ClientError>;
+    type GetObjectResponse: GetObjectResponse<
+        ClientError = Self::ClientError,
+        BackpressureHandle = Self::BackpressureHandle,
+    >;
     type PutObjectRequest: PutObjectRequest<ClientError = Self::ClientError>;
+    type BackpressureHandle: ClientBackpressureHandle + Send + Sync;
```
passaro, line 34: Do we need the associated type here? Cannot it be defined only in `GetObjectResponse`?

passaro (commented):
I have a couple of suggestions, but I like the improvement.

mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -160,18 +195,24 @@ pub struct S3GetObjectResponse {
...
     requested_checksums: bool,
     enable_backpressure: bool,
+    backpressure_handle: Option<S3BackpressureHandle>,
     headers: Headers,
     /// Next offset of the data to be polled from [poll_next]
     next_offset: u64,
     /// Upper bound of the current read window. When backpressure is enabled, [S3GetObjectRequest]
     /// can return data up to this offset *exclusively*.
-    read_window_end_offset: u64,
+    read_window_end_offset: Arc<AtomicU64>,
```
passaro, line 204: Do we need to keep a copy here? Couldn't we just use the one in `backpressure_handle`? I also suspect `enable_backpressure` may be redundant. Isn't it just `backpressure_handle.is_some()`?

monthonk: Yes, you're right. Updated for both CRT client and the mock client.

passaro (approved):

PR #1202, opened 2024-12-17
Remove workaround for macOS builds due to pkg-config issues

dannycjones:
In November, macOS runners had an issue with the `pkg-config` package and we introduced a workaround: https://github.com/awslabs/mountpoint-s3/pull/1158/

We see now that the fix has been released and we should no longer need the workaround: https://github.com/actions/runner-images/issues/10984.

No, CI change only.

No.

---


passaro (approved):
Thanks for tidying that up!

PR #1205, opened 2024-12-20
Update CRT `HeadersError` enum to include header name

dannycjones:
We recently saw an error in #1199 where "Header not found" was emitted, but its really unclear what header was missing.

    2024-12-12T18:33:59.379478Z  WARN flush{req=1609 ino=2 fh=1 pid=29257 name="testfile_100M.bin"}: mountpoint_s3::fuse: flush failed: put failed: put request failed: Client error: Internal S3 client error: Header not found

This change updates the `HeadersError::HeaderNotFound` enum variant to contain a copy of the header name, such that error messages can emit it for debugging purposes.

It may make more sense to have all the header names we use statically defined somewhere, such that we could include a static reference to the header and avoid allocating for an error message. However, we don't expect there to be any performance regression introduced by this change. This move to static values could be made later.

Header not found and invalid header value errors will now include the header name when printing the error message.

The enum variants change meaning any code using the enum may be impacted.

Not for Mountpoint itself. I have added a change log entry to `mountpoint-s3-crt` since it is a breaking API change.

---

### DIFF
--- a/mountpoint-s3-crt/src/http/request_response.rs
+++ b/mountpoint-s3-crt/src/http/request_response.rs
@@ -66,8 +66,8 @@ unsafe impl Sync for Headers {}
 #[derive(Debug, Error, PartialEq, Eq)]
 pub enum HeadersError {
     /// The header was not found
-    #[error("Header not found")]
-    HeaderNotFound,
+    #[error("Header {0:?} not found")]
+    HeaderNotFound(OsString),
 
     /// Internal CRT error
     #[error("CRT error: {0}")]
@@ -78,14 +78,12 @@ pub enum HeadersError {
     Invalid(OsString),
 }
 
-// Convert CRT error into HeadersError, mapping the HEADER_NOT_FOUND to HeadersError::HeaderNotFound.
-impl From<Error> for HeadersError {
-    fn from(err: Error) -> Self {
-        if err == (aws_http_errors::AWS_ERROR_HTTP_HEADER_NOT_FOUND as i32).into() {
-            Self::HeaderNotFound
-        } else {
-            Self::CrtError(err)
-        }
+/// Try to convert the CRT [Error] into [HeadersError::HeaderNotFound], or return [HeadersError::CrtError].
+fn try_as_header_not_found(err: Error, header_name: &OsStr) -> HeadersError {
+    if err.raw_error() == (aws_http_errors::AWS_ERROR_HTTP_HEADER_NOT_FOUND as i32) {
+        HeadersError::HeaderNotFound(header_name.to_owned())
+    } else {
+        HeadersError::CrtError(err)
     }
 }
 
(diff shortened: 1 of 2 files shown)


mountpoint-s3-crt/src/http/request_response.rs
```diff
@@ -263,7 +274,7 @@ impl Iterator for HeadersIterator<'_> {
             let header = self
                 .headers
                 .get_index(self.offset)
-                .expect("HeadersIterator: failed to get next header");
+                .expect("headers at any offset smaller than original count should always exist given mut access");
```
passaro, line 277: This is not true: look at how `Headers` implements `Clone`.

mountpoint-s3-crt/src/http/request_response.rs
```diff
@@ -78,14 +78,12 @@ pub enum HeadersError {
...
-// Convert CRT error into HeadersError, mapping the HEADER_NOT_FOUND to HeadersError::HeaderNotFound.
-impl From<Error> for HeadersError {
-    fn from(err: Error) -> Self {
-        if err == (aws_http_errors::AWS_ERROR_HTTP_HEADER_NOT_FOUND as i32).into() {
-            Self::HeaderNotFound
-        } else {
-            Self::CrtError(err)
-        }
+/// Try to convert the CRT [Error] into [HeadersError::HeaderNotFound], or return [HeadersError::CrtError].
+fn try_as_header_not_found(err: Error, header_name: &OsStr) -> HeadersError {
```
passaro, line 82: nit: this could be a method on `HeadersError`

mountpoint-s3-crt/src/http/request_response.rs
```diff
@@ -78,14 +78,12 @@ pub enum HeadersError {
     Invalid(OsString),
```
passaro, line 78: We should add the header name here as well.

mountpoint-s3-crt/src/http/request_response.rs
passaro: Nothing to do with this change, but we should review this type.

dannycjones: Good catch! I've opened issue https://github.com/awslabs/mountpoint-s3/issues/1209 to track.

As discussed offline, we'll go ahead with this change and track addressing the incorrect safety comments in a future change.

passaro (dismissed):
LGTM!

mountpoint-s3-crt/src/http/request_response.rs
```diff
@@ -66,25 +66,30 @@ unsafe impl Sync for Headers {}
...
-// Convert CRT error into HeadersError, mapping the HEADER_NOT_FOUND to HeadersError::HeaderNotFound.
-impl From<Error> for HeadersError {
-    fn from(err: Error) -> Self {
-        if err == (aws_http_errors::AWS_ERROR_HTTP_HEADER_NOT_FOUND as i32).into() {
-            Self::HeaderNotFound
+impl HeadersError {
+    /// Try to convert the CRT [Error] into [HeadersError::HeaderNotFound], or return [HeadersError::CrtError].
+    fn try_convert(err: Error, header_name: &OsStr) -> HeadersError {
+        if err.raw_error() == (aws_http_errors::AWS_ERROR_HTTP_HEADER_NOT_FOUND as i32) {
+            HeadersError::HeaderNotFound(header_name.to_owned())
```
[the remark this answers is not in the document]
dannycjones: Added a note into the struct rustdoc.

passaro (approved):

PR #1213, opened 2025-01-08
Create file-system-benchmarks.rs and update fs_latency_bench.sh accordingly

renanmagagnin:
Creation of a benchmarking binary for recreating and measuring the latencies of real-world usage patterns of Mountpoint. This version includes only a small file creation benchmark which measures the sequence of the file system operations: lookup, open, write (of one byte), and flush. The latency measurement captures the total duration and is averaged multiple iterations to ensure representativeness.

The benchmarking binary is used in the `fs_latency_bench.sh` script and the the new results are included in the final results of the script, ultimately being added to the [benchmarking GitHub pages dashboard](https://awslabs.github.io/mountpoint-s3/dev/latency_bench/).

No.

No.

---

### DIFF
--- a/mountpoint-s3/Cargo.toml
+++ b/mountpoint-s3/Cargo.toml
@@ -48,6 +48,8 @@ tracing = { version = "0.1.41", features = ["log"] }
 tracing-log = "0.2.0"
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
 
+statistical = { version = "1.0", optional = true }
+
 [target.'cfg(target_os = "linux")'.dependencies]
 procfs = { version = "0.17.0", default-features = false }
 
@@ -105,3 +107,8 @@ required-features = ["mock"]
 [[bin]]
 name = "mount-s3-log-analyzer"
 path = "src/bin/mount-s3-log-analyzer.rs"
+
+[[bin]]
+name = "file-system-benchmarks"
+path = "src/bin/file-system-benchmarks.rs"
+required-features = ["statistical"]
--- a/mountpoint-s3/Cargo.toml
+++ b/mountpoint-s3/Cargo.toml
@@ -48,8 +48,6 @@ tracing = { version = "0.1.41", features = ["log"] }
 tracing-log = "0.2.0"
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
 
-statistical = { version = "1.0", optional = true }
-
 [target.'cfg(target_os = "linux")'.dependencies]
 procfs = { version = "0.17.0", default-features = false }
 
@@ -111,4 +109,3 @@ path = "src/bin/mount-s3-log-analyzer.rs"
 [[bin]]
 name = "file-system-benchmarks"
 path = "src/bin/file-system-benchmarks.rs"
-required-features = ["statistical"]
--- a/mountpoint-s3/src/bin/file-system-benchmarks.rs
+++ b/mountpoint-s3/src/bin/file-system-benchmarks.rs
@@ -1,14 +1,13 @@
 use anyhow::{anyhow, Result};
 use clap::{Parser, ValueEnum};
-use serde::Serialize;
+use serde::{Serialize, Serializer};
 use serde_json::json;
-use statistical::mean;
-use std::{
-    fs::{self, File, OpenOptions},
-    io::{BufWriter, Write},
-    path::PathBuf,
-    time::Instant,
-};
+
+use std::fs::{self, File, OpenOptions};
+use std::io::{BufWriter, Write};
+use std::path::PathBuf;
+use std::time::Duration;
+use std::time::Instant;
 
 #[derive(Parser, Debug, Clone, ValueEnum)]
 enum BenchmarkType {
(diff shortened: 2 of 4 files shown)


mountpoint-s3/Cargo.toml
```diff
@@ -105,3 +107,8 @@ required-features = ["mock"]
 [[bin]]
 name = "mount-s3-log-analyzer"
 path = "src/bin/mount-s3-log-analyzer.rs"
+
+[[bin]]
+name = "file-system-benchmarks"
+path = "src/bin/file-system-benchmarks.rs"
+required-features = ["statistical"]
```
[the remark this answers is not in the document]
dannycjones: Note, this current behavior depends on this behavior of Rust that will be removed in the next Rust edition (2024): https://doc.rust-lang.org/edition-guide/rust-2024/cargo-remove-implicit-features.html

mountpoint-s3/Cargo.toml
```diff
@@ -48,6 +48,8 @@ tracing = { version = "0.1.41", features = ["log"] }
 tracing-log = "0.2.0"
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
 
+statistical = { version = "1.0", optional = true }
```
dannycjones, line 51: This crate is not maintained, so I'm not super fond of using it. Maybe we can use something that has seen more activity, or just calculate our own mean since that's all we're using.

If we eliminate the dependency, it also simplifies the optional dependency issue.

dannycjones (changes requested):
Nice, I have put a few comments but I'm happy to see us building out this tool. Happy to chat!

mountpoint-s3/Cargo.toml
renanmagagnin: I removed the dependency. 👍

mountpoint-s3/src/bin/file-system-benchmarks.rs
```diff
@@ -0,0 +1,172 @@
...
+};
+
+#[derive(Parser, Debug, Clone, ValueEnum)]
+enum BenchmarkType {
+    OneByteFile,
+    All,
+}
+
+#[derive(Parser, Debug)]
+struct CliArgs {
```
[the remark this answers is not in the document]
dannycjones: I think the current desc is fine - I was trying to highlight that it specifically won't be able to measure a FUSE `open`, since we cannot trigger them independently.

mountpoint-s3/src/bin/file-system-benchmarks.rs
```diff
@@ -0,0 +1,172 @@
...
+    let mut total_latency_samples = vec![];
+
+    for file_number in 1..=num_files {
+        let mut elapsed_total_ms: f64 = 0.0;
+        let path = format!("{mount_dir_str}/bench_file_{file_number}");
+
+        // Perform and time the lookup operation
+        let start = Instant::now();
+        let _ = fs::metadata(path.clone());
+        let elapsed_ms = start.elapsed().as_nanos() as f64 / NANOS_PER_MILLI;
```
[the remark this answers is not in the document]
dannycjones: Discussed this change would cause loss of precision, so we'll add a `DurationExt` trait for it until its stable in std.

mountpoint-s3/src/bin/file-system-benchmarks.rs
```diff
@@ -0,0 +1,219 @@
...
+    let CliArgs {
+        mount_dir,
+        out_file,
+        benchmark_type,
+        detailed,
+    } = CliArgs::parse();
+    let mount_dir_os_str = mount_dir.into_os_string();
+    let mount_dir_str = mount_dir_os_str
+        .to_str()
+        .ok_or_else(|| anyhow!("Invalid UTF-8 in mount directory path"))?;
```
dannycjones, line 201: Really I think we should have just passed through the `&Path`. We can do things like [`Path::join`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.join) to get new paths as you use elsewhere (instead of using `format!(..)`.

mountpoint-s3/src/bin/file-system-benchmarks.rs
```diff
@@ -0,0 +1,219 @@
...
+    fn as_millis_f64_temp(&self) -> f64 {
+        const NANOS_PER_MILLI: f64 = 1_000_000.0;
+        self.as_nanos() as f64 / NANOS_PER_MILLI
+    }
+}
+
+fn mean(v: &[f64]) -> f64 {
+    let len = v.len() as f64;
+    v.iter().fold(0.0, |acc: f64, elem| acc + *elem) / len
+}
```
dannycjones, line 71: It may be worth a unit test or two for this (just to back ourselves up)

dannycjones (approved):
LGTM, there's a few minor things but I think this can be merged

mountpoint-s3/src/bin/file-system-benchmarks.rs
renanmagagnin: Much better, done!

dannycjones (approved):
Ship it!

PR #1216, opened 2025-01-09
Fix Clippy warnings on Rust 1.84

muddyfish:
Rust 1.84 has just released and with it, new clippy lints have been added.

We're also updating to Rust 1.84 explicitly rather than returning to stable to avoid the issues we had with the most recent release

No

No

---


dannycjones (approved):
LGTM.

I think we agreed to pin the Rust version for the in-progress release, but ready to merge after.

passaro (approved):
LGTM!

PR #1222, opened 2025-01-13
Fix Proptest inflight_write invalidation after PutObject to same element

c-hagem:
Adds a failing proptest, and a potential fix for it. The root cause seems to have been that if after a file has been opened, and then modified by i.e., a remote PutObject, the old ino is stale, but will be opened in the inflight-write regardless. 

This code change fixes this issue by removing such a inflight write. For reviewers: 
* Maybe this code should go into inflight_writes (i.e. add a ``invalidate`` method to invalidate based on a PathBuf)?

---

### DIFF
--- a/mountpoint-s3/Cargo.toml
+++ b/mountpoint-s3/Cargo.toml
@@ -47,6 +47,7 @@ time = { version = "0.3.37", features = ["macros", "formatting"] }
 tracing = { version = "0.1.41", features = ["log"] }
 tracing-log = "0.2.0"
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
+log = "0.4.22"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 procfs = { version = "0.17.0", default-features = false }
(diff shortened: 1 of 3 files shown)


mountpoint-s3/Cargo.toml
```diff
@@ -47,6 +47,7 @@ time = { version = "0.3.37", features = ["macros", "formatting"] }
 tracing = { version = "0.1.41", features = ["log"] }
 tracing-log = "0.2.0"
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
+log = "0.4.22"
```
dannycjones, line 50: We shouldn't need this

dannycjones (commented):
I don't know if this quite represents what we want.

If the file has been created (`mknod`) and opened for writing (`open`), it should be possible to write to that file and upload the object even though it would then not be visible in the file system after. Moreover, the end state of the file system should mean that remotely in S3 we should see the file even though it will not be visible in Mountpoint.

We need to somehow tolerate the open returning `ESTALE` - maybe we can discuss in-person how feasible that might be.

PR #1232, opened 2025-01-14
Add detailed rustdoc to reftests

dannycjones:
The reference tests for Mountpoint can be quite complex, especially for those unfamiliar both with the tests themselves or the idea of reference-based testing.

This change adds more detailed rustdoc with the goal to ramp up new readers with the reftests, give an overview of what the tests are doing, and point the reader to resources for learning more.

No, all code documentation changes.

No, code doc changes only.

---


c-hagem (approved):
I think these are great additions. I was wondering if we should go into more depth with regards to how the reference construction / the comparison works (but maybe this can wait until we decide what, if anything, we would change there)?

PR #1233, opened 2025-01-16
Customise benchmark page

vladem:
Currently the benchmark page [loads](https://github.com/awslabs/mountpoint-s3/blob/gh-pages/dev/bench/index.html#L107) the chart.js library from CDN, this PR changes that and adds CSP to ensure that no other resources are loaded from external locations.

No.

No.

---


muddyfish:
I'm not convinced on the `static_files` name scheme - perhaps `benchmark_website_static_files` or similar? Perhaps in the root directory as those files have little to do with github actions itself

muddyfish (approved):

PR #1235, opened 2025-01-17
Add support for CRC64-NVME checksum algorithm

passaro:
This change adds support for the CRC64-NVME checksum algorithm when using relevant operations in `mountpoint-s3-client` and when appending to existing objects through Mountpoint (using `--incremental-upload` mode).

No.

Yes.

---


mountpoint-s3-client/tests/put_object_single.rs
```diff
@@ -805,3 +832,54 @@ async fn test_append_with_matching_checksum(checksum_algorithm: aws_sdk_s3::type
         .await
         .expect("append should succeed");
 }
+
+#[cfg(feature = "s3express_tests")]
+#[tokio::test]
+async fn test_append_crc64() {
```
dannycjones, line 838: This test is for append where original object has not specified a checksum, so it was computed service-side. I think the name here should be specific for that.

dannycjones (commented):
LGTM so far.

I see we already say we need a changelog entry in the PR desc. For sure for a new feature, but we also update CRC64 struct name in CRT.

dannycjones (commented):
Discussed in-person, we want to extend test coverage as well for GetObject, HeadObject, GetObjectAttributes, ListObjectsV2. There will bit a little bit of missing impl on some of these.

dannycjones (approved):
LGTM!

PR #1240, opened 2025-01-21
Update S3ClientConfig to support configurable EventLoop thread count

ericjheinz:
For our use case, we run many mountpoint-s3 clients on a single machine and want to restrict the number of threads each client uses in order to reduce heap fragmentation and CPU contention.

No, this only allows for overriding the default value.

No, there are no breaking changes.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2571,7 +2571,7 @@ dependencies = [
 
 [[package]]
 name = "mountpoint-s3-client"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
  "async-io",
  "async-lock",
--- a/mountpoint-s3-client/CHANGELOG.md
+++ b/mountpoint-s3-client/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Other Changes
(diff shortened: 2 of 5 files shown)


passaro:
@ericjheinz, DCO is failing: you need to sign your commit (`--signoff` flag).

dannycjones (changes requested):
@ericjheinz please can you add an entry to the change log (https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-client/CHANGELOG.md) to share that it is now possible to configure the event loop thread count?

As Ale suggested, please review the DCO so that we can merge once above is addressed.

Otherwise, lgtm!

ericjheinz:
Update commit messages to include sign off and added new commit to update changelog

dannycjones (dismissed):

dannycjones (dismissed):
This change adds a new behavior to the client, so it needs a version bump as well.

This change conflicts with the `main` branch as well. Since we want this change merging in fast, I'll address these issues.

passaro (approved):
LGTM!

PR #1241, opened 2025-01-22
Add expected bucket owner test for shared XZ cache

vladem:
Add the expected bucket owner test. The test attempts to use a bucket in another AWS account as a cache and verifies that:

- write to the cache is not done if the expected bucket owner check is enabled
- write to the cache is done if the check is disabled
- write to the cache is done if the cache bucket belongs to the expected account

No.

No.

---

### DIFF
--- a/mountpoint-s3/tests/fuse_tests/cache_test.rs
+++ b/mountpoint-s3/tests/fuse_tests/cache_test.rs
@@ -1,12 +1,17 @@
 use crate::common::cache::CacheTestWrapper;
 use crate::common::fuse::create_fuse_session;
 use crate::common::fuse::s3_session::create_crt_client;
-use crate::common::s3::{get_test_bucket, get_test_prefix};
+use crate::common::s3::{
+    get_bucket_owner, get_external_express_bucket, get_test_bucket, get_test_endpoint_config, get_test_prefix,
+};
 
+use futures::executor::block_on;
 use mountpoint_s3::data_cache::{DataCache, DiskDataCache, DiskDataCacheConfig};
 use mountpoint_s3::object::ObjectId;
 use mountpoint_s3::prefetch::caching_prefetch;
-use mountpoint_s3_client::S3CrtClient;
+use mountpoint_s3_client::config::S3ClientConfig;
+use mountpoint_s3_client::error::ObjectClientError;
+use mountpoint_s3_client::{S3CrtClient, S3RequestError};
 
 use fuser::BackgroundSession;
 use rand::{Rng, RngCore, SeedableRng};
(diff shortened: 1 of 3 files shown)


mountpoint-s3/tests/fuse_tests/cache_test.rs
```diff
@@ -288,6 +290,67 @@ where
     assert!(block.is_none());
 }
 
+/// A test that checks that data is not written to the bucket owned by an unexpected account
+#[test_case(get_express_bucket(), true, true; "bucket owned by the expected account")]
+#[test_case(get_external_express_bucket(), true, false; "bucket owned by another account")]
+#[test_case(get_external_express_bucket(), false, false; "bucket owned by another account, not checked")]
+#[cfg(feature = "s3express_tests")]
+fn express_cache_expected_bucket_owner(cache_bucket: String, owner_checked: bool, owner_matches: bool) {
```
dannycjones, line 298: I think we should add another feature to gate this test (and any that require a second account to be provisioned and configured).

In other words, I expect it to look more like this:

```suggestion
/// A test that checks that data is not written to the bucket owned by an unexpected account
#[test_case(get_express_bucket(), true, true; "bucket owned by the expected account")]
#[test_case(get_external_express_bucket(), true, false; "bucket owned by another account")]
#[test_case(get_external_express_bucket(), false, false; "bucket owned by another account, not checked")]
#[cfg(all(feature = "s3express_tests", feature = "second_account_tests"))]
fn express_cache_expected_bucket_owner(cache_bucket: String, owner_checked: bool, owner_matches: bool) {
```

mountpoint-s3/tests/fuse_tests/cache_test.rs
```diff
@@ -288,6 +290,67 @@ where
...
+    let prefix = get_test_prefix("express_expected_bucket_owner");
+    let cache = ExpressDataCache::new(client.clone(), Default::default(), &bucket, &cache_bucket);
+    let cache_valid = block_on(cache.verify_cache_valid());
+    if owner_checked && !owner_matches {
+        match cache_valid {
+            Err(ObjectClientError::ClientError(S3RequestError::Forbidden(..))) => (),
+            _ => panic!("expected S3RequestError::Forbidden, got: {:?}", cache_valid),
+        }
+    } else {
+        cache_valid.unwrap(); // Ok(()) is expected
```
dannycjones, line 327: nit: This is a use for `expect(..)`

```suggestion
        cache_valid.expect("should succeed if not enforcing bucket owner");
```

dannycjones (changes requested):

mountpoint-s3/tests/fuse_tests/cache_test.rs
vladem: added the feature flag here: https://github.com/awslabs/mountpoint-s3/pull/1243, going to rebase and use the flag in this one after https://github.com/awslabs/mountpoint-s3/pull/1243 is merged

dannycjones (approved):
LGTM!

Issue #831, opened 2024-04-01
Option for disabling negative metadata entry cache


nguyenminhdungpg:
If I use --cache option, will **mountpoint-s3** not cache metadata of a requested file that is not existing? 
In **s3fs**, there is option disable_noobj_cache that help me to achieve this.

Thank you very much

dannycjones:
Hey @nguyenminhdungpg. Mountpoint will indeed cache the non-existence of a file. This behavior was introduced in [Mountpoint v1.5.0](https://github.com/awslabs/mountpoint-s3/releases/tag/mountpoint-s3-1.5.0).

It sounds like you would prefer that Mountpoint would not cache these entries, and instead go to S3 each time if it has not seen that the file exists. Can you tell us a bit more about your use case?

nguyenminhdungpg:
Hi @dannycjones thank you for your information.
My usecase for using disable_noobj_cache option of s3fs at this time is very weird and it is not good. There are 2 system runs in parallel and one requests file in s3fs mounted folder before the other finished adding the file to S3. Then the non-existences of the file and maybe its parent folders are cached, requests to the file with high frequency can make it (and its parent folders) seem to be not-existed forever.

nguyenminhdungpg:
Hi @dannycjones 
Is there any news about how to flag off negative cache?

Currently I am setting TTL to 6 months but with negative cache is always enabled, some logics in my service are broken.

Thank you.

monthonk:
Hey, we still don't have any updates to share yet. For your use case, I wonder why you set the TTL to be really long? We usually recommend using long TTL for static bucket that its content doesn't change that often. Is it possible to reduce TTL and then you can change logic in your service to do retry?

Another option you have is calling `readdir` to discover the new files because we don't serve `readdir` results from the cache, but it's more expensive operation.

nguyenminhdungpg:
Hi @monthonk , thank you for your information.

I have a service to upload files to S3 using AWS SDK. File will be stored in folder tree with multi level style. File will not be modified or deleted. mount-s3 is used to serve files for my website and my website is very high traffics. My CDN cache is working not well, it still miss many requests and load on mount-s3 is very high at peak time. That's why I set mount-s3 TTL that long, I need files  loaded as fast as possible.

mount-s3 negative cache does not only cache metadata of files, also folder. So when my service check if a file or folder is existing, folders' non-existence are cached by negative cache and break the logic of next requests (a lot of requests).
[...]

PR #1246, opened 2025-01-25
Add negative metadata cache ttl

notoriaga:
Adds a new CLI argument `--negative-cache-ttl` that lets you set the TTL for negative metadata entries separately from `--metadata-ttl`. My use case is a write once read many bucket. Objects do not get deleted from this bucket, and new objects are added every few minutes. I'd like to be able to set `--metadata-ttl indefinite` and `--negative-cache-ttl 60` to effectively utilize the caching while still being able to pick up new objects. There is an open issue for this here - https://github.com/awslabs/mountpoint-s3/issues/831

No, if `--negative-cache-ttl` is omitted the existing behavior is maintained (use `--metadata-ttl` or the default file_ttl).

Because this is a new feature I believe it would require both.

---

### DIFF
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -319,6 +319,14 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
 
     #[clap(
         long,
+        help = "Time-to-live (TTL) for cached negative entries in seconds [default: same as metadata TTL]",
+        value_name = "SECONDS|indefinite|minimal",
+        help_heading = CACHING_OPTIONS_HEADER,
+    )]
+    pub negative_cache_ttl: Option<TimeToLive>,
+
+    #[clap(
+        long,
         help = "Maximum size of the cache directory in MiB [default: preserve 5% of available space]",
         value_name = "MiB",
         value_parser = value_parser!(u64),
@@ -934,6 +942,11 @@ where
     }
     tracing::trace!("using metadata TTL setting {metadata_cache_ttl:?}");
     filesystem_config.cache_config = CacheConfig::new(metadata_cache_ttl);
+    if let Some(negative_cache_ttl) = args.negative_cache_ttl {
+        filesystem_config.cache_config = filesystem_config
+            .cache_config
+            .with_negative_cache_ttl(negative_cache_ttl);
+    }
 
     match (args.disk_data_cache_config(), args.express_data_cache_config()) {
         (None, Some((config, bucket_name, cache_bucket_name))) => {
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -323,7 +323,7 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
         value_name = "SECONDS|indefinite|minimal",
         help_heading = CACHING_OPTIONS_HEADER,
     )]
-    pub negative_cache_ttl: Option<TimeToLive>,
+    pub negative_metadata_ttl: Option<TimeToLive>,
 
     #[clap(
         long,
@@ -942,7 +942,7 @@ where
     }
     tracing::trace!("using metadata TTL setting {metadata_cache_ttl:?}");
     filesystem_config.cache_config = CacheConfig::new(metadata_cache_ttl);
-    if let Some(negative_cache_ttl) = args.negative_cache_ttl {
+    if let Some(negative_cache_ttl) = args.negative_metadata_ttl {
         filesystem_config.cache_config = filesystem_config
             .cache_config
             .with_negative_cache_ttl(negative_cache_ttl);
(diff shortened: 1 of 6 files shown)


mountpoint-s3/src/cli.rs
```diff
@@ -317,6 +317,14 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
     )]
     pub metadata_ttl: Option<TimeToLive>,
 
+    #[clap(
+        long,
+        help = "Time-to-live (TTL) for cached negative entries in seconds [default: same as metadata TTL]",
+        value_name = "SECONDS|indefinite|minimal",
+        help_heading = CACHING_OPTIONS_HEADER,
+    )]
+    pub negative_cache_ttl: Option<TimeToLive>,
```
passaro, line 326: I suggest `negative_metadata_ttl` for consistency with the other option.

mountpoint-s3/src/fs/config.rs
```diff
@@ -87,6 +87,8 @@ pub struct CacheConfig {
     pub file_ttl: Duration,
     /// How long the kernel will cache metadata for directories
     pub dir_ttl: Duration,
+    /// How long the kernel will cache negative entries
```
passaro, line 90: File and directory TTLs are sent to the kernel, the negative cache is implemented only in Mountpoint.
```suggestion
    /// How long the file system will cache negative entries
```

mountpoint-s3/tests/fuse_tests/lookup_test.rs
```diff
@@ -198,6 +199,7 @@ fn lookup_with_negative_cache(creator_fn: impl TestSessionCreator) {
                 serve_lookup_from_cache: true,
                 dir_ttl: Duration::from_secs(600),
                 file_ttl: Duration::from_secs(600),
+                negative_ttl: Duration::from_secs(600),
```
passaro, line 202:
```suggestion
                negative_cache_ttl: Duration::from_secs(600),
```

mountpoint-s3/src/fs/config.rs
```diff
@@ -130,14 +133,33 @@ impl CacheConfig {
...
+                negative_cache_ttl: ttl,
                 ..Default::default()
             },
         }
     }
+
+    pub fn with_negative_cache_ttl(self, negative_cache_ttl: TimeToLive) -> Self {
+        match negative_cache_ttl {
+            TimeToLive::Minimal => Self {
+                negative_cache_ttl: Self::default().negative_cache_ttl,
```
passaro, line 152: Just setting the TTL here may not be enough. We use the same `serve_lookup_from_cache` flag both for the "positive" cache and the negative cache ([here](https://github.com/awslabs/mountpoint-s3/blob/b8338af6019cbe2cbecdccf2805edafefef6a766/mountpoint-s3/src/superblock.rs#L792), in particular). I suspect we need to introduce a separate flag and ensure it is set to false on `Minimal` here.

passaro (commented):
Hi @notoriaga,  thank you for opening this pull request!

I've got a few suggestions, but the next step would be to add a couple of tests. See existing tests for the negative cache, e.g. `test_negative_lookup_with_caching` in `superblock.rs` and `lookup_with_negative_cache` in `lookup_test.rs`. Happy to help if you need more guidance.

mountpoint-s3/src/fs/config.rs
passaro: I still see an issue here where even when setting `--negative-metadata-ttl minimal`, the negative cache is still needlessly populated with a very short TTL. My suggestion is to add an independent flag to `CacheConfig`, e.g. `use_negative_cache`, and use it in `update_from_remote` to replace `serve_lookup_from_cache` (`mountpoint-s3/mountpoint-s3/src/superblock.rs:792`).

passaro (commented):
Nice. I have only one remaining suggestion.

notoriaga:
> Nice. I have only one remaining suggestion.

Sorry for the delay! I think I got what you mean. One thing I'm not sure about is I also set `use_negative_cache = true` when calling `with_negative_metadata_ttl` with something other than minimal. I guess that would let you enable negative metadata caching but not the normal caching. Not sure if that makes sense to do.

mountpoint-s3/tests/fs.rs
```diff
@@ -192,13 +192,7 @@ async fn test_read_dir_nested(prefix: &str) {
...
 async fn test_lookup_negative_cached() {
     let fs_config = S3FilesystemConfig {
-        cache_config: CacheConfig {
-            serve_lookup_from_cache: true,
-            dir_ttl: Duration::from_secs(600),
-            file_ttl: Duration::from_secs(600),
-            negative_cache_ttl: Duration::from_secs(600),
-            ..Default::default()
-        },
+        cache_config: CacheConfig::new(TimeToLive::Duration(Duration::from_secs(600))),
```
passaro, line 195: That's better! Thanks

passaro (approved):
Looks good. Thank you for your contribution!

PR #1247, opened 2025-01-27
Add debug logging to FUSE flush to make upload completion clearer

dannycjones:
In some edge cases, Mountpoint will not be able to complete the MPU before the file is closed. For instance, we will not complete uploads where no bytes have been written since it can be common for applications to fork and result in file descriptor being closed before writing begins.

If an application relies on close completing before another system queries S3, it could lead to a race condition where the object is not yet in S3.

While this is an edge case, this change adds debug logging which can help identify when this behavior occurs.

Logging change only.

No, logging change only.

---

### DIFF
--- a/mountpoint-s3/src/fs.rs
+++ b/mountpoint-s3/src/fs.rs
@@ -850,15 +850,22 @@ where
             FileHandleState::Write(write_state) => write_state,
         };
 
-        let result = write_state.complete_if_in_progress(&file_handle.full_key).await;
+        let complete_result = write_state.complete_if_in_progress(&file_handle.full_key).await;
         metrics::gauge!("fs.current_handles", "type" => "write").decrement(1.0);
-        if let Ok(true) = result {
-            debug!(key = ?&file_handle.full_key, "upload completed async after file was closed");
-        }
 
-        // Errors won't actually be seen by the user because `release` is async,
-        // but it's the right thing to do.
-        result.map(|_| ())
+        match complete_result {
+            Ok(upload_completed_async) => {
+                if upload_completed_async {
+                    debug!(key = ?&file_handle.full_key, "upload completed async after file was closed");
+                }
+                Ok(())
+            }
+            Err(e) => {
+                // Errors won't actually be seen by the user because `release` is async,
+                // but it's the right thing to do so we'll return it.
+                Err(e)
+            }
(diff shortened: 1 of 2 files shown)


mountpoint-s3/src/fs/handles.rs
```diff
@@ -329,16 +327,23 @@ where
...
                 ..
-            } => Self::complete_append(request, key, handle, initial_etag).await,
-            UploadState::MPUInProgress { request, handle, .. } => Self::complete_upload(request, key, handle).await,
-            UploadState::Failed(_) | UploadState::Completed => Ok(()),
+            } => Self::complete_append(request, key, handle, initial_etag)
+                .await
+                .map(|()| true),
+            UploadState::MPUInProgress { request, handle, .. } => {
+                Self::complete_upload(request, key, handle).await.map(|()| true)
+            }
```
passaro, line 345: nit: I'd prefer something like this:
```suggestion
            } => {
              Self::complete_append(request, key, handle, initial_etag).await?;
                Ok(true)
              }
            UploadState::MPUInProgress { request, handle, .. } => {
                Self::complete_upload(request, key, handle).await?;
                Ok(true)
            }
```

passaro (commented):
Happy with the change, just a couple of nits with `map`s.

mountpoint-s3/src/fs/handles.rs
dannycjones: That's a lot nicer, thanks

mountpoint-s3/src/fs.rs
```diff
@@ -852,9 +852,13 @@ where
...
         let result = write_state.complete_if_in_progress(&file_handle.full_key).await;
         metrics::gauge!("fs.current_handles", "type" => "write").decrement(1.0);
+        if let Ok(true) = result {
+            debug!(key = ?&file_handle.full_key, "upload completed async after file was closed");
+        }
+
         // Errors won't actually be seen by the user because `release` is async,
         // but it's the right thing to do.
-        result
+        result.map(|_| ())
```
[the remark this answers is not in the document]
dannycjones: Yeah, I think that's better, and more consistent with the approach in the other comment too.

passaro (approved):
LGTM!

PR #1248, opened 2025-01-27
Export missing types in mountpoint-s3-client API

passaro:
The S3 client API in `mountpoint-s3-client` used many types directly from `mountpoint-s3-crt`. This change makes sure that all types used in the API are re-exported in `mountpoint-s3-client`, so that users do not need to depend on `mountpoint-s3-crt` directly.

No functional changes.

Yes, it includes an entry in the `client` changelog.

---


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -13,16 +13,18 @@ use std::time::{Duration, Instant};
 
 use futures::future::{Fuse, FusedFuture};
 use futures::FutureExt;
-use mountpoint_s3_crt::auth::credentials::{
+pub use mountpoint_s3_crt::auth::credentials::{
     CredentialsProvider, CredentialsProviderChainDefaultOptions, CredentialsProviderProfileOptions,
+    CredentialsProviderStaticOptions,
```
erkanonl, line 18: I noticed that CredentialsProviderChainDefaultOptions and CredentialsProviderProfileOptions are not being used in mountpoint-s3, maybe they can stay as private?

Rest is lgtm!

passaro: We are not re-exporting them in `lib.rs` so the `pub` here only slightly extends visibility, but you are right. Fixed it now.

erkanonl (approved):

PR #1249, opened 2025-01-29
Use custom endpoint and region setting for all fork tests

c-hagem:
Some of our fork tests did not set the custom endpoint for the binary, if this is provided. This change should fix all of them to use a custom endpoint and to set the region, if that is configured.

Does not need a changelog entry, as it only affects tests.

---

### DIFF
--- a/mountpoint-s3/tests/fuse_tests/fork_test.rs
+++ b/mountpoint-s3/tests/fuse_tests/fork_test.rs
@@ -75,13 +75,14 @@ fn run_in_background_with_passed_fuse_fd() -> Result<(), Box<dyn std::error::Err
     );
 
     let mut cmd = Command::cargo_bin("mount-s3")?;
-    let child = cmd
-        .arg(&bucket)
+    cmd.arg(&bucket)
         .arg(format!("/dev/fd/{}", fd.as_raw_fd()))
         .arg(format!("--prefix={prefix}"))
-        .arg(format!("--region={region}"))
-        .spawn()
-        .expect("unable to spawn child");
+        .arg(format!("--region={region}"));
+    if let Some(endpoint_url) = get_test_endpoint_url() {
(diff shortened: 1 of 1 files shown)


mountpoint-s3/tests/fuse_tests/fork_test.rs
```diff
@@ -263,11 +270,13 @@ fn run_in_background_fail_on_mount() -> Result<(), Box<dyn std::error::Error>> {
     // the mount would fail from error 403 on HeadBucket
     let bucket = get_test_bucket_forbidden();
     let mount_point = assert_fs::TempDir::new()?;
+    let region = get_test_region();
 
     let mut cmd = Command::cargo_bin("mount-s3")?;
     cmd.arg(&bucket).arg(mount_point.path()).arg("--auto-unmount");
     if let Some(endpoint_url) = get_test_endpoint_url() {
         cmd.arg(format!("--endpoint-url={endpoint_url}"));
+        cmd.arg(format!("--region={region}"));
```
unexge, line 279: Should this be outside of this if block? Don't we want to set the region even if there is no endpoint url?

unexge (approved):
LGTM. Left one comment, not sure if that was intentional or not

mountpoint-s3/tests/fuse_tests/fork_test.rs
c-hagem: I wasn't sure if this could make our tests less resilient, if we always add the region setting (since we might also want to test that the region is auto-resolved correctly [not sure if this makes sense]) -- however for custom endpoints I expected the auto-resolution to fail.

unexge: Makes sense, I think we can keep it as-is then.

PR #1250, opened 2025-01-30
Purge file from the negative cache on create

c-hagem:
This change adds code to remove entries from the negative cache once the new file is created.

Does not need a changelog entry, as it's only a small bugfix.

---

### DIFF
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 * Add support for appending to objects originally uploaded with a CRC64-NVME checksum. ([#1235](https://github.com/awslabs/mountpoint-s3/pull/1235))
 
+### Other Changes
+* Fix an issue where file names were not expelled from negative cache.
+
 ## v1.14.0 (January 10, 2025)
 
 ### New features
--- a/mountpoint-s3/src/superblock.rs
+++ b/mountpoint-s3/src/superblock.rs
@@ -367,6 +367,8 @@ impl Superblock {
             let inode = self
                 .inner
                 .create_inode_locked(&parent_inode, &mut parent_state, name, kind, state, true)?;
+            // Expel inode from negative cache, if it is present
+            self.inner.negative_cache.remove(parent_inode.ino(), name);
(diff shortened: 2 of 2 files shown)


muddyfish:
I think it probably does want a changelog entry but under the 'other changes' list

mountpoint-s3/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 * Add support for appending to objects originally uploaded with a CRC64-NVME checksum. ([#1235](https://github.com/awslabs/mountpoint-s3/pull/1235))
 
+### Other Changes
+* Fix an issue where file names were not expelled from negative cache.
```
passaro, line 6: We should include a link to this PR, like in the other entries:
```suggestion
* Fix an issue where file names were not expelled from negative cache. ([#1250](https://github.com/awslabs/mountpoint-s3/pull/1250))
```

mountpoint-s3/src/superblock.rs
```diff
@@ -367,6 +367,8 @@ impl Superblock {
             let inode = self
                 .inner
                 .create_inode_locked(&parent_inode, &mut parent_state, name, kind, state, true)?;
+            // Expel inode from negative cache, if it is present
+            self.inner.negative_cache.remove(parent_inode.ino(), name);
```
passaro, line 371: Should we consider updating the cache in `create_inode_locked`? It would require changes to the `update_from_remote` code path, but it could be a better trade off.

mountpoint-s3/CHANGELOG.md
passaro: Also:
* why "file names"? Doesn't it apply to directories as well?
* we should qualify when the issue occurs.

passaro (commented):
Could you add a test reproducing the issue?

passaro:
nit: I'd use "purge" rather than  "expel" in the PR/changelog/comments

c-hagem:
Regarding tests, I was not able to reproduce this issue easily as a Rust test -- I tried to re-create the Python script I used in Rust, however that test worked as expected. I'll try this again.

Issue #1236, opened 2025-01-17
Update readdirplus metric to use histogram instead of counter


dannycjones:
Today, the `readdirplus.entries` metric will simple count all the entries returned in the interval (roughly 5 seconds). However, it would make more sense to return some statistics on how many entries were returned in a `readdirplus` request when there was more than one in a given interval.

The metric is emitted here: https://github.com/awslabs/mountpoint-s3/blob/ef20898e36247aa358417da999efeb65e2fe195f/mountpoint-s3/src/fuse.rs#L228

We should update this code to emit the metric using a histogram instead.

PR #1254, opened 2025-02-05
Update readdir and readdirplus metrics to use histogram instead of counter

mansi153:
Update readdir and readdirplus APIs' `fuse.readdir[plus].entries` metric to use histogram instead of counter, as it would make more sense to record statistics on how many entries were returned in a readdir[plus] request when there was more than one in a given interval, than recording the total readdir[plus] entries per interval.

Addresses https://github.com/awslabs/mountpoint-s3/issues/1236.

Yes, the `fuse.readdir[plus].entries` metric type has been changed.

No, it is only updating a metric. Since metric names and availability are considered unstable, this does not need a changelog entry or version change.

---

### DIFF
--- a/mountpoint-s3-fs/src/fuse.rs
+++ b/mountpoint-s3-fs/src/fuse.rs
@@ -235,7 +235,7 @@ where
         match block_on(self.fs.readdirplus(parent, fh, offset, replier).in_current_span()) {
             Ok(_) => {
                 reply.ok();
-                metrics::counter!("fuse.readdirplus.entries").increment(count as u64);
+                metrics::histogram!("fuse.readdirplus.entries").record(count as f64);
             }
             Err(e) => fuse_error!("readdirplus", reply, e, self, req),
         }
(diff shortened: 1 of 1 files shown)


mountpoint-s3/src/fuse.rs
```diff
@@ -225,7 +225,7 @@ where
         match block_on(self.fs.readdirplus(parent, fh, offset, replier).in_current_span()) {
             Ok(_) => {
                 reply.ok();
-                metrics::counter!("fuse.readdirplus.entries").increment(count as u64);
+                metrics::histogram!("fuse.readdirplus.entries").record(count as f64);
```
dannycjones, line 228: We should make the same change for the `readdir` operation, as well as `readdirplus`.

dannycjones (dismissed):
This looks good. Some operating systems use `readdir` instead of `readdirplus`, we should cover that one too. Otherwise, looks good to ship!

PR #1257, opened 2025-02-07
Enforce sse on writes to the xz cache

vladem:
With this PR, flags `--sse` and `--sse-kms-key-id` will also apply to object uploads to the xz cache. 

Yes, bumped the version.

Yes, added an entry.

---


mountpoint-s3/src/data_cache/express_data_cache.rs
```diff
@@ -217,17 +224,38 @@ where
...
 
-        self.client
-            .put_object_single(
-                &self.bucket_name,
-                &object_key,
-                &block_metadata.to_put_object_params(),
-                data,
-            )
+        let mut params = block_metadata.to_put_object_params();
+        let (sse_type, key_id) = self
```
passaro, line 228: Could we extract most of this logic into a helper function?

mountpoint-s3/tests/fuse_tests/cache_test.rs
```diff
@@ -6,6 +6,7 @@ use crate::common::s3::{get_test_bucket, get_test_prefix};
 use mountpoint_s3::data_cache::{DataCache, DiskDataCache, DiskDataCacheConfig};
 use mountpoint_s3::object::ObjectId;
 use mountpoint_s3::prefetch::caching_prefetch;
+use mountpoint_s3::ServerSideEncryption;
```
passaro, line 9: add `cfg` or move to a local use if possible

doc/CONFIGURATION.md
```diff
@@ -195,6 +195,9 @@ New objects can be uploaded using different server-side encryption (SSE) setting
 > [!IMPORTANT]
 > Mountpoint currently accepts only **KMS key ARN** as the value for `--sse-kms-key-id` argument. AWS KMS [defines](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#key-id) several other key identifiers, including key ID, key alias name and key alias ARN, which are not supported by Mountpoint.
 
+> [!IMPORTANT]
+> CLI arguments `--sse` and `--sse-kms-key-id` can be used to override or enforce SSE type of the objects uploaded to the [shared cache](#shared-cache). We advise against overriding the encryption type to `aws:kms`, when the bucket default encryption is `AES256`, which will result in read failures of cache blocks. For more information on behaviour of overriding SSE type for objects in S3 Express bucket see [documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-express-specifying-kms-encryption.html).
```
passaro, line 199: What do you think of something like this?

```suggestion
> [!IMPORTANT]
> When Mountpoint is configured to use a [shared cache](#shared-cache) in [Amazon S3 Express One Zone](https://aws.amazon.com/s3/storage-classes/express-one-zone/), the CLI arguments `--sse` and `--sse-kms-key-id` also apply to the objects uploaded to the shared cache bucket and can be used to enforce or override their encryption settings. However, we recommend configuring encryption settings to match the default for the shared cache bucket, since some overriding configurations are not supported in Amazon S3 Express One Zone and may result in degraded performance when Mountpoint fails to read from the cache. For more information on overriding SSE settings for objects in Amazon S3 Express One Zone see [documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-express-specifying-kms-encryption.html).
```

passaro (commented):
The change looks fine. Just a suggestion for the doc change.

PR #1262, opened 2025-02-10
Add timer metric over endpoint rule engine resolution

dannycjones:
Today, Mountpoint evaluates the rule engine with every request despite always using the same endpoint in its default configuration.

This change adds a histogram metric so that we can begin to have visibility on how long this is taking, and how this can be distributed.

No, it adds a new metric for endpoint resolution.

No, metrics are not considered a stable feature. This is a new metric addition only, no change or removal.

---


dannycjones:
I'd be tempted to write some criterion benchmarks for the resolver, but I'll leave that for now. I think this can just get us some quick visibility.

PR #1264, opened 2025-02-11
Add metrics for FUSE worker idle and total count

dannycjones:
Before this change, there was no visibility into how many FUSE worker threads within Mountpoint had been created nor any indication if they are all busy or not.

This change adds both a count for the number of FUSE worker threads that have been spawned, as well as a count measuring how many are currently considered idle.

With the current metric implementation, these are guages and are only emitted when the value is updated. They are emitted in the logs as follows:

    2025-02-11T13:00:14.647185Z  INFO mountpoint_s3::metrics: fuse.workers.idle_count: 3
    2025-02-11T13:00:14.647227Z  INFO mountpoint_s3::metrics: fuse.workers.total_count: 4
    2025-02-11T13:00:19.659416Z  INFO mountpoint_s3::metrics: fuse.workers.idle_count: 3
    2025-02-11T13:00:24.672336Z  INFO mountpoint_s3::metrics: fuse.workers.idle_count: 3
    2025-02-11T13:00:29.685867Z  INFO mountpoint_s3::metrics: fuse.workers.idle_count: 3

This adds a new metric only.

No, new metric only. Metrics aren't advertised as a stable feature.

---


mountpoint-s3/src/fuse/session.rs
```diff
@@ -182,11 +193,17 @@ impl<W: Work> WorkerPool<W> {
         else {
             return Ok(false);
         };
-        self.state.idle_worker_count.fetch_add(1, Ordering::SeqCst);
+
+        let new_count = old_count + 1;
+        let idle_worker_count = self.state.idle_worker_count.fetch_add(1, Ordering::SeqCst) + 1;
+        metrics::gauge!(METRIC_NAME_FUSE_WORKERS_TOTAL).set(new_count as f64);
```
muddyfish, line 199: Shouldn't this also be set in `start`?

dannycjones: It is set when `start` is called, since `start` will invoke this method.

PR #1266, opened 2025-02-11
Add initial version of benchmark experiment runner

dannycjones:
In order to investigate performance in Mountpoint, we want to be able to vary different parameters. In fact, it can be very useful to vary these parameters together to see how performance (such as sequential read throughput) changes as we vary two parameters together.

This change introduces a new benchmark running script which uses the Python framework Hydra to enumerate combinations of parameters, and then execute some function with each combination. The script manages the lifecycle of the `mount-s3` file system and collecting data into an output folder.

The change currently does not reuse the FIO definitions used by our regression benchmarks. In the mid-term, these should be reconciled.

This pull request (PR) supersedes a previous PR: https://github.com/awslabs/mountpoint-s3/pull/986.

No, this adds a new benchmark runner and benchmark definitions. This does not impact the Mountpoint file system.

No, no impact to Mountpoint file system or crates.

---


benchmark/benchmark.py
```diff
@@ -0,0 +1,223 @@
...
+        "start_time": start_time,
+    }
+
+    try:
+        mount_dir = tempfile.mkdtemp(suffix=".mountpoint-s3")
+        mp_version = _mount_mp(cfg, metadata, mount_dir)
+        mounted = True
+        metadata["mp_version"] = mp_version
+    except subprocess.SubprocessError as e:
+        log.error(f"Error during mounting: {e}")
```
muddyfish, line 198: `try/except/else` appears to be what we want here, and could mean we drop the `mounted` variable

benchmark/benchmark.py
```diff
@@ -0,0 +1,223 @@
...
+
+    if mounted:
```
renanmagagnin, line 200: Instead of having the mount flag and testing it here and before unmounting, wouldn't it be better to do these two steps inside of the first try block and raise/handle specific custom error types? Could also maybe make the success variable unnecessary.

benchmark/benchmark.py
```diff
@@ -0,0 +1,223 @@
...
+        try:
+            # TODO: Add resource monitoring during FIO job
+            _run_fio(cfg, mount_dir)
+            success = True
```
muddyfish, line 204: `try/except/else` also is probably wanted here as well

benchmark/benchmark.py
muddyfish: Though in production quality code, we'd probably want this to be a `with` block

benchmark/README.md
```diff
@@ -0,0 +1,43 @@
...
+You might use AWS profiles or set some credentials in the environment.
+
+To run the experiment, you can execute a command like this:
+
+```
+uv run benchmark.py -- s3_bucket=amzn-s3-demo-bucket
+```
+
+This will run the default experiment, including many different configuration combinations.
+Output is written to `multirun/` within directories for the date, time, and job run.
```
renanmagagnin, line 43: Maybe it could be nice to include an example output?

benchmark/benchmark.py
```diff
@@ -0,0 +1,223 @@
...
+    start_time = datetime.now(tz=timezone.utc)
+    metadata = {
+        "start_time": start_time,
+    }
+
+    try:
+        mount_dir = tempfile.mkdtemp(suffix=".mountpoint-s3")
+        mp_version = _mount_mp(cfg, metadata, mount_dir)
+        mounted = True
+        metadata["mp_version"] = mp_version
```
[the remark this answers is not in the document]
dannycjones: I'm not sure what you mean here, can you rephrase?

benchmark/README.md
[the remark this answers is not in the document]
dannycjones: Added a short desc on how the benchmark works, where its defined

benchmark/benchmark.py
dannycjones: I don't follow what we want the `else` for?

benchmark/benchmark.py
```diff
@@ -79,36 +117,42 @@ def _mount_mp(
...
 
     log.info(f"Mounting S3 bucket {bucket} with args: %s; env: %s", subprocess_args, subprocess_env)
-    metadata["mount_s3_command"] = " ".join(subprocess_args)
-    metadata["mount_s3_env"] = subprocess_env
-    output = subprocess.check_output(subprocess_args, env=subprocess_env)
+    try:
+        output = subprocess.check_output(subprocess_args, env=subprocess_env)
+    except subprocess.CalledProcessError as e:
+        log.error(f"Error during mounting: {e}")
+        raise MountError()
```
muddyfish, line 124: This is probably what you want: `raise MountError() from e`

benchmark/benchmark.py
```diff
@@ -0,0 +1,223 @@
...
+        bucket,
+        mount_dir,
+        "--log-metrics",
+        "--allow-overwrite",
+        "--allow-delete",
+        f"--log-directory={MP_LOGS_DIRECTORY}",
+    ]
+    subprocess_env = {}
+
+    if cfg['s3_prefix'] is not None:
```
sahityadg, line 97: Is it better to pass mountpoint args as they are passed to benchmarks than have another mapping? For example, why not specify --prefix instead of s3_prefix.

benchmark/benchmark.py
dannycjones: I think its implicit but updating it to be explicit anyway

benchmark/benchmark.py
```diff
@@ -0,0 +1,243 @@
...
+    ]
+    subprocess_env = {
+        "PATH": os.environ["PATH"],
+        "APP_WORKERS": str(cfg['application_workers']),
+        "SIZE_GIB": str(100),
+        "DIRECT": str(1 if cfg['direct_io'] else 0),
+        "UNIQUE_DIR": datetime.now(tz=timezone.utc).isoformat(),
+        # TODO: Confirm assumption that `libaio` should make direct IO go faster.
+        # TODO: Review if we should use sync or psync. We use `sync` in other benchmarks.
+        "IO_ENGINE": str("libaio" if cfg['direct_io'] else "psync"),
```
muddyfish, line 162: Nit: `str` call is not needed

benchmark/benchmark.py
```diff
@@ -0,0 +1,243 @@
...
+        "--eta=never",
+        "--output-format=json",
+        f"--output={fio_output_filepath}",
+        f"--directory={mount_dir}",
+        fio_job_filepath,
+    ]
+    subprocess_env = {
+        "PATH": os.environ["PATH"],
+        "APP_WORKERS": str(cfg['application_workers']),
+        "SIZE_GIB": str(100),
```
muddyfish, line 157: Why not "100"? Same below

benchmark/benchmark.py
```diff
@@ -0,0 +1,243 @@
...
+    old_log_dir = path.join(logs_directory, dir_entries[0])
+    new_log_path = "mountpoint-s3.log"
+    log.debug(f"Renaming {old_log_dir} to {new_log_path}")
+    os.rename(old_log_dir, new_log_path)
+    os.rmdir(logs_directory)
+
+
+def _write_metadata(metadata: dict[str, any]) -> None:
+    with open("metadata.json", "w") as f:
+        json_str = json.dumps(metadata, default=str)
```
muddyfish, line 202: Nit: `json.dump` here writes directly to a file object

muddyfish (approved):

muddyfish:
LGTM

benchmark/benchmark.py
dannycjones: I'll merge this for now, we can still change in future if needed.

PR #1269, opened 2025-02-18
Fuser fork submodule

ujinho:
### Overview
To be prepared for further refactoring we want to ease the supporting effort for the FUSER fork we're using.
As a first step we want to embed it as a git submodule which will be using the existing `fuser/fork` branch as a remote. 
This will allow us to keep the current sync flow with the original FUSER [repo](https://github.com/cberner/fuser) but skip the manual sync step in our `main` branch. 

We will be able to update the fork by running `git submodule update --remote mountpoint-s3-fuser` which is more idiomatic. So we still be able to test changes locally w/o publishing the fork.

*NB* We will need to commit the changes (dirty indices) after submodule's update.

This change is completely internal and does not impact customer-facing behavior.

This change does not require version or changelog changes.

---


.gitmodules
```diff
@@ -31,3 +31,7 @@
 [submodule "aws-c-s3"]
 	path = mountpoint-s3-crt-sys/crt/aws-c-s3
 	url = https://github.com/awslabs/aws-c-s3.git
+[submodule "mountpoint-s3-fuser"]
+	path = mountpoint-s3-fuser
+	url = git@github.com:awslabs/mountpoint-s3.git
```
passaro, line 36: Why not use https?
```suggestion
	url = https://github.com/awslabs/mountpoint-s3.git
```

passaro (commented):
We need to update all workflows so that they checkout the submodules.

E.g. `Formatting` and `Licenses` will need:
```yaml
      - name: Checkout code
        uses: actions/checkout@v4
        with:
          submodules: true
```

passaro (commented):
Looks good to me, but you need to `--signoff` all your commits - that's why DCO is failing.

passaro (approved):

PR #1271, opened 2025-02-19
chore(clippy): remove exclusions for false positive rules

ujinho:
### Overview
We removed exclusions for false positive Clippy rules as they were fixed and this workaround is no longer needed.

No, this is a small boilerplate change.

Change log and version changes are not needed.


passaro (approved):
LGTM!

PR #1280, opened 2025-02-24
Fix benchmark script column names for sweeped params

dannycjones:
If the column names are not defined in the Hydra script, they will be prefixed by `+`. In order to keep things simple on the analysis side, we'll name the columns with placeholder values which will be replaced when running the script with multiple experiments.

It changes output of the benchmark script only.

No, no changelog entry or version change as no change to Mountpoint itself.

---

### DIFF
--- a/benchmark/conf/config.yaml
+++ b/benchmark/conf/config.yaml
@@ -12,7 +12,8 @@ write_part_size: 16777216 # to allow for uploads of 100GiB
 
 metadata_ttl: "indefinite"
 
-fio_benchmarks: sequential_read  #, random_read, sequential_write
+fio_benchmarks:
+  - sequential_read
 
 # Path to Mountpoint binary. Recommended to use an absolute path.
 mountpoint_binary: !!null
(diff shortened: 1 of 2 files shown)


benchmark/conf/config.yaml
```diff
@@ -24,6 +24,13 @@ upload_checksums: !!null
 
 iterations: 1
 
+# Define defaults for these columns, but script will run as MULTIRUN by default
+fuse_threads: !!null
+application_workers: 1
+direct_io: false
+fio_benchmark: "sequential_read"
+iteration: 1
```
sahityadg, line 32: nit: Should we use iterations and fio_benchmarks to define defaults and remove overrides, as they are also setting the same value anyway.

dannycjones: For iterations, I think using something static makes sense since it doesn't actually depend on the `iterations` value. (It should be `0` though, I'll update that.)

I think this makes a lot of sense for `fio_benchmark`, I'll update that.

PR #1294, opened 2025-02-28
Fork submodule revert

ujinho:
As we decided to U-turn from the submodule approach and back to keeping FUSER fork in the `vendor/fuser` folder we need to revert these changes.

No, this change only affects sync loop.

No

---


.github/workflows/bench.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       uses: actions-rust-lang/setup-rust-toolchain@v1
       with:
         # setup-rust-toolchain sets "-D warnings" by default, and Rust treats any warning as compile error.
-        # We need to this currently because `mountpoint-s3-fuser` contains some warnings and it breaks the build.
+        # We need to this currently because `vendor/fuser` contains some warnings and it breaks the build.
```
passaro, line 60: Let's keep the new directory name

Issue #1268, opened 2025-02-18
The file was not readable while being written


whitejdae:
### Mountpoint for Amazon S3 version

v1.12

### Describe the running environment

- The local kubernetes cluster uses the version: 1.31
- Use Kustomize to install
- The operating system that the pod runs is:
> # cat /etc/os-release
> NAME="Rocky Linux"
> VERSION="9.3 (Blue Onyx)"
> ID="rocky"
> ID_LIKE="rhel centos fedora"
> VERSION_ID="9.3"
> PLATFORM_ID="platform:el9"
> PRETTY_NAME="Rocky Linux 9.3 (Blue Onyx)"
> ANSI_COLOR="0;32"
> LOGO="fedora-logo-icon"
> CPE_NAME="cpe:/o:rocky:rocky:9::baseos"
> HOME_URL="https://rockylinux.org/"
> BUG_REPORT_URL="https://bugs.rockylinux.org/"
> SUPPORT_END="2032-05-31"
> ROCKY_SUPPORT_PRODUCT="Rocky-Linux-9"
> ROCKY_SUPPORT_PRODUCT_VERSION="9.3"
> REDHAT_SUPPORT_PRODUCT="Rocky Linux"
> REDHAT_SUPPORT_PRODUCT_VERSION="9.3"

### Mountpoint options

```shell
Default loading configuration：https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/install.md#deploy-driver
```

### What happened?

- The java program configures the GC log parameters to mount to s3, but s3 must wait until the GC log is written before uploading to s3, and an error message is reported

> Feb 17 22:10:39 k8s-node04 mount-s3[1683809]: [WARN] mountpoint_s3::cli: failed to detect network throughput. Using 10 gbps as throughput. Use --maximum-throughput-gbps CLI flag to configure a target throughput appropriate for the instance. Detection failed due to: failed to get instance type: IMDS query failed: Unknown CRT error
> Feb 17 22:11:08 k8s-node04 mount-s3[1683809]: [WARN] getxattr{req=76 ino=8 name="security.capability"}: mountpoint_s3::fuse: getxattr failed: operation not supported by Mountpoint
[...]

PR #1297, opened 2025-03-03
Disallow append on non-express buckets

vladem:
Fail early (on mount) with a descriptive error message, when `--incremental-upload` flag is used with standard buckets. We've seen in https://github.com/awslabs/mountpoint-s3/issues/1268 when an error reported on flush can be ignored by certain workloads.
___
Decision is made based on endpoint resolution [infer_s3_personality](https://github.com/awslabs/mountpoint-s3/blob/0c51de8dbd7f57d1a368ddcf522d09c7d89ecd34/mountpoint-s3/src/cli.rs#L1370). This may erroneously assume that the bucket is S3 Standard if either endpoint resolution failed or `auth_scheme` can not be parsed. In case of an error user may bypass it by specifying `--bucket-type=directory` flag. 

As an alternative, we can infer if the bucket is a directory bucket from the name: [docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/directory-bucket-naming-rules.html). Then we can assume that all directory buckets support `--incremental-upload`.

Yes, `mount-s3 s3-standard-bucket <folder> --incremental-upload` will start failing on mount.

Yes, added.

---

### DIFF
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -5,6 +5,10 @@
 * Reduce memory usage when using the `--prefix` flag. ([#1303](https://github.com/awslabs/mountpoint-s3/pull/1303))
 * Add support for endpoint override in credential providers. ([aws-c-auth#263](https://github.com/awslabs/aws-c-auth/pull/263/))
 
+### Breaking changes
+
+* Disallow the usage of `--incremental-upload` command-line argument on non S3 Express buckets ([#1297](https://github.com/awslabs/mountpoint-s3/pull/1297)).
+
 ## v1.15.0 (February 27, 2025)
 
 ### New features
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 ### Breaking changes
 
-* Disallow the usage of `--incremental-upload` command-line argument on non S3 Express buckets ([#1297](https://github.com/awslabs/mountpoint-s3/pull/1297)).
+* Disallow the usage of `--incremental-upload` command-line argument on non S3 Express buckets. ([#1297](https://github.com/awslabs/mountpoint-s3/pull/1297))
 
 ## v1.15.0 (February 27, 2025)
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2753,7 +2753,7 @@ dependencies = [
 
 [[package]]
 name = "mountpoint-s3"
-version = "1.15.1"
+version = "1.16.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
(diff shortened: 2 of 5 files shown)


mountpoint-s3/src/cli.rs
```diff
@@ -877,6 +877,12 @@ where
 
     let (client, runtime, s3_personality) = client_builder(&args)?;
 
+    if args.incremental_upload && !matches!(s3_personality, S3Personality::ExpressOneZone) {
+        return Err(anyhow!(
+            "--incremental-upload is only supported for S3 Express One Zone buckets"
+        ));
+    }
```
dannycjones, line 884: This should be below the log on line 887 that describes the detected bucket type

mountpoint-s3/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Breaking changes
+
+* Disallow the usage of `--incremental-upload` command-line argument on non S3 Express buckets ([#1297](https://github.com/awslabs/mountpoint-s3/pull/1297)).
```
dannycjones, line 5: very nitpick: the period should be before the PR link to be consistent with prev changelog.

(only putting this since we need to make the other change)

mountpoint-s3/src/cli.rs
dannycjones, line 884: I think the personality check should live in the impl of `S3Personality` to match things like `is_list_ordered`, etc..

dannycjones (changes requested):
just some minor things.

I think the logging is important, in case there's ever any changes to the way customers access S3 Express One Zone. That will provide a pointer that the logic has updated, and prompt to check for an update to Mountpoint.

One thing making me a little uncomfortable is that we don't have any tests for the `S3Personality` type. I wonder if its worth writing some quick ones that check for a basic endpoint config and bucket name, do we get the expected personality? i.e. here's a directory bucket name, yes it maps correct. here's a bucket name and an outposts endpoint, yes it maps correctly.

mountpoint-s3/src/cli.rs
```diff
@@ -877,6 +877,12 @@ where
 
     let (client, runtime, s3_personality) = client_builder(&args)?;
 
+    if args.incremental_upload && !matches!(s3_personality, S3Personality::ExpressOneZone) {
+        return Err(anyhow!(
+            "--incremental-upload is only supported for S3 Express One Zone buckets"
```
passaro, line 882: This is what we use in [docs](https://github.com/awslabs/mountpoint-s3/blob/main/doc/SEMANTICS.md#reading-and-writing-files):
```suggestion
            "--incremental-upload is only supported on directory buckets in S3 Express One Zone"
```

passaro (commented):
We should also mention the limitation in the help for `--incremental-upload`.

PR #1301, opened 2025-03-04
Add Python lints to CI

dannycjones:
Before this change, there is no linting or style checks enforced on Python code outside of manual review.

This change introduces both using [Ruff](https://docs.astral.sh/ruff/), a linter/checker written by the same organization owning [uv](https://docs.astral.sh/uv/) which we use as the package manager/runner in `benchmark/` project.

No change to Mountpoint. No functional changes to the Python scripts. Two new GitHub CI jobs are introduced to add checks on the Python code.

No, no impact to Mountpoint itself nor its crates.

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -308,3 +308,46 @@ jobs:
         uses: actions/checkout@v4
       - name: Run cargo deny
         uses: EmbarkStudios/cargo-deny-action@v2
+
+
+  benchmark-script-checks:
+    name: Benchmark script checks
+    runs-on: ubuntu-22.04
+    defaults:
+      run:
+        working-directory: ./benchmark/
+
+    steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -308,3 +308,46 @@ jobs:
...
+        uses: actions/checkout@v4
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: "./benchmark/.python-version"
+      - name: Format check
+        # TODO: Remove preservation of quote style
+        run: uvx ruff format --diff . --config "format.quote-style='preserve'"
```
dannycjones, line 331: This currently ignores quote style just to avoid making this a massive diff.

PR #1305, opened 2025-03-06
Reduce memory used to store inode names

passaro:
Each inode currently stores two separate strings for the key and the name (always contained in the key string), resulting in redundant memory usage. This change introduces a new `ValidKey` type which avoids the duplication by only storing the key and the offset of the name for O(1) retrieval.
`ValidKey` (and the related type `ValidName`) also enforce validation for the name and the whole key at construction time, allowing calling code to rely on the strings to be well-formed.

No.

No.

---


mountpoint-s3/src/superblock/path.rs
```diff
@@ -0,0 +1,218 @@
...
+
+impl<'a> TryFrom<&'a str> for ValidName<'a> {
+    type Error = InodeError;
+
+    fn try_from(value: &'a str) -> Result<Self, Self::Error> {
+        Self::parse_str(value)
+    }
+}
+
+impl Deref for ValidName<'_> {
```
vladem, line 123: a bit non-intuitive that `*valid_name` and `&valid_name` are both `&str`, where do we need `Deref` implementation?

vladem (approved):
lgtm, a couple of questions, non-blocking

mountpoint-s3/src/superblock/path.rs
passaro: `Deref` also gives us all the methods on `&str`, see [Deref coercion](https://doc.rust-lang.org/std/ops/trait.Deref.html#deref-coercion).

vladem (approved):

PR #1309, opened 2025-03-07
Pin the ring crate version as version 0.17.13 breaks build on cloud

ujinho:
We pin `ring = "=0.17.12"` because the latest version ring = "=0.17.13" breaks the build on Cloud Desktops.
Update was done previously because we'd want to eliminate vulnerabilities.
Pinned version is also fine from RustSec point of view ((link)[https://rustsec.org/advisories/RUSTSEC-2025-0009.html])

This change does not impact the current behavior 

No

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2603,6 +2603,7 @@ dependencies = [
  "rand",
  "rand_chacha",
  "regex",
+ "ring",
  "serde",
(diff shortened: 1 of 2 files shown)


passaro:
Mountpoint does not depend on `ring`, except for integration tests (dev-dependencies).

PR #1332, opened 2025-03-21
Add mock-mount-s3 to benchmark/ scripts

dannycjones:
This change allows us to run our benchmark scripts in `benchmark/` using the `mock-mount-s3` binary, which presents a Mountpoint file system backed by an in-memory mock S3 client.

This change itself incorporates quite a few changes (which may have been better suited as separate commits). There are some changes to accommodate configuration of part sizes in `mock-mount-s3`, removal of throughput limits (which is useful for benchmarking!), and finally adding the configuration options to the benchmarking scripts.

This change does include some hardcoded objects being added to `mock-mount-s3` which can accomodate the benchmarking scripts. This means that if the object keys change, the files will be created by FIO and "uploaded" / populated in memory, which probably isn't what you want.

No, there are no changes to main Mountpoint code.

No, no behavior changes new or existing.

---

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -76,9 +76,11 @@ def _mount_mp(
         ]
 
         if stub_mode == "s3_client":
+            # `mock-mount-s3` requires bucket to be prefixed with `sthree-` to verify we're not actually reaching S3
             logging.debug("using mock-mount-s3 due to `stub_mode`, bucket will be prefixed with \"sthree-\"")
-            mountpoint_args.append("--bin=mock-mount-s3")
             bucket = f"sthree-{cfg['s3_bucket']}"
+
+            mountpoint_args.append("--bin=mock-mount-s3")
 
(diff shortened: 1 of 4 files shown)


benchmark/benchmark.py
```diff
@@ -63,22 +63,30 @@ def _mount_mp(
...
             "cargo",
             "run",
             "--quiet",
             "--release",
-            "--",
+            "--features=mock",
         ]
+
+        if stub_mode == "s3_client":
+            logging.debug("using mock-mount-s3 due to `stub_mode`, bucket will be prefixed with \"sthree-\"")
```
muddyfish, line 79: Why? Just to make sure we're not actually touching s3? (should be included as a comment)

muddyfish:
The Python bits look good to me, I don't have too much context on the Rust though

dannycjones:
@muddyfish could you take a look at this PR? We don't really have anyone specific with more context on the Rust bits.

muddyfish (approved):

Issue #1331, opened 2025-03-21
Race condition on get_object path


IsaevIlya:
### Mountpoint for Amazon S3 version

mountpoint-s3-client v0.13.0

### AWS Region

us-east-1

### Describe the running environment

Running on g5.8xlarge instance (AMI- amazon/ubuntu/images/hvm-ssd-gp3/ubuntu-noble-24.04-amd64-server-20250115) in private VPC that access S3 through gateway. Instance use long term credentials exposed via environment variables.

### Mountpoint options

```shell
We are using mountpoint-s3-client directly
```

### What happened?

While investigating an issue where users occasionally receive empty responses for GET requests (see https://github.com/awslabs/s3-connector-for-pytorch/issues/293), we identified a potential race condition in the `get_object` logic. The issue may arise when the `part_receiver` is not yet ready [at this point](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-client/src/s3_crt_client/get_object.rs#L242), but the subsequent `request` call [here](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-client/src/s3_crt_client/get_object.rs#L247) completes before data transfer begins. This leads to the stream being closed [at this line](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-client/src/s3_crt_client/get_object.rs#L248), despite no data having been transferred to the client.

### Relevant log output

```shell
I have modified source code of mountpoint-s3-client to get logs presented below. I'm going to add diff with changes after logs.

2025-03-20T10:07:46.976405Z DEBUG _mountpoint_s3_client::mountpoint_s3_client_inner: Executing initial request key="ILSVRC/Data/CLS-LOC/val/ILSVRC2012_val_00013144.JPEG" params=GetObjectParams { range: None, if_match: None, checksum_mode: None }
2025-03-20T10:07:46.976626Z DEBUG get_object{id=343 bucket="imagenet-ds" key="ILSVRC/Data/CLS-LOC/val/ILSVRC2012_val_00013144.JPEG" range=None if_match=None}: mountpoint_s3_client::s3_crt_client::get_object: new request
2025-03-20T10:07:47.049884Z TRACE s3getobjectresponse{key=ILSVRC/Data/CLS-LOC/val/ILSVRC2012_val_00013144.JPEG}: mountpoint_s3_client::s3_crt_client::get_object: poll_next called, request terminated: false
[...]

PR #1334, opened 2025-03-26
Fix race condition in GetObject that could result in empty responses

passaro:
Address an issue in the `Stream` implementation for `S3GetObjectResponse` that could immediately return `None` (i.e. terminate the stream) when detecting that the meta request had completed, before returning previously received parts. Reported in #1331.

We changed the mechanism used to extract the response body parts and the request completion from the meta request callbacks. Instead of multiple independent channels, we introduced a single channel that supports multiple `S3GetObjectEvent`s. The events in the new channel match the order in which the callbacks are invoked, which is guaranteed by the CRT. The events channel also includes the `Headers` event, avoiding the need of a separate channel to await for the headers to be returned.

Additionally, we simplified the group of `make_meta_request` methods in `S3CrtClientInner`: `make_meta_request_with_callbacks` is the most customizable, used directly by `get_object` and `put_object`, while the simpler `make_meta_request` and `make_meta_request_with_body` are used by the other requests.

No.

Bug fix entry and increase patch version.

---


mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -69,41 +69,49 @@ impl S3CrtClient {
...
-                // If the MPU was not created, the request must have failed.
-                result?;
-                return Err(S3RequestError::internal_failure(S3PutObjectRequestError::CreateMultipartUploadFailed).into());
-            }
-        };
-
-        // Guaranteed when select_biased! executes the CreateMPU branch.
-        assert!(!request.is_terminated());
+        // Wait for CreateMultipartUpload to complete, or return error.
+        mpu_created.await.unwrap()?;
```
IsaevIlya, line 109: Would we wait indefinitely for `mpu_created` here? Is there any timeout under the hood or knowledge about  `CreateMultipartUpload` that assure us that we won't wait forever?

passaro: See from line 93 above: if we get a result instead of the CreateMultipartUpload completing, it will be an error and we send it to this channel.

IsaevIlya (dismissed):

mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -69,41 +69,49 @@ impl S3CrtClient {
...
+                            _ = sender.send(Ok(()));
                         }
                     }
                 },
-                |_| None,
                 on_headers,
+                |_, _| {},
+                |_| None,
+                move |result| {
+                    if let Some(sender) = on_failure_sender.lock().unwrap().take() {
```
unexge, line 93: Maybe using an atomic bool instead of `Arc<Mutex<Option<...>>>` could be simpler here

passaro: Not sure what you have in mind.

Another approach I considered was to introduce an "event" channel like in `get_object`, but for `MpuCreated` and `Result`. However, it would require more changes to `S3PutObjectRequest` and we are considering removing the wait for CreateMultipartUpload anyway, so I decided to go for the less intrusive change.

mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -229,29 +223,35 @@ impl GetObjectResponse for S3GetObjectResponse {
...
-        match this.request.poll(cx) {
-            Poll::Ready(Ok(_)) => Poll::Ready(None),
-            Poll::Ready(Err(e)) => Poll::Ready(Some(Err(e))),
+        match this.event_receiver.poll_next(cx) {
+            Poll::Ready(None) => Poll::Ready(None),
+            Poll::Ready(Some(S3GetObjectEvent::BodyPart(offset, part))) => {
+                *this.next_offset = offset + part.len() as u64;
+                Poll::Ready(Some(Ok((offset, part))))
+            }
+            Poll::Ready(Some(S3GetObjectEvent::Headers(_))) => Poll::Pending,
```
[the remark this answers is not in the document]
passaro: In fact, it could happen before. I've now updated the `on_headers` callback to only send the headers once and added an assertion here.

mountpoint-s3-client/src/s3_crt_client/put_object.rs
unexge: I was thinking of having only one channel for all results and keeping track of errors in a `AtomicBool`, but I think that way the initial `CreateMultipartUploadFailed` error would be emitted once the caller first polls this method rather than initially from this method.

unexge (dismissed):

passaro:
I have rebased and taken out the `make_meta_request` refactor. Now this change only includes the race condition fix. I'll open a separate PR for the other changes.

adpeace (commented):
lgtm, one minor comment about an inaccurate comment that can be addressed separately.

adpeace (approved):

PR #1337, opened 2025-03-31
Refactor internal S3 client methods to initiate meta-requests

passaro:
Re-organize the group of internal methods in the S3 client used to initiate CRT meta-requests. Follow-up to the changes in #1334.

With this change, `S3CrtClientInner` provides 4 methods:
* `meta_request_with_callbacks`: the lowest-level method, which allows more customization through callbacks. It is the basis for the other methods and it is used directly by `get_object` and `put_object`.
* `meta_request_with_body_payload`: simpler variant that returns the response body. Used by `list_object` and `get_object_attributes`.
* `meta_request_with_headers_payload`: variant returning the headers, but no body. Used by `head_object` and `put_object_single`.
* `meta_request_without_payload`: simplest variant, only returns error.

No.

No. Internal change only.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -758,7 +758,12 @@ impl S3CrtClientInner {
             parse_meta_request_error,
             move |result| {
                 // On success, extract the headers.
-                _ = tx.send(result.map(|_| on_result.lock().unwrap().take().unwrap()))
+                _ =
+                    tx.send(result.and_then(|_| {
+                        on_result.lock().unwrap().take().ok_or_else(|| {
+                            S3RequestError::internal_failure(ResponseHeadersError::MissingHeaders).into()
+                        })
+                    }))
             },
         )?;
         Ok(S3MetaRequest {
@@ -829,6 +834,13 @@ impl S3CrtClientInner {
     fn next_request_counter(&self) -> u64 {
         self.next_request_counter.fetch_add(1, Ordering::SeqCst)
     }
+}
+
+/// Failure retrieving headers
+#[derive(Debug, Error)]
+enum ResponseHeadersError {
+    #[error("response headers are missing")]
+    MissingHeaders,
 }
 
 /// S3 operation supported by this client.
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -509,60 +509,14 @@ impl S3CrtClientInner {
         })
     }
 
-    fn new_meta_request_options(message: S3Message, operation: S3Operation) -> MetaRequestOptions {
-        let mut options = MetaRequestOptions::new();
-        if let Some(checksum_config) = message.checksum_config {
-            options.checksum_config(checksum_config);
-        }
-        if let Some(signing_config) = message.signing_config {
-            options.signing_config(signing_config);
-        }
-        options
-            .message(message.inner)
-            .endpoint(message.uri)
-            .request_type(operation.meta_request_type());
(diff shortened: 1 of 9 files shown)


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -780,7 +718,78 @@ impl S3CrtClientInner {
             parse_meta_request_error,
             move |result| _ = tx.send(result.map(|_| std::mem::take(&mut *body.lock().unwrap()))),
```
muddyfish, line 719: I know this isn't changing, but is panicking really the best option here compared to say just failing this specific request with an internal error?

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -780,7 +718,78 @@ impl S3CrtClientInner {
...
+                // then on_meta_request_result will send an error.
+                if (200..300).contains(&status) {
+                    *on_headers.lock().unwrap() = Some(headers.clone());
+                }
+            },
+            |_, _| {},
+            parse_meta_request_error,
+            move |result| {
+                // On success, extract the headers.
+                _ = tx.send(result.map(|_| on_result.lock().unwrap().take().unwrap()))
```
muddyfish, line 761: Same here about unwrapping - is it better than logging and returning some kind of internal error?

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -733,45 +687,29 @@ impl S3CrtClientInner {
...
-        parse_meta_request_error: impl FnOnce(&MetaRequestResult) -> Option<E> + Send + 'static,
-    ) -> Result<S3HttpRequest<Vec<u8>, E>, S3RequestError> {
-        let options = Self::new_meta_request_options(message, operation);
-        self.make_simple_http_request_from_options(options, request_span, |_| {}, parse_meta_request_error, |_, _| ())
-    }
-
-    /// Make an HTTP request using this S3 client that returns the body on success or invokes the
-    /// given callback on failure. See [make_simple_http_request] for arguments.
-    fn make_simple_http_request_from_options<E: std::error::Error + Send + 'static>(
+    /// The `parse_meta_request_error` callback is invoked on failed requests. It should return `None`
```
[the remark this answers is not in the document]
passaro: That callback only parses request-specific errors, then we apply the generic error-handling logic.

mountpoint-s3-client/src/s3_crt_client.rs
passaro: `unwrap` here is just for the lock, it will fail if another thread has panicked while holding the lock

mountpoint-s3-client/src/s3_crt_client.rs
passaro: Yes. Fixed that.

muddyfish (dismissed):

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -780,7 +718,83 @@ impl S3CrtClientInner {
...
+                // Only store headers if we have a 2xx status code. If we only get other status codes,
+                // then on_meta_request_result will send an error.
+                if (200..300).contains(&status) {
+                    *on_headers.lock().unwrap() = Some(headers.clone());
+                }
+            },
+            |_, _| {},
+            parse_meta_request_error,
+            move |result| {
+                // On success, extract the headers.
```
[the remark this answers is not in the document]
passaro: We always send a `Result<Headers, _>` to `tx`. Hopefully it's clearer now.

unexge (approved):

PR #1340, opened 2025-04-01
Create a package with custom extension

vladem:
Add a `--pkg-extensions` CLI flag which specifies which packages to build and how to name them. Package type is inferred from top-level component of the provided extension. 

Also add a validation script for such packages.

No.

No.

---

### DIFF
--- a/package/README.md
+++ b/package/README.md
@@ -32,8 +32,7 @@ Then install the depdencies:
 
     sudo yum install fuse fuse-devel make cmake3 clang git pkg-config dpkg fakeroot rpmdevtools tar python3
 
-If you don't want to compile a DEB, you can skip `dpkg` and `fakeroot`, and use `--pkg-extensions` flag of the build script to specify which package to build.
-If you don't want to compile an RPM, you can skip `rpmdevtools`, and use `--pkg-extensions` flag of the build script to specify which package to build.
+You can skip `dpkg` and `fakeroot` if you don't want to compile a DEB, and `rpmdevtools` if you don't want to compile an RPM. Use the `--pkg-extensions` flag on the build script to specify which package to build.
 
 You'll also need Rust, and the [cargo-about] tool to generate third-party attribution documents:
 
--- a/package/README.md
+++ b/package/README.md
@@ -32,8 +32,8 @@ Then install the depdencies:
 
     sudo yum install fuse fuse-devel make cmake3 clang git pkg-config dpkg fakeroot rpmdevtools tar python3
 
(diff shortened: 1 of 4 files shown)


package/package.py
```diff
@@ -338,14 +338,26 @@ def build(args: argparse.Namespace) -> str:
...
     artifacts = []
-    if not args.no_rpm:
-        artifacts.append(build_rpm(metadata, package_dir))
-    if not args.no_suse_rpm:
-        artifacts.append(build_rpm(metadata, package_dir, "suse"))
-    if not args.no_deb:
-        artifacts.append(build_deb(metadata, package_dir))
-    artifacts.append(build_package_archive(metadata, package_dir))
+    for ext in args.pkg_extensions:
+        pkg_type = infer_package_type(ext)
```
passaro, line 354: nit: the `infer_package_type` function seems a bit redundant. We could just have the `if..elif`s below check the extension directly, e.g. `ext.endswith("rpm"):`

passaro: and still fallback on the exception if there is no match.

package/README.md
```diff
@@ -32,8 +32,8 @@ Then install the depdencies:
 
     sudo yum install fuse fuse-devel make cmake3 clang git pkg-config dpkg fakeroot rpmdevtools tar python3
 
-If you don't want to compile a DEB, you can skip `dpkg` and `fakeroot`, and pass the `--no-deb` flag to the build script.
-If you don't want to compile an RPM, you can skip `rpmdevtools`, and pass the `--no-rpm` flag to the build script.
+If you don't want to compile a DEB, you can skip `dpkg` and `fakeroot`, and use `--pkg-extensions` flag of the build script to specify which package to build.
+If you don't want to compile an RPM, you can skip `rpmdevtools`, and use `--pkg-extensions` flag of the build script to specify which package to build.
```
passaro, line 36: Let's simplify:
```suggestion
You can skip `dpkg` and `fakeroot` if you don't want to compile a DEB, and `rpmdevtools` if you don't want to compile an RPM. Use the `--pkg-extensions` flag on the build script to specify which package to build.
```

passaro (commented):
A couple of minor suggestions.

package/package.py
```diff
@@ -42,9 +41,9 @@ class BuildMetadata:
     def artifact_name(self, extension: str):
         return f"mount-s3-{self.version_string}-{self.arch_name}.{extension}"
 
-    def spec_file_name(self, distr: Optional[str] = None):
-        if distr:
-            return f"mount-s3-{distr}.spec"
+    def spec_file_name(self, ext: str):
+        if "suse" in ext:
```
muddyfish, line 45: Why is this changing?

PR #1346, opened 2025-04-02
Reduce memory usage for strings in inode metadata

passaro:
Reduce memory usage for strings included in inode metadata, like object keys, etags, and inode names. Using a `Box<str>` instead of a `String` ensures that no slack capacity is wasted and saves the `usize` field to keep track of the buffer capacity.

No functional changes.

Yes.

---

### DIFF
--- a/mountpoint-s3-fs/src/superblock/inode.rs
+++ b/mountpoint-s3-fs/src/superblock/inode.rs
@@ -317,7 +317,7 @@ pub struct InodeStat {
     /// Time of last access
     pub atime: OffsetDateTime,
     /// Etag for the file (object)
-    pub etag: Option<String>,
+    pub etag: Option<Box<str>>,
     /// Inodes corresponding to S3 objects with GLACIER or DEEP_ARCHIVE storage classes
     /// are only readable after restoration. For objects with other storage classes
     /// this field should be always `true`.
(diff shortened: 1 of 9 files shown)


mountpoint-s3-fs/src/superblock/inode.rs
```diff
@@ -344,9 +344,9 @@ impl InodeStat {
     /// restored, and so we override their permissions to 000 and reject reads to them. We also warn
     /// the first time we see an object like this, because FUSE enforces the 000 permissions on our
     /// behalf so we might not see an attempted `open` call.
-    fn is_readable(storage_class: Option<String>, restore_status: Option<RestoreStatus>) -> bool {
+    fn is_readable(storage_class: Option<&str>, restore_status: Option<RestoreStatus>) -> bool {
```
vladem, line 347: `storage_class` is still stored as `String` right? (don't see it changed to `Box<str>`)

mountpoint-s3-fs/src/superblock/path.rs
```diff
@@ -35,12 +35,19 @@ impl ValidKey {
...
         };
 
-        let mut key = self.as_ref().to_owned();
-        let name_offset = key.len();
+        let name_offset = self.key.len();
+        // Allocate the new string with the correct capacity.
+        let mut key =
+            String::with_capacity(name_offset + name.len() + if kind == InodeKind::Directory { 1 } else { 0 });
+        key.push_str(&self.key);
         key.push_str(&name);
```
vladem, line 43: > ensures that no slack capacity is wasted

regarding this aspect from the description, do you think slack capacity was added here, when pushing to the string? my assumption would be that initial allocation `self.as_ref().to_owned()` would use exactly the right amount of memory, so for `etag` and `storage_class` the saving is exactly one `usize` (previously used to track capacity)

vladem (commented):
a couple of questions and I'd wait for benchmark results, but otherwise LGTM

mountpoint-s3-fs/src/superblock/inode.rs
passaro: We do not store it in the inodes, it's only used here to determine the `is_readable` flag.

mountpoint-s3-fs/src/superblock/path.rs
passaro: Yes, the slack capacity savings are only for keys. For `etag` and the children names we only save the `usize` field.

vladem (approved):

PR #1348, opened 2025-04-03
`GetObjectResponse` returns part content as `Bytes` rather than `Box<[u8]>`

passaro:
Modify the `GetBodyPart` type streamed from `GetObjectResponse` so that it exposes the part content as a `Bytes` type, rather than as a `Box<[u8]>`. This is an API breaking change for `mountpoint-s3-client`, which will require minor adjustments for users consuming the part content. The switch to `Bytes` will enable the introduction of different buffer allocation strategies in future releases.

No functional changes, but it is a minor API breaking change.

Yes, it changes `mountpoint-s3-client` public API.

---


mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -98,7 +99,10 @@ impl S3CrtClient {
                     }
                 },
                 move |offset, data| {
-                    _ = part_sender.unbounded_send(S3GetObjectEvent::BodyPart(offset, data.into()));
+                    _ = part_sender.unbounded_send(S3GetObjectEvent::BodyPart(GetBodyPart {
+                        offset,
+                        data: Bytes::copy_from_slice(data),
```
muddyfish, line 104: Do we need to copy here?

passaro: Because `data`'s lifetime is restricted to this callback. We are considering ways in which we could take ownership of the underlying buffer in the future and at that point we will be able to leverage [`Bytes::from_owner`](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.from_owner) to avoid the copy.

muddyfish (approved):

PR #1355, opened 2025-04-08
Fix an issue where GetObject requests may not be cancelled

passaro:
The change in #1334 introduced an issue where a GetObject request would still run to completion if the future returned by the `get_object` method in `S3CrtClient` was dropped before being ready.

In Mountpoint, this would affect random read workloads where dropped prefetcher requests would not always be cancelled, resulting in reduced throughput and increased memory usage.

No.

Bug fix entry.

---

### DIFF
--- a/mountpoint-s3-client/CHANGELOG.md
+++ b/mountpoint-s3-client/CHANGELOG.md
@@ -7,6 +7,9 @@
 
 ### Other changes
 
+* Fix an issue where GetObject requests may not be cancelled.
+  ([#1355](https://github.com/awslabs/mountpoint-s3/pull/1355))
+
 ## v0.13.2 (April 1, 2025)
 
 * Fix race condition in GetObject that could result in empty responses.
--- a/mountpoint-s3-client/src/s3_crt_client.rs
+++ b/mountpoint-s3-client/src/s3_crt_client.rs
@@ -2,8 +2,8 @@ use std::cmp;
 use std::ffi::{OsStr, OsString};
 use std::future::Future;
 use std::num::NonZeroUsize;
-use std::ops::Deref;
 use std::ops::Range;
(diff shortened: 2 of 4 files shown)


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -1039,19 +1039,46 @@ impl<T: Send, E: Send> Future for S3MetaRequest<T, E> {
...
-    }
-}
-
 impl<T: Send, E: Send> FusedFuture for S3MetaRequest<T, E> {
     fn is_terminated(&self) -> bool {
         self.receiver.is_terminated()
     }
 }
 
+/// Wrapper for a [MetaRequest] that cancels it on drop.
```
unexge, line 1048: Are there any reason why we don't want this behavior to bake into `MetaRequest` itself?

passaro: `MetaRequest` is essentially a smart pointer for an `aws_s3_meta_request`. It implements `Clone` (`aws_s3_meta_request_acquire`) and `Drop` (`aws_s3_meta_request_release`). There are cases where we internally use multiple clones and don't want them to cancel the request on drop.

unexge: It may be worth documenting that `MetaRequest` won't be cancelled on drop and the callers must ensure it's dropped.

mountpoint-s3-client/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 
 ### Other changes
 
+* Fix an issue where GetObject requests may not be cancelled.
```
unexge, line 10: Should we also have a changelog entry in `mountpoint-s3` crate?

unexge (approved):

PR #1362, opened 2025-04-10
Add `fstab` CLI parser

muddyfish:
Introduce support for invoking Mountpoint with fstab style arguments:
```
./mount-s3 example-bucket /mnt/mountpoint -o rw,auto-unmount,allow-root
```

No breaking changes

Yes, but not yet. This is still WIP

---


mountpoint-s3/src/cli.rs
```diff
@@ -0,0 +1,250 @@
...
+    version = build_info::FULL_VERSION,
+)]
+pub struct AppCliArgs {
+    #[clap(flatten)]
+    pub cli_args: CliArgs,
+}
+
+#[derive(Parser, Debug)]
+struct FsTabCliArgs {
+    #[clap(help = "Name of bucket to mount")]
```
passaro, line 24: I don't see where this `help`  is used.

mountpoint-s3/src/cli.rs
```diff
@@ -0,0 +1,250 @@
...
+            .filter(|option| !Self::filter_option(option))
+            .map(|option| {
+                if option.starts_with("-") {
+                    option
+                } else {
+                    format!("--{}", option)
+                }
+            });
+        let cli_arg_list = [
+            env::args_os().nth(0).unwrap_or("mount-s3".into()),
```
passaro, line 55: Could we avoid the binary name? See https://docs.rs/clap/latest/clap/struct.Command.html#method.no_binary_name

mountpoint-s3/src/cli.rs
```diff
@@ -0,0 +1,250 @@
...
+    pub options: String,
+}
+
+impl FsTabCliArgs {
+    pub fn try_to_cli_args(self) -> anyhow::Result<CliArgs> {
+        let cli_arg_list = self.build_cli_arg_list()?;
+        Self::build_cli_args(cli_arg_list)
+    }
+
+    fn build_cli_arg_list(&self) -> anyhow::Result<impl Iterator<Item = OsString>> {
```
passaro, line 43: Why not `Item = String`? Or alternative keep everything `OsString`?

mountpoint-s3/src/cli.rs
```diff
@@ -0,0 +1,250 @@
...
+        help_heading = MOUNT_OPTIONS_HEADER,
+        conflicts_with="read_only",
+    )]
+    pub read_write: bool,
+    #[clap(long = "", default_value = "true", hide = true)]
+    pub foreground: bool,
+}
+
+pub fn get_cli_args() -> anyhow::Result<CliArgs> {
+    let is_fstab = env::args_os().len() == 5 && env::args_os().nth(3) == Some("-o".parse().unwrap());
```
passaro, line 145: Not too sure about this. How do we report errors in each case? Do we want to show the help?

passaro: Also what's `"-o".parse().unwrap()`?

passaro (commented):
A few initial thoughts.

mountpoint-s3-fs/src/cli.rs
```diff
@@ -41,27 +41,43 @@ use crate::prefix::Prefix;
...
+pub const CACHING_OPTIONS_HEADER: &str = "Caching options";
+pub const ADVANCED_OPTIONS_HEADER: &str = "Advanced options";
 
 #[derive(Parser, Debug)]
 pub struct ContextParams {
     pub full_version: String,
 }
 
+#[derive(Parser, Debug)]
+pub struct CliArgs {
```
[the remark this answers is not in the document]
unexge: +1, I feel like it would be cleaner to make `CliArgs` and `FstabArgs` orthogonal, and just parse `FstabArgs` and convert them to `CliArgs`.

mountpoint-s3/src/cli.rs
```diff
@@ -0,0 +1,250 @@
...
+        let fstab_cli_args = FsTabCliArgs::try_parse_from(args).unwrap();
+        let cli_args = fstab_cli_args.try_to_cli_args().unwrap();
+
+        assert!(cli_args.foreground);
+        assert_eq!(cli_args.base.mount_point.to_str(), Some("/mnt/test"));
+        assert_eq!(cli_args.base.bucket_name, "demo_s3_bucket");
+        assert_eq!(cli_args.base.prefix.unwrap().as_str(), "foo/bar,baz/");
+        assert_eq!(cli_args.base.uid, Some(2));
+        assert_eq!(cli_args.base.gid, None);
+        assert!(cli_args.read_only);
```
dannycjones, line 210: What happens if I do `ro=recursive` or `ro=vfs`?

I think on this point we want to document the intent here: we take options, parse everything we recognise and support(!) and then reject anything we can't support. I'd even add a test case to cover what happens if a customer provides `ro=vfs`. What do negative cases look like for customers trying to configure things they shouldn't?

mountpoint-s3-fs/src/cli.rs
```diff
@@ -554,23 +560,26 @@ impl CliArgs {
...
         }
 
         #[cfg(target_os = "linux")]
         if matches!(mount_point, MountPoint::FileDescriptor(_)) {
-            let passed_mount_options = &[(self.read_only, "--read-only"), (self.auto_unmount, "--auto-unmount")]
-                .iter()
-                .filter(|o| o.0)
-                .map(|o| o.1)
-                .collect::<Vec<_>>();
+            let passed_mount_options = &[
```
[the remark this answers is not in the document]
muddyfish: I don't think so - right now at least, if we're using `fstab` mode or not isn't forwarded to `CliArgs` after conversion.

mountpoint-s3/src/cli.rs
```diff
@@ -0,0 +1,250 @@
...
+            return Err(anyhow!("Unexpected end of string after '\\'"));
+        }
+        fragments.push(&string[prev_idx..]);
+        unescaped.push(fragments.join(""));
+        Ok(unescaped)
+    }
+}
+
+#[derive(Parser, Debug)]
+#[clap(disable_help_flag = true)]
```
dannycjones, line 122: Why do we disable help. Would it make sense to include it anyway?

mountpoint-s3/src/cli.rs
```diff
@@ -0,0 +1,250 @@
+use crate::build_info;
+
+use anyhow::anyhow;
+use clap::{CommandFactory, FromArgMatches, Parser};
+use mountpoint_s3_fs::cli::{CliArgs, CliArgsBase, MOUNT_OPTIONS_HEADER};
+use std::env;
+use std::ffi::OsString;
+
+// TODO: Extract `mountpoint_s3_fs::cli::CliArgs` to this crate.
```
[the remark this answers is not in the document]
dannycjones: What's the remaining work here for the `TODO`, and does anyone own it? Would it be reasonable to make the move now?

dannycjones (commented):
Some initial feedback, I need to go back and review the design doc also.

mountpoint-s3/src/cli.rs
```diff
@@ -0,0 +1,250 @@
...
+        CliArgs::from_arg_matches(&arg_matches).map_err(|e| anyhow!(e))
+    }
+
+    fn filter_option(option: &str) -> bool {
+        // Same as from https://github.com/s3fs-fuse/s3fs-fuse/blob/22ca6ba6ee9ffb27602046e6f4334ec4e9b0b7d7/src/s3fs.cpp#L5340-L5345
+        // and https://github.com/libfuse/sshfs/blob/ed0825440c48895b7e20cc1440bbafd8d9c88eb8/sshfs.c#L533-L538
+        ["auto", "noauto", "user", "nouser", "users", "_netdev"].contains(&option) || option.starts_with("x-")
+    }
+
+    fn split_commas(string: &str) -> anyhow::Result<Vec<String>> {
```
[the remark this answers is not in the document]
muddyfish: Added some comments

mountpoint-s3/src/cli.rs
```diff
@@ -0,0 +1,250 @@
...
+#[derive(Parser, Debug)]
+struct FsTabCliArgs {
+    #[clap(help = "Name of bucket to mount")]
+    pub bucket_name: String,
+
+    #[clap(
+        help = "Directory or FUSE file descriptor to mount the bucket at",
+        value_name = "DIRECTORY"
+    )]
+    pub mount_point: String,
```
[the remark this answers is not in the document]
muddyfish: The current approach uses the presence of the `-o` argument to swap over to the alternative fstab CLI parser. 

In theory there's nothing stopping the usage of a FUSE file descriptor here, but there's no realistic reason for this - if someone's using this for fstab there's no way to get a file descriptor in here. However it's plausible that someone may want to write a script which would open the FD and pass on all the arguments to MP as-is.

mountpoint-s3-fs/src/cli.rs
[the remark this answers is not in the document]
muddyfish: I'm going to remove this and do preprocessing and post-processing the arguments to simplify the code here. Duplicating the `CliArgs` structure may still be a good idea here, but given we still need to migrate it out of the `mountpoint-s3-fs` crate, I'd like to wait until then to make the choice

mountpoint-s3/src/cli.rs
```diff
@@ -0,0 +1,250 @@
...
+        // `read_write` is only used to be exclusive with read-only. We don't actually use it for anything else as it's the default.
+        let _ = arg_matches.try_remove_one::<bool>("read_write");
+        CliArgs::from_arg_matches(&arg_matches).map_err(|e| anyhow!(e))
+    }
+
+    fn filter_option(option: &str) -> bool {
+        // Same as from https://github.com/s3fs-fuse/s3fs-fuse/blob/22ca6ba6ee9ffb27602046e6f4334ec4e9b0b7d7/src/s3fs.cpp#L5340-L5345
+        // and https://github.com/libfuse/sshfs/blob/ed0825440c48895b7e20cc1440bbafd8d9c88eb8/sshfs.c#L533-L538
+        ["auto", "noauto", "user", "nouser", "users", "_netdev"].contains(&option) || option.starts_with("x-")
+    }
```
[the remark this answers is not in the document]
dannycjones: Does the check for `x-` prefix need to be case insensitive?

mountpoint-s3/src/cli.rs
```diff
@@ -0,0 +1,32 @@
...
+struct AppCliArgs {
+    #[clap(flatten)]
+    pub cli_args: CliArgs,
+}
+
+pub fn get_cli_args() -> anyhow::Result<CliArgs> {
+    let is_fstab = env::args_os().len() == 5 && env::args_os().nth(3) == Some(OsString::from("-o"));
+
+    if is_fstab {
+        let args = FsTabCliArgs::parse();
```
unexge, line 26: It's probably very unlikely but – should we print something like "fstab mode detected" in this block to ensure people running into this block accidentally can quickly understand what's going on?

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,227 @@
...
+pub struct FsTabCliArgs {
+    pub bucket_name: String,
+    #[clap(value_name = "DIRECTORY")]
+    pub mount_point: String,
+    #[clap(short = 'o')]
+    pub options: String,
+}
+
+impl FsTabCliArgs {
+    pub fn try_to_cli_args(self) -> anyhow::Result<CliArgs> {
```
unexge, line 17: In Rust, if the method consumes `self`, it usually called as `into`, so maybe: `try_into_cli_args`?

mountpoint-s3-fs/src/cli.rs
```diff
@@ -59,8 +59,18 @@ pub struct ContextParams {
     group(
         ArgGroup::new("cache_group")
             .multiple(true),
-    ),
+    )
 )]
+#[command(after_help = concat!(
```
unexge, line 64: Wonder if there is a way to generate this from `FsTabCliArgs`?

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,227 @@
...
+use std::env;
+use std::ffi::OsString;
+
+#[derive(Parser, Debug)]
+pub struct FsTabCliArgs {
+    pub bucket_name: String,
+    #[clap(value_name = "DIRECTORY")]
+    pub mount_point: String,
+    #[clap(short = 'o')]
+    pub options: String,
```
unexge, line 13: Should we make this a `FsTabOptions` and use a [value parser](https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_parser)?

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,227 @@
...
+            (&self.mount_point).into(),
+        ]
+        .into_iter()
+        .chain(options.into_iter().map(|s| s.into()));
+
+        Ok(cli_arg_list)
+    }
+
+    fn build_cli_args(cli_arg_list: impl Iterator<Item = OsString>) -> anyhow::Result<CliArgs> {
+        let mut cli_args = CliArgs::try_parse_from(cli_arg_list).map_err(|e| anyhow!(e))?;
```
unexge, line 53: I feel like the parsing logic should be: parse options into `FsTabCliArgs` and convert `FsTabCliArgs` into `CliArgs` rather than converting `FsTabCliArgs` back to a string and then trying to parse `CliArgs` from there.

mountpoint-s3/src/fstab.rs
muddyfish: In the most recent commit, I removed `FsTabOptions` and opted to go with a simpler approach of pre&post processing of the CLI arguments prior to parsing them as `CliArgs` .

muddyfish: I could still feasibly use a value parser to split by commas and do all the pre-processing however if we think that's a good idae

mountpoint-s3/src/fstab.rs
muddyfish: I'd not noticed it consumed `self` and I don't think it needs to. I'll change it to taking an `&self`

passaro (commented):
A few more comments (I have yet to look at `split_commas`):

mountpoint-s3-fs/src/cli.rs
```diff
@@ -59,8 +59,18 @@ pub struct ContextParams {
...
+    "\nAlternative fstab structure:\n",
+    "  mount-s3 <BUCKET> <DIRECTORY> -o <OPTIONS>\n\n",
+    "Arguments:\n",
+    "  <BUCKET_NAME>\n          Name of bucket to mount, with s3:// URIs supported\n",
+    "  <DIRECTORY>\n          Location to mount bucket at\n",
+    "  <OPTIONS>\n",
+    "          fstab style options. Comma separated list of CLI options, with backslash escapes for commas, backslashes, and double quotes.\n",
+    "          No need to use `--` to prefix arguments."
```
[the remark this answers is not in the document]
muddyfish: No need to use - currently adding `--` is supported if wanted, but is not required

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,230 @@
...
+impl FsTabCliArgs {
+    /// Parse the args we've been given into an iterator of options to pass to the main CliArgs parser
+    /// Filters out arguments that aren't meant for Mountpoint, and add `--` to any argument that's not prefixed with `-`.
+    fn into_cli_arg_list(self) -> anyhow::Result<impl Iterator<Item = OsString>> {
+        let mut options: Vec<String> = self
+            .options
+            .into_iter()
+            .filter(|option| !Self::filter_option(option))
+            .map(|option| {
+                if option.starts_with("-") {
```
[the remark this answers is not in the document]
muddyfish: Yes

mountpoint-s3/tests/cli.rs
```diff
@@ -301,3 +301,42 @@ fn verify_new_part_size_config_conflict_with_old_one(
 
     Ok(())
 }
+
+#[cfg(feature = "fstab")]
```
unexge, line 305: Should we add some end-to-end tests that integrates with real fstab? Maybe we can spawn a container that configures Mountpoint using fstab or systemd.

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,230 @@
...
+use mountpoint_s3_fs::cli::CliArgs;
+use std::env;
+use std::ffi::OsString;
+
+#[derive(Parser, Debug)]
+pub struct FsTabCliArgs {
+    pub bucket_name: String,
+    #[clap(value_name = "DIRECTORY")]
+    pub mount_point: String,
+    // Needs to be explicit full name because of https://github.com/clap-rs/clap/issues/4808
```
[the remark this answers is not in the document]
muddyfish: I used the value parser as it was suggested in this comment: https://github.com/awslabs/mountpoint-s3/pull/1362#discussion_r2039929334

I feel it does look nicer this way, but I'm happy either way

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// Options prefixed with `x-` are 'comments' in fstab, and can be ignored by us.
+    fn filter_option(option: &str) -> bool {
+        // Same as from https://github.com/libfuse/sshfs/blob/ed0825440c48895b7e20cc1440bbafd8d9c88eb8/sshfs.c#L533-L538
+        ["auto", "noauto", "user", "nouser", "users", "_netdev"].contains(&option) || option.starts_with("x-")
+    }
+
+    fn rename_option(option: String) -> String {
+        match option.as_str() {
+            "--ro" => "--read-only".to_string(),
+            "--rw" => "--read-write".to_string(),
```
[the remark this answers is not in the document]
muddyfish: I was renaming so we could support both `rw`, `--rw`, `read-write`, and `--read-write`. If we're happy just supporting `rw`, then we can remove this line

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,230 @@
...
+                if option.starts_with("-") {
+                    option
+                } else {
+                    format!("--{}", option)
+                }
+            })
+            .map(Self::rename_option)
+            .collect();
+
+        Self::validate_options(&options)?;
```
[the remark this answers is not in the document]
muddyfish: Yes, but what if someone passes `read-only`, and `rw` gets added without the customer knowing?

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,230 @@
...
+        let mut cli_args = CliArgs::try_parse_from(cli_arg_list).map_err(|e| anyhow!(e))?;
+        cli_args.foreground = true;
+        Ok(cli_args)
+    }
+}
+
+impl FsTabCliArgs {
+    /// Parse the args we've been given into an iterator of options to pass to the main CliArgs parser
+    /// Filters out arguments that aren't meant for Mountpoint, and add `--` to any argument that's not prefixed with `-`.
+    fn into_cli_arg_list(self) -> anyhow::Result<impl Iterator<Item = OsString>> {
```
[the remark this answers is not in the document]
muddyfish: I was originally using `OsString` as I was doing fewer transformations over the original data, but now I feel that just using `String` is better. I'll swap it over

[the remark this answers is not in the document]
muddyfish: I did consider inlining to `try_from`, and thought that it would be better as two separate functions. Happy to change that if you think it's better to inline

mountpoint-s3/tests/cli.rs
muddyfish: I'd want to do that in a separate PR as that would be quite a lot of additional effort on top of what's going on here

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,228 @@
...
+        assert_eq!(cli_args.bucket_name, "demo_s3_bucket");
+        assert_eq!(cli_args.prefix.unwrap().as_str(), "foo/bar,baz/");
+        assert_eq!(cli_args.uid, Some(2));
+        assert_eq!(cli_args.gid, None);
+        assert!(cli_args.read_only);
+    }
+
+    proptest! {
+        #[test]
+        fn proptest_split_commas_errors_correctly_and_reconstructs(string: String) {
```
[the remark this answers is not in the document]
muddyfish: Lets do this in another PR

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,252 @@
...
+        let cli_arg_list = [
+            env::args().nth(0).unwrap_or("mount-s3".to_string()),
+            self.bucket_name,
+            self.mount_point,
+        ]
+        .into_iter()
+        .chain(
+            self.options
+                .into_iter()
+                .filter(|option| !Self::filter_option(option))
```
passaro, line 45: nit: if you invert the meaning of `filter_option` to be in line with the `filter` predicate, you should be able to do:
```suggestion
                .filter(Self::filter_option)
```

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,252 @@
...
+                .map(Self::rename_option),
+        )
+        .collect();
+        Ok(cli_arg_list)
+    }
+
+    /// Remove options that are for systemd/the fstab parser
```
passaro, line 52: See previous comment and the docs for `filter`: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter

mountpoint-s3/src/lib.rs
```diff
@@ -1,16 +1,43 @@
...
+use std::ffi::OsString;
 
 // TODO: Extract `mountpoint_s3_fs::cli::CliArgs` to this crate.
-// TODO: Stabilise the library API
 #[derive(Parser, Debug)]
 #[clap(
     name = "mount-s3",
     about = "Mountpoint for Amazon S3",
     version = build_info::FULL_VERSION,
+    after_help = if cfg!(feature = "fstab") {concat!(
```
passaro, line 16: Could you extract it into a constant (and a multiline string maybe)?

mountpoint-s3/src/main.rs
```diff
@@ -1,11 +1,15 @@
-pub mod build_info;
-
-use clap::Parser;
+use mountpoint_s3::build_info;
+use mountpoint_s3::try_parse_cli_args;
 
 fn main() -> anyhow::Result<()> {
-    let args = mountpoint_s3::AppCliArgs::parse();
+    let cli_args = try_parse_cli_args()
```
passaro, line 5: `match` may be clearer here

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+        if let Some(disallowed_option) = options
+            .iter()
+            .find(|x| ["foreground", "read-only"].contains(&x.as_str()))
+        {
+            return Err(Command::error(
+                &mut Self::command(),
+                ErrorKind::InvalidValue,
+                format!(
+                    "Cannot use '{}' flag when using fstab style arguments.",
```
unexge, line 90: So we only support `ro` for `--read-only` and disallow directly using `ready-only` right? Would be useful to note that in the error message

unexge (approved):

mountpoint-s3/src/fstab.rs
```diff
@@ -0,0 +1,252 @@
...
+        let cli_arg_list = [
+            env::args().nth(0).unwrap_or("mount-s3".to_string()),
+            self.bucket_name,
+            self.mount_point,
+        ]
+        .into_iter()
+        .chain(
+            self.options
+                .into_iter()
+                .filter(|option| Self::option_allowed(option))
```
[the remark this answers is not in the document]
muddyfish: I tried that, and it doesn't work because `option` is an `&String`, and `Self::option_allowed` takes an `&str`. For some reason the closure can convert it implicitly, but without it it doesn't. I can't change the function signature to take `&String` because that upsets clippy, and Alex and I decided this was the simplest solution

mountpoint-s3/src/fstab.rs
muddyfish: Will edit in the next PR

PR #1347, opened 2025-04-03
Allow changing log level dynamically with `USR2` signal

unexge:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

PR #1364, opened 2025-04-10
Apply `EnvFilter` as a global filter rather than a per-layer filter

unexge:
Currently, the `EnvFilter` – the filter that decides which spans and events should be processed or ignored – applied to each layer (i.e., file layer, syslog layer, and console layer). This would allow us to decide what spans and events should be processed at layer-level, but currently we use the same filter at each layer. 

This PR updates logging configuration to apply the filter at global-level rather than at layer-level. The main motivation is adding things like https://github.com/awslabs/mountpoint-s3/pull/1347/commits/e821cf6fbcd24f473ddfb8619ff1a3bd849b6fd3, where we could have only one global filter, and we could tweak that filter in some cases. Doing this with per-layer filter would also be okay, but it would require a bit more effort to handle/decide. 

No

No

---

### DIFF
--- a/mountpoint-s3-fs/src/logging.rs
+++ b/mountpoint-s3-fs/src/logging.rs
@@ -14,10 +14,9 @@ use time::format_description::FormatItem;
 use time::macros;
 use time::OffsetDateTime;
 use tracing::Span;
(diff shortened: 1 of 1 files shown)


dannycjones (approved):

passaro (approved):
LGTM!

PR #1366, opened 2025-04-11
Inline ObjectInfo to ReaddirEntry

vladem:
The type `object_client::ObjectInfo` is `#[non_exhaustive]` so it cannot be constructed outside of the defining crate. In future we may want to construct `ReaddirEntry`-s without using the client, but from data stored on disk. As a workaround we inline ObjectInfo to ReaddirEntry.

No. This change is an implementation detail of the `readdir` module.

No.

---

### DIFF
--- a/mountpoint-s3-fs/src/superblock/readdir.rs
+++ b/mountpoint-s3-fs/src/superblock/readdir.rs
@@ -44,8 +44,9 @@
 use std::cmp::Ordering;
 use std::collections::VecDeque;
 
-use mountpoint_s3_client::types::ObjectInfo;
+use mountpoint_s3_client::types::{ObjectInfo, RestoreStatus};
 use mountpoint_s3_client::ObjectClient;
+use time::OffsetDateTime;
 use tracing::{error, trace, warn};
 
 use crate::sync::{Arc, AsyncMutex, Mutex};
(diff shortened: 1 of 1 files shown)


mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -200,13 +201,63 @@ impl ReaddirHandle {
...
+    /// Objects in flexible retrieval storage classes (such as GLACIER and DEEP_ARCHIVE) are only
+    /// accessible after restoration.
+    pub restore_status: Option<RestoreStatus>,
+
+    /// Entity tag of this object.
+    pub etag: String,
+}
+
+impl ReaddirObjectInfo {
+    fn from(client_object_info: ObjectInfo) -> Self {
```
passaro, line 235: This looks like a `From` implementation

mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -200,13 +201,63 @@ impl ReaddirHandle {
...
 enum ReaddirEntry {
-    RemotePrefix { name: String },
-    RemoteObject { name: String, object_info: ObjectInfo },
-    LocalInode { lookup: LookedUp },
+    RemotePrefix {
+        name: String,
+    },
+    RemoteObject {
+        name: String,
+        object_info: ReaddirObjectInfo,
```
passaro, line 256: If `ReaddirObjectInfo` is single-use, I'd rather embed the relevant fields here.

passaro (commented):
Is it worth adding a new type, rather than just copy the required fields?

mountpoint-s3-fs/src/superblock/readdir.rs
vladem: good point, it also doesn't have to be `pub` since will likely be constructed inside this module

mountpoint-s3-fs/src/superblock/readdir.rs
vladem: constructed in one place, so decided it doesn't help to have a separate function at all

mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -204,9 +212,30 @@ impl ReaddirHandle {
...
-    RemoteObject { name: String, object_info: ObjectInfo },
-    LocalInode { lookup: LookedUp },
+    RemotePrefix {
+        name: String,
+    },
+    RemoteObject {
+        /// Last component of the S3 Key.
+        name: String,
+        /// S3 Key for this object.
+        full_key: String,
```
passaro, line 222: nit: is the key this needed?

passaro (approved):
Minor nit (non blocking)

passaro (changes requested):
Title and description need updating

mountpoint-s3-fs/src/superblock/readdir.rs
vladem: it's used for a couple of warnings:
```
warn!("{} has an invalid name and will be unavailable", next.description());
warn!(
    "{} is omitted because another {} exist with the same name",
    entry.description(),
    last_entry.description(),
);
```

we don't log the `full_key` for `RemotePrefix` (only `name`) so may be we can ditch it:
```
Self::RemotePrefix { name } => {
    format!("directory '{name}'")
}
Self::RemoteObject { name, full_key, .. } => {
    format!("file '{}' (full key {:?})", name, full_key)
}
```

but I think parent inode is not available at the point of logging, so there's no simple way to restore `full_key` from `name`

passaro: No worries, I had missed that.

passaro (approved):

PR #1367, opened 2025-04-11
Allow changing log level dynamically with `USR2` signal 

unexge:
This PR makes Mountpoint capable of changing log verbosity dynamically with `USR2` Unix signal. The users can send a `USR2` signal to Mountpoint process, e.g., `kill -USR2 <mount-s3-pid>`, to toggle between the following log verbosity levels:
  1. Default logging verbosity
  2. Debug logging for all except CRT (i.e., `debug,awscrt=off`)
  3. Debug logging for all (i.e., `debug,awscrt=debug`)
  4. Trace logging for all except CRT (i.e., `trace,awscrt=off`)
  5. Trace logging for all (i.e., `trace,awscrt=trace`)

No breaking change, a new runtime behavior with `USR2` Unix signal.

Yes, will update.

---

### DIFF
--- a/doc/LOGGING.md
+++ b/doc/LOGGING.md
@@ -42,6 +42,17 @@ To enable more verbose logging for the AWS Common Runtime that Mountpoint uses t
 
 For finer-grained control over log verbosity, Mountpoint uses the `MOUNTPOINT_LOG` environment variable, which overrides the verbosity options above. The `MOUNTPOINT_LOG` environment variable uses the [`tracing-subscriber` directive syntax](https://docs.rs/tracing-subscriber/0.3.17/tracing_subscriber/filter/struct.EnvFilter.html), and can be used to control log verbosity on a per-subject basis. For example, setting `MOUNTPOINT_LOG` to `trace` enables all trace-level logs, while `trace,awscrt=warn` enables trace-level logs for all log subjects except `awscrt`, which has only warning-level logging enabled.
 
+#### Changing logging verbosity on runtime
+
+Mountpoint allows changing logging verbosity dynamically on runtime using `USR2` Unix signal, e.g., `kill -USR2 <mount-s3-pid>`.
+
+Mountpoint toggles between the following verbosity levels each time it receives a `USR2` signal:
+  1. Default logging verbosity
+  2. Debug logging for all except CRT (i.e., `debug,awscrt=off`)
+  3. Debug logging for all (i.e., `debug,awscrt=debug`)
+  4. Trace logging for all except CRT (i.e., `trace,awscrt=off`)
+  5. Trace logging for all (i.e., `trace,awscrt=trace`)
+
 ## Metrics
 
 Mountpoint optionally collects metrics measuring various values across different components.
--- a/doc/LOGGING.md
+++ b/doc/LOGGING.md
@@ -44,10 +44,10 @@ For finer-grained control over log verbosity, Mountpoint uses the `MOUNTPOINT_LO
 
 #### Changing logging verbosity on runtime
 
-_This is an unstable interface and might be subject to change in the future._
-
 > [!WARNING]
-> Default action of `SIGUSR2` POSIX signal is to terminate the process.
+> This is an unstable interface and might be subject to change in the future.
+>
+> The default action of `SIGUSR2` POSIX signal is to terminate the process.
 > Ensure Mountpoint version supports `SIGUSR2` signal before sending it, as otherwise it might terminate the process. Mountpoint v1.17.0 onward supports `SIGUSR2` signal as long as `--no-log` options is not passed.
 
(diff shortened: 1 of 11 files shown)


doc/LOGGING.md
```diff
@@ -42,6 +42,17 @@ To enable more verbose logging for the AWS Common Runtime that Mountpoint uses t
 
 For finer-grained control over log verbosity, Mountpoint uses the `MOUNTPOINT_LOG` environment variable, which overrides the verbosity options above. The `MOUNTPOINT_LOG` environment variable uses the [`tracing-subscriber` directive syntax](https://docs.rs/tracing-subscriber/0.3.17/tracing_subscriber/filter/struct.EnvFilter.html), and can be used to control log verbosity on a per-subject basis. For example, setting `MOUNTPOINT_LOG` to `trace` enables all trace-level logs, while `trace,awscrt=warn` enables trace-level logs for all log subjects except `awscrt`, which has only warning-level logging enabled.
 
+#### Changing logging verbosity on runtime
```
muddyfish, line 45: I think it would be worth adding a disclaimer here that this interface isn't stable and intended only for humans to use, so we can change it in the future

doc/LOGGING.md
```diff
@@ -42,6 +42,17 @@ To enable more verbose logging for the AWS Common Runtime that Mountpoint uses t
...
+
+Mountpoint allows changing logging verbosity dynamically on runtime using `USR2` Unix signal, e.g., `kill -USR2 <mount-s3-pid>`.
+
+Mountpoint toggles between the following verbosity levels each time it receives a `USR2` signal:
+  1. Default logging verbosity
```
muddyfish, line 50: Should we add a 'whatever logging level we were initially started with'? I'm not entirely sure if that's what's already implemented, but if so this needs re-wording

mountpoint-s3-fs/src/logging.rs
```diff
@@ -100,14 +122,11 @@ fn install_panic_hook() {
...
-    // Create the logging config from the MOUNTPOINT_LOG environment variable or the default config
-    // if that variable is unset.
-    let env_filter =
-        EnvFilter::try_from_env("MOUNTPOINT_LOG").unwrap_or_else(|_| EnvFilter::new(&config.default_filter));
+fn init_tracing_subscriber(config: LoggingConfig) -> anyhow::Result<LoggingHandle> {
     // Don't create the files or subscribers if we'll never emit any logs
-    if env_filter.max_level_hint() == Some(LevelFilter::OFF) {
-        return Ok(());
+    let default_filter = make_default_filter(config.default_filter.clone())();
+    if default_filter.max_level_hint() == Some(LevelFilter::OFF) {
```
unexge, line 128: We can disable this optimization and set up the listeners all the time maybe to support going from no log to some logging?

mountpoint-s3-fs/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+* Allow changing log level dynamically with `USR2` signal. ([#1367](https://github.com/awslabs/mountpoint-s3/pull/1367))
```
muddyfish, line 3: Should include link to docs here

doc/LOGGING.md
```diff
@@ -42,6 +42,25 @@ To enable more verbose logging for the AWS Common Runtime that Mountpoint uses t
 
 For finer-grained control over log verbosity, Mountpoint uses the `MOUNTPOINT_LOG` environment variable, which overrides the verbosity options above. The `MOUNTPOINT_LOG` environment variable uses the [`tracing-subscriber` directive syntax](https://docs.rs/tracing-subscriber/0.3.17/tracing_subscriber/filter/struct.EnvFilter.html), and can be used to control log verbosity on a per-subject basis. For example, setting `MOUNTPOINT_LOG` to `trace` enables all trace-level logs, while `trace,awscrt=warn` enables trace-level logs for all log subjects except `awscrt`, which has only warning-level logging enabled.
 
+#### Changing logging verbosity on runtime
```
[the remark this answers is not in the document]
passaro: Also update the changelogs

mountpoint-s3-fs/src/logging.rs
passaro: Is there a performance impact?

passaro (commented):
Change looks good. I think it only needs a new version number.

mountpoint-s3-fs/src/logging.rs
unexge: I expect it to be minimal if there is even. I can try doing simple benchmarks to see if it increases memory usage or decreases throughput. The upside would be going from no log to some level of logging in case something goes wrong with a particular instance, which I feel like it could be useful.

muddyfish:
LGTM - will let Alex approve

PR #1374, opened 2025-04-17
Isolate metadata cache configuration in CliArgs

passaro:
Minor change to take the metadata cache configuration out of the `mount` function.

No changes.

No.

---


mountpoint-s3-fs/src/cli.rs
```diff
@@ -465,6 +465,37 @@ impl CliArgs {
...
+                TimeToLive::Duration(Duration::from_secs(60))
+            } else {
+                TimeToLive::Minimal
+            }
+        });
+        if matches!(metadata_cache_ttl, TimeToLive::Duration(Duration::ZERO)) {
+            const ZERO_TTL_WARNING: &str = "The '--metadata-ttl 0' setting is no longer supported, is now interpreted as 'minimal', and will be removed in a future release. Use '--metadata-ttl minimal' instead";
+            tracing::warn!("{}", ZERO_TTL_WARNING);
+            if !self.foreground {
+                // Ensure warning is visible even when not redirecting logs to stdout.
```
unexge, line 481: I know this logic didn't change with this PR, but I'm a bit confused about this block. If the application is not foreground (i.e., deamonized) then it won't have its stdout/stderr channels right? What's the point of having `eprintln!` in that case? I feel like this if should be reverted

passaro: In foreground mode, we direct the logs to stdout, so the `tracing::warn` above will be visible. In "background"/"deamonized" mode, we need to explicitly write it to stderr. Note that in the latter mode, the std channels are only closed after the mount succeeds, see line 698.

unexge (dismissed):

passaro:
Rebase to solve merge confict.

PR #1375, opened 2025-04-17
Move FuseSessionConfig out of cli

passaro:
Refactor `FuseSessionConfig` out of the `cli` module.

No changes.

No.

---


mountpoint-s3-fs/src/fuse/config.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+impl FuseSessionConfig {
+    pub fn new(
+        mount_point: MountPoint,
+        read_only: bool,
+        auto_unmount: bool,
+        allow_root: bool,
+        allow_other: bool,
+        max_threads: usize,
```
unexge, line 25: Should we make this a struct? Given that we have four boolean parameters next to each other, it would be easy to mix the ordering and configure it incorrectly.

mountpoint-s3-fs/src/fuse/config.rs
```diff
@@ -0,0 +1,223 @@
...
+    ) -> anyhow::Result<Self> {
+        let fs_name = String::from("mountpoint-s3");
```
unexge, line 27: Maybe we should move fs_name to a constant? I assume we might be using it other parts of the codebase as well, but we probably repeat it in multiple places

unexge (approved):
Left some comments but LGTM. Do we need a CHANGELOG entry?

unexge (approved):

PR #1377, opened 2025-04-21
Implement creation of the internal manifest

vladem:
Implement creation of an SQLite database from an iterator of manifest entries (in future reading from a file; now from RAM in tests). For more context see [branch](https://github.com/vladem/mountpoint-s3/pull/7/files) where manifest is used to load metadata of the objects.   

No, only used in tests.

No, only used in tests.

---


mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -0,0 +1,242 @@
...
+        )?;
+
+        Ok(())
+    }
+
+    pub(super) fn create_index(&self) -> Result<()> {
+        let conn = self.conn.lock().expect("lock must succeed");
+        conn.execute("CREATE UNIQUE INDEX idx_key ON s3_objects (key)", ())?;
+
+        conn.execute("CREATE INDEX idx_parent_key ON s3_objects (parent_key, key)", ())?;
```
muddyfish, line 70: Why? I don't think we ever filter using `parent_key` at all

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -0,0 +1,242 @@
...
+            insert(self, &conn, &mut insert_buffer, &mut shadowed)?;
+        }
+
+        Ok(shadowed)
+    }
+
+    fn insert_batch_locked(&self, conn: &MutexGuard<'_, Connection>, entries: &[DbEntry]) -> Result<()> {
+        conn.execute_batch("BEGIN TRANSACTION;")?;
+        let mut stmt =
+            conn.prepare("INSERT OR REPLACE INTO s3_objects (key, parent_key, etag, size) VALUES (?1, ?2, ?3, ?4)")?;
```
muddyfish, line 133: Should we consider using batch insertions? https://stackoverflow.com/questions/1609637/how-to-insert-multiple-rows-in-sqlite/1609688#1609688

mountpoint-s3-fs/src/manifest/db.rs
vladem: this PR just creates the DB, so all of the actual useful queries are not here, but here is how we [will](https://github.com/vladem/mountpoint-s3/pull/7/files#diff-c4e56cb844e7f3cd61b6fbcafc093275be050e5217317ebdfe15575d2f6f3bf5R73) do a `readdir` in future:
```
SELECT key, etag, size FROM s3_objects WHERE parent_key = ?1 ORDER BY key LIMIT ?2, ?3
```

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -0,0 +1,265 @@
...
+
+        Ok(())
+    }
+
+    pub(super) fn insert_batch(&self, entries: &[DbEntry]) -> Result<()> {
+        let conn = self.conn.lock().expect("lock must succeed");
+        self.insert_batch_locked(&conn, entries)
+    }
+
+    /// Iterates over a sorted list of S3 object keys, infers and inserts the directories,
```
passaro, line 80: It's not clear to me what is happening here. Do we need to report shadowing in this detail? Let's consider other approaches.

passaro (commented):
I've left a few comments, but let's first discuss in person whether we need to report warnings and how to handle directories.

mountpoint-s3-fs/src/manifest/db.rs
vladem: I believe insertion is already happening in batches since we execute all of the statements inside one transaction, but I've tried inserting multiple rows in one statement (haven't noticed significant improvement so decided to leave as-is for simplicity):
```rust
    fn insert_batch_locked(&self, conn: &MutexGuard<'_, Connection>, entries: &[DbEntry]) -> Result<()> {
        let prepare_stmt = |entries_num| {
            let placeholders = std::iter::repeat("(?, ?, ?, ?)")
                .take(entries_num)
                .collect::<Vec<_>>()
                .join(", ");
            let query = format!(
                "INSERT OR REPLACE INTO s3_objects (key, parent_key, etag, size) VALUES {}",
                placeholders
            );
            conn.prepare(&query)
        };
        let parent_keys: Vec<_> = entries.iter().map(|entry| entry.parent_key().to_string()).collect();
        let params: Vec<_> = entries.iter().enumerate().map(|(idx, entry)| vec![
            &entry.full_key as &dyn rusqlite::ToSql,
            &parent_keys[idx] as &dyn rusqlite::ToSql,
            &entry.etag as &dyn rusqlite::ToSql,
            &entry.size as &dyn rusqlite::ToSql,
        ] ).flatten().collect();

        let mut stmt = prepare_stmt(entries.len())?;
        stmt.execute(&*params)?;
        Ok(())
    }
```

results (1M keys and 8000 batch size):
- multiple rows per statement: 10.05s
- single row per statement all in transaction: 9.06s

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -0,0 +1,265 @@
...
+#[derive(Debug, Clone)]
+pub(super) struct Db {
+    conn: Arc<Mutex<Connection>>,
+}
+
+impl Db {
+    pub(super) fn new(manifest_db_path: &Path) -> Result<Self> {
+        let conn = Connection::open(manifest_db_path)?;
+        let mode: String = conn.query_row("PRAGMA journal_mode=off", [], |row| row.get(0))?;
+        assert_eq!(&mode, "off");
```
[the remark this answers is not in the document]
vladem: this is where configuration is done, this pragma changes the `journal_mode` and expects a certain value after it (see [journal_mode doc](https://www.sqlite.org/pragma.html#pragma_journal_mode)), the assertion will fail if in future releases of SQLite this pragma does not take effect (see [pragma doc](https://www.sqlite.org/pragma.html))

passaro (approved):

PR #1379, opened 2025-04-24
Add CRT memory limit config to prefetcher and uploader benchmarks

dannycjones:
In some benchmarking, we want to experiment with adjusting the CRT's memory limiter to observe the change in throughput performance.

This change introduces CLI flags to the benchmark scripts (examples) that allows us to directly configure the CRT memory limiter.

No.

No, impacts benchmarking scripts only.

---

### DIFF
--- a/mountpoint-s3-fs/examples/prefetch_benchmark.rs
+++ b/mountpoint-s3-fs/examples/prefetch_benchmark.rs
@@ -57,9 +57,12 @@ pub struct CliArgs {
     )]
     pub maximum_throughput_gbps: Option<u64>,
 
+    #[arg(long, help = "Override value for CRT memory limit in gibibytes", value_name = "GiB")]
+    pub crt_memory_limit_gib: Option<u64>,
+
     #[clap(
         long,
-        help = "Maximum memory usage target in MiB",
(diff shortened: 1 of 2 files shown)


mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -57,9 +57,12 @@ pub struct CliArgs {
     )]
     pub maximum_throughput_gbps: Option<u64>,
 
+    #[arg(long, help = "Override value for CRT memory limit in gibibytes", value_name = "GiB")]
+    pub crt_memory_limit_gib: Option<u64>,
+
     #[clap(
         long,
-        help = "Maximum memory usage target in MiB",
+        help = "Maximum memory usage target for Mountpoint's memory limiter [default: 95% of total system memory]",
```
sahityadg, line 65: Shouldn't we use CRT's defaults here? Why 95%?

dannycjones: This line is updating the help for Mountpoint's memory limiter (independent of CRT). I'm not sure 95% is a good idea, but I'm not looking to adjust it in this PR.

sahityadg (approved):

PR #1383, opened 2025-04-25
Use manifest in readdir and lookup operations

vladem:
Use metadata stored in an sqlite database instead of s3, when performing lookup and readdir.

No, only used in tests.

No, only used in tests.

---

### DIFF
--- a/mountpoint-s3-fs/src/manifest.rs
+++ b/mountpoint-s3-fs/src/manifest.rs
@@ -1,4 +1,8 @@
+use crate::superblock::InodeError;
+use db::Db;
+use std::{collections::VecDeque, path::Path};
 use thiserror::Error;
+use tracing::{error, trace};
 
 mod builder;
 mod db;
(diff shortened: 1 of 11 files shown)


mountpoint-s3-fs/tests/fuse_tests/manifest_test.rs
```diff
@@ -0,0 +1,283 @@
...
+    sdk_client
+        .put_object()
+        .bucket(bucket)
+        .key(&full_key)
+        .body(ByteStream::from(content))
+        .send()
+        .await
+        .expect("put object must succeed");
+
+    let head_resp = sdk_client
```
passaro, line 273: Is head_object required? put_object already returns the Etag

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -98,6 +102,11 @@ impl Superblock {
             config.cache_config.negative_cache_ttl,
         );
 
+        let manifest = config
+            .manifest_db_path
+            .as_ref()
+            .map(|manifest_db_path| Manifest::new(manifest_db_path).expect("manifest must be created"));
```
passaro, line 108: We should not panic here. Have you considered adding the pre-built `Manifest` to `config`, instead of the path? That would shift the burden of initializing it to the caller (only when they use the manifest).

The alternative is to return an error here, but it would be cumbersome to change the signature only for this edge case.

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -40,6 +49,38 @@ impl Db {
         })
     }
 
+    /// Queries rows from the DB representing either the file or a directory (typically a single row returned)
```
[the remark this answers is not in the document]
vladem: yeah, I think you're right, we should assume that a single entry is always returned here (previously the query searched for a file or directory, but now they both don't have `/` in the end)

mountpoint-s3-fs/src/manifest.rs
```diff
@@ -14,4 +18,141 @@ pub enum ManifestError {
...
+        let mut full_path = parent_full_path;
+        full_path.push_str(name);
+
+        // search for an entry
+        let mut db_entries = self.db.select_entries(&full_path)?;
+        let Some(db_entry) = db_entries.pop() else {
+            return Ok(None);
+        };
+        if !db_entries.is_empty() {
+            Err(ManifestError::TooManyRows)
```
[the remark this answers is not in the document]
vladem: responded below, fixed

mountpoint-s3-fs/src/manifest.rs
```diff
@@ -14,4 +20,138 @@ pub enum ManifestError {
...
+    },
+}
+
+impl TryFrom<DbEntry> for ManifestEntry {
+    type Error = ManifestError;
+
+    fn try_from(db_entry: DbEntry) -> Result<Self, Self::Error> {
+        if db_entry.full_key.ends_with('/') {
+            Err(ManifestError::InvalidRow)
+        } else if db_entry.etag.is_none() && db_entry.size.is_none() {
```
muddyfish, line 46: Match statement can remove some unwraps and make this clearer

mountpoint-s3-fs/src/manifest.rs
```diff
@@ -14,4 +20,138 @@ pub enum ManifestError {
...
+        parent_full_path: String,
+        name: &str,
+    ) -> Result<Option<ManifestEntry>, ManifestError> {
+        trace!("using manifest to lookup {} in {}", name, parent_full_path);
+        let mut full_path = parent_full_path;
+        full_path.push_str(name);
+
+        // search for an entry and validate it
+        let db_entry = self.db.select_entries(&full_path)?;
+        let Some(db_entry) = db_entry else {
```
muddyfish, line 86: Could this be refactored into a `match` statement? This is rather hard to parse right now

mountpoint-s3-fs/src/manifest.rs
```diff
@@ -14,4 +20,138 @@ pub enum ManifestError {
...
+    /// Offset of the next child to search for in the database
+    next_offset: usize,
+    /// Max amount of entries to read from the database at once
+    batch_size: usize,
+    /// Database has no more entries
+    finished: bool,
+}
+
+impl ManifestIter {
+    fn new(db: Db, _bucket: &str, parent_key: &str) -> Result<Self, InodeError> {
```
muddyfish, line 115: Why does this return a result if we can never return an error here?

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -650,6 +656,46 @@ impl SuperblockInner {
...
+        parent_ino: InodeNo,
+        name: &str,
+    ) -> Result<Option<RemoteLookup>, InodeError> {
+        let parent = self.get(parent_ino)?;
+        if parent.kind() != InodeKind::Directory {
+            return Err(InodeError::NotADirectory(parent.err()));
+        }
+
+        let parent_full_path = self.full_key_for_inode(&parent);
+        let mount_time = OffsetDateTime::now_utc(); // todo: mount time
```
muddyfish, line 672: Missing todo

mountpoint-s3-fs/src/manifest.rs
vladem: good point, fixed (previously it made a query to db here)

passaro (approved):

PR #1385, opened 2025-04-25
Disable ANSI colors where not supported for fs and client examples

dannycjones:
Simple change - currently, redirecting the logs to a file will keep ANSI colors. With this change, the scripts will automatically turn off ANSI colors when the standard error output is redirected.

This change is not urgent as users can turn off ANSI colors using `NO_COLOR=1`.

For relevant examples/benchmarks only, ANSI color will be disabled when not supported (i.e. not console output).

No, benchmarking/example change only.

---

### DIFF
--- a/mountpoint-s3-client/examples/client_benchmark.rs
+++ b/mountpoint-s3-client/examples/client_benchmark.rs
@@ -22,8 +22,8 @@ fn init_tracing_subscriber() {
 
     let subscriber = Subscriber::builder()
         .with_env_filter(EnvFilter::from_default_env())
+        .with_ansi(supports_color::on(supports_color::Stream::Stderr).is_some())
         .with_writer(std::io::stderr)
-        .with_ansi(false)
         .finish();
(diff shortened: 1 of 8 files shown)


mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -22,8 +22,8 @@ fn init_tracing_subscriber() {
 
     let subscriber = Subscriber::builder()
         .with_env_filter(EnvFilter::from_default_env())
+        .with_ansi(supports_color::on(supports_color::Stream::Stderr).is_some())
```
dannycjones, line 25: This is lifted from how we configure it for Mountpoint proper: https://github.com/awslabs/mountpoint-s3/blob/c336f951a4934724f2975f76df24e22c0d299afc/mountpoint-s3-fs/src/logging.rs#L161

unexge (approved):

PR #1386, opened 2025-04-25
Parse manifest from csv

vladem:
Add an iterator parsing a CSV file and some tests for it. 

No, only used in tests.

No, only used in tests.

---

### DIFF
--- a/mountpoint-s3-fs/Cargo.toml
+++ b/mountpoint-s3-fs/Cargo.toml
@@ -91,3 +91,7 @@ s3_tests = []
 s3express_tests = []
 shuttle = []
 second_account_tests = []
+
+[[example]]
+name = "manifest_to_sqlite"
+required-features = ["manifest"]
--- a/mountpoint-s3-fs/examples/manifest_to_sqlite.rs
+++ b/mountpoint-s3-fs/examples/manifest_to_sqlite.rs
@@ -1,5 +1,6 @@
 use clap::Parser;
-use std::path::PathBuf;
+use mountpoint_s3_fs::manifest::{create_db, CsvReader};
+use std::{fs::File, io::BufReader, path::PathBuf, time::Instant};
 
 #[derive(Parser, Debug)]
 #[clap(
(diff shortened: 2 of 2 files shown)


mountpoint-s3-fs/src/manifest/csv_reader.rs
```diff
@@ -0,0 +1,79 @@
...
+        self.reader.next().map(|item| item.map_err(ManifestError::from))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use test_case::test_case;
+
+    #[test_case("\"key1\",\"\"\"etag1\"\"\",1024\n\"key2\",\"\"\"etag2\"\"\",2048\n", &[
```
muddyfish, line 32: Could we use raw string literals here? This is very hard to read

mountpoint-s3-fs/Cargo.toml
```diff
@@ -48,6 +48,7 @@ tracing = { version = "0.1.41", features = ["log"] }
 tracing-log = "0.2.0"
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
 signal-hook = "0.3.17"
+csv = "1.3.1"
```
passaro, line 51: We should make the dependency optional and only enable it with the "manifest" feature. Same with `rusqlite`.

mountpoint-s3-fs/src/manifest/csv_reader.rs
passaro: Agreed. Maybe even constants with raw multiline literals.

passaro (approved):
The change looks good to me.

About the feature flag: shall we put the whole `manifest` module under the feature flag? It would allow to make the rusqlite dependency optional as well. Happy to leave it for a separate PR, though.

mountpoint-s3-fs/examples/manifest_to_sqlite.rs
passaro, line 1: We should add the example to `Cargo.toml` and set `required-features = ["manifest"]`

passaro (changes requested):

PR #1388, opened 2025-04-25
Remove Prefetch trait

passaro:
Simplify the type signature of `S3Filesystem` and related types by removing the `Prefetch` trait and replacing it with a single `Prefetcher` implementation, which has an `ObjectClient` generic parameter. 

No.

No.

---

### DIFF
--- a/mountpoint-s3-fs/src/prefetch/part_stream.rs
+++ b/mountpoint-s3-fs/src/prefetch/part_stream.rs
@@ -12,27 +12,22 @@ use crate::async_util::Runtime;
 use crate::checksums::ChecksummedBytes;
 use crate::mem_limiter::MemoryLimiter;
 use crate::object::ObjectId;
-use crate::prefetch::backpressure_controller::{new_backpressure_controller, BackpressureConfig};
-use crate::prefetch::part::Part;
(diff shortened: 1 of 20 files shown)


mountpoint-s3-fs/src/prefetch/part_stream.rs
```diff
@@ -161,30 +156,62 @@ impl Debug for RequestRange {
...
+    pub fn client(&self) -> &Client {
+        self.inner.client()
+    }
+}
+
+impl<Client> Debug for PartStream<Client> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("PartStream").finish()
+    }
+}
```
dannycjones, line 191: nit: prefer using [`finish_non_exhaustive`](https://doc.rust-lang.org/std/fmt/struct.DebugStruct.html#method.finish_non_exhaustive) when we skip some fields?

```suggestion
impl<Client> Debug for PartStream<Client> {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct("PartStream").finish_non_exhaustive()
    }
}
```

dannycjones (approved):
Looks good to me, thanks for focusing on simplifying this

PR #1400, opened 2025-05-02
Introduce MountpointConfig

passaro:
Extract the configuration and the logic to create a new Mountpoint FUSE session out of the `cli` module.

No.

No.

---

### DIFF
--- a/mountpoint-s3-fs/src/config.rs
+++ b/mountpoint-s3-fs/src/config.rs
@@ -1,0 +1,97 @@
+use anyhow::Context as _;
+use futures::executor::block_on;
+use mountpoint_s3_client::ObjectClient;
+
+use crate::data_cache::{DataCacheConfig, DiskDataCache, ExpressDataCache, MultilevelDataCache};
+use crate::fuse::config::FuseSessionConfig;
+use crate::fuse::session::FuseSession;
+use crate::fuse::S3FuseFilesystem;
+use crate::prefetch::{Prefetcher, PrefetcherBuilder};
+use crate::s3::config::S3Path;
+use crate::sync::Arc;
+use crate::{Runtime, S3Filesystem, S3FilesystemConfig};
+
+/// Configuration for a Mountpoint session
+#[derive(Debug)]
+pub struct MountpointConfig {
+    fuse_session_config: FuseSessionConfig,
+    data_cache_config: DataCacheConfig,
+    filesystem_config: S3FilesystemConfig,
+}
+
+impl MountpointConfig {
+    pub fn new(fuse_session_config: FuseSessionConfig) -> Self {
+        Self {
+            fuse_session_config,
+            data_cache_config: Default::default(),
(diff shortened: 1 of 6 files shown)


mountpoint-s3-fs/src/config.rs
```diff
@@ -0,0 +1,97 @@
...
+    data_cache_config: DataCacheConfig,
+    filesystem_config: S3FilesystemConfig,
+}
+
+impl MountpointConfig {
+    pub fn new(fuse_session_config: FuseSessionConfig) -> Self {
+        Self {
+            fuse_session_config,
+            data_cache_config: Default::default(),
+            filesystem_config: Default::default(),
```
unexge, line 27: Do we need to initialize them to default values? Seems like we always provide all configurations currently, wouldn't it be cleaner to accept all in `new` method here? My concern is that it might be easy to miss calling configuration methods and leaving them to default values

unexge (commented):
Does this change require a changelog entry?

mountpoint-s3-fs/src/config.rs
```diff
@@ -0,0 +1,97 @@
...
+            &s3_path.prefix,
+            self.filesystem_config,
+        );
+
+        let fuse_fs = S3FuseFilesystem::new(fs);
+        FuseSession::new(fuse_fs, self.fuse_session_config)
+    }
+}
+
+fn create_prefetcher_builder<Client>(
```
muddyfish, line 67: Should this be a method of `create_prefetcher_builder`?

mountpoint-s3-fs/src/config.rs
```diff
@@ -0,0 +1,97 @@
...
+/// Configuration for a Mountpoint session
+#[derive(Debug)]
+pub struct MountpointConfig {
+    fuse_session_config: FuseSessionConfig,
+    data_cache_config: DataCacheConfig,
+    filesystem_config: S3FilesystemConfig,
+}
+
+impl MountpointConfig {
+    pub fn new(fuse_session_config: FuseSessionConfig) -> Self {
```
muddyfish, line 23: Why is `fuse_session_config` the only argument that's required here?

mountpoint-s3-fs/src/config.rs
passaro: Updated

mountpoint-s3-fs/src/config.rs
passaro: Updated

mountpoint-s3-fs/src/config.rs
passaro: Not sure what you mean. It's a private helper function for now. We may consider exposing it directly at some point and review where it should live then.

passaro:
@unexge, about a changelog entry: I plan to add one for the next PR, which will (finally) move `CliArgs` out of the crate.

muddyfish (approved):

PR #1402, opened 2025-05-06
Remove manifest from the released executable

vladem:
Remove the code using `rusqlite` from the released executable. Implementation of the manifest using this crate becomes gated behind the `manifest` feature flag.

No, only used in tests.

No, only used in tests.

---

### DIFF
--- a/mountpoint-s3-fs/src/superblock.rs
+++ b/mountpoint-s3-fs/src/superblock.rs
@@ -39,6 +39,7 @@ use tracing::{debug, error, trace, warn};
 use crate::fs::error_metadata::{ErrorMetadata, MOUNTPOINT_ERROR_CLIENT};
 use crate::fs::CacheConfig;
 use crate::logging;
+#[cfg(feature = "manifest")]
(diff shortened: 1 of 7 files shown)


mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -98,6 +99,7 @@ impl ReaddirHandle {
             }
         };
 
+        #[cfg(feature = "manifest")]
```
passaro, line 102: The code duplication here is not ideal. We should consider refactoring to avoid it. E.g. by extracting `iter` creation into its own function.

Could be a follow up, though.

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -596,11 +598,14 @@ impl SuperblockInner {
         let lookup = match lookup {
             Some(lookup) => lookup?,
             None => {
+                #[cfg(feature = "manifest")]
```
passaro, line 601: Same comment as for the duplication in `readdir`.

passaro (approved):

PR #1414, opened 2025-05-13
Fstab cliargs roundtrip tests

tadiwa-aizen:
### Fstab cliargs roundtrip tests

This PR adds a property-based test that ensures roundtrip conversion between CliArgs and FsTabCliArgs behaves as expected. Specifically, we:

Implemented a custom FstabCompatibleCliArgs strategy for generating valid CliArgs inputs.

Serialise these into fstab-style CLI arguments.

Parse them back into CliArgs through the FsTabCliArgs path.

Assert equality with the original input.

This responds to a prior review comment requesting a test for round-trip parsing of CLI arguments.

No, this change does not impact runtime behavior. It only adds non-breaking test code under #[cfg(test)].

No changelog entry is required. This is an internal test-only enhancement and does not affect functionality or the public interface.

---

### DIFF
--- a/mountpoint-s3/src/fstab.rs
+++ b/mountpoint-s3/src/fstab.rs
@@ -150,6 +150,7 @@ mod tests {
     use super::*;
     use proptest::prelude::*;
     use proptest::{prop_assert_eq, proptest};
+    use proptest_derive::Arbitrary;
     use test_case::test_case;
 
     #[test_case("no commas", Some(["no commas"].to_vec()))]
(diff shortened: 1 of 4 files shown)


mountpoint-s3/src/fstab.rs
```diff
@@ -251,4 +253,68 @@ mod tests {
...
+
+    // This addresses the PR suggestion to test roundtrip from CliArgs -> FsTabCliArgs -> CliArgs
+    proptest! {
+        #[test]
+        fn cli_args_roundtrip_from_fstab_args(original in any::<FstabCompatibleCliArgs>()) {
+            let args = serialize_to_fstab_args(&original);
+            let fstab = FsTabCliArgs::try_parse_from(&args).unwrap();
+            let roundtripped: CliArgs = fstab.try_into().unwrap();
+
+            prop_assert_eq!(FstabCompatibleCliArgs::from(roundtripped), original);
```
muddyfish, line 317: Nit: does `roundtripped.into()` work here?

mountpoint-s3/src/fstab.rs
```diff
@@ -251,4 +253,68 @@ mod tests {
...
+    struct FstabCompatibleCliArgs {
+        #[proptest(regex = "bucket-[a-z]{3,10}")]
+        bucket_name: String,
+        #[proptest(regex = "/mnt/test-[a-z]{1,5}")]
+        mount_point: String,
+        uid: u32,
+        allow_delete: bool,
+        allow_other: bool,
+        debug: bool,
+        #[proptest(value = "false")]
```
muddyfish, line 268: What's the point of this if it's always false?

mountpoint-s3/src/fstab.rs
```diff
@@ -251,4 +253,68 @@ mod tests {
...
+
+        args.push("-o".to_string());
+        args.push(options.join(","));
+        args
+    }
+
+    // This addresses the PR suggestion to test roundtrip from CliArgs -> FsTabCliArgs -> CliArgs
+    proptest! {
+        #[test]
+        fn cli_args_roundtrip_from_fstab_args(original in any::<FstabCompatibleCliArgs>()) {
```
muddyfish, line 312: Nit: does this work? `original: FstabCompatibleCliArgs`

mountpoint-s3/src/fstab.rs
```diff
@@ -251,4 +253,68 @@ mod tests {
...
+        if cli_args.debug {
+            options.push("debug".to_string());
+        }
+
+        args.push("-o".to_string());
+        args.push(options.join(","));
+        args
+    }
+
+    // This addresses the PR suggestion to test roundtrip from CliArgs -> FsTabCliArgs -> CliArgs
```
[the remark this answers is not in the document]
tadiwa-aizen: I'll have it as 
`    // Test roundtrip conversion between CLI argument formats`

mountpoint-s3/src/fstab.rs
tadiwa-aizen: Okay, ive updated it so that read_only is now exercised in prop tests. I also updated `serialize_to_fstab_args` to include "`ro`" when read_only is true.

mountpoint-s3/src/fstab.rs
tadiwa-aizen: So ,Currently we have:

`impl From<CliArgs> for FstabCompatibleCliArgs `

(which allows FstabCompatibleCliArgs::from(cli_args))

But we don't have the reverse implementation. From what i gather, your comment is suggesting we could potentially use .into() which would require implementing:

`   impl From<FstabCompatibleCliArgs> for CliArgs` or  `  impl Into<FstabCompatibleCliArgs> for CliArgs`

I think the  current approach might be better becasue

- CliArgs is a more complex type with more fields so converting from `FstabCompatibleCliArgs` to `CliArgs` would require setting default values for many fields
- The current `From<CliArgs> `for `FstabCompatibleCliArgs` implementation is simpler because it's extracting a subset of fieldsn and not the whole thing

muddyfish: We tested this offline and found `.into` isn't appropriate here

muddyfish (approved):

PR #1415, opened 2025-05-13
Add test demonstrating MP behavior with ABAC IAM policies

dannycjones:
This adds new tests to demonstrate/document the behavior of Mountpoint when trying to implement attribute-based access control (ABAC). The purpose here is to simply demonstrate the behavior, so that we can understand current state/options.

No, new test only.

No, new test only.

---


muddyfish (approved):

mountpoint-s3-fs/tests/fuse_tests/iam_perm_test.rs
```diff
@@ -0,0 +1,189 @@
...
+#[test_case(false; "without metadata cache")]
+fn get_object_req_existing_tags(metadata_cache: bool) {
+    let bucket = get_test_bucket();
+
+    let cache_config = if metadata_cache {
+        let one_hour = Duration::from_secs(60 * 60);
+        CacheConfig::new(mountpoint_s3_fs::fs::TimeToLive::Duration(one_hour))
+    } else {
+        CacheConfig {
+            // Avoid any caching in MP, though this could hides weird inconsistencies!
```
vladem, line 38: this test shows that disabling caching results in more immediate failure: on open (lookup) instead of read; but what about concurrent `ListObjectsV2` or `HeadObject` requests? may directory lookup depend on the order in which those requests finished, yielding success or failure "randomly"?

dannycjones: If ListObjectsV2 doesn't return any entry that would shadow, then we'll wait for HeadObject - which in this case will eventually return the EIO. (I added a sleep of 5 seconds locally to verify what happens when listing returns first.)

I think ultimately with caching, if the open happens right after the lookup then sure there could be inconsistency.

Does that answer the concern? If not, can you share a more concrete example?

dannycjones: I think you might be asking about the `lookup` of `dir1`. This is OK because `HeadObject` will return 404 if the object doesn't exist, as the condition will match. I read the condition in the policy as "If the object exists, then it must have these tags. If the object does not exist, condition matches as there's no existing tags to compare."

vladem: I think shadowed (case 3.) is the corner case here. Would be great if we could confirm, to maybe document in future:

1. lookup("dir"), "dir/a.txt" exists, "dir" does not exist => HeadObject -> 404, ListObjectsV2 -> 200 => lookup ok
1. lookup("a.txt"), no "a.txt/*", "a.txt" exists =>  HeadObject -> 200/403, ListObjectsV2 -> 404 => lookup ok / EIO (deterministic, based on permissions)
1. lookup("shadowed"), "shadowed/a.txt" exists, "shadowed" exists => HeadObject -> 200/403, ListObjectsV2 -> 200 => lookup ok / EIO (not deterministic based on response order of HeadObject/ListObjectsV2?) 

But it's more of a documentation concern, so probably out of scope of this PR!

vladem (approved):

mountpoint-s3-fs/tests/fuse_tests/iam_perm_test.rs
dannycjones: In the interest of simplicity, I'll leave it as it is but yes that does seem like another sharp edge.

PR #1416, opened 2025-05-13
Logging errors in a structured format

vladem:
Implement an `FileErrorLogger` and use it to write errors occurring during supported fuse calls in `S3FuseFilesystem`. This logging is not enabled in the main executable, only in the `mount_from_config` example.

No, only used in the example.

No, only used in the example.

---

### DIFF
--- a/mountpoint-s3-fs/examples/mount_from_config.rs
+++ b/mountpoint-s3-fs/examples/mount_from_config.rs
@@ -17,11 +17,12 @@ use mountpoint_s3_fs::{
         session::FuseSession,
         ErrorCallback,
     },
-    logging::{event_log::EventLogger, init_logging, LoggingConfig, LoggingHandle},
+    logging::{event_log::LogErrorCallback, init_logging, LoggingConfig, LoggingHandle},
     manifest::{ingest_manifest, Manifest},
     metrics::{self, MetricsSinkHandle},
     prefix::Prefix,
     s3::config::{ClientConfig, PartConfig, Region, S3Path},
+    sync::Arc,
     MountpointConfig, Runtime, S3FilesystemConfig,
 };
 use serde::{Deserialize, Serialize};
@@ -210,13 +211,19 @@ fn setup_logging(config: &ConfigOptions) -> Result<(LoggingHandle, MetricsSinkHa
     Ok((logging, metrics))
 }
 
-fn mount_filesystem(config: &ConfigOptions, manifest: Manifest, error_callback: ErrorCallback) -> Result<FuseSession> {
+fn mount_filesystem(
+    config: &ConfigOptions,
+    manifest: Manifest,
+    error_callback: Arc<dyn ErrorCallback + Send + Sync>,
+) -> Result<FuseSession> {
     // Create the Mountpoint configuration
     let mp_config = MountpointConfig::new(
         config.build_fuse_session_config()?,
         config.build_filesystem_config(manifest)?,
         config.build_data_cache_config(),
-    );
+    )
+    .error_callback(Some(error_callback));
+
     // Get S3 Path
     let s3_path = config.build_s3_path()?;
 
@@ -229,7 +236,7 @@ fn mount_filesystem(config: &ConfigOptions, manifest: Manifest, error_callback: 
 
     // Create and run the FUSE session
     let fuse_session = mp_config
-        .create_fuse_session(s3_path, client, runtime, Some(error_callback))
+        .create_fuse_session(s3_path, client, runtime)
         .context("Failed to create FUSE session")?;
 
     Ok(fuse_session)
@@ -249,16 +256,16 @@ fn main() -> Result<()> {
     // Read the config
     let config = load_config(&args.config).context("Failed to load config")?;
     // Set up the event log
-    let event_logger = EventLogger::new(&config.event_log_dir).context("Failed to create an event log")?;
+    let error_callback =
+        Arc::new(LogErrorCallback::new(&config.event_log_dir).context("Failed to create an event log")?);
     // Set up logging
     let (_logging, _metrics) = setup_logging(&config).context("Failed to setup logging")?;
     // Process manifests if needed
-    let temporary_dir =
-        tempdir_in(&config.metadata_store_dir).map_err(|err| anyhow!(err).context("Failed to create manifest"))?;
+    let temporary_dir = tempdir_in(&config.metadata_store_dir).context("Failed to create manifest")?;
     let manifest = process_manifests(&config, temporary_dir.path()).context("Failed to create manifest")?;
     // Build all configurations
     let fuse_session =
-        mount_filesystem(&config, manifest, event_logger.log_error_callback()).context("Failed to mount filesystem")?;
+        mount_filesystem(&config, manifest, error_callback.clone()).context("Failed to mount filesystem")?;
     // Join the session and wait until it completes
     fuse_session.join().context("Failed to join session")?;
     Ok(())
--- a/mountpoint-s3-fs/examples/mount_from_config.rs
+++ b/mountpoint-s3-fs/examples/mount_from_config.rs
@@ -1,3 +1,5 @@
+use nix::sys::signal::{self, Signal};
+use nix::unistd::Pid;
 use std::{
     fs::File,
     io::BufReader,
@@ -15,14 +17,13 @@ use mountpoint_s3_fs::{
     fuse::{
         config::{FuseOptions, FuseSessionConfig, MountPoint},
         session::FuseSession,
-        ErrorCallback,
+        ErrorLogger,
     },
-    logging::{event_log::LogErrorCallback, init_logging, LoggingConfig, LoggingHandle},
+    logging::{error_logger::FileErrorLogger, init_logging, LoggingConfig, LoggingHandle},
     manifest::{ingest_manifest, Manifest},
     metrics::{self, MetricsSinkHandle},
     prefix::Prefix,
     s3::config::{ClientConfig, PartConfig, Region, S3Path},
-    sync::Arc,
     MountpointConfig, Runtime, S3FilesystemConfig,
 };
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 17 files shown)


mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -235,14 +247,18 @@ fn main() -> Result<()> {
...
+    // Set up the event log
+    let event_logger = EventLogger::new(&config.event_log_dir).context("Failed to create an event log")?;
     // Set up logging
-    let (_logging, _metrics) = setup_logging(&config)?;
+    let (_logging, _metrics) = setup_logging(&config).context("Failed to setup logging")?;
     // Process manifests if needed
-    let temporary_dir = tempdir()?;
-    let manifest = process_manifests(&config, temporary_dir.path())?;
+    let temporary_dir =
+        tempdir_in(&config.metadata_store_dir).map_err(|err| anyhow!(err).context("Failed to create manifest"))?;
```
passaro, line 257: `map_err` should not be required. You should be able to call `.context(...)` on any `Error`.

mountpoint-s3-fs/src/config.rs
```diff
@@ -38,6 +38,7 @@ impl MountpointConfig {
         s3_path: S3Path,
         client: Client,
         runtime: Runtime,
+        error_callback: Option<ErrorCallback>,
```
passaro, line 41: Couldn't this be a field in `MountpointConfig`?

mountpoint-s3-fs/src/fuse.rs
```diff
@@ -36,12 +38,19 @@ macro_rules! event {
 /// Handle an error in a FUSE handler. This logs the appropriate error message and then calls
 /// `reply` on the given replier with the error's corresponding errno.
 macro_rules! fuse_error {
-    ($name:literal, $reply:expr, $err:expr) => {{
+    ($name:literal, $reply:expr, $err:expr, $error_callback:expr, $request_id:expr) => {{
```
passaro, line 41: Consider taking (references to) `S3FuseFilesystem` and `Request` directly, if it simplifies the calling code.

mountpoint-s3-fs/src/fuse.rs
```diff
@@ -36,12 +38,19 @@ macro_rules! event {
...
         let err = $err;
         event!(err.level, "{} failed with errno {}: {:#}", $name, err.to_errno(), err);
         ::metrics::counter!("fuse.op_failures", "op" => $name).increment(1);
+        if let Some(error_callback) = $error_callback {
+            let error_callback: &ErrorCallback = error_callback;
+            error_callback(&err, $name, $request_id);
+        }
         $reply.error(err.to_errno());
     }};
+    ($name:literal, $reply:expr, $err:expr) => {
```
passaro, line 51: Shouldn't we use the new form everywhere?

mountpoint-s3-fs/src/logging/event_log.rs
```diff
@@ -0,0 +1,224 @@
...
+        let (tx, rx) = mpsc::sync_channel(max_inflight_events);
+        let file = File::create(log_directory.as_ref().join(EVENT_LOG_FILE_NAME))?;
+        let writer_thread = thread::spawn(|| Self::write_to_file(rx, file));
+        Ok(Self {
+            events_sender: Some(tx),
+            writer_thread: Some(writer_thread),
+        })
+    }
+
+    pub fn log_error_callback(&self) -> crate::fuse::ErrorCallback {
```
passaro, line 52: Why return a new callback? Shouldn't we just return the original on construction? We would not even need a multi-producer channel.

mountpoint-s3-fs/src/logging/event_log.rs
```diff
@@ -0,0 +1,224 @@
...
+        // todo: overhead of [SyncSender::clone]?
```
passaro, line 53: Obsolete comment?

mountpoint-s3-fs/src/logging/event_log.rs
```diff
@@ -0,0 +1,224 @@
...
+
+impl Drop for EventLogger {
+    /// [Self::writer_thread] terminates when there are no [SyncSender]-s left. We assume that no [ErrorCallback] will be alive
+    /// when [EventLogger] is dropped so, last [SyncSender] is dropped just before joining the thread.
+    fn drop(&mut self) {
+        self.events_sender.take();
+        self.writer_thread
+            .take()
+            .expect("writer must be set")
+            .join()
```
passaro, line 107: This will block forever if `EventLogger` is dropped at the "wrong" time.

mountpoint-s3-fs/src/config.rs
vladem: The field `S3Filesystem::config` is private, so either:

1) make the field public
2) add a getter to `S3Filesystem` for `config` field (or for `error_callback`)
3) store it in `S3FuseFilesystem` directly (as implemented now)

What do you think will be more appropriate?

mountpoint-s3-fs/src/logging/event_log.rs
```diff
@@ -0,0 +1,239 @@
...
+            .expect("must be able to send an event");
+    }
+
+    fn write_to_file(events_receiver: Receiver<Event>, file: File) {
+        let mut writer = BufWriter::new(file);
+        for event in events_receiver {
+            if let Err(err) = event.write(&mut writer) {
+                tracing::error!("failed to write to the event log: {}, event: {:?}", err, event);
+                // trigger graceful shutdown (with umount) by sending a signal to self
+                signal::kill(Pid::this(), Signal::SIGINT).expect("kill must succeed");
```
passaro, line 100: Consider adding a callback so we can move this logic to the caller.

mountpoint-s3-fs/src/logging/event_log.rs
```diff
@@ -0,0 +1,239 @@
+use nix::sys::signal::{self, Signal};
+use nix::unistd::Pid;
+use std::fs::File;
+use std::io::BufWriter;
+use std::path::Path;
+
+/// Provides callback ([LogErrorCallback]) for logging events in a structured format to a file.
```
passaro, line 7: This could be the rustdoc for `LogErrorCallback`.

passaro: Let's also review naming.

mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -198,13 +211,19 @@ fn setup_logging(config: &ConfigOptions) -> Result<(LoggingHandle, MetricsSinkHa
     Ok((logging, metrics))
 }
 
-fn mount_filesystem(config: &ConfigOptions, manifest_path: Manifest) -> Result<FuseSession> {
+fn mount_filesystem(
+    config: &ConfigOptions,
+    manifest: Manifest,
+    error_callback: Arc<dyn ErrorCallback + Send + Sync>,
```
passaro, line 217: Why `Arc` and not `Box`?

passaro (changes requested):

mountpoint-s3-fs/src/fuse.rs
```diff
@@ -19,6 +20,10 @@ use fuser::{
 pub mod config;
 pub mod session;
 
+pub trait ErrorCallback: std::fmt::Debug {
```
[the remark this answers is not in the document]
vladem: `ErrorLogger` as a trait name looks fine, then we can have `FileEventLogger` as a name for the struct implementing the trait

on a side note, `ErrorCallback` is a bit more generic for the trait, which would also make sense if errors were recorded in memory and served via a webserver

mountpoint-s3-fs/src/config.rs
```diff
@@ -30,13 +30,13 @@ impl MountpointConfig {
...
-            error_callback: None,
+            error_logger: None,
         }
     }
 
-    /// Set the [Self::error_callback] field
-    pub fn error_callback(mut self, error_callback: Option<Arc<dyn ErrorCallback + Send + Sync>>) -> Self {
-        self.error_callback = error_callback;
+    /// Set the [Self::error_logger] field
+    pub fn error_logger(mut self, error_logger: Option<Box<dyn ErrorLogger + Send + Sync>>) -> Self {
```
passaro, line 38: Do we need the `Option` here?

Would something like this work?
```suggestion
    pub fn error_logger(mut self, error_logger: impl ErrorLogger + Send + Sync) -> Self {
```

mountpoint-s3-fs/src/fuse.rs
```diff
@@ -20,7 +19,10 @@ use fuser::{
 pub mod config;
 pub mod session;
 
-pub trait ErrorCallback: std::fmt::Debug {
+/// The trait may be implemented to log failed fuse operations.
```
passaro, line 22:
```suggestion
/// A trait that can be implemented to log errors returned by fuse operations.
```

mountpoint-s3-fs/src/fuse.rs
```diff
@@ -20,7 +19,10 @@ use fuser::{
...
+pub trait ErrorLogger: std::fmt::Debug {
+    /// This method will be invoked for each failed fuse operation with an exception of those
+    /// operations which are not implemented.
```
passaro, line 25:
```suggestion
    /// Log an error returned by a fuse operation.
    ///
    /// This method will be invoked immediately before a fuse operation returns an error response, with the exception of those
    /// operations which are not implemented.
```

mountpoint-s3-fs/src/logging/error_logger.rs
```diff
@@ -1,13 +1,23 @@
...
 use std::io::BufWriter;
 use std::path::Path;
 
-/// Provides callback ([LogErrorCallback]) for logging events in a structured format to a file.
+use crate::fs::error_metadata::{MOUNTPOINT_ERROR_INTERNAL, MOUNTPOINT_ERROR_LOOKUP_NONEXISTENT};
+use crate::fs::Error;
+use crate::fuse::ErrorLogger;
+use crate::logging::log_file_name_time_suffix;
+use crate::sync::mpsc::{self, Receiver, SyncSender};
+use serde::{Deserialize, Serialize};
```
passaro, line 10: nit: follow the common layout, `std`, other, `crate`

mountpoint-s3-fs/src/logging.rs
```diff
@@ -29,6 +29,7 @@ mod testing;
 mod envfilter;
 use envfilter::{toggleable, ToggleableHandle};
 
+pub mod error_logger;
```
passaro, line 32: Should we put this under a new feature flag?

passaro (changes requested):

passaro (approved):
LGTM

PR #1419, opened 2025-05-14
Add fstab to GitHub CI

muddyfish:
Add integration tests to Github CI

Tests passed here: https://github.com/awslabs/mountpoint-s3/actions/runs/15135395318/job/42545885483

- fstab now forces runs in the background instead of the foreground
- When ran through fstab, Mountpoint now ignores the `nodev` and `nosuid` options, as we default to these capabilities.

No breaking changes, as fstab is not released.

No

---


mountpoint-s3/tests/fstab/common.sh
```diff
@@ -0,0 +1,53 @@
+function build_mountpoint() {
+  build_out=$(cargo build --bin mount-s3 --release --features=fstab --message-format=json-render-diagnostics)
+
+  MOUNTPOINT_ROOT=$(dirname "$(which "$0")")/../../..
```
dannycjones, line 4:
```suggestion
  PROJECT_ROOT=$(dirname "$(which "$0")")/../../..
```

dannycjones (commented):
Mainly just nitpick comments.

Why do we need to switch Mountpoint from foreground to background?

mountpoint-s3/src/fstab.rs
```diff
@@ -24,7 +24,7 @@ impl TryFrom<FsTabCliArgs> for CliArgs {
         let cli_arg_list = fstab_cli_args.into_cli_arg_list()?;
 
         let mut cli_args = CliArgs::try_parse_from(cli_arg_list)?;
-        cli_args.foreground = true;
+        cli_args.foreground = false;
```
unexge, line 27: Does fstab have a health check mechanism? Or is it more like fire and forget? This might cause problems if it tries to monitor spawned process as now it would deamonise itself and exit

muddyfish: If we have `foreground` true, it times out waiting for the mount to complete, even though Mountpoint itself is healthy. As far as I understand, it's basically fire and forget and there's no additional health check after we've forked.

mountpoint-s3/tests/fstab/common.sh
```diff
@@ -0,0 +1,53 @@
...
+  GENERATOR_BIN="/usr/lib/systemd/system-generators/systemd-fstab-generator"
+  SYSTEMD_MOUNT_DIR=/etc/systemd/system/
+  UNIT_SOURCE_DIR=$OUTPUT_DIR/normal
+
+  echo "$FSTAB_CONTENT" > "$SYSTEMD_FSTAB"
+
+  rm -r "$OUTPUT_DIR" || true
+  mkdir -p "$OUTPUT_DIR"/{normal,early,late}
+
+  $GENERATOR_BIN "$UNIT_SOURCE_DIR" "$OUTPUT_DIR/early" "$OUTPUT_DIR/late" || true
```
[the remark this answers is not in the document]
muddyfish: It fails to symlink some files after actually generating, which doesn't matter for our usecase, so I just added the `|| true` to make it work

mountpoint-s3/tests/fstab/basic_checks.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
```
unexge, line 1: Have we tried writing these tests with Rust? Not sure how hard would it be but at least we could re-use some existing utilities and run more exhaustive tests. `mountpoint-s3/tests/fork_test.rs` might be a good inspiration

mountpoint-s3/tests/fstab/common.sh
```diff
@@ -0,0 +1,53 @@
+function build_mountpoint() {
+  build_out=$(cargo build --bin mount-s3 --release --features=fstab --message-format=json-render-diagnostics)
+
+  MOUNTPOINT_ROOT=$(dirname "$(which "$0")")/../../..
+  MOUNTPOINT_PATH=$(printf "%s" "$build_out" | jq -js '[.[] | select(.reason == "compiler-artifact") | select(.executable != null)] | last | .executable')
+  echo "Mountpoint path: $MOUNTPOINT_PATH"
+}
+
+function spawn_mounts() {
```
unexge, line 9: So this means we're just starting Mountpoint on already running system, I also remember that fstab also allows you to mount filesystems during startup. Are there big differences between mounting at startup vs. mounting at runtime? Should we also have tests for startup as well?

unexge: Another thing I remember is systemd/fstab is not behaving well if filesystems starts to fail during startup, and I think there was an option to not fail mount to allow system to start, should we also add test-case for that?

muddyfish: Github's CI doesn't allow you to reboot the host during CI, so this is the best we can do without significant effort. From my testing on Fedora, you don't actually need `nofail` to make sure the system boots even with bad Mountpoint configuration, but I don't think it's worth tempting fate with the actual docs we're writing.

mountpoint-s3/tests/fstab/basic_checks.sh
muddyfish: I considered it, but the fact we need to invoke `systemd-fstab-generator` and interact with systemd a bunch made me decide against it.

mountpoint-s3/tests/fstab/common.sh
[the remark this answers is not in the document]
dannycjones: Do you have an example of the errors?

In any case, I guess the script will fail later when Mountpoint isn't mounted.

mountpoint-s3/src/fstab.rs
```diff
@@ -57,7 +57,11 @@ impl FsTabCliArgs {
     /// Options prefixed with `x-` are 'comments' in fstab, and can be ignored by us.
     fn option_allowed(option: &str) -> bool {
         // Same as from https://github.com/libfuse/sshfs/blob/ed0825440c48895b7e20cc1440bbafd8d9c88eb8/sshfs.c#L533-L538 with addition of `rw` argument
```
dannycjones, line 59: this comment is stale now :(

dannycjones (approved):
LGTM, one nit on the comment being outdated on the ignored args

dannycjones (approved):

mountpoint-s3/tests/fstab/common.sh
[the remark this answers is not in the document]
muddyfish: Renamed to `spawn_mounts.sh`, but the script does automatic cleanup at exit so running as a script won't work easily

passaro (approved):
LGTM

PR #1420, opened 2025-05-14
Adding fstab tag to user agent headers

tadiwa-aizen:
### What changed and why?

- This PR adds a new mp-fstab tag to the user agent header when Mountpoint is launched via an fstab entry.
- Introduces an is_fstab field to CliArgs, which is set to true when parsing arguments from fstab.
- This allows downstream consumers (like the product team) to detect and analyze fstab-based usage of Mountpoint for Amazon S3, supporting product analytics and future UX improvements.

Example Request Header
image

- No breaking changes.
- The only impact is the addition of the mp-fstab tag in the user agent header for fstab-based mounts.
- All other mounting methods and user agent construction remain unchanged.

No

---

### DIFF
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### New features
+
+* Add `mp-fstab` tag to the User-Agent header when Mountpoint is launched via an fstab entry. This enables detection and analysis of fstab-based usage in product analytics. ([#1420](https://github.com/awslabs/mountpoint-s3/pull/1420))
+
 ## v1.17.0 (May 12, 2025)
 
 ### New features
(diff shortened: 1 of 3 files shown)


mountpoint-s3/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### New features
+
+* Add `mp-fstab` tag to the User-Agent header when Mountpoint is launched via an fstab entry. This enables detection and analysis of fstab-based usage in product analytics. ([#1420](https://github.com/awslabs/mountpoint-s3/pull/1420))
```
dannycjones, line 5: This is a minor analytics change, and doesn't warrant a change log entry IMO. If it did, I'd recommend putting it under `### Other changes` anyway since this isn't a new feature that would encourage customers to upgrade.

dannycjones (commented):
Just a comment on change log entry

dannycjones (approved):

PR #1424, opened 2025-05-16
Add errno check to FS mock S3 tests

dannycjones:
Simple update to the test to check error number. We check this in other parts of the code, but this provides integration testing using the mock S3 HTTP server tests.

No, test change only.

No.

---


dannycjones:
Switched out the expected error code, this would be the output:

```
thread 'test_lookup_throttled_mock::head_object' panicked at mountpoint-s3-fs/tests/mock_s3_tests.rs:78:5:
assertion `left == right` failed: lookup failure due to HTTP 503 must return EPERM
  left: EIO
 right: EPERM
```

unexge (approved):

PR #1425, opened 2025-05-19
Update prefetcher wait_for_read_window_increment to drain queue

dannycjones:
When reviewing the prefetcher logic, the `wait_for_read_window_increment` method call in the S3 part stream was identified as a potential issue. The logic currently pulls only one increment from the queue of read window increments when large amounts of data are being fetched. Today, this is likely not to cause an issue as the read increments are much larger than the size of the parts emitted by `part_stream`. However, it would cause issues if there were changes here in future that resulted in increments smaller than those parts.

This change updates the method to drain all available increments and return the new value to the caller. This ensures that the backpressure mechanism doesn't wait for each part before processing only one window increment event. A new test is added to verify this behavior is guaranteed.

There should be no impact, other than fixing logic that currently is unlikely to introduce performance changes outside of very large parts sizes.

No, there is no known customer impact of the current issue.

---


mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -245,14 +245,17 @@ impl BackpressureLimiter {
     ) -> Result<Option<u64>, PrefetchReadError<E>> {
         // There is already enough read window so no need to block
         if self.read_window_end_offset > offset {
-            // Check the read window incrementing queue to see there is an early request to increase read window
-            let new_read_window_offset = if let Ok(len) = self.read_window_incrementing_queue.try_recv() {
-                self.read_window_end_offset += len as u64;
-                Some(self.read_window_end_offset)
+            // Drain increment queue
+            let old_offset = self.read_window_end_offset;
+            while let Ok(len) = self.read_window_incrementing_queue.try_recv() {
```
vladem, line 250: this looks correct with the assumption that `read_window_incrementing_queue` may contain multiple increments, but such situations looks erroneous to me

if `read_window_incrementing_queue` was instead used to receive 2 events: `DataRead(last_read_offset)`, `PartQueueStall` it would make more sense to expect multiple items in the channel here, but that may imply significant refactoring

vladem (approved):

PR #1428, opened 2025-05-20
 run-as-user ( uid switching implementation)

tadiwa-aizen:
### What changed and why?

This PR introduces support for user-mode mounting by adding a new CLI argument `--run-as-user.` This addresses the problem that, when invoked via /etc/fstab, mount-s3 runs as root on the host, which can lead to unnecessary privilege over provisioning.

The new logic does the following:

Accepts a `--run-as-user`=<username> argument. Converts the username to a UID and GID using a syscall.

Uses privilege dropping (via setuid and setgid) to run the mountpoint process as the specified user. This allows mount-s3 to run as a specific unprivileged user even when launched by root, such as via fstab or systemd.

**Most importantly addresses the issue with fstab where mountpoint would automatically launch from root or 'sudo'**

No breaking changes expected. This feature is opt-in only via the `--as-user flag`. Default behavior (root execution) remains unchanged when the flag is not used.

 Yes and Yes :)

###  Test Summary
Tested manually with the following sequence to confirm that:

- The mount is created by root but the process switches to testuser1.
- Only the specified user (or root) has access to the mount. 
- Unauthorized users are correctly denied access.

1. Mount as a user

`sudo ./target/release/mount-s3 --region eu-west-1 tadiwaom-mountpoint-test ~/mnt/mount-test --run-as-user=testuser1`

Result: Success. Bucket is mounted at `/home/tadiwaom/mnt/mount-test.
`
2. Verify mount
`mount | grep mount-test`
Result: Mount is visible with fuse and default permissions.

3. Access as current user
`ls ~/mnt/mount-test`
Result: Permission denied — as expected.

4. Access as testuser1
`sudo -u testuser1 ls ~/mnt/mount-test`
Result: Permission denied (as tadiwaom is not allowed to sudo into testuser1) — confirms access is restricted.

5. Access as root
`sudo ls ~/mnt/mount-test`
Result: Success — confirms root access works.

Process Verification
Checked the mountpoint process identity:
```
ps aux | grep mount-s3
ps -o user= -p <PID>
ps -f -p <PID>
```
Result:

Process is running with UID 43319 (testuser1)

Confirms successful user switch

---

### DIFF
--- a/mountpoint-s3/src/run.rs
+++ b/mountpoint-s3/src/run.rs
@@ -13,11 +13,15 @@ use mountpoint_s3_fs::logging::init_logging;
 use mountpoint_s3_fs::s3::S3Personality;
 use mountpoint_s3_fs::{metrics, MountpointConfig, Runtime};
 use nix::sys::signal::Signal;
-use nix::unistd::getgid;
-use nix::unistd::getuid;
-use nix::unistd::setresgid;
-use nix::unistd::setresuid;
-use nix::unistd::ForkResult;
+use nix::unistd::{getgid, getuid, ForkResult};
+
+// Linux/Android have setresgid/setresuid for more secure privilege dropping
+#[cfg(any(target_os = "linux", target_os = "android"))]
+use nix::unistd::{setresgid, setresuid};
+
+// macOS and other platforms use the simpler setgid/setuid
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
+use nix::unistd::{setgid, setuid};
 
 use crate::cli::CliArgs;
 use crate::{build_info, parse_cli_args};
@@ -438,15 +442,35 @@ fn drop_privileges_if_needed(args: &CliArgs) -> anyhow::Result<()> {
 
                 // Step 1: Change group first
                 tracing::info!("Changing GID from {} to {}", current_gid, target_gid);
-                if let Err(e) = setresgid(target_gid, target_gid, target_gid) {
-                    tracing::error!(error=?e, "Failed to set GID with setresgid");
-                    return Err(anyhow!("Failed to set GID with setresgid: {}", e));
+                #[cfg(any(target_os = "linux", target_os = "android"))]
+                {
+                    if let Err(e) = setresgid(target_gid, target_gid, target_gid) {
+                        tracing::error!(error=?e, "Failed to set GID with setresgid");
+                        return Err(anyhow!("Failed to set GID with setresgid: {}", e));
+                    }
+                }
+                #[cfg(not(any(target_os = "linux", target_os = "android")))]
+                {
+                    if let Err(e) = setgid(target_gid) {
+                        tracing::error!(error=?e, "Failed to set GID with setgid");
+                        return Err(anyhow!("Failed to set GID with setgid: {}", e));
+                    }
                 }
 
                 // Step 2: Change user
-                if let Err(e) = setresuid(target_uid, target_uid, target_uid) {
-                    tracing::error!(error=?e, "Failed to set UID with setresuid");
-                    return Err(anyhow!("Failed to set UID with setresuid: {}", e));
+                #[cfg(any(target_os = "linux", target_os = "android"))]
+                {
+                    if let Err(e) = setresuid(target_uid, target_uid, target_uid) {
+                        tracing::error!(error=?e, "Failed to set UID with setresuid");
+                        return Err(anyhow!("Failed to set UID with setresuid: {}", e));
+                    }
+                }
+                #[cfg(not(any(target_os = "linux", target_os = "android")))]
+                {
+                    if let Err(e) = setuid(target_uid) {
+                        tracing::error!(error=?e, "Failed to set UID with setuid");
+                        return Err(anyhow!("Failed to set UID with setuid: {}", e));
+                    }
                 }
 
                 //  Set up environment variables for the target user


mountpoint-s3/src/run.rs
```diff
@@ -185,8 +192,19 @@ where
     let bucket_description = args.bucket_description();
     tracing::debug!("using S3 personality {s3_personality:?} for {bucket_description}");
 
+    let mut filesystem_config = args.filesystem_config(sse.clone(), s3_personality);
```
dannycjones, line 195: I'm thinking `CliArgs::filesystem_config` should set the UID and GID itself.

dannycjones (commented):
Just some early comments

This will need a version bump and change log entry, this is a new feature!

mountpoint-s3/src/cli.rs
```diff
@@ -388,6 +388,14 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
     )]
     pub bind: Option<Vec<String>>,
 
+    #[clap(
+        long,
+        help = "Run as the specified user instead of root",
+        help_heading = MOUNT_OPTIONS_HEADER,
+        value_name = "USERNAME"
+    )]
+    pub as_user: Option<String>,
```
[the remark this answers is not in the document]
muddyfish: I was thinking `--run-as` but `--run-as-user` is also good.

mountpoint-s3/src/run.rs
```diff
@@ -185,8 +192,19 @@ where
...
+
+    if let Some(switcher) = &user_switcher {
+        filesystem_config.uid = switcher.uid().as_raw();
+        filesystem_config.gid = switcher.gid().as_raw();
+    }
+
+    //  Switch user before mounting
+    if let Some(switcher) = user_switcher {
```
muddyfish, line 203: Why not merge these two if statements together?

mountpoint-s3/src/run.rs
```diff
@@ -254,3 +272,51 @@ fn create_pid_file() -> anyhow::Result<()> {
...
+    uid: Uid,
+    gid: Gid,
+}
+
+impl UserSwitcher {
+    pub fn new(username: &str) -> anyhow::Result<Self> {
+        let username = CString::new(username)?;
+
+        // Get user info
+        let passwd_ptr = unsafe { getpwnam(username.as_ptr()) };
```
muddyfish, line 286: Why is this unsafe, and what makes us think this particular call is safe?

muddyfish: Is this a better approach? https://docs.rs/nix/latest/nix/unistd/struct.User.html#method.from_name
Would it avoid usage of unsafe?

mountpoint-s3/src/run.rs
```diff
@@ -184,8 +184,9 @@ where
     let bucket_description = args.bucket_description();
     tracing::debug!("using S3 personality {s3_personality:?} for {bucket_description}");
 
-    let s3_path = args.s3_path();
     let filesystem_config = args.filesystem_config(sse.clone(), s3_personality);
+
```
muddyfish, line 188: I'm assuming this was left over from a previous change and is no longer needed?

mountpoint-s3/src/cli.rs
```diff
@@ -511,9 +519,46 @@ impl CliArgs {
...
+                }
+            }
+        }
+
         filesystem_config
     }
 
+    fn resolve_user_group(username: &str) -> anyhow::Result<(u32, u32)> {
+        // Get user info using the safe nix API
+        let user = nix::unistd::User::from_name(username)?.ok_or_else(|| anyhow!("User '{}' not found", username))?;
```
muddyfish, line 553: Nit: Could we import `User` here and just do `User::from_name` (and the same for Group below?)

mountpoint-s3/src/cli.rs
```diff
@@ -511,9 +519,46 @@ impl CliArgs {
...
+                    // Handle errors
+                    if let Err(e) = setgid(gid) {
+                        tracing::error!(error=?e, "Failed to set GID");
+                    }
+                    if let Err(e) = setuid(uid) {
+                        tracing::error!(error=?e, "Failed to set UID");
+                    }
+                }
+                Err(e) => {
+                    tracing::error!(error=?e, username=%username, "Failed to resolve user/group");
```
muddyfish, line 543: Does this exit? It probably should exit right if we can't set the user

mountpoint-s3/src/cli.rs
```diff
@@ -511,9 +519,46 @@ impl CliArgs {
...
+
+        if let Some(username) = &self.run_as_user {
+            match Self::resolve_user_group(username) {
+                Ok((uid, gid)) => {
+                    filesystem_config.uid = uid; // Sets filesystem operations to run as this user
+                    filesystem_config.gid = gid;
+                    tracing::info!("Running Mountpoint as user: {}", username);
+
+                    // Convert to proper types for setuid/setgid
+                    let gid = Gid::from_raw(gid);
```
muddyfish, line 531: Perhaps we could make `resolve_user_group` return Gid/Uid and do `to_raw` once each rather than doing a back and forth

mountpoint-s3/src/cli.rs
```diff
@@ -511,9 +519,46 @@ impl CliArgs {
         filesystem_config.cache_config = self.cache_config();
         filesystem_config.mem_limit = self.mem_limit();
         filesystem_config.use_upload_checksums = self.should_use_upload_checksum(s3_personality);
+
+        if let Some(username) = &self.run_as_user {
+            match Self::resolve_user_group(username) {
+                Ok((uid, gid)) => {
+                    filesystem_config.uid = uid; // Sets filesystem operations to run as this user
```
muddyfish, line 526: Could we have a test to see how this interacts with the `uid` and `gid` flags?

mountpoint-s3/src/cli.rs
```diff
@@ -511,9 +519,40 @@ impl CliArgs {
...
+                    filesystem_config.gid = gid.as_raw();
+                    tracing::info!("Running Mountpoint as user: {}", username);
+
+                    // Handle errors
+                    if let Err(e) = setgid(gid) {
+                        tracing::error!(error=?e, "Failed to set GID");
+                    }
+                    if let Err(e) = setuid(uid) {
+                        tracing::error!(error=?e, "Failed to set UID");
+                    }
```
dannycjones, line 536: Sorry, I didn't explain well earlier. I think this method should only handle preparing the `S3FilesystemConfig` and should not have side effects on the process (such as setting UID/GID). I think that should happen elsewhere - perhaps very early in Mountpoint's lifetime?

We need to handle errors here. Logging/tracing is not enough, we should exit with failure.

mountpoint-s3/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## Unreleased
 
+## v1.18.0 (May 13, 2025)
```
dannycjones, line 3: drop this, we'll add during release.

the important thing is that we add the new features section

PR #1429, opened 2025-05-21
Add "nofail" to list of ignored arguments in fstab

muddyfish:
Added "nofail" to the list of ignored arguments with fstab. Whilst `systemd` removed `nofail` when launching Mountpoint, `mount` did not, which meant `mount -a` could fail when systemd launched Mountpoint fine.

Yes, Mountpoint no longer crashes if given `nofail` with fstab

No

---


passaro (approved):
LGTM

Issue #1389, opened 2025-04-28
Cache folder size keep increasing even max-cache-size exceeded


nguyenminhdungpg:
### Mountpoint for Amazon S3 version

mount-s3 1.16.2

### Describe the running environment

N/A

### Mountpoint options

```shell
N/A
```

### What happened?

I have max-cache-size is around 400000 MiB. Traffics to read file is too high so when cache folder size exceeds the max-cache-size, it keeps increasing. 
I guess the cause is in the logic: "By default, Mountpoint will limit the maximum size of the local cache such that the free space on the file system does not fall below max-cache-size, and will automatically evict the least recently used content from the local cache when caching new content." The item evicted is smaller than the new cache content, so it keep increasing the cache folder size.

What will happen if the cache folder size exceeds available size of the disk?

### Relevant log output

```shell

```

nguyenminhdungpg:
@dannycjones this is urgent, could you give me some quick response. Thank you.

vladem:
Hi @nguyenminhdungpg, thanks for reporting this. We don't expect Mountpoint to crash when the disk is full, but it may result in downgraded performance due to errors on writing/reading from cache. I can see how Mountpoint using more space than configured can be an issue and happy to investigate that. A couple of questions:

1. How much beyond the limit of 400GiB do you observe Mountpoint to use for cache?
2. Does disk usage keeps increasing or saturates at some value above the configured limit?
3. Do you observe any errors in logs, in particular in regards to cache block eviction (`unable to evict block`)? 

Please also share Mountpoint debug logs capturing the moment when Mountpoint uses more space than configured so we can investigate the issue further (may be enabled with `--debug` CLI flag).

___ 

In regards to `If the item evicted is smaller than the new cache content, the cache folder continue to increasing its size.`, that shouldn't be an issue, since Mountpoint evict multiple blocks until the limit is met again:
https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs#L330

nguyenminhdungpg:
Hi @vladem , thank you very much.

Let me tell you more detail info. My system has a very high traffics to read S3 file, millions per 24hour. In front of mount-s3, I have several layer of cache, so the reading requests sent to mount-s3 is around 1 -> 2 millions. (uploaded files are saved into S3 bucket in another sub-system, 100 thousands files per day).
[...]

PR #1433, opened 2025-05-23
Ensure cache blocks are written atomically

passaro:
Address an issue with cache block reads failing while a concurrent write is in progress, observed for example in #1389 (see log entries in [comment](https://github.com/awslabs/mountpoint-s3/issues/1389#issuecomment-2861696762)). This change modifies `put_block` to write to a temporary file first and then rename to the expected cache block file name. 

In addition, this PR also addresses concurrency issues in tracking block usage data for eviction: updates to `UsageInfo` were not previously synchronized correctly with the operations on disk and we could end up recording a new block write when in fact the block had been concurrently deleted. Now we lock `UsageInfo` while performing file system operations.

No.

Bug fix entry.

---

### DIFF
--- a/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
+++ b/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
@@ -2,7 +2,7 @@
 
 use std::fs;
 use std::io::{ErrorKind, Read, Write};
-use std::os::unix::fs::{DirBuilderExt, OpenOptionsExt};
+use std::os::unix::fs::{DirBuilderExt, PermissionsExt};
 use std::path::{Path, PathBuf};
 use std::time::Instant;
 
@@ -14,6 +14,7 @@ use bytes::Bytes;
 use linked_hash_map::LinkedHashMap;
 use mountpoint_s3_client::checksums::crc32c::{self, Crc32c};
 use sha2::{Digest, Sha256};
+use tempfile::NamedTempFile;
 use thiserror::Error;
 use tracing::{trace, warn};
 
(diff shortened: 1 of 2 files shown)


mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -1083,4 +1069,52 @@ mod tests {
...
+                        let bytes: Bytes = vec![0u8; block_size].into();
+                        data_cache
+                            .put_block(cache_key, block_idx, block_offset, bytes.into(), object_size)
+                            .await
+                            .expect("put_block should succeed");
+                    }
+                })
+                .unwrap();
+            handles.push(handle);
+        }
```
dannycjones, line 1111: Did this demonstrate the issue before the fix? I'm wondering if we should have something a bit more sophisticated, like a set of threads just reading non-stop either getting a miss or hit, while another set writes the blocks non-stop.

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -365,21 +366,18 @@ impl DiskDataCache {
...
-            .create(true)
-            .truncate(true)
-            .mode(0o600)
-            .open(path.as_ref())?;
-        file.write_all(CACHE_VERSION.as_bytes())?;
-        let bytes_written = block.write(&mut file)?;
-        Ok(bytes_written)
+        temp_file.write_all(CACHE_VERSION.as_bytes())?;
+        let bytes_written = block.write(&mut temp_file)?;
+        Ok((temp_file, bytes_written))
```
dannycjones: The logging here doesn't make sense anymore. We're writing to a temp location.

Maybe we include temp path and final path?

dannycjones (commented):
Nice, the `NamedTempFile` approach is quite clean

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
passaro: Yes, it fails with `get_block should not return error` before the fix.

PR #1434, opened 2025-05-23
Add support for passing S3 URIs as part of the bucket name field

muddyfish:
Allows invoking Mountpoint with an S3 URI in the 'bucket name' parameter

- When using an S3 URI, a prefix can also be supplied. When it is, the `--prefix` option cannot be given.
- Allows using an S3 URI with the `--cache-xz` parameter, but without a prefix.
- Documentation entry for the feature was introduced

Yes, the 'bucket name' and 'cache-xz' parameters now can take S3 URIs.
There are no breaking changes.

Changelog entry was made. Needs minor version bump.

---

### DIFF
--- a/mountpoint-s3-fs/src/prefix.rs
+++ b/mountpoint-s3-fs/src/prefix.rs
@@ -3,7 +3,7 @@ use std::str::FromStr;
 
 use thiserror::Error;
 
-#[derive(Error, Debug)]
+#[derive(Error, Debug, PartialEq)]
 pub enum PrefixError {
     #[error("prefix must end in '/'")]
     MissingFinalDelimiter,
@@ -24,6 +24,10 @@ impl Prefix {
                 path: prefix.to_owned(),
             })
         }
+    }
+
+    pub fn empty() -> Self {
+        Self::new("").unwrap()
     }
 
     pub fn as_str(&self) -> &str {
--- a/mountpoint-s3-fs/src/prefix.rs
+++ b/mountpoint-s3-fs/src/prefix.rs
@@ -27,7 +27,7 @@ impl Prefix {
     }
 
     pub fn empty() -> Self {
-        Self::new("").unwrap()
+        Self { path: "".to_owned() }
     }
 
     pub fn as_str(&self) -> &str {
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -16,7 +16,6 @@ use mountpoint_s3_fs::prefix::Prefix;
 use mountpoint_s3_fs::s3::config::{ClientConfig, PartConfig, S3Path};
 use mountpoint_s3_fs::s3::S3Personality;
 use mountpoint_s3_fs::{autoconfigure, metrics, S3FilesystemConfig};
-use regex::Regex;
 use sysinfo::{RefreshKind, System};
 
 use crate::build_info;
@@ -35,7 +34,7 @@ Alternative fstab style:
 
 Arguments:
   <BUCKET_NAME>
-          Name of bucket to mount, with s3:// URIs supported
+          Name of bucket, or an S3 URI, to mount
   <DIRECTORY>
           Location to mount bucket at
   <OPTIONS>
@@ -54,7 +53,7 @@ Arguments:
     after_help = if cfg!(feature = "fstab") {FSTAB_DOCS} else {""},
 )]
 pub struct CliArgs {
-    #[clap(help = "Name of bucket to mount", value_parser = parse_bucket_name)]
+    #[clap(help = "Name of bucket, or an S3 URI, to mount")]
     pub bucket_name: String,
 
     #[clap(
@@ -342,7 +341,6 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
         help = "Enable caching of object content to the specified bucket on S3 Express One Zone (same region only)",
         help_heading = CACHING_OPTIONS_HEADER,
         value_name = "BUCKET",
-        value_parser = parse_bucket_name,
         group = "cache_group",
     )]
     pub cache_xz: Option<String>,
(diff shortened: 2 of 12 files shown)


mountpoint-s3/src/cli.rs
```diff
@@ -553,19 +551,30 @@ impl CliArgs {
         cache_config
     }
 
-    fn cache_express_bucket_name(&self) -> Option<&str> {
+    fn cache_express_bucket_name(&self) -> Option<Result<S3Path, S3PathError>> {
         if let Some(bucket_name) = &self.cache_xz {
-            return Some(bucket_name);
+            let s3_path = S3Path::new(bucket_name.to_owned(), Prefix::new("").unwrap());
```
vladem, line 556: what's the point of allowing s3 path in `--cache-xz` argument? specifying prefix is not allowed there, is it just for uniformity?

mountpoint-s3/src/cli.rs
```diff
@@ -54,7 +53,7 @@ Arguments:
     after_help = if cfg!(feature = "fstab") {FSTAB_DOCS} else {""},
 )]
 pub struct CliArgs {
-    #[clap(help = "Name of bucket to mount", value_parser = parse_bucket_name)]
+    #[clap(help = "Name of bucket to mount")]
     pub bucket_name: String,
```
dannycjones, line 57: Should we update the help here? Do we want to encourage any adoption of S3 URIs or discourage it?

```suggestion
    #[clap(help = "Name of bucket, or an S3 URI, to mount")]
    pub bucket_name: String,
```

mountpoint-s3/src/cli.rs
```diff
@@ -553,19 +554,31 @@ impl CliArgs {
...
-    fn cache_express_bucket_name(&self) -> Option<&str> {
+    fn cache_express_bucket_name(&self) -> Option<anyhow::Result<S3Path>> {
         if let Some(bucket_name) = &self.cache_xz {
-            return Some(bucket_name);
+            let s3_path = S3Path::new(bucket_name.to_owned(), Prefix::new("").unwrap())
+                .map_err(|e| anyhow!(e).context("could not parse cache-xz argument"));
+            return Some(s3_path);
         }
         None
     }
```
dannycjones, line 564: This should probably return `Result<Option<S3Path>>`, not `Option<Result<S3Path>>`?

mountpoint-s3/src/cli.rs
```diff
@@ -553,19 +554,31 @@ impl CliArgs {
         cache_config
     }
 
-    fn cache_express_bucket_name(&self) -> Option<&str> {
+    fn cache_express_bucket_name(&self) -> Option<anyhow::Result<S3Path>> {
         if let Some(bucket_name) = &self.cache_xz {
-            return Some(bucket_name);
+            let s3_path = S3Path::new(bucket_name.to_owned(), Prefix::new("").unwrap())
+                .map_err(|e| anyhow!(e).context("could not parse cache-xz argument"));
```
dannycjones, line 560: If we want `Prefix::new("")` to always be valid, should we just add `Prefix::empty()`?

dannycjones (commented):
All of these comments are more implementation focused. Ultimately, we need to get the customer experience right here so let's ensure we're aligned there. I'll take another look.

dannycjones (commented):
I think a lot of the complexity here is because we've moved argument validation from the Clap parser into how we access the fields.

mountpoint-s3-fs/src/prefix.rs
```diff
@@ -26,6 +26,10 @@ impl Prefix {
         }
     }
 
+    pub fn empty() -> Self {
+        Self::new("").unwrap()
```
unexge, line 30: Maybe just
```suggestion
        Self {
            path: String::new(),
        }
```
to avoid `unwrap`

mountpoint-s3/src/cli.rs
```diff
@@ -448,12 +446,24 @@ impl CliArgs {
...
+                S3Path::from_s3_uri(self.bucket_name.to_owned()).map_err(|e| anyhow!(e))
+            } else {
+                Err(anyhow!("explicit prefix option not allowed with S3 URI"))
+            }
+        } else {
+            S3Path::from_bucket_prefix(self.bucket_name.to_owned(), prefix.clone()).map_err(|e| anyhow!(e))
+        }
+        .map_err(|e| {
+            e.context(format!(
+                "could not parse `BUCKET_NAME` and `prefix`. Found bucket name: '{}' and prefix: '{}'",
```
unexge, line 462: Could this be confusing if the customer is using S3 URIs? Maybe we should have separate context in each block (S3 URI vs S3 Bucket/Prefix)

mountpoint-s3/src/cli.rs
```diff
@@ -54,7 +53,7 @@ Arguments:
     after_help = if cfg!(feature = "fstab") {FSTAB_DOCS} else {""},
 )]
 pub struct CliArgs {
-    #[clap(help = "Name of bucket to mount", value_parser = parse_bucket_name)]
+    #[clap(help = "Name of bucket, or an S3 URI, to mount")]
     pub bucket_name: String,
```
unexge, line 57: As per @dannycjones's comment, maybe this should be 
```suggestion
    pub bucket_name: S3Path,
```
and we should keep value parser here? So we don't have to change each method to return a `Result`?

dannycjones (commented):
I want to try to get to a point we're happy with test cases, since we can refactor internals later.

Can we disallow S3 URIs for cache bucket name to simplify in the first iteration?

mountpoint-s3/src/cli.rs
muddyfish: We can't change this to `S3Path` directly here, as it needs to take the prefix argument into account

dannycjones (dismissed):
pretty much looks good to me, happy to merge

PR #1440, opened 2025-05-28
Update prefetch and backpressure documentation, minor code changes for clarity

dannycjones:
This change should not change any functionality, and only modifies document comments or rewrites code for clarity and to demonstrate assumptions.

No.

No behavior changes expected, no need for any changelog or version change.

---

### DIFF
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -633,8 +633,11 @@ impl UploadChecksum {
 pub trait ClientBackpressureHandle {
     /// Increment the flow-control read window, so that response data continues downloading.
     ///
-    /// The client should read data that is included within the read window,
-    /// even if that means reading beyond `len` (such as a full part).
+    /// The client should read data up to and including the end of the read window,
+    /// even if that means fetching and returning data beyond the end of the window
+    /// to fulfil a part-sized GetObject request.
+    /// As an example, a call with `len = 1` could trigger an 8MiB GetObject request
+    /// if the given client fetches data in requests of 8MiB ranges.
     fn increment_read_window(&mut self, len: usize);
 
     /// Move the upper bound of the read window to the given offset if it's not already there.
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -632,6 +632,9 @@ impl UploadChecksum {
 /// Maintain a smaller window to limit the amount of data buffered in memory.
 pub trait ClientBackpressureHandle {
     /// Increment the flow-control read window, so that response data continues downloading.
+    ///
+    /// The client should read data that is included within the read window,
+    /// even if that means reading beyond `len` (such as a full part).
     fn increment_read_window(&mut self, len: usize);
 
     /// Move the upper bound of the read window to the given offset if it's not already there.
--- a/mountpoint-s3-fs/examples/prefetch_benchmark.rs
+++ b/mountpoint-s3-fs/examples/prefetch_benchmark.rs
@@ -162,8 +162,9 @@ fn main() {
                         let mut offset = 0;
                         while offset < size {
                             let bytes = request.read(offset, args.read_size).await.unwrap();
-                            offset += bytes.len() as u64;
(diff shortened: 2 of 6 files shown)


mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -162,8 +162,9 @@ fn main() {
                         let mut offset = 0;
                         while offset < size {
                             let bytes = request.read(offset, args.read_size).await.unwrap();
-                            offset += bytes.len() as u64;
-                            received_bytes.fetch_add(bytes.len() as u64, Ordering::SeqCst);
+                            let length = bytes.len() as u64;
+                            offset += length;
+                            received_bytes.fetch_add(length, Ordering::SeqCst);
```
dannycjones, line 167: I have some logging changes I'm using for benchmarking, this makes it a bit nicer since I need to reuse the length again.

mountpoint-s3-fs/src/prefetch/part_stream.rs
```diff
@@ -437,11 +443,14 @@ fn read_from_request<'a, Client: ObjectClient + 'a>(
             if excess_bytes > 0 {
                 metrics::histogram!("s3.client.read_window_excess_bytes").record(excess_bytes as f64);
             }
-            // Blocks if read window increment if it's not enough to read the next offset
+
+            // When we detect an updated read window end offset, pass this signal on to the S3 client.
+            // TODO:
+            //   It does not make sense to 'block' here. In reality, we don't actually block here anyway.
```
vladem, line 449: alternatively we may have a separate async task responsible for advancing the read window based on values from the channel, while this task will yield data as soon as it is available in request

the issue with `backpressure controller should communicate directly with the client` is that the requests do not yet exist when `BackpressureController` is instantiated, so we will have to spawn both requests early (with zero read window?) in order to have `client_backpressure_handle` in the right place

mountpoint-s3-client/src/object_client.rs
```diff
@@ -632,6 +632,9 @@ impl UploadChecksum {
 /// Maintain a smaller window to limit the amount of data buffered in memory.
 pub trait ClientBackpressureHandle {
     /// Increment the flow-control read window, so that response data continues downloading.
+    ///
+    /// The client should read data that is included within the read window,
```
vladem, line 636: is this comment meant to convey that client may return more data than expected since the read window is always aligned to the `read_part_size`?

vladem (approved):
LGTM, added some clarifying questions

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -31,25 +31,40 @@ pub struct BackpressureConfig {
...
     read_window_updater: Sender<usize>,
+    /// Amount by which the producer should be producing data ahead of [Self::next_read_offset].
     preferred_read_window_size: usize,
     min_read_window_size: usize,
     max_read_window_size: usize,
+    /// Multiplier by which [Self::preferred_read_window_size] is scaled.
     read_window_size_multiplier: usize,
-    /// Upper bound of the current read window. The request can return data up to this
-    /// offset *exclusively*. This value must be advanced to continue fetching new data.
+    /// Upper bound of the current read window, relative to the S3 object.
```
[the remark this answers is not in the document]
dannycjones: The object. I had originally put "(not meta request)" and dropped it thinking it was just me, I will put it back since it's still unclear. 😄

mountpoint-s3-fs/src/prefetch/part_stream.rs
dannycjones: Makes sense, I was expecting there to be some juggling to do. I do wonder if it would be better yet to avoid introducing another async task, but I think we need to think about it a bit more.

I'll rework the comment a bit more so it's less prescriptive and more exploring ideas.

mountpoint-s3-client/src/object_client.rs
dannycjones: No, this comment was worded a bit badly. I just want to convey that if you were to send 1 byte here, it would trigger an 8MiB GetObject (or whatever the part size is). I'll reword!

vladem (approved):

PR #1443, opened 2025-05-30
Fstab tests: Ignore empty directory

c-hagem:
Currently, our fstab tests can fail if run in an environment where the output dir does not exist.
This PR changes the `rm -r` call to a `rm -rf` to ignore cases where the directory is empty.

---

### DIFF
--- a/mountpoint-s3/tests/fstab/spawn_mounts.sh
+++ b/mountpoint-s3/tests/fstab/spawn_mounts.sh
@@ -10,7 +10,7 @@ function spawn_mounts() {
 
   # systemd-fstab-generator takes 3 arguments, which are paths to put results in
   # It only actually puts anything in the first directory, but it appears standard to call these normal, early, and late.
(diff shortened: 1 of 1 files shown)


mountpoint-s3/tests/fstab/spawn_mounts.sh
```diff
@@ -10,7 +10,7 @@ function spawn_mounts() {
 
   # systemd-fstab-generator takes 3 arguments, which are paths to put results in
   # It only actually puts anything in the first directory, but it appears standard to call these normal, early, and late.
-  rm -r "$OUTPUT_DIR" || true
+  rm -rf "$OUTPUT_DIR" || true
```
unexge, line 13: I assume we don't need `|| true` anymore with this change?

c-hagem: Yes, deleted it.

PR #1450, opened 2025-05-30
Fix example for using fstab in user data

muddyfish:
The previous example for using fstab with user data failed to install Mountpoint occasionally on AL2023 hosts, and appears to be impacted by this bug: https://github.com/amazonlinux/amazon-linux-2023/issues/397 & https://repost.aws/questions/QU_tj7NQl6ReKoG53zzEqYOw/amazon-linux-2023-issue-with-installing-packages-with-cloud-init.

Updating the example user data script to retry installing Mountpoint if it fails.

No

No

---


unexge (approved):
Nice catch!

PR #1456, opened 2025-06-05
Add Slack notifications for PRs and issues

muddyfish:
Adds a Slack notifier URL workflow (copied from Pytorch connector)

No

No

---


.github/workflows/notify_issue.yml
```diff
@@ -0,0 +1,20 @@
+name: Issue Slack Notifier
+
+on:
+  issues:
+    types: [opened, reopened, edited]
```
dannycjones: Why for `edited`?

.github/workflows/notify_pr.yml
```diff
@@ -0,0 +1,20 @@
+name: Pull Request Slack Notifier
+
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize]
```
dannycjones: Why on `reopened`, `synchronize`?

.github/workflows/notify_issue.yml
dannycjones, line 1: nit, non-blocking: add missing new lines

dannycjones, line 1: nit, non-blocking: maybe we should try and combine all the slack jobs into a single workflow

.github/workflows/notify_pr.yml
```diff
@@ -0,0 +1,20 @@
...
+      - name: Send notification to Slack
+        uses: slackapi/slack-github-action@v2.1.0
+        with:
+          webhook: ${{ secrets.SLACK_WEBHOOK_URL_PR }}
+          webhook-type: incoming-webhook
+          payload: |
+            {
+              "action": "${{ github.event.action }}",
+              "url": "${{ github.event.pull_request.html_url || github.event.head_commit.url }}"
+            }
```
dannycjones: When would we actually want to link to the head commit?

.github/workflows/notify_pr.yml
tadiwa-aizen, line 1: About using  `pull_request_target`

From my understanding , we need  pull_request_target to access the `SLACK_WEBHOOK_URL_PR` secret
But pull_request_target creates a security vulnerability because it runs with full repository permissions which i guess could apply even for PRs from forks?

Worst case scenario could be like this?:
1)Attacker forks mountpoint-s3
2)Modifies the workflow to exfiltrate secrets instead of sending to Slack:

3)Opens a PR - the malicious workflow runs automatically

4)Steals the Slack webhook URL

Then the attacker could:
Spamour team's Slack channel or even Impersonate AWS team communications to spread misinformation



`A suggested solution?`

on:
  pull_request_target:
    types: [opened]
// `Add a job condition to only run for trusted users/orgs`
if: github.event.pull_request.head.repo.full_name == github.repository

muddyfish: We need to use `pull_request_target` for this because `pull_request` removes secrets. However, `pull_request_target` runs the workflow from the context of the main repository rather than the fork, which means that PRs changing this workflow wouldn't change the actual workflow ran before it merges. 

Changing the PR so it only runs for this repo/PRs from team members means we don't get notified when someone creates an issue/PR which defeats the point of this change.

.github/workflows/notify_pr.yml
muddyfish: `reopened` so we learn if a previously closed PR needs attention again (so we don't miss it)
`synchronize` so we know when someone's updated a PR - it fires when new commits are added.

.github/workflows/notify_issue.yml
muddyfish: I'd be happy to remove the edited flag, but the idea was such that we'd know if someone edited an issue. It doesn't matter too much

.github/workflows/notify_pr.yml
tadiwa-aizen: 🫡 Understood

.github/workflows/notify_pr.yml
dannycjones: Do we really want `synchronize`?

renanmagagnin (dismissed):

PR #1457, opened 2025-06-06
Update MP client_benchmark to support CRT backpressure

mansi153:
Update MP client_backmark to support CRT backpressure. Extend the benchmark to optionally enable read-backpressure in CRT, and configure the initial read window size. This test aims to simulate the read-ahead capability of the prefetcher, making it easier to baseline the performance against the prefetcher benchmark.

No, changes to the benchmark only.

No, changes to the benchmark only.

---

### DIFF
--- a/mountpoint-s3-client/examples/client_benchmark.rs
+++ b/mountpoint-s3-client/examples/client_benchmark.rs
@@ -47,7 +47,7 @@ fn run_benchmark(
                 let received_size_clone = Arc::clone(&received_size);
                 scope.spawn(|| {
                     futures::executor::block_on(async move {
-                        let received_obj_len = Arc::new(AtomicU64::new(0));
+                        let mut received_obj_len = 0u64;
                         let mut request = client
                             .get_object(bucket, key, &GetObjectParams::new())
                             .await
@@ -55,9 +55,11 @@ fn run_benchmark(
                         let mut backpressure_handle = request.backpressure_handle().cloned();
                         if enable_backpressure {
                             if let Some(backpressure_handle) = backpressure_handle.as_mut() {
-                                backpressure_handle.ensure_read_window(client.initial_read_window_size().unwrap() as u64);
+                                let initial_read_window_size = client.initial_read_window_size().expect("initial size always set when backpressure is enabled");
+                                backpressure_handle.ensure_read_window(initial_read_window_size as u64);
                             }
                         }
+
                         let mut request = pin!(request);
                         loop {
                             match request.next().await {
@@ -70,14 +72,14 @@ fn run_benchmark(
                                         "consuming data",
                                     );
                                     received_size_clone.fetch_add(part_len as u64, Ordering::SeqCst);
-                                    received_obj_len.fetch_add(part_len as u64, Ordering::SeqCst);
+                                    received_obj_len += part_len as u64;
                                     if enable_backpressure {
                                         if let Some(backpressure_handle) = backpressure_handle.as_mut() {
                                             tracing::info!(
                                                 target: "benchmarking_instrumentation",
                                                 preferred_read_window_size = ?client.initial_read_window_size(),
-                                                prev_read_window_end_offset = ?(client.initial_read_window_size().unwrap() as u64 + received_obj_len.load(Ordering::Relaxed) - part_len as u64),
-                                                new_read_window_end_offset = ?(client.initial_read_window_size().unwrap() as u64 + received_obj_len.load(Ordering::Relaxed)),
+                                                prev_read_window_end_offset = ?(client.initial_read_window_size().unwrap() as u64 + received_obj_len - part_len as u64),
+                                                new_read_window_end_offset = ?(client.initial_read_window_size().unwrap() as u64 + received_obj_len),
                                                 part_len = part_len,
                                                 "advancing read window",
                                             );
@@ -152,15 +154,15 @@ struct CliArgs {
         visible_alias = "memory-limit-gb"
     )]
     crt_memory_limit_gb: u64,
-    #[arg(long, help = "Part size for multi-part GET", default_value = "8388608")]
+    #[arg(long, help = "Part size in bytes for multi-part GET", default_value = "8388608")]
     part_size: usize,
     #[arg(long, help = "Number of benchmark iterations", default_value = "1")]
     iterations: usize,
     #[arg(long, help = "Number of concurrent downloads", default_value = "1")]
     downloads: usize,
-    #[arg(long, help = "Enable CRT backpressure mode", default_value = "false")]
+    #[arg(long, help = "Enable CRT backpressure mode")]
     enable_backpressure: bool,
-    #[arg(long, help = "Initial window size", default_value = "0")]
+    #[arg(long, help = "Initial window size in bytes", default_value = "0")]
     initial_window_size: usize,
 }
 
--- a/mountpoint-s3-client/examples/client_benchmark.rs
+++ b/mountpoint-s3-client/examples/client_benchmark.rs
@@ -189,7 +189,14 @@ fn main() {
             }
             let client = S3CrtClient::new(config).expect("couldn't create client");
 
-            run_benchmark(client, args.iterations, args.downloads, &bucket, &key, args.enable_backpressure);
+            run_benchmark(
+                client,
+                args.iterations,
+                args.downloads,
+                &bucket,
+                &key,
+                args.enable_backpressure,
+            );
         }
         Client::Mock { object_size } => {
             const BUCKET: &str = "bucket";
@@ -204,7 +211,14 @@ fn main() {
 
             client.add_object(KEY, MockObject::ramp(0xaa, object_size as usize, ETag::for_tests()));
 
-            run_benchmark(client, args.iterations, args.downloads, BUCKET, KEY, args.enable_backpressure);
+            run_benchmark(
+                client,
+                args.iterations,
+                args.downloads,
+                BUCKET,
+                KEY,
+                args.enable_backpressure,
+            );
         }
     }
 }
--- a/mountpoint-s3-client/examples/client_benchmark.rs
+++ b/mountpoint-s3-client/examples/client_benchmark.rs
@@ -162,8 +162,12 @@ struct CliArgs {
     downloads: usize,
     #[arg(long, help = "Enable CRT backpressure mode")]
     enable_backpressure: bool,
-    #[arg(long, help = "Initial window size in bytes", default_value = "0")]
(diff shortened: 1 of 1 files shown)


mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -128,6 +158,10 @@ struct CliArgs {
     iterations: usize,
     #[arg(long, help = "Number of concurrent downloads", default_value = "1")]
     downloads: usize,
+    #[arg(long, help = "Enable CRT backpressure mode", default_value = "false")]
+    enable_backpressure: bool,
```
dannycjones: should be able to use the implied default (`false`)

```suggestion
    #[arg(long, help = "Enable CRT backpressure mode")]
    enable_backpressure: bool,
```

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -128,6 +158,10 @@ struct CliArgs {
...
+    #[arg(long, help = "Initial window size", default_value = "0")]
+    initial_window_size: usize,
```
dannycjones:
```suggestion
    #[arg(long, help = "Initial window size in bytes", default_value = "0")]
    initial_window_size: usize,
```

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -46,21 +47,50 @@ fn run_benchmark(
...
+                                            backpressure_handle.increment_read_window(part_len);
+                                        }
+                                    }
                                 }
                                 Some(Err(e)) => {
                                     tracing::error!(error = ?e, "request failed");
                                     break;
                                 }
-                                None => break,
+                                _ => break,
```
dannycjones, line 95: leave this as `None`

```suggestion
                                None => break,
```

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -46,21 +47,50 @@ fn run_benchmark(
...
                         let mut request = client
                             .get_object(bucket, key, &GetObjectParams::new())
                             .await
                             .expect("couldn't create get request");
+                        let mut backpressure_handle = request.backpressure_handle().cloned();
+                        if enable_backpressure {
+                            if let Some(backpressure_handle) = backpressure_handle.as_mut() {
+                                backpressure_handle.ensure_read_window(client.initial_read_window_size().unwrap() as u64);
+                            }
+                        }
```
dannycjones: I'd just take the backpressure if supplied. The handle shouldn't be made available anyway if backpressure is off.

```suggestion
                        if let Some(backpressure_handle) = request.backpressure_handle() {
                            let initial_read_window_size = client.initial_read_window_size().expect("initial size always set when backpressure is enabled");
                            backpressure_handle.ensure_read_window(initial_read_window_size as u64);
                        }
```

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -46,21 +47,50 @@ fn run_benchmark(
...
+                                                preferred_read_window_size = ?client.initial_read_window_size(),
+                                                prev_read_window_end_offset = ?(client.initial_read_window_size().unwrap() as u64 + received_obj_len.load(Ordering::Relaxed) - part_len as u64),
+                                                new_read_window_end_offset = ?(client.initial_read_window_size().unwrap() as u64 + received_obj_len.load(Ordering::Relaxed)),
+                                                part_len = part_len,
+                                                "advancing read window",
+                                            );
+
+                                            backpressure_handle.increment_read_window(part_len);
+                                        }
+                                    }
```
dannycjones: same again: https://github.com/awslabs/mountpoint-s3/pull/1457#discussion_r2135631495

```rust
if let Some(backpressure_handle) = request.backpressure_handle() { .. }
```

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -46,21 +47,50 @@ fn run_benchmark(
...
                             match request.next().await {
                                 Some(Ok(part)) => {
-                                    received_size_clone.fetch_add(part.data.len() as u64, Ordering::SeqCst);
+                                    let part_len = part.data.len();
+                                    tracing::info!(
+                                        target: "benchmarking_instrumentation",
+                                        received_obj_len = ?received_obj_len,
+                                        part_len = part_len,
+                                        "consuming data",
+                                    );
```
dannycjones, line 73: I haven't added any of these tracing points yet in other layer benchmarks. I am wondering if we should omit for now for merging into `main`.

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -46,21 +47,50 @@ fn run_benchmark(
                 let received_size_clone = Arc::clone(&received_size);
                 scope.spawn(|| {
                     futures::executor::block_on(async move {
+                        let received_obj_len = Arc::new(AtomicU64::new(0));
```
dannycjones: We don't need an atomic here, this isn't shared between threads

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: Thanks, done!

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: Thanks, done!

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: Why not have a catch-all here, given we're not handling the `None` case in a special way? Am I missing something?

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: Replied below with the explanation. Hopefully I'm not misunderstanding something here.

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: Sorry, that's how I originally tried it (and again now), but it gets quite complicated to borrow the reference due to the pinning of request (which is need because we need to `await` on parts):

```
error[E0596]: cannot borrow data in dereference of Pin<&mut <impl ObjectClient + Clone + Send as ObjectClient>::GetObjectResponse> as mutable
  --> mountpoint-s3-client/examples/client_benchmark.rs:75:76
   |
75 | ...                   if let Some(backpressure_handle) = request.backpressure_handle() {
   |                                                          ^^^^^^^ cannot borrow as mutable
   |
   = help: trait DerefMut is required to modify through a dereference, but it is not implemented for Pin<&mut <impl ObjectClient + Clone + Send as ObjectClient>::GetObjectResponse>

For more information about this error, try rustc --explain E0596.
error: could not compile mountpoint-s3-client (example "client_benchmark") due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

"Borrowing the handle before the pinning happens and keeping it around for later" is a simple workaround, because fixing it otherwise seems to require a bunch of refactoring and is probably also impossible given we can't change anything in the BackpressureHandle signature. I'd prefer to keep it as is to avoid complicating the benchmark script too much. Happy to incorporate if you have any ideas though.

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: I think it's a useful trace point to have for future testing. Since we're tracing these with a specific target and within a benchmark script itself, it might not be too noisy and I'd prefer to keep it here instead of having to add it again later when backtracking logs for analyses.

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: Thanks, done!

mountpoint-s3-client/examples/client_benchmark.rs
dannycjones: I'd lean towards being explicit since we'll be made aware if the API changes in some way where we may need to handle additional cases. This may be subjective, but I'd also avoid changing from `None` to `_` now since its unrelated to this PR.

mountpoint-s3-client/examples/client_benchmark.rs
dannycjones: Ah, yeah, sorry I didn't check this one.

I think borrowing the handle before the pin is fine. The goal here is to eliminate the amount of conditions around backpressure handles, to keep it to a minimum. The workaround is adding only one line (and removing the need to pass `enable_backpressure` around).

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -122,12 +154,16 @@ struct CliArgs {
...
+    #[arg(long, help = "Part size in bytes for multi-part GET", default_value = "8388608")]
     part_size: usize,
     #[arg(long, help = "Number of benchmark iterations", default_value = "1")]
     iterations: usize,
     #[arg(long, help = "Number of concurrent downloads", default_value = "1")]
     downloads: usize,
+    #[arg(long, help = "Enable CRT backpressure mode")]
+    enable_backpressure: bool,
+    #[arg(long, help = "Initial window size in bytes", default_value = "0")]
+    initial_window_size: usize,
```
dannycjones: An initial read window size of `0` doesn't make sense and will result in `EmptyReadWindow` errors.

I'd propose that we make this parameter an `Option<usize>` which is required when `enable_backpressure` is set. When using it, we can `expect("read window size is required when backpressure is enabled")`.

I'd also propose updating the name - this is actually just a read window size and is used beyond the initial request start. (Sorry, should have mentioned this first time around.)

```suggestion
    #[arg(long, help = "Enable CRT backpressure mode", requires = "initial_window_size")]
    enable_backpressure: bool,
    #[arg(long, help = "Initial read window size in bytes, used to dictate how far ahead we request data from S3")]
    read_window_size: Option<usize>,
```

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -46,21 +47,52 @@ fn run_benchmark(
                 let received_size_clone = Arc::clone(&received_size);
                 scope.spawn(|| {
                     futures::executor::block_on(async move {
+                        let mut received_obj_len = 0u64;
```
dannycjones, line 50: if we don't specify the type, we won't need to annotate with `as u64` later.

```suggestion
                        let mut received_obj_len = 0;
```

 a couple of the type casts throughout should be removable.

c-hagem (approved):

PR #1461, opened 2025-06-09
Minor refactor to prefetch_benchmark

dannycjones:
Introducing some minor refactoring to `prefetch_benchmark` before adding some more significant changes (- adding caching support). This change also introduces `anyhow::Result` to properly format errors when running the benchmark, including sharing additional context and error sources.

This is mainly a refactor. It does change error handling - errors are now properly returned and formatted using `anyhow`, rather than via panics.

No, refactor only.

---

### DIFF
--- a/mountpoint-s3-fs/examples/prefetch_benchmark.rs
+++ b/mountpoint-s3-fs/examples/prefetch_benchmark.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 use std::thread;
 use std::time::Instant;
 
+use anyhow::Context;
 use clap::{value_parser, Parser};
 use futures::executor::block_on;
 use mountpoint_s3_client::config::{EndpointConfig, RustLogAdapter, S3ClientConfig};
@@ -19,7 +20,7 @@ use tracing_subscriber::EnvFilter;
 
 /// Like `tracing_subscriber::fmt::init` but sends logs to stderr
 fn init_tracing_subscriber() {
-    RustLogAdapter::try_init().expect("unable to install CRT log adapter");
+    RustLogAdapter::try_init().expect("should succeed as first and only adapter init call");
 
     let subscriber = Subscriber::builder()
         .with_env_filter(EnvFilter::from_default_env())
@@ -27,7 +28,9 @@ fn init_tracing_subscriber() {
         .with_writer(std::io::stderr)
         .finish();
 
-    subscriber.try_init().expect("unable to install global subscriber");
+    subscriber
+        .try_init()
+        .expect("should succeed as first and only subscriber init call");
 }
 
 #[derive(Parser, Debug)]
@@ -101,7 +104,7 @@ pub struct CliArgs {
     pub bind: Option<Vec<String>>,
 }
 
-fn main() {
+fn main() -> anyhow::Result<()> {
     init_tracing_subscriber();
     let _metrics_handle = mountpoint_s3_fs::metrics::install();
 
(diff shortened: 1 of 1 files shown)


mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -110,48 +110,35 @@ fn main() {
...
-    for i in 0..args.iterations {
-        let runtime = Runtime::new(client.event_loop_group());
-        let manager =
-            Prefetcher::default_builder(client.clone()).build(runtime, mem_limiter.clone(), Default::default());
-        let received_bytes = Arc::new(AtomicU64::new(0));
+    let runtime = Runtime::new(client.event_loop_group());
+
+    for iteration in 0..args.iterations {
+        let manager = {
+            let client = client.clone();
```
passaro: Why the change here? This should be inlined as before.

mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -176,10 +163,31 @@ fn main() {
...
-            i,
+            iteration,
             received_size,
             elapsed.as_secs_f64(),
             (received_size as f64) / elapsed.as_secs_f64() / (1024 * 1024 * 1024 / 8) as f64
         );
     }
 }
+
+fn make_s3_client_from_args(args: &CliArgs) -> Arc<S3CrtClient> {
```
passaro: `S3CrtClient` is `Clone`, no need to wrap it in `Arc`

mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -176,10 +163,31 @@ fn main() {
...
+    if let Some(limit_gib) = args.crt_memory_limit_gib {
+        client_config = client_config.memory_limit_in_bytes(limit_gib * 1024 * 1024 * 1024);
+    }
+    if let Some(part_size) = args.part_size {
+        client_config = client_config.part_size(part_size as usize);
+    }
+    if let Some(interfaces) = &args.bind {
+        client_config = client_config.network_interface_names(interfaces.clone());
+    }
+    Arc::new(S3CrtClient::new(client_config).unwrap())
```
passaro: I'd rather propagate the error here and restore the `expect` at the call site.

passaro (changes requested):

mountpoint-s3-fs/examples/prefetch_benchmark.rs
dannycjones: It's a bit wonky because the CRT client init error is not made public. I will try with returning `impl Error` instead but happy to discuss.

I'm also going to just update this benchmark to return an `anyhow::Result` so we can propagate errors properly rather than panicking.

mountpoint-s3-fs/examples/prefetch_benchmark.rs
dannycjones: Nice spot, happy to remove 😄

mountpoint-s3-fs/examples/prefetch_benchmark.rs
dannycjones: It's part of the next PR, but I'll inline it for this one.

mountpoint-s3-fs/examples/prefetch_benchmark.rs
passaro: Could also have returned anyhow here. We should make that error public, in any case!

PR #1462, opened 2025-06-10
Use multi-threaded FuseSession in tests

passaro:
Tests for the `mountpoint-s3-fs` crates now use the multi-threaded `FuseSession` used in `mount-s3` rather than the single-threaded `BackgroundSession` from `fuser`.

No, it only affects tests.

No.

---

### DIFF
--- a/mountpoint-s3-fs/src/superblock.rs
+++ b/mountpoint-s3-fs/src/superblock.rs
@@ -1084,6 +1084,8 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
                 trace!(ino, "removing inode from superblock");
                 let parent_ino = removed_inode.parent();
 
+                // Remove from the parent. Nothing to do if the parent has already been removed,
+                // e.g. when FORGETs are handled out-of-order.
                 if let Some((parent, _)) = inodes.get_mut(&parent_ino) {
                     let mut parent_state = parent.get_mut_inode_state_no_check();
                     let InodeKindData::Directory {
@@ -1101,10 +1103,6 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
                         }
                     }
                     writing_children.remove(&ino);
-                } else {
-                    // Should be impossible for this to fail (VFS inodes reference their parent, so
-                    // children need to be freed first), but let's not crash in a `forget` function...
-                    debug_assert!(false, "children should be forgotten before parents");
                 }
 
                 drop(inodes);


mountpoint-s3-fs/src/fuse/session.rs
```diff
@@ -24,7 +24,7 @@ pub struct FuseSession {
     on_close: Vec<OnClose>,
 }
 
-type OnClose = Box<dyn FnOnce()>;
+type OnClose = Box<dyn FnOnce() + Send>;
```
muddyfish, line 27: Why do we need to change the constraint here?

mountpoint-s3-fs/tests/fuse_tests/prefetch_test.rs
```diff
@@ -79,6 +79,7 @@ fn prefetch_test_etag(
         TestSessionConfig {
             part_size,
             initial_read_window_size,
+            max_worker_threads: 1, // avoid stale inode issues. (FIXME)
```
vladem, line 82: any idea on what kind of stale inode issues?

mountpoint-s3-fs/tests/common/fuse.rs
```diff
@@ -150,8 +153,13 @@ impl Drop for TestSession {
         // If the session created with a pre-existing mount (e.g., with `pass_fuse_fd`),
         // this will unmount it explicitly...
         drop(self.mount.take());
-        // ...if not, the background session will have a mount here, and dropping it will unmount it.
-        self.session.take();
+        // ...if not, the fuse session will unmount on shutdown.
+        if let Some(session) = self.session.take() {
+            session.shutdown_fn()();
+            if let Err(error) = session.join() {
+                tracing::warn!(?error, "error while unmounting");
```
muddyfish, line 160: I'm assuming this is fine for tests?

passaro: Yes. It means clean up failed, but it does not affect the tests.

mountpoint-s3-fs/src/fuse/session.rs
passaro: This is so `FuseSession` can be `Send`. We use that in tests and it's not an unreasonable constraint here.

muddyfish (approved):

mountpoint-s3-fs/tests/fuse_tests/prefetch_test.rs
passaro: There seems to be an issue with the second open failing sometimes with `ESTALE`.

PR #1464, opened 2025-06-12
Otlp exporter

15skumar:
This PR adds an implementation of OpenTelemetry Exporting of metrics through the OpenTelemetry protocol (OTLP). Changes are: a new OtlpMetricsExporter struct which handles exporting metrics to an OTLP endpoint, and integration of the OTLP exporter with the existing metrics system. 

Testing:
I tested the implementation with a test otlp_metrics() in metrics.rs and ran a docker container running the OpenTelemetry Collector at the default port

docker run -d --name otel-collector \
  -p 4318:4318 -p 4317:4317 \
  -v $(pwd)/collector-config.yaml:/etc/otelcol/config.yaml \
  otel/opentelemetry-collector-contrib:latest

Once I ran the test, I verified that the test metrics can be viewed in the collector logs. (viewed using 'docker logs otel-collector'). Here is a screenshot of an example of a test metric collected at the endpoint: 
Screenshot 2025-06-18 at 15 32 16

---

### DIFF
--- a/mountpoint-s3-fs/src/metrics.rs
+++ b/mountpoint-s3-fs/src/metrics.rs
@@ -65,9 +65,8 @@ pub fn install(otlp_config: Option<OtlpConfig>) -> Result<MetricsSinkHandle, Box
     };
 
     let recorder = MetricsRecorder { sink };
-    metrics::set_global_recorder(recorder).map_err(|e| {
-        Box::<dyn std::error::Error>::from(format!("Failed to set global metrics recorder: {}", e))
-    })?;
+    metrics::set_global_recorder(recorder)
+        .map_err(|e| Box::<dyn std::error::Error>::from(format!("Failed to set global metrics recorder: {}", e)))?;
 
     Ok(handle)
 }
@@ -104,9 +103,7 @@ impl MetricsSink {
         let otlp_exporter = if let Some(config) = otlp_config {
             // Basic validation of the endpoint URL
             if !config.endpoint.starts_with("http://") && !config.endpoint.starts_with("https://") {
-                return Err(
-                    "Invalid OTLP endpoint configuration: endpoint must start with http:// or https://".into()
-                );
+                return Err("Invalid OTLP endpoint configuration: endpoint must start with http:// or https://".into());
             }
 
             match OtlpMetricsExporter::new(&config) {
--- a/mountpoint-s3-fs/src/metrics.rs
+++ b/mountpoint-s3-fs/src/metrics.rs
@@ -33,7 +33,7 @@ pub const TARGET_NAME: &str = "mountpoint_s3_fs::metrics";
 /// done with their work; metrics generated after shutting down the sink will be lost.
 ///
 /// Panics if a sink has already been installed.
-pub fn install(otlp_config: Option<OtlpConfig>) -> Result<MetricsSinkHandle, Box<dyn std::error::Error>> {
+pub fn install(otlp_config: Option<OtlpConfig>) -> anyhow::Result<MetricsSinkHandle> {
     let sink = Arc::new(MetricsSink::new(otlp_config)?);
     let mut sys = System::new();
 
@@ -66,7 +66,7 @@ pub fn install(otlp_config: Option<OtlpConfig>) -> Result<MetricsSinkHandle, Box
 
     let recorder = MetricsRecorder { sink };
     metrics::set_global_recorder(recorder)
-        .map_err(|e| Box::<dyn std::error::Error>::from(format!("Failed to set global metrics recorder: {}", e)))?;
+        .map_err(|e| anyhow::anyhow!("Failed to set global metrics recorder: {}", e))?;
 
     Ok(handle)
 }
@@ -98,12 +98,14 @@ struct MetricsSink {
 }
 
 impl MetricsSink {
-    fn new(otlp_config: Option<OtlpConfig>) -> Result<Self, Box<dyn std::error::Error>> {
+    fn new(otlp_config: Option<OtlpConfig>) -> anyhow::Result<Self> {
         // Initialise the OTLP exporter if a config is provided
         let otlp_exporter = if let Some(config) = otlp_config {
             // Basic validation of the endpoint URL
             if !config.endpoint.starts_with("http://") && !config.endpoint.starts_with("https://") {
-                return Err("Invalid OTLP endpoint configuration: endpoint must start with http:// or https://".into());
+                return Err(anyhow::anyhow!(
+                    "Invalid OTLP endpoint configuration: endpoint must start with http:// or https://"
+                ));
             }
 
             match OtlpMetricsExporter::new(&config) {
@@ -117,10 +119,10 @@ impl MetricsSink {
 
                     // If the user explicitly requested metrics export but it failed,
                     // we should return an error rather than silently continuing without metrics
-                    return Err(format!(
+                    return Err(anyhow::anyhow!(
                         "Failed to initialize OTLP metrics exporter: {}. If metrics export is not required, omit the OTLP configuration.", 
                         e
-                    ).into());
+                    ));
                 }
             }
         } else {
--- a/mountpoint-s3-fs/src/metrics.rs
+++ b/mountpoint-s3-fs/src/metrics.rs
@@ -93,7 +93,7 @@ fn poll_process_metrics(sys: &mut System) {
 #[derive(Debug)]
 struct MetricsSink {
     metrics: DashMap<Key, Metric>,
-    otlp_exporter: Option<OtlpMetricsExporter>, 
+    otlp_exporter: Option<OtlpMetricsExporter>,
 }
 
 impl MetricsSink {
@@ -102,9 +102,11 @@ impl MetricsSink {
         let otlp_exporter = if let Some(config) = otlp_config {
             // Basic validation of the endpoint URL
             if !config.endpoint.starts_with("http://") && !config.endpoint.starts_with("https://") {
-                return Err(format!("Invalid OTLP endpoint configuration: endpoint must start with http:// or https://"));
+                return Err(
+                    "Invalid OTLP endpoint configuration: endpoint must start with http:// or https://".to_string(),
+                );
             }
-            
+
             match OtlpMetricsExporter::new(&config) {
                 Ok(exporter) => {
                     tracing::info!("OpenTelemetry metrics export enabled to {}", config.endpoint);
@@ -113,7 +115,7 @@ impl MetricsSink {
                 Err(e) => {
                     let error_msg = format!("Failed to initialise OTLP exporter: {}", e);
                     tracing::error!("{}", error_msg);
-                    
+
                     // If the user explicitly requested metrics export but it failed,
                     // we should return an error rather than silently continuing without metrics
                     return Err(format!("Failed to initialize OTLP metrics exporter: {}. If metrics export is not required, omit the OTLP configuration.", e));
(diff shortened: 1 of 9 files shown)


collector-config.yaml
```diff
@@ -0,0 +1,45 @@
+receivers:
```
muddyfish: Should this be committed?

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -90,12 +93,27 @@ fn poll_process_metrics(sys: &mut System) {
 #[derive(Debug)]
 struct MetricsSink {
     metrics: DashMap<Key, Metric>,
+    otlp_exporter: Option<OtlpMetricsExporter>, // optional OTLP exporter
 }
 
 impl MetricsSink {
-    fn new() -> Self {
+    fn new(otlp_endpoint: Option<&str>) -> Self {
```
muddyfish: Perhaps we should add a config struct for this so it's easier to add/change fields

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -90,12 +93,27 @@ fn poll_process_metrics(sys: &mut System) {
...
+        // Initialise the OTLP exporter if an endpoint is provided
+        // check syntax and if error case works
+        let otlp_exporter = otlp_endpoint.and_then(|endpoint| match OtlpMetricsExporter::new(endpoint) {
+            Ok(exporter) => {
+                tracing::info!("OpenTelemetry metrics export enabled to {}", endpoint);
+                Some(exporter)
+            }
+            Err(e) => {
+                tracing::error!("Failed to initialise OTLP exporter: {}", e);
```
muddyfish: Do we want to continue or error out in this case?

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -310,4 +359,85 @@ mod tests {
             }
         });
     }
+
+    #[test]
+    #[ignore] // Mark as ignored so it doesn't run in normal test runs
```
muddyfish: This test appears to just export metrics and then not have any validation steps. Ideally tests should fail if they don't work properly

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -310,4 +359,85 @@ mod tests {
...
+            .with_level(true)
+            .with_file(true)
+            .with_line_number(true)
+            .with_test_writer()
+            .set_default();
+
+        info!("Starting OTLP metrics test...");
+
+        // Initialize metrics with an OTLP endpoint
+        let sink = Arc::new(MetricsSink::new(Some("http://localhost:4318/v1/metrics")));
```
muddyfish: Nit: This should probably be an environment variable

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -310,4 +359,85 @@ mod tests {
...
+            histogram.record(10.0);
+            histogram.record(20.0);
+            histogram.record(30.0);
+            info!("Recorded histogram values: 10.0, 20.0, 30.0");
+
+            // Publish metrics immediately to verify initial values
+            info!("Publishing initial metrics...");
+            sink.publish();
+
+            // Sleep to allow metrics to be exported
```
muddyfish, line 471: Exported from where?

mountpoint-s3-fs/src/metrics_otel.rs
```diff
@@ -0,0 +1,167 @@
...
+            .with_protocol(Protocol::HttpBinary)
+            .with_endpoint(endpoint)
+            .build()?;
+
+        // Create a meter provider with the OTLP Metric Exporter that will collect and export metrics at regular intervals
+        let meter_provider = opentelemetry_sdk::metrics::SdkMeterProvider::builder()
+            // The default interval is 60 seconds so we use a PeriodicReader to allow us to specify a custom interval duration
+            .with_reader(
+                opentelemetry_sdk::metrics::PeriodicReader::builder(exporter)
+                    .with_interval(Duration::from_secs(5)) // This sets a 5-second export interval
```
muddyfish: Should be configurable

mountpoint-s3-fs/src/metrics_otel.rs
```diff
@@ -0,0 +1,167 @@
...
+                    .build(),
+            )
+            .build();
+        global::set_meter_provider(meter_provider);
+
+        // Get a meter
+        let meter = global::meter("mountpoint-s3");
```
muddyfish, line 63: What does this do?

mountpoint-s3-fs/src/metrics_otel.rs
```diff
@@ -0,0 +1,167 @@
...
+
+        Ok(Self { meter })
+    }
+
+    /// Record a counter metric in OTel format
+    pub fn record_counter(&self, key: &Key, value: u64, attributes: &[KeyValue]) {
+        let counter = self
+            .meter
+            .u64_counter(key.name().to_string())
+            .with_description("Mountpoint counter metric")
```
muddyfish: These are not useful descriptions.

mountpoint-s3-fs/src/metrics_otel.rs
15skumar: The set_meter_provider function makes the configured meter provider the default provider that will be used throughout the application for all OpenTelemetry metrics. 

The next line obtains a meter instance from the global meter provider that was set up. The meter is used to create specific metrics (counters, gauges, histograms) and record values to them.

I'll add comments in the code to make this more clear.

collector-config.yaml
15skumar: No, it shouldn't as it was just for local testing purposes. I have deleted it now, thank you for pointing it out!

mountpoint-s3-fs/src/metrics.rs
15skumar: Yes makes sense, I have added a new struct for the configuration of the otlp exporter so that fields can be specified and added in the future if needed.

mountpoint-s3-fs/src/metrics_otel.rs
15skumar: I have now added a field in the new config struct to allow interval to be configurable. Thank you!

mountpoint-s3-fs/src/metrics.rs
15skumar: The aim of this test was to send metrics to the endpoint, and I locally verified that metrics were sent by running a collector. 
Have discussed that we can have a look in the future at writing a full test to verify it within the test.

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -90,12 +93,28 @@ fn poll_process_metrics(sys: &mut System) {
 #[derive(Debug)]
 struct MetricsSink {
     metrics: DashMap<Key, Metric>,
+    otlp_exporter: Option<OtlpMetricsExporter>, // optional OTLP exporter
```
renanmagagnin: nit: comment seems unnecessary

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -90,12 +93,28 @@ fn poll_process_metrics(sys: &mut System) {
...
 }
 
 impl MetricsSink {
-    fn new() -> Self {
+    fn new(otlp_config: Option<OtlpConfig>) -> Self {
+        // Initialise the OTLP exporter if a config is provided
+        let otlp_exporter = otlp_config
```
renanmagagnin: We should probably handle the None otlp_config separately with it's dedicated tracing. This would provide more clarity and possibly avoid confusing with error cases

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -310,4 +360,86 @@ mod tests {
             }
         });
     }
+
+    #[test]
+    #[ignore]
```
renanmagagnin, line 403: Does it make sense for us to have tests that are used for manual testing and will be ignored in the code base? Not clear what the standard for this is. If we decide to keep them, maybe we should make the purpose clear with comments

renanmagagnin: Also, should we have a test for the case where the config is none?

mountpoint-s3-fs/src/metrics.rs
15skumar: Since the user is explicitly requesting metrics export, it makes sense to return an error if the exporter fails to initialise for any reason. I have changed the code to do this, and have also added basic endpoint url validation.

mountpoint-s3-fs/src/metrics.rs
15skumar: I have added explicit logging now for the case when no OTLP configuration is provided, thank you for the suggestion

mountpoint-s3-fs/src/metrics.rs
15skumar: I have added comments to make the purpose of the manual tests clear, and have added a test for the case of no otlp config

mountpoint-s3-fs/src/metrics_otel.rs
15skumar: I've removed the generic descriptions as they weren't providing specific information about the metrics and don't appear in CloudWatch. If we need descriptions in the future, we can add more specific ones that describe what each metric actually measures.

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -30,8 +33,8 @@ pub const TARGET_NAME: &str = "mountpoint_s3_fs::metrics";
 /// done with their work; metrics generated after shutting down the sink will be lost.
 ///
 /// Panics if a sink has already been installed.
-pub fn install() -> MetricsSinkHandle {
-    let sink = Arc::new(MetricsSink::new());
+pub fn install(otlp_config: Option<OtlpConfig>) -> Result<MetricsSinkHandle, String> {
```
c-hagem: Would it make sense to have an error type here instead of returning a string? (not really sure what the convention for this would be)

c-hagem (approved):

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -62,9 +65,10 @@ pub fn install() -> MetricsSinkHandle {
     };
 
     let recorder = MetricsRecorder { sink };
-    metrics::set_global_recorder(recorder).unwrap();
+    metrics::set_global_recorder(recorder)
```
muddyfish, line 68: Would using anyhow here be better?

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -90,13 +94,45 @@ fn poll_process_metrics(sys: &mut System) {
...
 
 impl MetricsSink {
-    fn new() -> Self {
-        Self {
+    fn new(otlp_config: Option<OtlpConfig>) -> Result<Self, Box<dyn std::error::Error>> {
+        // Initialise the OTLP exporter if a config is provided
+        let otlp_exporter = if let Some(config) = otlp_config {
+            // Basic validation of the endpoint URL
+            if !config.endpoint.starts_with("http://") && !config.endpoint.starts_with("https://") {
+                return Err("Invalid OTLP endpoint configuration: endpoint must start with http:// or https://".into());
```
muddyfish: Same here - `anyhow` might be more readable

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -90,13 +94,45 @@ fn poll_process_metrics(sys: &mut System) {
...
+            }
+
+            match OtlpMetricsExporter::new(&config) {
+                Ok(exporter) => {
+                    tracing::info!("OpenTelemetry metrics export enabled to {}", config.endpoint);
+                    Some(exporter)
+                }
+                Err(e) => {
+                    let error_msg = format!("Failed to initialise OTLP exporter: {}", e);
+                    tracing::error!("{}", error_msg);
```
muddyfish: Any reason why this is split on two lines?

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -115,12 +151,43 @@ impl MetricsSink {
...
+                // Convert labels to OpenTelemetry KeyValue pairs
+                let attributes: Vec<KeyValue> = key
+                    .labels()
+                    .map(|label| KeyValue::new(label.key().to_string(), label.value().to_string()))
+                    .collect();
+
+                // Record the metric based on its type
+                match metric {
+                    Metric::Counter(counter) => {
+                        if let Some((value, _)) = counter.load_and_reset() {
```
muddyfish, line 174: Does this interfere with how metrics are printed with the `--log-metrics` parameter?

15skumar: Yes it does, and I have implemented a solution for that in another branch for which I was planning to open a PR for next.

mountpoint-s3-fs/src/metrics.rs
15skumar: Yes using anyhow is more readable and cleaner code, i have changed the code to use anyhow for error handling now.

muddyfish (approved):

PR #1474, opened 2025-06-23
Add support to collect perf stat counters in benchmark.py

sahityadg:
Add support to collect perf stat counters in benchmark.py

No

No, only affects benchmark.py

---


benchmark/benchmark.py
```diff
@@ -258,26 +260,56 @@ def _get_ec2_instance_id() -> Optional[str]:
...
+                return pid
+
+        raise RuntimeError(f"No mount-s3 process found for path: {mount_dir}")
+
+    except subprocess.CalledProcessError as e:
+        log.error("Failed to execute ps command")
+        raise RuntimeError("Could not search for mount-s3 process") from e
+    except (ValueError, IndexError) as e:
+        log.error("Failed to parse ps output")
+        raise RuntimeError("Could not determine mount-s3 PID") from e
```
dannycjones: Mountpoint already has a method to output its own PID: https://github.com/awslabs/mountpoint-s3/blob/7c023072cea67aa617d85170594eb8fc2a1db0f7/mountpoint-s3/src/run.rs#L241-L255

Is querying it here preferred in case Mountpoint is spawned in its own namespace? If so, maybe we should remove this code from MP itself in the future. (It was added only for attaching `perf`.) What do you think?

benchmark/benchmark.py
```diff
@@ -258,26 +260,56 @@ def _get_ec2_instance_id() -> Optional[str]:
...
+
+
 class ResourceMonitoring:
-    def __init__(self, with_bwm: bool):
+    def __init__(self, target_pid, with_bwm: bool, with_perf_stat: bool):
```
dannycjones, line 282:
```suggestion
    def __init__(self, target_pid: Optional[int], with_bwm: bool, with_perf_stat: bool):
```

or `int | None`

dannycjones (dismissed):
looks good! nothing blocking

benchmark/benchmark.py
adpeace: The pidfile option isn't stable right now, though.  Should we make it stable, merge this change as-is, then update the implementation to use the stable pidfile support?  I agree the method of "searching" for the right PID is not ideal.  Or use the unstable option now, and change this implementation when we make a stable version?

sahityadg: I missed the pid file option, switched to using that one for now to unblock the change.

benchmark/benchmark.py
sahityadg: There are other tools we want to add that will take the pid (like numa stats) . So I will keep the current version instead of making it an option.

dannycjones: Ah my bad, not sure why I thought it was optional. The main thing here was just to maintain the type annotations.

dannycjones (approved):

adpeace (approved):

PR #1475, opened 2025-06-23
Move `reader_count` out of inode

c-hagem:
Moves the reader count out of the inode and instead stores the reader counts for all inodes with non-zero reader count in a HashMap (that is protected by a lock).

This should not have breaking changes, it could potentially reduce unlikely issues with the reader count getting messed up in highly concurrent scenarios involving re-creation of inodes with the same number.

Does not need a Changelog entry or version change.

---

### DIFF
--- a/mountpoint-s3-fs/src/superblock.rs
+++ b/mountpoint-s3-fs/src/superblock.rs
@@ -404,7 +404,7 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
 
     /// Create a new handle for a file being read. The handle can be used to update the state of
     /// the inflight read and commit it once finished.
-    pub async fn read(&self, ino: InodeNo) -> Result<ReadHandle, InodeError> {
+    pub async fn read(&self, ino: InodeNo) -> Result<ReadHandle<OC>, InodeError> {
         trace!(?ino, "read");
 
         let inode = self.inner.get(ino)?;
@@ -1187,7 +1187,7 @@ impl<OC: ObjectClient + Send + Sync> SuperblockInner<OC> {
         name: ValidName,
         remote: Option<RemoteLookup>,
     ) -> Result<LookedUp, InodeError> {
-        let mut parent_state = parent.get_mut_inode_state()?.state;
+        let mut parent_state = parent.get_mut_inode_state()?;
         let inode = match &parent_state.kind_data {
             InodeKindData::File { .. } => unreachable!("we know parent is a directory"),
             InodeKindData::Directory { children, .. } => children.get(name.as_ref()).cloned(),
--- a/mountpoint-s3-fs/src/superblock.rs
+++ b/mountpoint-s3-fs/src/superblock.rs
@@ -1399,6 +1399,8 @@ impl InodeMap {
 }
 
 #[derive(Debug, Default)]
+/// A wrapper around a `HashMap<InodeNo, u32>` to manage reader counts for Inodes
+/// Ensures that the inodes are locked for writing while performing these operations.
 struct ReaderCountMap {
     map: HashMap<InodeNo, u32>,
 }
--- a/mountpoint-s3-fs/src/superblock.rs
+++ b/mountpoint-s3-fs/src/superblock.rs
@@ -242,7 +242,7 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
             bucket: bucket.to_owned(),
             prefix: prefix.clone(),
             inodes: RwLock::new(inodes),
-            reader_counts: RwLock::new(ReaderCountMap::new()),
+            reader_counts: Default::default(),
             negative_cache,
             next_ino: AtomicU64::new(2),
             mount_time,
@@ -1404,11 +1404,7 @@ struct ReaderCountMap {
 }
 
 impl ReaderCountMap {
-    fn new() -> Self {
-        ReaderCountMap { map: HashMap::new() }
-    }
-
-    fn is_anyone_reading(&self, locked_inode: &InodeLockedForWriting) -> bool {
+    fn has_readers(&self, locked_inode: &InodeLockedForWriting) -> bool {
         self.map.get(&locked_inode.ino).is_some_and(|&count| count > 0)
     }
 
--- a/mountpoint-s3-fs/src/superblock.rs
+++ b/mountpoint-s3-fs/src/superblock.rs
@@ -111,6 +111,7 @@ struct SuperblockInner<OC: ObjectClient + Send + Sync> {
     bucket: String,
     prefix: Prefix,
     inodes: RwLock<InodeMap>,
+    reader_counts: RwLock<ReaderCountMap>,
     negative_cache: NegativeCache,
     cached_rename_support: RenameCache,
     next_ino: AtomicU64,
@@ -239,6 +240,7 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
             bucket: bucket.to_owned(),
             prefix: prefix.clone(),
             inodes: RwLock::new(inodes),
+            reader_counts: RwLock::new(ReaderCountMap::new()),
             negative_cache,
             next_ino: AtomicU64::new(2),
             mount_time,
@@ -404,7 +406,7 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
         trace!(?ino, "read");
 
         let inode = self.inner.get(ino)?;
-        ReadHandle::new(inode)
+        ReadHandle::new(self.inner.clone(), inode)
     }
 
     /// Start a readdir stream for the given directory inode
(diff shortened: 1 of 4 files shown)


mountpoint-s3-fs/src/superblock/inode.rs
```diff
@@ -27,6 +27,19 @@ const ROOT_INODE_NO: InodeNo = crate::fs::FUSE_ROOT_INODE;
 // 200 years seems long enough
 const NEVER_EXPIRE_TTL: Duration = Duration::from_secs(200 * 365 * 24 * 60 * 60);
 
+#[derive(Debug)]
+pub struct InodeLockedForWriting<'a> {
```
passaro: These should both have the same visibility as `InodeState`, `pub(super)`.

mountpoint-s3-fs/src/superblock/inode.rs
```diff
@@ -27,6 +27,19 @@ const ROOT_INODE_NO: InodeNo = crate::fs::FUSE_ROOT_INODE;
...
+    pub ino: InodeNo,
+    pub state: RwLockWriteGuard<'a, InodeState>,
```
passaro, line 36: Have you considered implementing `Deref` and `DerefMut` with `Target == InodeState`? They would avoid all the `.state` changes in this PR.

Same for `InodeLockedForReading` and `Deref`.

mountpoint-s3-fs/src/superblock/inode.rs
```diff
@@ -27,6 +27,19 @@ const ROOT_INODE_NO: InodeNo = crate::fs::FUSE_ROOT_INODE;
...
+}
+
+#[derive(Debug)]
+#[allow(dead_code)]
```
passaro: If you are not using `ino` right now, add the attribute there. Also, consider using `expect`, which will fail if/when we start using it, forcing us to remove the attribute.

mountpoint-s3-fs/src/superblock/inode.rs
```diff
@@ -27,6 +27,19 @@ const ROOT_INODE_NO: InodeNo = crate::fs::FUSE_ROOT_INODE;
 // 200 years seems long enough
 const NEVER_EXPIRE_TTL: Duration = Duration::from_secs(200 * 365 * 24 * 60 * 60);
 
+#[derive(Debug)]
```
passaro: rustdoc

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1446,6 +1449,36 @@ impl InodeMap {
     }
 }
 
+#[derive(Debug, Default)]
+struct ReaderCountMap {
+    map: HashMap<InodeNo, u32>,
+}
+
+impl ReaderCountMap {
+    fn new() -> Self {
```
passaro: Do we need this when we have `Default`?

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1446,6 +1449,36 @@ impl InodeMap {
...
+        ReaderCountMap { map: HashMap::new() }
+    }
+
+    fn is_anyone_reading(&self, locked_inode: &InodeLockedForWriting) -> bool {
```
passaro: Maybe `has_readers`?

passaro: And `add_reader`/`remove_reader` for the other methods?

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1446,6 +1449,36 @@ impl InodeMap {
...
+
+    fn decrease_reader_count(&mut self, locked_inode: &InodeLockedForWriting) {
+        if let Some(count) = self.map.get_mut(&locked_inode.ino) {
+            if *count > 0 {
+                *count -= 1;
+            }
+            if *count == 0 {
+                self.map.remove(&locked_inode.ino);
+            }
+        }
```
passaro: You can use `entry` here as well, e.g.:
```suggestion
        if let Entry::Occupied(mut entry) = self.map.entry(locked_inode.ino) {
            let count = entry.get_mut();
            if *count > 0 {
                *count -= 1;
            }
            if *count == 0 {
                entry.remove();
            }
        }
```

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1446,6 +1449,36 @@ impl InodeMap {
     }
 }
 
+#[derive(Debug, Default)]
```
passaro, line 1403: rustdoc

mountpoint-s3-fs/src/superblock/inode.rs
c-hagem: Adjusted that

mountpoint-s3-fs/src/superblock/inode.rs
c-hagem: Also implemented into for `RwLockWriteGuard<'a, InodeState>`, thanks for the suggestion!

mountpoint-s3-fs/src/superblock/inode.rs
c-hagem: Added the attribute

mountpoint-s3-fs/src/superblock/inode.rs
c-hagem: Added

mountpoint-s3-fs/src/superblock.rs
c-hagem: Nope -- no longer require it

mountpoint-s3-fs/src/superblock.rs
c-hagem: I left the counts, as I think it might be better to be explicit here (we are not modifying any list of readers but just increase / decrease a counter) -- happy to disagree and commit though

mountpoint-s3-fs/src/superblock.rs
c-hagem: Used that -- thanks!

mountpoint-s3-fs/src/superblock.rs
passaro: nit: rustdoc goes above the attributes

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1450,16 +1451,14 @@ impl InodeMap {
 }
 
 #[derive(Debug, Default)]
+/// A wrapper around a `HashMap<InodeNo, u32>` to manage reader counts for Inodes
+/// Ensures that the inodes are locked for writing while performing these operations.
```
passaro: The fact that it wraps a hash map is not too interesting (and visible 2 lines below). I'd rather mention that it only stores entries with > 0 readers.

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1450,16 +1451,14 @@ impl InodeMap {
...
 }
 
 impl ReaderCountMap {
-    fn new() -> Self {
-        ReaderCountMap { map: HashMap::new() }
-    }
-
-    fn is_anyone_reading(&self, locked_inode: &InodeLockedForWriting) -> bool {
+    fn has_readers(&self, locked_inode: &InodeLockedForWriting) -> bool {
         self.map.get(&locked_inode.ino).is_some_and(|&count| count > 0)
```
passaro: Isn't the `count > 0` check redundant?

mountpoint-s3-fs/src/superblock/inode.rs
passaro: nit: order

mountpoint-s3-fs/src/superblock/inode.rs
```diff
@@ -28,18 +29,48 @@ const ROOT_INODE_NO: InodeNo = crate::fs::FUSE_ROOT_INODE;
...
+impl<'a> From<InodeLockedForWriting<'a>> for RwLockWriteGuard<'a, InodeState> {
+    fn from(locked: InodeLockedForWriting<'a>) -> Self {
+        locked.state
+    }
+}
+
 #[derive(Debug)]
-#[allow(dead_code)]
-pub struct InodeLockedForReading<'a> {
+/// Read-locked inode state with its inode number, ensuring other operations use the correct inode number.
```
passaro, line 53: nit: order

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -193,11 +196,11 @@ impl<'a> RenameLockGuard<'a> {
         // Take read lock
         let ancestor_check = Self::dst_is_ancestor(&superblock_inner.inodes.read().unwrap(), source_parent, dst_ino);
         if ancestor_check || src_ino > dst_ino {
-            dst_parent_lock = Some(dest_parent.get_mut_inode_state()?);
-            src_parent_lock = source_parent.get_mut_inode_state()?;
+            dst_parent_lock = Some(dest_parent.get_mut_inode_state()?.into());
```
passaro: minor: not too convinced this `.into()` is worth it. It's not transparent and even longer than `.state`. Also not used consistently, e.g. line 189.

 Why not just use `InodeLockedForWriting` in `RenameLockGuard`?

mountpoint-s3-fs/src/superblock.rs
passaro: The type is called `ReaderCountMap` and the methods do not take a "reader", so I don't think it's ambiguous.

I do prefer the shorter names, but it's your choice.

mountpoint-s3-fs/src/superblock.rs
c-hagem: Yes

mountpoint-s3-fs/src/superblock.rs
c-hagem: Changed that.

mountpoint-s3-fs/src/superblock/inode.rs
c-hagem: Changed the order

mountpoint-s3-fs/src/superblock.rs
c-hagem: Adjusted

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1183,7 +1187,7 @@ impl<OC: ObjectClient + Send + Sync> SuperblockInner<OC> {
         name: ValidName,
         remote: Option<RemoteLookup>,
     ) -> Result<LookedUp, InodeError> {
-        let mut parent_state = parent.get_mut_inode_state()?;
+        let mut parent_state = parent.get_mut_inode_state()?.state;
```
passaro: Not needed?

c-hagem: Fixed that

PR #1476, opened 2025-06-23
Move object client into Superblock

c-hagem:
This PR moves the client into the Superblock, thus a superblock will always interact with the same instantiation of an `ObjectClient + Send + Sync`. 

No, this change does not impact existing behaviour, as is only an internal re-organisation. 

No, does not need a Changelog entry, as it only moves around where we store the client.

---

### DIFF
--- a/mountpoint-s3-fs/src/fs.rs
+++ b/mountpoint-s3-fs/src/fs.rs
@@ -57,12 +57,12 @@ where
 {
     config: S3FilesystemConfig,
     client: Client,
-    superblock: Superblock,
+    superblock: Superblock<Client>,
     prefetcher: Prefetcher<Client>,
     uploader: Uploader<Client>,
     bucket: String,
     next_handle: AtomicU64,
-    dir_handles: AsyncRwLock<HashMap<u64, Arc<DirHandle>>>,
+    dir_handles: AsyncRwLock<HashMap<u64, Arc<DirHandle<Client>>>>,
     file_handles: AsyncRwLock<HashMap<u64, Arc<FileHandle<Client>>>>,
 }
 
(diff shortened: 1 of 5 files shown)


mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -128,7 +128,7 @@ impl ReaddirHandle {
     /// Return the next inode for the directory stream. If the stream is finished, returns
     /// `Ok(None)`. Does not increment the lookup count of the returned inodes: the caller
     /// is responsible for calling [`remember()`] if required.
-    pub async fn next<OC: ObjectClient>(&self, client: &OC) -> Result<Option<LookedUp>, InodeError> {
+    pub async fn next(&self, client: &OC) -> Result<Option<LookedUp>, InodeError> {
```
passaro: Why are we still passing the client here?

mountpoint-s3-fs/src/fs.rs
```diff
@@ -57,12 +57,12 @@ where
 {
     config: S3FilesystemConfig,
     client: Client,
```
passaro: Does `S3Filesystem` still use `client`?

mountpoint-s3-fs/src/superblock/readdir.rs
c-hagem: No real reason to -- I changed this, but I do not adjust this for `RemoteIter` (that takes `&impl ObjectClient`)

mountpoint-s3-fs/src/fs.rs
c-hagem: The FilehandleState still used `fs.client` -- however this is redundant, I removed it

passaro (approved):
LGTM

passaro:
Can you edit the description to restore the template questions?

```

```

passaro (changes requested):
Can you edit the description to restore the template questions?

```

```

PR #1477, opened 2025-06-23
Add unused `CredentialsProvider::new_delegate` provider

muddyfish:
Add unused `CredentialsProvider::new_delegate` provider.

`new_delegate` takes a closure which returns a future for credentials. This can be used to implement other credentials providers in the future.

No.

Changelog entry written for mountpoint-s3-client.

---

### DIFF
--- a/mountpoint-s3-client/CHANGELOG.md
+++ b/mountpoint-s3-client/CHANGELOG.md
@@ -1,6 +1,7 @@
-## Unreleased (v0.16.1)
+## Unreleased (v0.17.0)
 
 * Update to latest CRT dependencies.
+* Add `CredentialsProvider::new_delegate`, which can be used to implement custom credential providers in Rust.
 
 ## v0.16.0 (Jun 19, 2025)
 
--- a/mountpoint-s3-crt/src/auth/credentials.rs
+++ b/mountpoint-s3-crt/src/auth/credentials.rs
@@ -1,22 +1,29 @@
 //! AWS credentials providers
 
-use std::fmt::Debug;
-use std::ptr::NonNull;
-
 use mountpoint_s3_crt_sys::{
-    aws_credentials_provider, aws_credentials_provider_acquire, aws_credentials_provider_cached_options,
-    aws_credentials_provider_chain_default_options, aws_credentials_provider_new_anonymous,
+    aws_credentials, aws_credentials_new, aws_credentials_provider, aws_credentials_provider_acquire,
+    aws_credentials_provider_cached_options, aws_credentials_provider_chain_default_options,
+    aws_credentials_provider_delegate_options, aws_credentials_provider_new_anonymous,
     aws_credentials_provider_new_cached, aws_credentials_provider_new_chain_default,
-    aws_credentials_provider_new_profile, aws_credentials_provider_new_static,
+    aws_credentials_provider_new_delegate, aws_credentials_provider_new_profile, aws_credentials_provider_new_static,
     aws_credentials_provider_profile_options, aws_credentials_provider_release,
-    aws_credentials_provider_static_options,
+    aws_credentials_provider_static_options, aws_credentials_release, aws_on_get_credentials_callback_fn, AWS_OP_ERR,
+    AWS_OP_SUCCESS,
 };
+use std::ffi::OsStr;
+use std::fmt::Debug;
+use std::future::Future;
+use std::ptr::NonNull;
+use std::sync::Arc;
+use std::time::{SystemTime, UNIX_EPOCH};
 
 use crate::auth::auth_library_init;
 use crate::common::allocator::Allocator;
 use crate::common::error::Error;
 use crate::io::channel_bootstrap::ClientBootstrap;
-use crate::{CrtError as _, ToAwsByteCursor as _};
+use crate::io::event_loop::EventLoopGroup;
+use crate::io::futures::FutureSpawner;
+use crate::{CrtError as _, ToAwsByteCursor};
 
 /// Options for creating a default credentials provider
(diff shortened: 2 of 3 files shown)


mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -54,11 +61,113 @@ impl Debug for CredentialsProviderStaticOptions<'_> {
     }
 }
 
+/// Options for creating an owned set of credentials
```
passaro: Is the rustdoc up to date?

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -54,11 +61,113 @@ impl Debug for CredentialsProviderStaticOptions<'_> {
...
+pub struct Credentials {
+    credentials: *mut aws_credentials,
```
passaro: We should use `NonNull` here.

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -54,11 +61,113 @@ impl Debug for CredentialsProviderStaticOptions<'_> {
...
+    expiration: u64,
```
passaro: Do we need `expiration`? Isn't it accessible from `aws_credentials`?

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -54,11 +61,113 @@ impl Debug for CredentialsProviderStaticOptions<'_> {
...
+            .field("access_key_id", &"** redacted **")
+            .field("secret_access_key", &"** redacted **")
+            .field("expiration", &self.expiration)
+            .finish()
+    }
+}
+
+impl Credentials {
+    /// Builds some credentials from access_key, secret_access_key, and an optional session token + expiration.
+    pub fn build_credentials<S: AsRef<OsStr>>(
```
passaro: Why not just `new`?

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -54,11 +61,113 @@ impl Debug for CredentialsProviderStaticOptions<'_> {
...
+        access_key_id: S,
```
passaro: Here we are constraining all 3 strings to be of the same type, which may be awkward. We should either:
* just use `&OsStr` and let the caller handle it,
* or use `impl AsRef<OsStr>` on each parameter for complete flexibility.

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -54,11 +61,113 @@ impl Debug for CredentialsProviderStaticOptions<'_> {
...
+    }
+
+    fn as_ptr(&self) -> &*mut aws_credentials {
+        &self.credentials
+    }
+}
+
+impl Drop for Credentials {
+    fn drop(&mut self) {
+        // SAFETY: There's at least one reference until here
```
passaro: Why? We should mention that `aws_credentials_new` in the constructor gave us an `acquired` pointer.

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -155,7 +273,50 @@ impl CredentialsProvider {
...
         };
 
-        Ok(Self { inner })
+        Ok(Self {
+            inner,
+            on_delegate_callback: None,
+        })
+    }
+
+    /// Creates a delegate credential provider that always returns the given credentials
```
passaro: Which "given credentials"? We should mention the callback and the event loop group.

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -54,11 +61,113 @@ impl Debug for CredentialsProviderStaticOptions<'_> {
...
+        drop(credentials);
+    }
+}
+
+// SAFETY: Whilst the CRT doesn't explicitly say `callback_user_data` and `callback` are usable between threads,
+// the sts provider (and others) use the `user_data` in other threads via an event loop.
+unsafe impl Send for DelegateOnGetCredentialsReplier {}
+
+type OnGetCredentials = Box<dyn Fn(DelegateOnGetCredentialsReplier) + Send + Sync + 'static>;
+struct OnDelegateCallback(OnGetCredentials);
```
passaro: Do we need all this indirection? Isn't this always the closure defined at line 294? Wouldn't it be clearer to have that logic in the type directly?

mountpoint-s3-crt/src/auth/credentials.rs
```diff
@@ -54,11 +61,113 @@ impl Debug for CredentialsProviderStaticOptions<'_> {
...
+impl Drop for Credentials {
+    fn drop(&mut self) {
+        // SAFETY: There's at least one reference until here
+        unsafe {
+            aws_credentials_release(self.credentials);
+        }
+    }
+}
+
+type OnGetCredentialsCallbackFn =
```
passaro: We should add a recap of all the callbacks and their roles.

mountpoint-s3-client/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## Unreleased (v0.16.1)
+## Unreleased (v0.17.0)
```
passaro: Not sure it needs a version bump. But if it does, we have to update `Cargo.toml` and dependent crates.

mountpoint-s3-crt/src/auth/credentials.rs
muddyfish: It is not accessible

PR #1481, opened 2025-06-24
Avoid copy of data returned by GetObject

passaro:
Reduce memory fragmentation and peak usage by avoiding copying data returned by GetObject into newly allocated buffers. This change relies on the new CRT API integrated in #1430, which allows `S3CrtClient` to extend the lifetime of the buffers from the CRT memory pool when they are returned by GetObject. Callers of the `get_object` method are now responsible for dropping the returned `Bytes` instances in order for the buffers to be released back to the CRT memory pool. 

At the moment, the memory-limiting strategy used in the prefetcher component in Mountpoint does not cope well with the change and may end up starving the CRT of available buffers. For this specific use case, we introduced a temporary feature flag in the `mountpoint-s3-client` crate, `restore_buffer_copy` which restores the previous behavior, i.e. GetObject allocates and returns new buffers with a copy of the object content. As we rework this aspect of the prefetcher, we will likely remove the feature flag.

Yes. The buffers returned by GetObject will be borrowed from the internal memory pool.

Yes. Entry and new version number for the client crate.

---

### DIFF
--- a/mountpoint-s3-client/CHANGELOG.md
+++ b/mountpoint-s3-client/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 * Reduce memory fragmentation and peak usage by avoiding copying data returned by GetObject into newly allocated buffers.
   Callers of the `get_object` method are now responsible for returning the buffers to the internal memory pool by dropping
-  the received `Bytes` instances after use. Failure to do so may eventually lead to reduced throughput when the memory pool
-  reaches capacity.
+  the received `Bytes` instances after use. Failure to do so may eventually lead to reduced or zero throughput when the
+  memory pool reaches capacity.
   ([#1481](https://github.com/awslabs/mountpoint-s3/pull/1481))
 
 ## v0.16.0 (Jun 19, 2025)
--- a/mountpoint-s3-client/CHANGELOG.md
+++ b/mountpoint-s3-client/CHANGELOG.md
@@ -1,6 +1,12 @@
-## Unreleased (v0.16.1)
+## Unreleased (v0.17.0)
 
-* Update to latest CRT dependencies.
+### Breaking changes
+
+* Reduce memory fragmentation and peak usage by avoiding copying data returned by GetObject into newly allocated buffers.
+  Callers of the `get_object` method are now responsible for returning the buffers to the internal memory pool by dropping
+  the received `Bytes` instances after use. Failure to do so may eventually lead to reduced throughput when the memory pool
+  reaches capacity.
+  ([#1481](https://github.com/awslabs/mountpoint-s3/pull/1481))
 
 ## v0.16.0 (Jun 19, 2025)
 
--- a/mountpoint-s3-client/src/s3_crt_client/get_object.rs
+++ b/mountpoint-s3-client/src/s3_crt_client/get_object.rs
@@ -102,7 +102,10 @@ impl S3CrtClient {
                 move |offset, data| {
                     _ = part_sender.unbounded_send(S3GetObjectEvent::BodyPart(GetBodyPart {
                         offset,
-                        data: Bytes::copy_from_slice(data),
+                        data: Bytes::from_owner(
+                            data.to_owned_buffer()
+                                .expect("can acquire ownership of buffers from GetObject"),
+                        ),
                     }));
                 },
                 parse_get_object_error,
--- a/mountpoint-s3-client/src/s3_crt_client/get_object.rs
+++ b/mountpoint-s3-client/src/s3_crt_client/get_object.rs
@@ -101,7 +101,10 @@ impl S3CrtClient {
                     }
                 },
                 move |offset, data| {
-                    let body_part = make_body_part(offset, data);
+                    let body_part = GetBodyPart {
+                        offset,
+                        data: make_owned_bytes(data),
+                    };
                     _ = part_sender.unbounded_send(S3GetObjectEvent::BodyPart(body_part));
                 },
                 parse_get_object_error,
@@ -148,22 +151,16 @@ impl S3CrtClient {
 }
 
 #[cfg(not(feature = "restore_buffer_copy"))]
-fn make_body_part(offset: u64, data: &Buffer) -> GetBodyPart {
+fn make_owned_bytes(data: &Buffer) -> Bytes {
     let owned_buffer = data
         .to_owned_buffer()
         .expect("buffers returned from GetObject can always be acquired");
-    GetBodyPart {
-        offset,
-        data: Bytes::from_owner(owned_buffer),
-    }
+    Bytes::from_owner(owned_buffer)
 }
 
 #[cfg(feature = "restore_buffer_copy")]
-fn make_body_part(offset: u64, data: &Buffer) -> GetBodyPart {
-    GetBodyPart {
-        offset,
-        data: Bytes::copy_from_slice(data),
-    }
+fn make_owned_bytes(data: &Buffer) -> Bytes {
+    Bytes::copy_from_slice(data)
 }
 
 #[derive(Debug)]
--- a/mountpoint-s3-client/src/s3_crt_client/get_object.rs
+++ b/mountpoint-s3-client/src/s3_crt_client/get_object.rs
@@ -149,12 +149,12 @@ impl S3CrtClient {
 
 #[cfg(not(feature = "restore_buffer_copy"))]
 fn make_body_part(offset: u64, data: &Buffer) -> GetBodyPart {
+    let owned_buffer = data
+        .to_owned_buffer()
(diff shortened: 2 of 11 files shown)


mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -100,10 +101,8 @@ impl S3CrtClient {
                     }
                 },
                 move |offset, data| {
-                    _ = part_sender.unbounded_send(S3GetObjectEvent::BodyPart(GetBodyPart {
-                        offset,
-                        data: Bytes::copy_from_slice(data),
-                    }));
+                    let body_part = make_body_part(offset, data);
```
vladem: are we releasing the buffer here after dropping `data`?

mountpoint-s3-client/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
-## Unreleased (v0.16.1)
+## Unreleased (v0.17.0)
 
-* Update to latest CRT dependencies.
+### Breaking changes
+
+* Reduce memory fragmentation and peak usage by avoiding copying data returned by GetObject into newly allocated buffers.
+  Callers of the `get_object` method are now responsible for returning the buffers to the internal memory pool by dropping
+  the received `Bytes` instances after use. Failure to do so may eventually lead to reduced throughput when the memory pool
```
vladem: > Failure to do so may eventually lead to reduced throughput

may it lead to `0` throughput if the caller holds `Bytes` forever and never drops them?

mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -148,6 +147,25 @@ impl S3CrtClient {
     }
 }
 
+#[cfg(not(feature = "restore_buffer_copy"))]
+fn make_body_part(offset: u64, data: &Buffer) -> GetBodyPart {
+    GetBodyPart {
+        offset,
+        data: Bytes::from_owner(
+            data.to_owned_buffer()
+                .expect("can acquire ownership of buffers from GetObject"),
```
vladem: under which conditions may this expectation fail?

mountpoint-s3-crt/src/s3/buffer.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /// Create an instance without incrementing the reference count of the underlying ticket.
+    /// Returns `None` if `ticket` is `null`.
+    ///
+    /// # Safety
+    /// `ticket` must be valid for the lifetime of the returned `Self`.
+    pub unsafe fn new_unowned(ticket: *mut aws_s3_buffer_ticket) -> Option<ManuallyDrop<Self>> {
+        let inner = NonNull::new(ticket)?;
+        Some(ManuallyDrop::new(Self { inner }))
```
vladem: is `ManuallyDrop` a good fit here? caller of `new_unowned` may manually drop the returned value, while we haven't called `aws_s3_buffer_ticket_acquire`

may be `drop` could handle itself? e.g. based on a boolean flag:
```
pub(crate) struct BufferTicket {
    inner: NonNull<aws_s3_buffer_ticket>,
    owned: bool,
}
```

mountpoint-s3-crt/src/s3/buffer.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// SAFETY: slice must be a valid `aws_byte_cursor`.
+    pub(crate) fn new_unchecked(slice: &'slice aws_byte_cursor, ticket: Option<&'slice BufferTicket>) -> Self {
+        Self { slice, ticket }
+    }
+
+    /// Acquire ownership of the underlying memory pool buffer and return an [OwnedBuffer].
+    ///
+    /// Fails and returns [None] if no [BufferTicket] was provided.
+    pub fn to_owned_buffer(&self) -> Option<OwnedBuffer> {
+        let ticket = self.ticket?.clone();
```
vladem: may be a comment here, that we acquire ownership / call `aws_s3_buffer_ticket_acquire` here

mountpoint-s3-crt/src/s3/buffer.rs
vladem: or 2 structs: `BufferTicket` and `BufferTicketUnowned`?

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -558,19 +561,20 @@ unsafe extern "C" fn meta_request_headers_callback(
...
 ) -> i32 {
     // SAFETY: user_data always will be a MetaRequestOptionsInner since that's what we set it to
     // in MetaRequestOptions::new.
     let user_data = MetaRequestOptionsInner::from_user_data_ptr(user_data);
 
-    if let Some(callback) = user_data.on_body.as_mut() {
-        let slice: &[u8] = aws_byte_cursor_as_slice(&*body);
-        callback(range_start, slice);
+    if let Some(callback) = user_data.on_body_ex.as_mut() {
+        let ticket = BufferTicket::new_unowned(meta.ticket);
```
vladem: will we panic here if `meta.ticket` is nullptr?

mountpoint-s3-client/src/s3_crt_client/get_object.rs
vladem: I see, we actually don't acquire ownership in the first place, may be a comment with explanation would help

mountpoint-s3-client/src/s3_crt_client/get_object.rs
vladem: why `data.to_owned_buffer()` returns `Option`?

mountpoint-s3-crt/src/s3/buffer.rs
passaro: You are right, the "unowned" part is confusing. I've reworked `Buffer` to hold the `aws_s3_buffer_ticket` pointer directly and only build a `BufferTicket` wrapper when acquiring ownership.

mountpoint-s3-client/src/s3_crt_client/get_object.rs
muddyfish: `make_body_part` takes ownership of `data` here, either by copying the bytes, or by calling `to_owned_buffer`

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -558,19 +560,19 @@ unsafe extern "C" fn meta_request_headers_callback(
...
 ) -> i32 {
     // SAFETY: user_data always will be a MetaRequestOptionsInner since that's what we set it to
     // in MetaRequestOptions::new.
     let user_data = MetaRequestOptionsInner::from_user_data_ptr(user_data);
 
-    if let Some(callback) = user_data.on_body.as_mut() {
-        let slice: &[u8] = aws_byte_cursor_as_slice(&*body);
-        callback(range_start, slice);
+    if let Some(callback) = user_data.on_body_ex.as_mut() {
+        let buffer = Buffer::new_unchecked(&*body, &meta.ticket);
```
muddyfish: Ideally should have a safety comment saying that `buffer` is dropped before `body` and `meta.ticket`

mountpoint-s3-client/src/s3_crt_client/get_object.rs
passaro: About `data`: from its type (`&Buffer`), it should be clear that it's only borrowed. Since we need to return a `GetBodyPart` which contains `Bytes`, we need to get an "owner" buffer in one way or another.

I'm thinking it would be clearer if I had a `make_owned_bytes` function instead.

mountpoint-s3-crt/src/s3/client.rs
passaro: Good point.

I think you'll like [Rust 2024](https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html).

mountpoint-s3-crt/src/s3/buffer.rs
```diff
@@ -0,0 +1,134 @@
...
+
+use crate::aws_byte_cursor_as_slice;
+
+/// A slice of a data buffer.
+///
+/// Wrapper for a [aws_byte_cursor] pointing to (part of) the body returned by
+/// a meta request. If built with a non-null ticket, it can be used to
+/// acquire ownership of the underlying memory pool buffer.
+#[derive(Debug)]
+pub struct Buffer<'slice> {
```
vladem, line 18: Ideally the name of this struct would hint that this is unowned chunk of memory, but you've got  `A slice of a data buffer.` in the rust doc and "slice" implies no ownership, so I guess all good

vladem (approved):

muddyfish (approved):

PR #1498, opened 2025-07-01
Upgrade to Rust 2024

passaro:
Upgrade crates to [Rust 2024 Edition](https://doc.rust-lang.org/edition-guide/rust-2024/index.html).

Changes are for the most part:
* formatting changes applied by `cargo fmt`,
* adopting the new requirement of `unsafe` blocks (and `SAFETY` comments) in `unsafe` functions.

No user-visible changes.

Increased versions of library crates.

---


mountpoint-s3-client/tests/auth.rs
```diff
@@ -119,9 +119,11 @@ async fn test_profile_provider_static_async() {
     .unwrap();
     writeln!(config_file, "aws_access_key_id = {}", credentials.access_key_id()).unwrap();
 
-    // Set up the environment variables to use this new config file. This is only OK to do because
-    // this test is run in a forked process, so won't affect any other concurrently running tests.
-    std::env::set_var("AWS_CONFIG_FILE", config_file.path().as_os_str());
+    // Set up the environment variables to use thiseii new config file.
```
renanmagagnin: Typo?

renanmagagnin (dismissed):

mountpoint-s3-crt-sys/src/logging_shim.rs
```diff
@@ -55,17 +55,20 @@ extern "C" {
...
     log_level: aws_log_level::Type,
     subject: aws_log_subject_t,
     body: *mut aws_string,
     body_length: usize,
 ) -> libc::c_int {
-    if let Some(f) = LOGGER_LOG_FN {
-        f(logger, log_level, subject, body, body_length)
-    } else {
-        AWS_OP_SUCCESS
+    // SAFETY: `LOGGER_LOG_FN` is only mutated in `try_init`.
```
muddyfish: Should this include that this can only be called after `try_init`?

mountpoint-s3-crt/src/lib.rs
```diff
@@ -73,24 +74,37 @@ pub(crate) unsafe fn aws_byte_cursor_as_slice<'a>(cursor: &aws_byte_cursor) -> &
 pub(crate) trait CrtError: Sized {
     type Return;
 
-    /// Safety: This must only be used immediately on a pointer returned from the CRT, with no other
+    /// # Safety
+    /// This must only be used immediately on a pointer returned from the CRT, with no other
```
muddyfish, line 78: Should this also mention that there shouldn't be any `await`s or similar points where a different task could be scheduled on the same thread/the current task gets rescheduled?

mountpoint-s3-fs/tests/common/fuse.rs
```diff
@@ -299,7 +299,7 @@ pub mod mock_session {
         }
     }
 
-    fn create_test_client(client: Arc<MockClient>, prefix: &str) -> impl TestClient {
+    fn create_test_client(client: Arc<MockClient>, prefix: &str) -> impl TestClient + use<> {
```
muddyfish, line 302: What does `use<>` mean?

passaro: It means that the `impl TestClient` that we return does not capture any generic parameters (the lifetime of the `prefix` reference in this case). See [reference on precise capturing](https://doc.rust-lang.org/reference/types/impl-trait.html#r-type.impl-trait.generic-capture.precise).

renanmagagnin (approved):

PR #1502, opened 2025-07-02
Introduce builder pattern for mockclientconfig

c-hagem:
Use a builder pattern for MockClientConfig.

Does not impact existing behaviour as it only changes the way we build the structure.

---

### DIFF
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -70,54 +70,6 @@ pub struct MockClientConfig {
 }
 
 impl MockClientConfig {
-    /// Create a new builder for MockClientConfig
-    pub fn builder() -> MockClientConfigBuilder {
-        MockClientConfigBuilder::default()
-    }
-
-    /// Get the bucket name
-    pub fn bucket(&self) -> &str {
-        &self.bucket
-    }
-
-    /// Get the part size
-    pub fn part_size(&self) -> usize {
-        self.part_size
-    }
-
-    /// Get the unordered list seed
-    pub fn unordered_list_seed(&self) -> Option<u64> {
-        self.unordered_list_seed
-    }
-
-    /// Check if backpressure is enabled
-    pub fn enable_backpressure(&self) -> bool {
-        self.enable_backpressure
-    }
-
-    /// Get the initial read window size
-    pub fn initial_read_window_size(&self) -> usize {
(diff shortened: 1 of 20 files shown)


mountpoint-s3-client/src/mock_client.rs
```diff
@@ -61,18 +61,110 @@ static RAMP_BYTES: LazyLock<Vec<u8>> = LazyLock::new(|| ramp_bytes(0, RAMP_BUFFE
...
+}
+
+impl MockClientConfig {
+    /// Create a new builder for MockClientConfig
+    pub fn builder() -> MockClientConfigBuilder {
+        MockClientConfigBuilder::default()
+    }
+
+    /// Get the bucket name
+    pub fn bucket(&self) -> &str {
```
passaro: Are any of these method needed?

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -61,18 +61,110 @@ static RAMP_BYTES: LazyLock<Vec<u8>> = LazyLock::new(|| ramp_bytes(0, RAMP_BUFFE
...
+
+    /// Check if rename is enabled
+    pub fn enable_rename(&self) -> bool {
+        self.enable_rename
+    }
+}
+
+/// Builder for MockClientConfig
+#[derive(Debug, Default)]
+pub struct MockClientConfigBuilder {
```
passaro: This is exactly the same as `MockClientConfig`. Instead of a builder for the config, should we have the config **be** the builder for `MockClient`?

Something like:
```rust
let mock_client: MockClient = MockClient::config()
                .bucket(...)
                .part_size(...)
                .unordered_list_seed(...)
                .build();
```

or just keep `MockClientConfig` , e.g.:

```rust
let config = MockClientConfig::default()
                .bucket(...)
                .part_size(...)
                .unordered_list_seed(...);
let client = MockClient(config);
```

I like the first option, but the second may be more practical: it should not require too many changes.

c-hagem: Went with the first suggestion, should have updated all usages

mountpoint-s3-client/src/mock_client.rs
c-hagem: Nope, removed them

passaro (dismissed):

PR #1503, opened 2025-07-04
Extend prefetcher benchmarks to test multiple objects

sahityadg:
With this change, we can benchmark concurrent downloads of multiple objects at prefetcher.

No, only extends prefetch benchmarks. 

No

---

### DIFF
--- a/mountpoint-s3-fs/examples/prefetch_benchmark.rs
+++ b/mountpoint-s3-fs/examples/prefetch_benchmark.rs
@@ -208,7 +208,12 @@ fn make_s3_client_from_args(args: &CliArgs) -> anyhow::Result<S3CrtClient> {
         client_config = client_config.part_size(part_size as usize);
     }
     if let Some(interfaces) = &args.bind {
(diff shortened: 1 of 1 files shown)


mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -104,71 +109,81 @@ pub struct CliArgs {
...
-                        }
-                    })
-                });
+            let mut download_tasks = Vec::new();
+
+            for (object_id, size) in &object_metadata {
+                for _ in 0..args.downloads_per_object {
+                    let received_bytes = received_bytes.clone();
+                    let object_id = object_id.clone();
+                    let manager = Prefetcher::default_builder(client.clone()).build(
```
c-hagem, line 156: Is it correct to create the Prefetcher here? Shouldn't we have one Prefetcher the entire iteration?

sahityadg: Yeah, we should have one per iteration.

PR #1504, opened 2025-07-08
Extend prefetcher benchmarks to test multiple object downloads

sahityadg:
With this change, we can benchmark concurrent downloads of multiple objects at prefetcher.

This change also allows passing NICs as a comma separated list and adds a new parameter to limit the run time of the test 

No, only extends prefetch benchmarks.

No

---

### DIFF
--- a/mountpoint-s3-fs/examples/prefetch_benchmark.rs
+++ b/mountpoint-s3-fs/examples/prefetch_benchmark.rs
@@ -1,3 +1,4 @@
+use std::error::Error;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::thread;
@@ -12,7 +13,7 @@ use mountpoint_s3_client::{ObjectClient, S3CrtClient};
 use mountpoint_s3_fs::Runtime;
 use mountpoint_s3_fs::mem_limiter::MemoryLimiter;
 use mountpoint_s3_fs::object::ObjectId;
-use mountpoint_s3_fs::prefetch::{Prefetcher, PrefetcherConfig};
+use mountpoint_s3_fs::prefetch::{PrefetchGetObject, Prefetcher, PrefetcherConfig};
 use sysinfo::{RefreshKind, System};
 use tracing_subscriber::EnvFilter;
 use tracing_subscriber::fmt::Subscriber;
@@ -39,11 +40,11 @@ fn init_tracing_subscriber() {
     about = "Run workloads against the prefetcher component of Mountpoint. Fetched data is discarded."
 )]
 pub struct CliArgs {
-    #[clap(help = "S3 bucket name containing the S3 object to fetch")]
+    #[clap(help = "S3 bucket name containing the S3 objects to fetch")]
     pub bucket: String,
 
-    #[clap(help = "S3 object key to fetch")]
-    pub s3_key: String,
+    #[clap(help = "List of S3 object keys to fetch", num_args = 1.., value_delimiter = ',')]
+    pub s3_keys: Vec<String>,
 
     #[clap(
         long,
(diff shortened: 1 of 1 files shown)


mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -193,7 +264,12 @@ fn make_s3_client_from_args(args: &CliArgs) -> anyhow::Result<S3CrtClient> {
         client_config = client_config.part_size(part_size as usize);
     }
     if let Some(interfaces) = &args.bind {
-        client_config = client_config.network_interface_names(interfaces.clone());
+        let nics: Vec<String> = interfaces
```
c-hagem, line 267: Should we verify that this is the expected number of NICs?

mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -93,8 +94,20 @@ pub struct CliArgs {
...
+    )]
+    runtime: Option<u64>,
+
+    #[arg(
+        long,
+        help = "Number of concurrent downloads per object",
+        default_value_t = 1,
+        value_name = "N"
+    )]
+    downloads_per_object: usize,
```
c-hagem, line 110: Do we anticipate to use this with N != 1?

c-hagem (approved):

mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -104,79 +117,137 @@ pub struct CliArgs {
...
+    let mem_limiter = create_memory_limiter(&args, &client);
     let runtime = Runtime::new(client.event_loop_group());
 
-    for iteration in 0..args.iterations {
+    // Verify if all objects exist and collect metadata
+    let object_metadata: Vec<(ObjectId, u64)> = args
+        .s3_keys
+        .iter()
+        .map(|key| {
+            let head_result = block_on(client.head_object(bucket, key, &HeadObjectParams::new()))
```
muddyfish, line 147: Ideally we'd do this concurrently rather than sequentially

mountpoint-s3-fs/examples/prefetch_benchmark.rs
sahityadg: Not that I'm aware of. I want to remove it in a subsequent PR

mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -104,79 +117,137 @@ pub struct CliArgs {
...
+                            runtime_exceeded_clone,
+                        ));
+                        if let Ok(bytes_read) = result {
+                            received_bytes.fetch_add(bytes_read, Ordering::SeqCst);
+                        } else {
+                            // As object download failures can produce
+                            // misleading results, exit the benchmarks
+                            // to avoid confusion.
+                            eprintln!("Download failed: {:?}", result.err());
+                            eprintln!("Existing benchmarks due to download failure");
```
sahityadg: typo: Exiting

mountpoint-s3-fs/examples/prefetch_benchmark.rs
sahityadg: The client creation will fail when trying to bind to non-existent NICs. As that's taken care of, I don't think we  need to restrict the number of NICs.

mountpoint-s3-fs/examples/prefetch_benchmark.rs
sahityadg: Agree, I will follow up in a separate CR as that's not the most time taking operation for the tests we are interested in.

muddyfish (approved):

PR #1506, opened 2025-07-09
[Experimental] Support mounting multiple buckets with CSV manifest

vladem:
Implement `ManifestMetablock` which uses SQLite db as a source of object metadata. This implementation also allows mounting multiple buckets within a single filesystem. This change is not intended to be applied to the regular `mount-s3` binary, only to `mount_from_config` example.

No, change is enabled in example only.

Yes, of the `mountpoint-s3-fs` crate.

---

### DIFF
--- a/mountpoint-s3-fs/src/manifest/db.rs
+++ b/mountpoint-s3-fs/src/manifest/db.rs
@@ -82,7 +82,7 @@ impl Db {
     pub fn select_entry_by_id(&self, id: u64) -> Result<Option<DbEntry>> {
         let start = Instant::now();
         let conn = self.conn.lock().expect("lock must succeed");
-        metrics::histogram!("manifest.query.duration_us", "query" => "lock").record(start.elapsed().as_micros() as f64);
+        metrics::histogram!("manifest.connection_lock.duration_us").record(start.elapsed().as_micros() as f64);
 
         let start = Instant::now();
         let query =
@@ -103,7 +103,7 @@ impl Db {
     pub fn select_entry(&self, parent_id: u64, name: &str) -> Result<Option<DbEntry>> {
         let start = Instant::now();
         let conn = self.conn.lock().expect("lock must succeed");
-        metrics::histogram!("manifest.query.duration_us", "query" => "lock").record(start.elapsed().as_micros() as f64);
+        metrics::histogram!("manifest.connection_lock.duration_us").record(start.elapsed().as_micros() as f64);
 
         let start = Instant::now();
         let query = "SELECT id, parent_id, channel_id, parent_partial_key, name, etag, size FROM s3_objects WHERE parent_id = ?1 AND name = ?2";
@@ -123,7 +123,7 @@ impl Db {
     pub fn select_children(&self, parent_id: u64, next_offset: usize, batch_size: usize) -> Result<Vec<DbEntry>> {
         let start = Instant::now();
         let conn = self.conn.lock().expect("lock must succeed");
-        metrics::histogram!("manifest.query.duration_us", "query" => "lock").record(start.elapsed().as_micros() as f64);
+        metrics::histogram!("manifest.connection_lock.duration_us").record(start.elapsed().as_micros() as f64);
 
         let start = Instant::now();
         let query = "SELECT id, parent_id, channel_id, parent_partial_key, name, etag, size FROM s3_objects WHERE parent_id = ?1 ORDER BY name LIMIT ?2 OFFSET ?3";
--- a/mountpoint-s3-fs/CHANGELOG.md
+++ b/mountpoint-s3-fs/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Introduce `Metablock` abstraction and adjust the interface between `S3Filesystem` and implementors of `Metablock`. ([#1500](https://github.com/awslabs/mountpoint-s3/pull/1500), [#1488](https://github.com/awslabs/mountpoint-s3/pull/1488))
 * Upgrade to Rust 2024. ([#1498](https://github.com/awslabs/mountpoint-s3/pull/1498))
+* Introduce `ManifestMetablock` under the `manifest` feature flag, which allows loading S3 metadata from disk. ([#1506](https://github.com/awslabs/mountpoint-s3/pull/1506))
 
 ## v0.5.0 (Jun 19, 2025)
 


mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -188,18 +164,18 @@ fn load_config<P: AsRef<Path>>(path: P) -> Result<ConfigOptions> {
...
+            "At lest one channel must be specified and the number of channels must not exceed 20"
+        ));
     }
 
-    let channel = &config.channels[0];
-    let csv_path = &channel.manifest_path;
     // Generate manifest path and check if it exists
     let db_path = database_directory.join("metadata.db");
-    info!("Creating manifest for channel {}", channel.directory_name);
+    info!("Creating manifest for channels {:?}", config.channels);
```
c-hagem: I think I put this info log line here originally, but are we really creating a manifest here?

vladem: ack, hopefully made the log line more accurate

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -104,113 +152,62 @@ impl Db {
...
-        conn.execute("CREATE UNIQUE INDEX idx_key ON s3_objects (key)", ())?;
 
-        conn.execute("CREATE INDEX idx_parent_key ON s3_objects (parent_key, key)", ())?;
+        conn.execute("CREATE UNIQUE INDEX idx_parent_id ON s3_objects (parent_id, name)", ())?;
 
         Ok(())
     }
 
     pub fn insert_batch(&self, entries: &[DbEntry]) -> Result<()> {
         let mut conn = self.conn.lock().expect("lock must succeed");
```
c-hagem, line 191: Do we have any indication whether there is contention for this lock? (Perhaps better to leave that for later)

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -104,113 +152,62 @@ impl Db {
...
-    #[test_case("dir1/", &["dir1"]; "ends with /")]
-    fn test_infer_directories(dir_key: &str, inferred_dirs: &[&str]) {
-        let inferred_dirs: Vec<_> = inferred_dirs
-            .iter()
-            .map(|key| DbEntry {
-                full_key: key.to_string(),
-                etag: None,
-                size: None,
-            })
+        // assume channel ids is a contiguous sequence of integers starting from 0
```
c-hagem: Should we check / assert that this is invariant is upheld?

mountpoint-s3-fs/src/manifest/manifest_impl.rs
```diff
@@ -0,0 +1,238 @@
...
+        Ok(entry)
+    }
+
+    pub fn readd(&mut self, entry: ManifestEntry) {
+        self.next_offset -= 1;
+
+        self.entries.push_front(entry);
+    }
+
+    // TODO: test the out of order readdir
```
c-hagem: We should have a test for that, (I guess fuse test at file system level) as this is a behaviour difference from other metablock implementations.

mountpoint-s3-fs/src/manifest/manifest_impl.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub fn seek(&mut self, offset: usize) -> Result<(), ManifestError> {
+        if offset != self.next_offset {
```
c-hagem: If I understand correctly, this implementation performs more discarding of results for seeks at small offsets from the last offset compared to the superblock one? Do we expect this to be a significant performance hit? [I.e. to re-read offset - 2, we completely re-stream] 

I still wonder, if we should keep some interval around the current offset and allow easy streaming around it, but perhaps this is a premature optimisation.

mountpoint-s3-fs/src/manifest/manifest_impl.rs
```diff
@@ -0,0 +1,238 @@
...
+            metrics::counter!("manifest.readdir.out_of_order").increment(1);
+            self.entries.clear();
+            self.next_offset = offset;
+            self.finished = false;
```
c-hagem: This could be true, if we seek behind the end? (Does it matter?)

mountpoint-s3-fs/src/manifest/manifest_impl.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    /// Load next batch of entries from the database
+    fn search_next_entries(&mut self) -> Result<(), ManifestError> {
+        let db_entries = self
+            .db
+            .select_children(self.parent_id, self.next_offset, self.batch_size)?;
+
+        if db_entries.len() < self.batch_size {
+            self.finished = true;
```
c-hagem: Could in theory also be true if `db_entries.len() == self.batch_size` I guess if that is the last entry (I don't think we need to care?)

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,347 @@
...
+    fn manifest_entry_to_lookup(&self, manifest_entry: ManifestEntry) -> Result<Lookup, ManifestError> {
+        let channel_id = match &manifest_entry {
+            ManifestEntry::File { channel_id, .. } => *channel_id,
+            ManifestEntry::Directory { channel_id, .. } => *channel_id,
+        } as usize;
+        if channel_id >= self.channels.len() {
+            return Err(ManifestError::InvalidRow(manifest_entry.get_full_key().to_string()));
+        }
+        let channel = self.channels[channel_id].clone();
+        let lookup = match manifest_entry {
```
c-hagem: Could this be a `try_into` or differently named method on the `ManifestEntry`?

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,347 @@
...
+                size,
+                id,
+                full_key,
+                ..
+            } => Lookup::new(
+                id,
+                self.stat_for_file(&etag, size),
+                InodeKind::File,
+                true,
+                Some(S3Location {
```
c-hagem: Consider using `S3Location::new`.

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,347 @@
...
+            .insert(readdir_handle_id, Arc::new(Mutex::new(readdir_handle)));
+        Ok(readdir_handle_id)
+    }
+
+    async fn readdir<'a>(
+        &self,
+        parent: InodeNo,
+        fh: u64,
+        mut offset: i64,
+        _is_readdirplus: bool,
```
c-hagem, line 133: Maybe doc-comment why this is unused.

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,347 @@
...
+        mut replier: TryAddDirEntry<'a>,
+    ) -> Result<(), InodeError> {
+        // serve '.' and '..' entries
+        if offset < 1 {
```
c-hagem, line 147: I know we do this in superblock as well, but I wonder if logically equivalent `offset == 0` would be clearer.

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,347 @@
...
+            if replier(
+                InodeInformation::new(parent, self.stat_for_directory(), InodeKind::Directory, true),
+                ".".into(),
+                offset + 1,
```
c-hagem, line 151: Should we consider (for later perhaps) to change API to do the +1 conversion somewhere else?

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,347 @@
...
+        Ok(())
+    }
+
+    async fn releasedir(&self, fh: u64) -> Result<(), InodeError> {
+        self.readdir_handles.write().expect("lock must succeed").remove(&fh);
+        Ok(())
+    }
+
+    async fn start_reading(&self, _ino: InodeNo) -> Result<(), InodeError> {
+        // Assume getattr was just called to check for inode existence, so this is a no-op
```
c-hagem, line 194: Should we validate that this inode exists?

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,347 @@
...
+        Ok(())
+    }
+
+    async fn forget(&self, _ino: InodeNo, _n: u64) {
+        // Inodes are kept on disk for the lifetime of a mount (for feature lookup-s), so this is a no-op
+    }
+
+    async fn create(&self, dir: InodeNo, _name: &OsStr, _kind: InodeKind) -> Result<Lookup, InodeError> {
+        // For a read-only view, don't allow creation
+        Err(InodeError::InodeNotWritable(InodeErrorInfo {
```
c-hagem, line 209: Maybe we should have custom error that maps to `EPERM` here?

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,347 @@
...
+        }))
+    }
+
+    async fn start_writing(&self, ino: InodeNo, _mode: &WriteMode, _is_truncate: bool) -> Result<(), InodeError> {
+        // For a read-only view, don't allow writing
+        Err(InodeError::InodeNotWritable(InodeErrorInfo {
+            ino,
+            key: "".into(),
+            bucket: None,
+        })) // TODO: lookup the key?
```
c-hagem: I think we should distinguish between inode does not exist and exists -> we give information about the key (otherwise our error log may become confusing I fear).

c-hagem: Maybe we should re-view the error handling for these not implemented methods in general.

mountpoint-s3-fs/src/metablock/lookup.rs
```diff
@@ -16,6 +16,12 @@ pub struct Lookup {
 impl Lookup {
     /// Creates a new Lookup instance
     pub fn new(ino: InodeNo, stat: InodeStat, kind: InodeKind, is_remote: bool, location: Option<S3Location>) -> Self {
+        debug_assert!(
```
c-hagem, line 23: Do we still need this?

c-hagem (commented):
Nice! 

Added some comments, don't think a lot of them are blocking apart from reviewing the error handling and perhaps adding a test for out of order readdir (sorry, if didn't see that test)?

mountpoint-s3-fs/src/metablock/path.rs
```diff
@@ -118,6 +130,23 @@ impl From<ValidKey> for String {
     }
 }
 
+impl TryFrom<String> for ValidKey {
+    type Error = ValidKeyError;
+
+    /// Constructs a valid key performing checks.
+    fn try_from(value: String) -> Result<Self, Self::Error> {
```
passaro: Can we add tests for this? How does it handle strings ending in "/"?

Also, consider inlining the `validate` method, especially if it helps avoiding the additional scan to get `name_len`.

mountpoint-s3-fs/src/manifest/mod.rs
passaro, line 1: We prefer `manifest.rs` rather than `manifest/mod.rs`. Also, why move the types into `manifest_impl`?

mountpoint-s3-fs/src/manifest/manifest_impl.rs
vladem: this `self.finished` flag must must be set conservatively here: it's just an optimisation allowing us not to query db, when we **know** that there are no more entries, but when `db_entries.len() == self.batch_size` there may be something left

mountpoint-s3-fs/src/manifest/manifest_impl.rs
vladem: in that case we will query the db, get in response `0` entries and set the flag; we don't know the size of list upfront, so querying the db to find that out is the only way

mountpoint-s3-fs/src/manifest/metablock.rs
vladem: we currently call `start_reading` while serving the `open` fuse request; in the same function, just before creating the read handle, we [call](https://github.com/awslabs/mountpoint-s3/blob/fa6203366755e0009fe9b962cc21393999dc0b4a/mountpoint-s3-fs/src/fs.rs#L382) `getattr` on the metablock, which will check for inode existence 

I don't think it's necessary to repeat this check, especially taking into the account that it means an extra query to the database

mountpoint-s3-fs/src/manifest/mod.rs
vladem: > Also, why move the types into manifest_impl?

I find it convenient to have all public types / methods not mixed with private and to be concisely re-exported like this:

```rust
pub use builder::{ChannelConfig, ChannelManifest, create_db, ingest_manifest};
pub use csv_reader::CsvReader;
pub use db::DbEntry;
pub use manifest_impl::{Manifest, ManifestError};
pub use metablock::ManifestMetablock;
```

I guess it just somehow resembles `.h` files. I see that we still need to look into `manifest_impl.rs` to find public methods, of `Manifest` for example. This defeats the purpose.

Just put a thumbs up to this and I'll inline the whole `manifest_impl.rs` to `manifest.rs`?

mountpoint-s3-fs/src/manifest/metablock.rs
c-hagem: Yes -- let's not query the database twice for nothing.

mountpoint-s3-fs/src/metablock/lookup.rs
vladem: I think it's an important invariant that `kind` provided as an argument matches the one inferred from the `location.partial_key`, this assert doesn't cost anything in release build, so I'd rather keep it

mountpoint-s3-fs/src/manifest/metablock.rs
vladem: > I think we should distinguish between inode does not exist and exists 

Why does it matter for not implemented methods?

> Maybe we should re-view the error handling for these not implemented methods in general.

Agree, how about this?

```rust
pub enum InodeError {
    ...
    #[error("mutating operations are now allowed on a read-only metablock")]
    ReadonlyMetablock, // maps to libc::EPERM
}
```

mountpoint-s3-fs/src/manifest/metablock.rs
vladem: While the type is `offset: i64`, I think [this](https://github.com/libfuse/libfuse/blob/4166f2eb97da4e25a516abee3d6fe13b9ed77bc6/include/fuse_lowlevel.h#L750) gives us enough confidence that `offset >= 0`:

```c
	/**
	 * Read directory
	 ...
	 * If off_t is non-zero, then it will correspond to one of the off_t
	 * values that was previously returned by readdir() for the same
	 * directory handle. In this case, readdir() should skip over entries
```

So I have no objection against using `offset == 0` and `offset == 1` for clarity. I would insist on consistency with code in `Superblock` though. Should I change this in both places?

And I would add this check:

```rust
        if offset < 0 {
            return Err(InodeError::OutOfOrderReadDir { expected: 0, actual: offset, fh });
        }
```

mountpoint-s3-fs/src/manifest/metablock.rs
vladem: could be useful, a thing to keep in mind here is that we return offsets starting from `1` (`.` has offset `1`), which is counterintuitive but apparently required

mountpoint-s3-fs/src/metablock/path.rs
vladem: note that `ValidKey::validate` is used in `builder.rs`

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,286 @@
...
+    fn stat_for_directory(&self) -> InodeStat {
+        const NEVER_EXPIRE_TTL: Duration = Duration::from_secs(200 * 365 * 24 * 60 * 60);
+        InodeStat::for_directory(self.mount_time, NEVER_EXPIRE_TTL)
+    }
+}
+
+#[async_trait]
+impl Metablock for ManifestMetablock {
+    async fn lookup(&self, parent_ino: InodeNo, name: &OsStr) -> Result<Lookup, InodeError> {
+        let Some(name) = name.to_str().map(String::from) else {
```
vladem: `let name: ValidName = name.try_into()?;`

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,286 @@
...
+            return Err(InodeError::InvalidFileName(name.to_os_string()));
+        };
+        let Some(manifest_entry) = self.manifest.manifest_lookup(parent_ino, &name)? else {
+            return Err(InodeError::FileDoesNotExist(
+                name.to_string(),
+                InodeErrorInfo {
+                    ino: parent_ino,
+                    key: name.into(),
```
vladem: empty string and todo (review `InodeErrorInfo`)

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        // load entries from the manifest
+        let Some(readdir_handle) = self
+            .readdir_handles
+            .read()
+            .expect("lock must succeed")
+            .get(&fh)
+            .cloned()
+        else {
+            return Err(InodeError::NoSuchDirHandle { fh });
```
vladem: do this check before the `.` and `..`

mountpoint-s3-fs/src/manifest/manifest_impl.rs
```diff
@@ -0,0 +1,333 @@
...
+            None => Ok(None),
+        }
+    }
+
+    /// Create an iterator over directory's direct children
+    pub fn iter(&self, parent_id: u64) -> ManifestIter {
+        ManifestIter::new(self.db.clone(), parent_id)
+    }
+
+    pub fn channels(&self) -> Result<Vec<S3Path>, ManifestError> {
```
vladem: load_channels

mountpoint-s3-fs/src/manifest/manifest_impl.rs
```diff
@@ -0,0 +1,333 @@
...
+        // search for an entry and validate it
+        let db_entry = self.db.select_entry_by_id(ino)?;
+        match db_entry {
+            Some(db_entry) => Ok(Some(db_entry.try_into()?)),
+            None => Ok(None),
+        }
+    }
+
+    /// Create an iterator over directory's direct children
+    pub fn iter(&self, parent_id: u64) -> ManifestIter {
```
vladem: dir_iter

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -1,39 +1,46 @@
...
 use std::sync::{Arc, Mutex};
 use std::time::Instant;
 
-#[derive(Debug, Clone, PartialEq, Hash, Eq, Deserialize)]
+use crate::manifest::ManifestError;
+use crate::prefix::Prefix;
+use crate::s3::config::S3Path;
+
+#[derive(Debug, Clone, PartialEq, Hash, Eq, Deserialize, Default)]
 pub struct DbEntry {
```
vladem, line 17: separate struct for CSV?

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -88,13 +124,26 @@ impl Db {
 
     pub fn create_table(&self) -> Result<()> {
         let conn = self.conn.lock().expect("lock must succeed");
+        // NOTE: SUBSTR in SQLite starts indexing from 1 (name_offset column assumes left-most character has index 0)
         conn.execute(
             "CREATE TABLE s3_objects (
                 id          INTEGER   PRIMARY KEY,
                 key         TEXT      NOT NULL,
```
vladem: `parent_key` instead

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -88,13 +124,26 @@ impl Db {
...
-                parent_key  TEXT      NOT NULL,
+                name_offset INTEGER   NOT NULL,
+                name        TEXT      GENERATED ALWAYS AS (SUBSTR(key,name_offset + 1)) VIRTUAL,
```
vladem: actual string instead of generated

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,318 @@
...
+
+    let mut next_id = FUSE_ROOT_ID + 1;
+    for (channel_id, channel_manifest) in channel_manifests.into_iter().enumerate() {
+        let mut dir_ids: HashMap<String, u64> = Default::default(); // TODO: limit size of this hash map
+        let mut buffer = Vec::with_capacity(batch_size);
+
+        for entry in channel_manifest.entries {
+            // parse next entry and validate it
+            let mut entry = entry?;
+            match validate_db_entry(&entry) {
```
vladem: validate as part of conversion to DbEntry

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,318 @@
...
 
-/// Ingests a manifest into the database
-pub fn ingest_manifest(csv_path: &Path, db_path: &Path) -> Result<(), ManifestError> {
-    let file = File::open(csv_path).map_err(|err| ManifestError::CsvOpenError(csv_path.to_path_buf(), err))?;
-    let csv_reader = CsvReader::new(BufReader::new(file));
-    if db_path.exists() {
-        return Err(ManifestError::DbExists);
+/// Ingests directory `dir_key` and its parents to db (if not done yet), returns the ID of the `dir_key`
+fn ensure_dirs_inserted(
+    db: &Db,
```
vladem: use a helper struct to avoid passing db, dir_ids, next_id

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,318 @@
...
+    dir_key: &str,
+    channel_id: u64,
+) -> Result<u64, ManifestError> {
+    let mut insert_buffer = Vec::new();
+    let mut dir_key_len = 0;
+    let mut parent_id = FUSE_ROOT_ID;
+
+    for component in dir_key.split('/') {
+        dir_key_len += component.len() + 1; // includes the trailing '/'
+        let directory_key = &dir_key[..dir_key_len - 1];
```
vladem: use `ValidKey` to build parent keys

mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -193,19 +169,19 @@ fn load_config<P: AsRef<Path>>(path: P) -> Result<ConfigOptions> {
 
 /// Processes a manifest and stores database in `database_directory`
 fn process_manifests(config: &ConfigOptions, database_directory: &Path) -> Result<Manifest> {
-    // Error if no channels
-    if config.channels.len() != 1 {
-        return Err(anyhow!("Exactly one channel must be configured"));
+    // TODO: fail fast if buckets are in different regions?
+    if config.channels.is_empty() || config.channels.len() > 20 {
+        return Err(anyhow!(
+            "At lest one channel must be specified and the number of channels must not exceed 20"
```
passaro: typo: least

mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -193,19 +169,19 @@ fn load_config<P: AsRef<Path>>(path: P) -> Result<ConfigOptions> {
...
+        ));
     }
 
-    let channel = &config.channels[0];
-    let csv_path = &channel.manifest_path;
     // Generate manifest path and check if it exists
     let db_path = database_directory.join("metadata.db");
-    info!("Creating manifest for channel {}", channel.directory_name);
+    info!("Ingesting CSV manifests to the DB, channels {:?}", config.channels);
```
passaro: nit: into the DB

passaro: let's also review whether we use "DB", "database", or something more specific, e.g. "metadata store"?

mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -222,26 +198,26 @@ fn mount_filesystem(
...
-    let s3_path = config.build_s3_path()?;
-
     // Create the client and runtime
     let client_config = config.build_client_config()?;
     let client = client_config
         .create_client(None)
         .context("Failed to create S3 client")?;
     let runtime = Runtime::new(client.event_loop_group());
 
+    let metablock = Box::new(mountpoint_s3_fs::manifest::ManifestMetablock::new(manifest)?);
```
passaro: nit: import type

mountpoint-s3-fs/src/fs.rs
```diff
@@ -147,19 +145,11 @@ where
         client: Client,
         prefetch_builder: PrefetcherBuilder<Client>,
         runtime: Runtime,
-        bucket: &str,
-        prefix: &Prefix,
+        metablock: Box<dyn Metablock>,
```
passaro: Here and above, if possible, take a `impl Metablock` and box (or rather "arc") it when setting the field.

mountpoint-s3-fs/examples/fs_benchmark.rs
```diff
@@ -161,14 +161,16 @@ fn mount_file_system(
...
-        prefetcher_builder,
-        runtime,
+    let superblock = Box::new(Superblock::new(
+        client.clone(),
         bucket_name,
         &Default::default(),
-        filesystem_config,
-    );
+        SuperblockConfig {
+            cache_config: filesystem_config.cache_config.clone(),
```
passaro, line 169: Should we take these fields out of file system config? Happy to do it later though.

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,392 @@
...
+    etag: String,
+    size: usize,
+}
+
+impl InputManifestEntry {
+    pub fn new(partial_key: &str, etag: &str, size: usize) -> Result<Self, ManifestError> {
+        Self::from_owned(partial_key.to_string(), etag.to_string(), size)
+    }
+
+    pub fn from_owned(partial_key: String, etag: String, size: usize) -> Result<Self, ManifestError> {
```
passaro: Can't we merge these by using `impl Into<String>`?

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,392 @@
...
+    ) -> Result<DbEntry, ManifestError> {
+        debug_assert_eq!(self.partial_key.kind(), InodeKind::File);
+
+        // parse and validate the name
+        let name = match self.partial_key.rsplit_once('/') {
+            Some((_, name)) => name.to_string(),
+            None => self.partial_key.to_string(),
+        };
+        let _ = ValidName::parse_str(&name)
+            .map_err(|_| ManifestError::InvalidKey(ValidKeyError::InvalidKey(name.clone())))?;
```
passaro: Can't we just get the name?
```suggestion
        let name = self.partial_key.name().to_string();
```

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,392 @@
...
+#[derive(Debug, Deserialize)]
+pub struct ChannelConfig {
+    pub directory_name: String,
+    pub bucket_name: String,
+    #[serde(default)]
+    pub prefix: String,
+    pub manifest_path: PathBuf,
+}
+
+pub struct ChannelManifest<I> {
```
passaro: `I` -> `EntriesIterator`

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,392 @@
-use super::ManifestError;
+use std::collections::HashSet;
+use std::io::BufReader;
+use std::path::{Path, PathBuf};
+use std::{collections::HashMap, fs::File};
+
+use fuser::FUSE_ROOT_ID;
```
passaro: nit: we already import this constant somewhere (metablock?). Let's use it from there.

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,392 @@
...
-    Ok(())
-}
+        let components: Vec<_> = object_key.split('/').collect();
+        let mut parent_id = channel_root_id;
+        let mut parent_partial_key = ValidKey::root();
+
+        for component in components[..components.len() - 1].iter() {
+            let valid_name = ValidName::parse_str(component)
+                .map_err(|_| ManifestError::InvalidKey(ValidKeyError::InvalidKey(object_key.to_string())))?;
+            let partial_key = parent_partial_key.new_child(valid_name, InodeKind::Directory)?;
```
passaro: If `object_key` is valid, aren't all its components also valid?

mountpoint-s3-fs/src/manifest/csv_reader.rs
```diff
@@ -17,10 +20,32 @@ impl<R: Read> CsvReader<R> {
...
+        match csv_entry {
+            Ok(csv_entry) => Some(csv_entry.try_into()),
+            Err(err) => Some(Err(err)),
+        }
+    }
+}
+
+#[derive(Deserialize)]
+struct CsvEntry {
+    partial_key: String, // may hold a folder marker (e.g. "dir1/dir2/")
```
passaro: Directory markers are not accepted, right? Do you want to document that they may be present here and only rejected later on? I'd be a bit more explicit.

mountpoint-s3-fs/src/manifest/csv_reader.rs
```diff
@@ -17,10 +20,32 @@ impl<R: Read> CsvReader<R> {
...
 impl<R: Read> Iterator for CsvReader<R> {
-    type Item = Result<DbEntry, ManifestError>;
+    type Item = Result<InputManifestEntry, ManifestError>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.reader.next().map(|item| item.map_err(ManifestError::from))
+        let csv_entry = self
+            .reader
+            .next()
+            .map(|csv_entry| csv_entry.map_err(ManifestError::from))?;
```
passaro: `.map` here and `match` below are confusing. Consolidate on a single `match`

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -1,39 +1,36 @@
...
-    }
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DbEntry {
+    pub id: u64,
+    pub parent_id: u64,
+    pub channel_id: usize,
+    pub parent_partial_key: Option<String>, // not set for synthetic channel dirs
+    pub name: String,
+    pub etag: Option<String>, // not set for all dirs
+    pub size: Option<usize>,  // not set for all dirs
```
passaro: nit: "not set for dirs"

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -1,39 +1,36 @@
...
-        } else {
-            &key[..key.len() - last_component_len - 1]
-        }
-    }
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DbEntry {
+    pub id: u64,
+    pub parent_id: u64,
+    pub channel_id: usize,
+    pub parent_partial_key: Option<String>, // not set for synthetic channel dirs
```
passaro: When set, does it include the end "/"?

passaro: Also, why not use rustdocs?

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -88,13 +114,25 @@ impl Db {
 
     pub fn create_table(&self) -> Result<()> {
         let conn = self.conn.lock().expect("lock must succeed");
+        // NOTE: SUBSTR in SQLite starts indexing from 1 (name_offset column assumes left-most character has index 0)
```
passaro: leftover comment?

mountpoint-s3-fs/src/manifest/manifest_impl.rs
```diff
@@ -0,0 +1,318 @@
...
+    #[error("database exists")]
+    DbExists,
+    #[error("error opening manifest file at '{0}'")]
+    CsvOpenError(PathBuf, #[source] io::Error),
+    #[error("database error")]
+    DbError(#[from] rusqlite::Error),
+    #[error("key is invalid")]
+    InvalidKey(#[from] ValidKeyError),
+    #[error("folder marker {0}")]
+    FolderMarker(String),
```
passaro: directory marker

mountpoint-s3-fs/src/manifest/manifest_impl.rs
```diff
@@ -0,0 +1,318 @@
...
+
+use crate::metablock::{InodeInformation, InodeStat, Lookup, S3Location, ValidKey, ValidKeyError, ValidName};
+use crate::prefix::PrefixError;
+use crate::s3::config::S3Path;
+
+const NEVER_EXPIRE_TTL: Duration = Duration::from_secs(200 * 365 * 24 * 60 * 60);
+
+#[derive(Debug, Error)]
+pub enum ManifestError {
+    #[error("database exists")]
```
passaro: rephrase to make it clearer what is the error

mountpoint-s3-fs/src/manifest/manifest_impl.rs
```diff
@@ -0,0 +1,318 @@
...
+use super::db::{Db, DbEntry};
+
+use crate::metablock::{InodeInformation, InodeStat, Lookup, S3Location, ValidKey, ValidKeyError, ValidName};
+use crate::prefix::PrefixError;
+use crate::s3::config::S3Path;
+
+const NEVER_EXPIRE_TTL: Duration = Duration::from_secs(200 * 365 * 24 * 60 * 60);
+
+#[derive(Debug, Error)]
+pub enum ManifestError {
```
passaro: Should we split out ingestion errors from read errors? It may help getting to more user-friendly error messages.

mountpoint-s3-fs/src/manifest/manifest_impl.rs
```diff
@@ -0,0 +1,318 @@
...
+use time::OffsetDateTime;
+use tracing::error;
+
+use super::db::{Db, DbEntry};
+
+use crate::metablock::{InodeInformation, InodeStat, Lookup, S3Location, ValidKey, ValidKeyError, ValidName};
+use crate::prefix::PrefixError;
+use crate::s3::config::S3Path;
+
+const NEVER_EXPIRE_TTL: Duration = Duration::from_secs(200 * 365 * 24 * 60 * 60);
```
passaro: We should define this in the `expiry` module and reuse.

mountpoint-s3-fs/src/manifest/manifest_impl.rs
```diff
@@ -0,0 +1,318 @@
...
+                name: db_entry.name,
+                etag,
+                size,
+            }),
+            _ => Err(ManifestError::InvalidRow(db_entry.id)),
+        }
+    }
+}
+
+/// Manifest of all available objects in the bucket
```
passaro: Review rustdoc: why "all available"? multiple buckets?

mountpoint-s3-fs/src/manifest/manifest_impl.rs
```diff
@@ -0,0 +1,318 @@
...
+        ManifestIter::new(self.db.clone(), parent_id)
+    }
+
+    pub fn load_channels(&self) -> Result<Vec<S3Path>, ManifestError> {
+        self.db.load_channels()
+    }
+}
+
+#[derive(Debug)]
+pub struct ManifestIter {
```
passaro: Maybe `ManifestDirIter`

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,392 @@
...
+use std::{collections::HashMap, fs::File};
+
+use fuser::FUSE_ROOT_ID;
+use serde::Deserialize;
+
+use crate::fs::InodeKind;
+use crate::metablock::{ValidKey, ValidKeyError, ValidName};
+use crate::prefix::Prefix;
+use crate::s3::config::S3Path;
 use crate::{
```
passaro: Here and in other sub-modules: prefer `super` rather than `crate::manifest`.

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,286 @@
...
+            next_dir_handle_id: Default::default(),
+            readdir_handles: Default::default(),
+        })
+    }
+
+    fn manifest_entry_to_lookup(&self, manifest_entry: ManifestEntry) -> Result<Lookup, ManifestError> {
+        let channel_id = match &manifest_entry {
+            ManifestEntry::File { channel_id, .. } => *channel_id,
+            ManifestEntry::Directory { channel_id, .. } => *channel_id,
+        };
```
passaro: This calls for a `channel_id()` method on `ManifestEntry`.

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        match manifest_entry {
+            ManifestEntry::File { parent_id, .. } => Ok(parent_id),
+            ManifestEntry::Directory { parent_id, .. } => Ok(parent_id),
+        }
+    }
+
+    fn stat_for_directory(&self) -> InodeStat {
+        const NEVER_EXPIRE_TTL: Duration = Duration::from_secs(200 * 365 * 24 * 60 * 60);
```
passaro: See previous comment about this constant.

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -0,0 +1,286 @@
...
+        }
+
+        // load entries from the manifest
+        let mut readdir_handle = readdir_handle.lock().expect("lock must succeed");
+        let shifted_offset = (offset - 2) as usize; // shift offset accounting for '.' and '..'
+        readdir_handle.seek(shifted_offset)?; // typically no-op, but required for out-of-order requests
+        loop {
+            let Some(manifest_entry) = readdir_handle.next_entry()? else {
+                break;
+            };
```
passaro: nit:
```suggestion
        while let Some(manifest_entry) = readdir_handle.next_entry()? {
```

mountpoint-s3-fs/src/metablock/lookup.rs
passaro: What about `new_from_info_and_loc`?

mountpoint-s3-fs/src/metablock/path.rs
```diff
@@ -118,6 +120,33 @@ impl From<ValidKey> for String {
     }
 }
 
+impl TryFrom<String> for ValidKey {
+    type Error = ValidKeyError;
+
+    /// Constructs a valid key performing checks.
+    fn try_from(full_key: String) -> Result<Self, Self::Error> {
+        // validate
+        let components: Vec<_> = full_key.split_terminator('/').collect();
```
passaro: nit: skip the vector, just save the length of the last component

mountpoint-s3-fs/src/manifest/builder.rs
vladem: that is true, `ValidName::parse_str(component)` is technically redundant here

but this is with the assumption that splitting is done right, I can simplify if we add the following method to `ValidKey`:

```rust
    /// Path components for this key.
    ///
    /// For directories, this does not include empty component after the terminal '/'.
    pub fn components(&self) -> Vec<ValidName> {
        self.key.trim_end_matches('/').split('/').map(ValidName).collect()
    }
```

mountpoint-s3-fs/src/manifest/csv_reader.rs
vladem: skipped later on, not rejected, will document in a comment

mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -222,26 +201,26 @@ fn mount_filesystem(
     error_logger: impl ErrorLogger + Send + Sync + 'static,
 ) -> Result<FuseSession> {
     // Create the Mountpoint configuration
+    let fs_config = config.build_filesystem_config()?;
     let mp_config = MountpointConfig::new(
         config.build_fuse_session_config()?,
-        config.build_filesystem_config(manifest)?,
+        fs_config.clone(),
```
passaro: Can we remove the clone? And maybe just inline?

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -1,49 +1,73 @@
...
+    pub id: u64,
+    /// Identifier of the parent directory entry.
+    pub parent_id: u64,
+    /// Identifier of the S3 channel (bucket+prefix combination) this entry belongs to.
+    pub channel_id: usize,
+    /// Partial S3 key of the parent directory, not set for synthetic channel directories.
+    ///
+    /// Does not include S3 prefix when prefix is mounted. Always includes the trailing '/'.
+    ///
+    /// This field allows to reconstruct the full S3 key of the given entry without retrieving it's parent.
```
passaro: typo:
```suggestion
    /// This field allows to reconstruct the full S3 key of the given entry without retrieving its parent.
```

mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -193,19 +169,22 @@ fn load_config<P: AsRef<Path>>(path: P) -> Result<ConfigOptions> {
 
 /// Processes a manifest and stores database in `database_directory`
 fn process_manifests(config: &ConfigOptions, database_directory: &Path) -> Result<Manifest> {
-    // Error if no channels
-    if config.channels.len() != 1 {
-        return Err(anyhow!("Exactly one channel must be configured"));
+    // TODO: fail fast if buckets are in different regions?
```
passaro: remove todo

mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -193,19 +169,22 @@ fn load_config<P: AsRef<Path>>(path: P) -> Result<ConfigOptions> {
...
+    if config.channels.is_empty() || config.channels.len() > 20 {
```
passaro: Consider not enforcing the upper limit

mountpoint-s3-fs/src/fs/config.rs
```diff
@@ -2,16 +2,14 @@ use std::time::Duration;
...
-use crate::manifest::Manifest;
 use crate::mem_limiter::MINIMUM_MEM_LIMIT;
 use crate::metablock::WriteMode;
 use crate::prefetch::PrefetcherConfig;
 use crate::s3::S3Personality;
 
 use super::{ServerSideEncryption, TimeToLive};
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
passaro: nit: double-check if still required, remove if not

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,410 @@
...
+use super::{
+    CsvReader,
+    db::{Db, DbEntry},
 };
-use std::fs::File;
-use std::io::BufReader;
-use std::path::Path;
 
-pub fn create_db(
+/// [InputManifestError] represents errors occurring during the creation of metadata database.
```
passaro: tiny nit: here and in other docs, the self-link it not very useful.

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,410 @@
...
+    InvalidPrefix(#[from] PrefixError),
+    #[error("failed to write to the database")]
+    DbError(#[from] rusqlite::Error),
+    #[error("s3 key provided in the csv manifest is invalid")]
+    InvalidKey(#[from] ValidKeyError),
+}
+
+/// [ChannelConfig] represents per-channel configuration, when multiple buckets are mounted.
+///
+/// This struct is a part of the CLI interface and contains a path to the csv manifest: [manifest_path].
```
passaro: This line is confusing: we don't expose the manifest to the CLI at the moment. Remove.

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,410 @@
...
+            name: self.partial_key.name().to_string(),
+            etag: Some(self.etag),
+            size: Some(self.size),
+        }
+    }
+}
+
+/// Ingests a manifest into the database.
+///
+/// The expected file format is CSV with no header and 3 columns -- partial_key, etag, size.
```
passaro: Let's mention that these requirements apply to each CSV file provided in `channel_configs`.

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,410 @@
...
+    let mut builder = ManifestBuilder::new(db_path, batch_size)?;
+    builder.insert_channels(&channel_manifests)?;
+    builder.insert_entries(channel_manifests)?;
+    builder.create_index()
+}
+
+/// A private helper struct implementing methods for database creation
+struct ManifestBuilder {
+    db: Db,
+    dir_ids: HashMap<String, u64>, // TODO: limit size of this hash map
```
passaro: remove TODO

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,410 @@
...
+        &mut self,
+        object_key: &ValidKey,
+        channel_id: usize,
+        channel_root_id: u64,
+    ) -> Result<(u64, ValidKey), InputManifestError> {
+        debug_assert_eq!(object_key.kind(), InodeKind::File);
 
-    Ok(())
-}
+        let components: Vec<_> = object_key.components();
```
passaro, line 278: This is wasteful: in the common case the directories are already cached, so the new `Vec` (and the `ValidKey`) are discarded.

Happy to leave it for now and improve in a later PR.

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -54,163 +78,184 @@ impl Db {
...
-    /// Queries up to `batch_size` direct children of the directory with key `parent`, starting from `next_offset`
-    pub fn select_children(&self, parent: &str, next_offset: usize, batch_size: usize) -> Result<Vec<DbEntry>> {
+    /// Queries a row from the DB representing either a file or a directory.
+    ///
+    /// Looks up an entry by its parent ID and name, which uniquely identifies
+    /// an entry within its parent directory.
+    pub fn select_entry(&self, parent_id: u64, name: &str) -> Result<Option<DbEntry>> {
+        let start = Instant::now();
+        let conn = self.conn.lock().expect("lock must succeed");
+        metrics::histogram!("manifest.lookup_by_id.lock.elapsed_micros").record(start.elapsed().as_micros() as f64);
```
passaro: Wrong metric name. In this case, shouldn't we use the same name whenever we are waiting for the same lock?

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -54,163 +78,184 @@ impl Db {
...
+
+        let start = Instant::now();
+        let query = "SELECT id, parent_id, channel_id, parent_partial_key, name, etag, size FROM s3_objects WHERE parent_id = ?1 AND name = ?2";
+        tracing::debug!("executing {} with parameters {:?}", query, (parent_id, name,));
+        let mut stmt = conn.prepare(query)?;
+        let result = stmt.query_row((parent_id, name), |row: &Row| row.try_into()).optional();
+        metrics::histogram!("manifest.lookup_by_id.query.elapsed_micros").record(start.elapsed().as_micros() as f64);
```
passaro: Wrong metric name. I'd suggest reviewing all names to follow a common pattern, e.g.:
`manifest.query.<QUERY_NAME>.<METRIC>`.

mountpoint-s3-fs/src/superblock/inode.rs
```diff
@@ -2,19 +2,14 @@ use std::collections::{HashMap, HashSet};
 use std::ops::{Deref, DerefMut};
 use std::time::Duration;
 
-use crate::metablock::ValidKey;
 use crate::metablock::{InodeError, InodeErrorInfo, InodeKind, InodeNo, InodeStat};
+use crate::metablock::{NEVER_EXPIRE_TTL, ROOT_INODE_NO, ValidKey};
```
passaro: nit: merge

mountpoint-s3-fs/tests/common/fuse.rs
```diff
@@ -152,6 +159,71 @@ pub trait TestSessionCreator: FnOnce(&str, TestSessionConfig) -> TestSession {}
 // `FnOnce(...)` in place of `impl TestSessionCreator`.
 impl<T> TestSessionCreator for T where T: FnOnce(&str, TestSessionConfig) -> TestSession {}
 
+fn create_metablock<Client>(
```
passaro: unused?

mountpoint-s3-fs/src/manifest.rs
```diff
@@ -1,168 +1,11 @@
...
-            self.finished = true;
-        }
-
-        self.next_offset += db_entries.len();
-        self.entries.extend(db_entries);
-
-        Ok(())
-    }
-}
+pub use manifest_core::{Manifest, ManifestError};
```
passaro: nit: could we use just `core`?

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -1,62 +1,410 @@
...
+#[derive(Debug, Deserialize)]
+pub struct ChannelConfig {
+    pub directory_name: String,
+    pub bucket_name: String,
+    #[serde(default)]
+    pub prefix: String,
+    pub manifest_path: PathBuf,
+}
+
+/// [ChannelManifest] is a helper struct, primarily exposed for usage in tests.
```
passaro: This and `create_db` make an awkward interface and are only public because we need them in tests. We should consider refining the builder interface and make that public instead.

But let's do that in a future PR.

mountpoint-s3-fs/examples/mount_from_config.rs
vladem: not sure what do you mean by inline, pass `fs_config` as an argument to `create_fuse_session`?

passaro: You could inline the `fs_config` variable, i.e. just have `config.build_filesystem_config()?,` here.

passaro (dismissed):
LGTM!

mountpoint-s3-fs/src/manifest/db.rs
```diff
@@ -82,15 +82,16 @@ impl Db {
     pub fn select_entry_by_id(&self, id: u64) -> Result<Option<DbEntry>> {
         let start = Instant::now();
         let conn = self.conn.lock().expect("lock must succeed");
-        metrics::histogram!("manifest.query.lock.elapsed_micros").record(start.elapsed().as_micros() as f64);
+        metrics::histogram!("manifest.query.duration_us", "query" => "lock").record(start.elapsed().as_micros() as f64);
```
passaro: Not a blocker, but this is not a query. Maybe it should be a separate metric?

passaro (approved):
Ship it!

PR #1507, opened 2025-07-09
Refactor benchmark.py to extend to prefetcher and other benchmarks

sahityadg:
This change extracts fio and mountoint specific code from benchmark.py to specific modules to make it cleaner. It also separates the configuration into sections allowing us to have benchmark specific sweeper parameters. 

No, benchmark change only

No, benchmark change only

---

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -189,14 +189,14 @@ def run_experiment(cfg: DictConfig) -> None:
 
         # Mark success if we get here without exceptions
         metadata["success"] = True
-    except Exception as e:
-        log.error(f"Benchmark execution failed: {str(e)}")
+    except Exception:
+        log.error("Benchmark execution failed:", exc_info=True)
         raise
     finally:
         try:
             benchmark.post_process()
-        except Exception as e:
-            log.error(f"Post-processing failed: {str(e)}")
+        except Exception:
+            log.error("Post-processing failed:", exc_info=True)
         finally:
             write_metadata(metadata)
             metadata["end_time"] = datetime.now(tz=timezone.utc)
--- a/benchmark/benchmarks/fio_benchmark.py
+++ b/benchmark/benchmarks/fio_benchmark.py
@@ -60,19 +60,14 @@ class FioBenchmark(BaseBenchmark):
         subprocess_env.update(fio_env)
         log.debug("Subproces env: %s; env: %s", subprocess_env)
 
-        try:
-            with Popen(subprocess_args, env=subprocess_env) as process:
-                exit_code = process.wait()
-                if exit_code != 0:
-                    log.error(f"FIO process failed with exit code {exit_code}")
-                    raise CalledProcessError(exit_code, subprocess_args)
+        with Popen(subprocess_args, env=subprocess_env) as process:
+            exit_code = process.wait()
+            if exit_code != 0:
+                log.error(f"FIO process failed with exit code {exit_code}")
+                raise CalledProcessError(exit_code, subprocess_args)
 
-            # Store benchmark results in metadata
-            self.metadata["fio_output_file"] = self.fio_output_filepath
-
-        except Exception as e:
-            log.error(f"Benchmark failed: {e}")
-            raise
+        # Store benchmark results in metadata
+        self.metadata["fio_output_file"] = self.fio_output_filepath
 
     def post_process(self) -> Dict[str, Any]:
         cleanup_mp(self.mount_dir)
--- a/benchmark/benchmarks/mountpoint.py
+++ b/benchmark/benchmarks/mountpoint.py
@@ -15,14 +15,12 @@ MP_LOGS_DIRECTORY = "mp_logs/"
 
 
 def cleanup_mp(mount_dir):
-    try:
-        if mount_dir is not None:
-            log.info(f"Cleaning up {mount_dir}")
-            subprocess.check_output(["umount", mount_dir])
-            os.rmdir(mount_dir)
-            os.remove(f"{mount_dir}.pid")
-    except Exception:
-        log.error(f"Error cleaning up Mountpoint at {mount_dir}:", exc_info=True)
+    if mount_dir is not None:
+        log.info(f"Cleaning up {mount_dir}")
+        subprocess.check_output(["umount", mount_dir])
+        os.rmdir(mount_dir)
+        os.remove(f"{mount_dir}.pid")
+    log.error(f"Error cleaning up Mountpoint at {mount_dir}:", exc_info=True)
 
 
 def mount_mp(cfg: DictConfig, mount_dir: str) -> Dict[str, Any]:
@@ -72,8 +70,6 @@ def mount_mp(cfg: DictConfig, mount_dir: str) -> Dict[str, Any]:
         f"--log-directory={MP_LOGS_DIRECTORY}",
     ]
 
-    mp_env = {}
-
     if mp_config['prefix'] is not None:
         subprocess_args.append(f"--prefix={mp_config['prefix']}")
 
@@ -108,6 +104,7 @@ def mount_mp(cfg: DictConfig, mount_dir: str) -> Dict[str, Any]:
             )
         subprocess_args.append(f"--maximum-throughput-gbps={max_throughput}")
 
+    mp_env = {}
     if mp_config['mountpoint_max_background'] is not None:
         mp_env["UNSTABLE_MOUNTPOINT_MAX_BACKGROUND"] = str(mp_config['mountpoint_max_background'])
 
@@ -135,11 +132,7 @@ def mount_mp(cfg: DictConfig, mount_dir: str) -> Dict[str, Any]:
     subprocess_env.update(mp_env)
     log.debug("Subprocess env: %s", subprocess_env)
 
-    try:
-        output = subprocess.check_output(subprocess_args, env=subprocess_env)
-    except subprocess.CalledProcessError as e:
-        raise RuntimeError("Error during mounting") from e
-
+    output = subprocess.check_output(subprocess_args, env=subprocess_env)
     mountpoint_pid = get_mount_s3_pid(subprocess_env["UNSTABLE_MOUNTPOINT_PID_FILE"])
     log.info("Mountpoint pid: %d, output: %s", mountpoint_pid, output.decode("utf-8").strip())
 
@@ -153,18 +146,8 @@ def mount_mp(cfg: DictConfig, mount_dir: str) -> Dict[str, Any]:
 
 
 def get_mount_s3_pid(pid_file: str) -> int:
-    """
-    Get the process ID of the mount-s3 process from the PID file.
-    """
-    try:
-        with open(pid_file, 'r') as f:
-            pid = int(f.read().strip())
+    with open(pid_file, 'r') as f:
+        pid = int(f.read().strip())
 
-        log.debug(f"Read mount-s3 pid: {pid} from file: {pid_file}")
-
-        return pid
-
-    except FileNotFoundError:
-        raise RuntimeError(f"Mountpoint pid file not found: {pid_file}")
-    except Exception as e:
-        raise RuntimeError("Could not determine mountpoint pid") from e
+    log.debug(f"Read mount-s3 pid: {pid} from file: {pid_file}")
+    return pid


benchmark/benchmarks/fio_benchmark.py
```diff
@@ -0,0 +1,117 @@
...
+    def setup(self) -> Dict[str, Any]:
+        self.mount_dir = tempfile.mkdtemp(suffix=".mountpoint-s3")
+        self.mount_metadata = mount_mp(self.cfg, self.mount_dir)
+
+        fio_job_name = self.fio_config['fio_benchmark']
+        self.fio_output_filepath = f"fio.{fio_job_name}.json"
+
+        return self.mount_metadata
+
+    def run_benchmark(self) -> Dict[str, Any]:
```
muddyfish: This only ever returns `{"success": True}` or `{"success": False}`. It's also a bit weird that we're doing the exception handling here rather than in the caller

benchmark/benchmarks/fio_benchmark.py
```diff
@@ -0,0 +1,117 @@
...
+                    raise CalledProcessError(exit_code, subprocess_args)
+            return {"success": True}
+        except Exception as e:
+            log.error(f"Benchmark failed: {e}")
+            return {"success": False}
+
+    def post_process(self) -> Dict[str, Any]:
+        self.collect_logs()
+
+        try:
```
muddyfish: Should probably not be nested - only around the `os.rmdir/remove` pieces. Ideally wouldn't catch Exception here either

benchmark/benchmarks/fio_benchmark.py
```diff
@@ -0,0 +1,117 @@
...
+
+            log.debug(f"{self.mount_dir} unmounted")
+            os.rmdir(self.mount_dir)
+            os.remove(self.mount_metadata["mount_s3_env"]["UNSTABLE_MOUNTPOINT_PID_FILE"])
+        except Exception as e:
+            log.error(f"Error cleaning up Mountpoint at {self.mount_dir}: {e}")
+
+        return {}
+
+    def collect_logs(self) -> None:
```
muddyfish: Function name is not explicit in what this actually does. It renames an existing log file into the current directory (which feels a bit weird).

benchmark/benchmark.py
```diff
@@ -393,20 +162,34 @@ def run_experiment(cfg: DictConfig) -> None:
...
+            benchmark = FioBenchmark(cfg)
+        else:
+            raise ValueError(f"Unsupported benchmark type: {benchmark_type}")
 
-    _postprocessing(metadata)
+        # Run setup phase
+        setup_metadata = benchmark.setup()
+        target_pid = setup_metadata.get("target_pid")
+
+        with ResourceMonitoring.managed(target_pid, cfg.monitoring.with_bwm, cfg.monitoring.with_perf_stat):
```
muddyfish, line 187: Ideally, the acronyms here would be expanded, or using keyword arguments.

benchmark/benchmarks/base_benchmark.py
```diff
@@ -0,0 +1,55 @@
+from abc import ABC, abstractmethod
+import logging
+from typing import Dict, Any
+
+from omegaconf import DictConfig
+
+log = logging.getLogger(__name__)
```
muddyfish, line 6: Not used

benchmark/benchmarks/benchmark_config_parser.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self, cfg: DictConfig):
+        """
+        Initialize the config parser with the configuration.
+
+        Args:
+            cfg: The benchmark configuration
+        """
+        self.cfg = cfg
+
+    def get_common_config(self) -> Dict[str, Any]:
```
muddyfish, line 13: Ideally these functions would return a [`TypedDict`](https://typing.python.org/en/latest/spec/typeddict.html)

benchmark/benchmarks/fio_benchmark.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, cfg: DictConfig):
+        super().__init__(cfg)
+        self.config_parser = BenchmarkConfigParser(cfg)
+        self.common_config = self.config_parser.get_common_config()
+        self.fio_config = self.config_parser.get_fio_config()
+        self.mount_dir = None
+        self.mount_metadata = None
+        self.fio_output_filepath = None
+
+    def setup(self) -> Dict[str, Any]:
```
muddyfish, line 28: The `setup` and `post_process` functions look kind of like a context manager that manages a Mountpoint instance

benchmark/benchmarks/fio_benchmark.py
```diff
@@ -0,0 +1,117 @@
...
+                exit_code = process.wait()
+                if exit_code != 0:
+                    log.error(f"FIO process failed with exit code {exit_code}")
+                    raise CalledProcessError(exit_code, subprocess_args)
+            return {"success": True}
+        except Exception as e:
+            log.error(f"Benchmark failed: {e}")
+            return {"success": False}
+
+    def post_process(self) -> Dict[str, Any]:
```
muddyfish, line 72: This only returns the empty dict. Why?

benchmark/benchmarks/mountpoint.py
```diff
@@ -0,0 +1,176 @@
...
+    if mp_config['mountpoint_binary'] is None:
+        mountpoint_args = [
+            "cargo",
+            "run",
+            "--quiet",
+            "--release",
+            "--features=mock",
+        ]
+
+        if stub_mode == "s3_client":
```
muddyfish, line 47: Why is `stub_mode` only used if no custom MP binary?

benchmark/benchmarks/mountpoint.py
```diff
@@ -0,0 +1,176 @@
...
+            # `mock-mount-s3` requires bucket to be prefixed with `sthree-` to verify we're not actually reaching S3
+            logging.debug("using mock-mount-s3 due to `stub_mode`, bucket will be prefixed with \"sthree-\"")
+            bucket = f"sthree-{bucket}"
+
+            mountpoint_args.append("--bin=mock-mount-s3")
+
+        # End Cargo command, begin passing arguments to Mountpoint
+        mountpoint_args.append("--")
+    else:
+        mountpoint_args = [mp_config['mountpoint_binary']]
```
muddyfish, line 56: This isn't a great name - it's the args to invoke Mountpoint, not args to be given to Mountpoint

benchmark/benchmarks/mountpoint.py
```diff
@@ -0,0 +1,176 @@
...
+        "--allow-overwrite",
+        "--allow-delete",
+        f"--log-directory={MP_LOGS_DIRECTORY}",
+    ]
+    subprocess_env = os.environ.copy()
+
+    if mp_config['prefix'] is not None:
+        subprocess_args.append(f"--prefix={mp_config['prefix']}")
+
+    if mp_config['mountpoint_debug']:
```
muddyfish, line 76: Most of these look like they could be a dictionary of config option to MP argument.

benchmark/benchmark.py
```diff
@@ -23,259 +23,38 @@
...
+        with urllib.request.urlopen(token_request) as token_response:
+            token = token_response.read().decode()
+
+        metadata_url = "http://169.254.169.254/latest/meta-data/instance-id"
+        metadata_request = urllib.request.Request(metadata_url, headers={"X-aws-ec2-metadata-token": token})
+        with urllib.request.urlopen(metadata_request) as metadata_response:
+            instance_id = metadata_response.read().decode()
+
+        return instance_id
+    except Exception:
```
muddyfish, line 45: This should only catch network related exceptions, otherwise this will be extremely hard to debug (`NameError`) for example

benchmark/benchmark.py
```diff
@@ -23,259 +23,38 @@
...
-        raise RuntimeError(f"Mountpoint pid file not found: {pid_file}")
-    except Exception as e:
-        raise RuntimeError("Could not determine mountpoint pid") from e
+def write_metadata(metadata: Dict[str, Any]) -> None:
+    """Write metadata to a file."""
+    try:
+        with open("metadata.json", "w") as f:
+            json.dump(metadata, f, default=str)
+        log.debug("Metadata written to metadata.json")
+    except Exception:
```
muddyfish, line 56: Perhaps catching `OSError` would be better here

benchmark/benchmark.py
```diff
@@ -380,34 +159,47 @@ def managed(target_pid, with_bwm=False, with_perf_stat=False):
...
+
+        with ResourceMonitoring.managed(target_pid, cfg.monitoring.with_bwm, cfg.monitoring.with_perf_stat):
+            benchmark.run_benchmark()
 
-    _postprocessing(metadata)
-    log.info("Experiment ended")
+        # Mark success if we get here without exceptions
+        metadata["success"] = True
+    except Exception as e:
+        log.error(f"Benchmark execution failed: {str(e)}")
```
muddyfish: Perhaps use `exc_info=True` as previously in these two cases?

benchmark/benchmarks/fio_benchmark.py
```diff
@@ -0,0 +1,79 @@
...
+        fio_env = {}
+        fio_env["APP_WORKERS"] = str(self.common_config['application_workers'])
+        fio_env["SIZE_GIB"] = str(self.common_config['object_size_in_gib'])
+        fio_env["DIRECT"] = "1" if self.fio_config['direct_io'] else "0"
+        fio_env["UNIQUE_DIR"] = datetime.now(tz=timezone.utc).isoformat()
+        fio_env["IO_ENGINE"] = self.fio_config['fio_io_engine']
+        fio_env["RUN_TIME"] = str(self.common_config['run_time'])
+
+        log.info("Running FIO with args: %s; env: %s", subprocess_args, fio_env)
+        subprocess_env = os.environ.copy()
```
muddyfish, line 59: Nit: You could probably do `subprocess_env = os.environ | fio_env` here

benchmark/benchmarks/fio_benchmark.py
```diff
@@ -0,0 +1,79 @@
...
+            with Popen(subprocess_args, env=subprocess_env) as process:
+                exit_code = process.wait()
+                if exit_code != 0:
+                    log.error(f"FIO process failed with exit code {exit_code}")
+                    raise CalledProcessError(exit_code, subprocess_args)
+
+            # Store benchmark results in metadata
+            self.metadata["fio_output_file"] = self.fio_output_filepath
+
+        except Exception as e:
```
muddyfish: The caller of this function also has a try/catch block with similar logging. Is this necessary?

benchmark/benchmarks/fio_benchmark.py
```diff
@@ -0,0 +1,79 @@
...
+        subprocess_args = [
+            FIO_BINARY,
+            "--eta=never",
+            "--output-format=json",
+            f"--output={self.fio_output_filepath}",
+            f"--directory={self.mount_dir}",
+            fio_job_filepath,
+        ]
+
+        fio_env = {}
```
muddyfish, line 50: This should probably be a function `build_fio_env`

benchmark/benchmarks/mountpoint.py
```diff
@@ -0,0 +1,170 @@
...
+
+logging.basicConfig(level=os.environ.get('LOGLEVEL', 'INFO').upper())
+log = logging.getLogger(__name__)
+
+MOUNT_DIRECTORY = "s3"
+MP_LOGS_DIRECTORY = "mp_logs/"
+
+
+def cleanup_mp(mount_dir):
+    try:
```
muddyfish: Again, the caller of this function has a try/catch block. We don't need this many

benchmark/benchmarks/mountpoint.py
```diff
@@ -0,0 +1,170 @@
...
+        *mountpoint_args,
+        bucket,
+        mount_dir,
+        "--log-metrics",
+        "--allow-overwrite",
+        "--allow-delete",
+        f"--log-directory={MP_LOGS_DIRECTORY}",
+    ]
+
+    mp_env = {}
```
muddyfish: Nit: Could we move this down closer to where it's first mutated?

benchmark/benchmarks/mountpoint.py
```diff
@@ -0,0 +1,170 @@
...
+        if mount_dir is not None:
+            log.info(f"Cleaning up {mount_dir}")
+            subprocess.check_output(["umount", mount_dir])
+            os.rmdir(mount_dir)
+            os.remove(f"{mount_dir}.pid")
+    except Exception:
+        log.error(f"Error cleaning up Mountpoint at {mount_dir}:", exc_info=True)
+
+
+def mount_mp(cfg: DictConfig, mount_dir: str) -> Dict[str, Any]:
```
muddyfish, line 26: Ideally we'd split this into two functions for getting the args/env for MP, and then this one for launching it (and possibly getting the args from the other)

benchmark/benchmarks/mountpoint.py
```diff
@@ -0,0 +1,170 @@
...
+        with open(pid_file, 'r') as f:
+            pid = int(f.read().strip())
+
+        log.debug(f"Read mount-s3 pid: {pid} from file: {pid_file}")
+
+        return pid
+
+    except FileNotFoundError:
+        raise RuntimeError(f"Mountpoint pid file not found: {pid_file}")
+    except Exception as e:
```
muddyfish: Try making this Exception narrower.

benchmark/benchmark.py
sahityadg: I think it can fail for other reasons that may not fall under OSError, though they are unlikely to happen.

benchmark/benchmarks/mountpoint.py
sahityadg: agree, retained the existing names as much as possible to not introduce other issues. I could look at cleaning this up further later.

benchmark/benchmarks/mountpoint.py
sahityadg: yeah, I thought it might be easier to compare the changeset. Agree we could make this cleaner, I'll do this later.

benchmark/benchmarks/mountpoint.py
sahityadg: Moved the catch all exception handling to benchmark.py

PR #1512, opened 2025-07-11
Update MP client benchmarks to benchmark multiple object downloads 

sahityadg:
This change adds new parameters to client benchmarks to download multiple objects and to limit the duration of the test. 
This change also extends bind parameter to take a comma separated list of NICs.

No, client benchmarks only

No, client benchmarks only

---

### DIFF
--- a/mountpoint-s3-client/examples/client_benchmark.rs
+++ b/mountpoint-s3-client/examples/client_benchmark.rs
@@ -3,7 +3,7 @@ use std::pin::pin;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::thread;
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 use clap::{Parser, Subcommand};
 use futures::StreamExt;
@@ -38,13 +38,16 @@ fn run_benchmark(
     keys: &Vec<String>,
     enable_backpressure: bool,
     output_path: Option<&Path>,
+    max_duration: Option<Duration>,
 ) {
     let mut total_bytes = 0;
     let total_start = Instant::now();
     let mut iter_results = Vec::new();
+    let mut iteration = 0;
+    let timeout = total_start + max_duration.unwrap_or(Duration::MAX);
 
-    for i in 0..num_iterations {
-        let start = Instant::now();
+    while iteration < num_iterations && Instant::now() < timeout {
+        let iter_start = Instant::now();
         let received_size = Arc::new(AtomicU64::new(0));
 
         thread::scope(|scope| {
@@ -68,7 +71,7 @@ fn run_benchmark(
                         }
 
                         let mut request = pin!(request);
-                        loop {
+                        while Instant::now() < timeout {
                             match request.next().await {
                                 Some(Ok(part)) => {
                                     let part_len = part.data.len();
@@ -107,32 +110,42 @@ fn run_benchmark(
             }
         });
 
-        let elapsed = start.elapsed();
+        let elapsed = iter_start.elapsed();
         let received_size = received_size.load(Ordering::SeqCst);
         total_bytes += received_size;
         println!(
             "{}: received {} bytes in {:.2}s: {:.2} Gib/s",
-            i,
+            iteration,
             received_size,
             elapsed.as_secs_f64(),
             (received_size as f64) / elapsed.as_secs_f64() / (1024 * 1024 * 1024 / 8) as f64
         );
 
         iter_results.push(json!({
-            "iteration": i,
+            "iteration": iteration,
             "bytes": received_size,
             "duration_seconds": elapsed.as_secs_f64(),
         }));
+
+        iteration += 1;
     }
 
     let total_elapsed = total_start.elapsed();
+    println!(
+        "Total: received {} bytes in {:.2}s across {} iterations: {:.2} Gib/s",
+        total_bytes,
+        total_elapsed.as_secs_f64(),
+        iter_results.len(),
+        (total_bytes as f64) / total_elapsed.as_secs_f64() / (1024 * 1024 * 1024 / 8) as f64
+    );
+
     if let Some(output_path) = output_path {
         let ouput_file = std::fs::File::create(output_path).expect("Failed to create output_file: {output_path}");
         let results = json!({
             "summary": {
                 "total_bytes": total_bytes,
                 "duration_seconds": total_elapsed.as_secs_f64(),
-                "iterations": num_iterations,
+                "iterations": iter_results.len(),
             },
             "iterations": iter_results
         });
@@ -199,8 +212,19 @@ struct CliArgs {
         default_value = "0"
     )]
     initial_window_size: Option<usize>,
-    #[clap(long, help = "Output file to write the results to", value_name = "OUTPUT_FILE")]
+    #[arg(long, help = "Output file to write the results to", value_name = "OUTPUT_FILE")]
     output_file: Option<PathBuf>,
+    #[arg(
+        long,
+        help = "Maximum duration (in seconds) to run the benchmark",
+        value_name = "SECONDS",
+        value_parser = |arg: &str| -> Result<Duration, String> {
+            arg.parse::<u64>()
+            .map(Duration::from_secs)
+            .map_err(|e| format!("Invalid duration: {e}"))
+        }
+    )]
+    max_duration: Option<Duration>,
 }
 
 fn create_s3_client_config(region: &str, args: &CliArgs, bind: &Option<Vec<String>>) -> S3ClientConfig {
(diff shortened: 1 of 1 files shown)


mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -34,29 +34,32 @@ fn init_tracing_subscriber() {
 fn run_benchmark(
     client: impl ObjectClient + Clone + Send,
     num_iterations: usize,
-    num_downloads: usize,
     bucket: &str,
-    key: &str,
+    keys: &[String],
```
muddyfish: `IntoIterator<Item=String>`?

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -34,29 +34,32 @@ fn init_tracing_subscriber() {
...
+    while iteration < num_iterations && start_time.elapsed() < max_duration.unwrap_or(Duration::MAX) {
+        let iter_start = Instant::now();
         let received_size = Arc::new(AtomicU64::new(0));
 
         thread::scope(|scope| {
-            for _ in 0..num_downloads {
+            for key in keys {
                 let client = client.clone();
                 let received_size_clone = Arc::clone(&received_size);
+                let key_clone = key.clone();
```
muddyfish: If you use `IntoIterator`, then I don't think you need this

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -68,7 +71,7 @@ fn run_benchmark(
                         }
 
                         let mut request = pin!(request);
-                        loop {
+                        while start_time.elapsed() < max_duration.unwrap_or(Duration::MAX) {
```
muddyfish: If it matters, you could match with a timeout as well to ensure it exits fast

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -142,12 +155,12 @@ fn run_benchmark(
...
 enum Client {
-    #[command(about = "Download a key from S3")]
+    #[command(about = "Download keys from S3")]
     Real {
         #[arg(help = "Bucket name")]
         bucket: String,
-        #[arg(help = "Key name")]
-        key: String,
+        #[arg(help = "Comma-separated list of key names")]
+        keys: String,
```
muddyfish: I think you can use `value_delimiter` here and `keys: Vec<String>`

muddyfish (approved):

muddyfish (approved):

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -34,29 +34,34 @@ fn init_tracing_subscriber() {
 fn run_benchmark(
     client: impl ObjectClient + Clone + Send,
     num_iterations: usize,
-    num_downloads: usize,
     bucket: &str,
-    key: &str,
+    keys: &Vec<String>,
```
passaro: It is unusual to pass a reference to a vector (of `String`s even). Prefer `&[&str]`, or `Vec<String>` if you really want to take ownership of both the `String`s and the `Vec`.

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -34,29 +34,34 @@ fn init_tracing_subscriber() {
...
     let total_start = Instant::now();
     let mut iter_results = Vec::new();
+    let mut iteration = 0;
+    let timeout: Instant = total_start
+        .checked_add(max_duration.unwrap_or(Duration::from_secs(86400)))
+        .expect("Duration overflow error");
 
-    for i in 0..num_iterations {
-        let start = Instant::now();
+    while iteration < num_iterations && Instant::now() < timeout {
```
passaro, line 51: Subjective: I'd prefer the `for` here and the time check as a separate early exit.

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -107,32 +112,42 @@ fn run_benchmark(
...
+            "iteration": iteration,
             "bytes": received_size,
             "duration_seconds": elapsed.as_secs_f64(),
         }));
+
+        iteration += 1;
     }
 
     let total_elapsed = total_start.elapsed();
+    println!(
```
passaro, line 135: Should we treat the timeout case differently?

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -142,17 +157,22 @@ fn run_benchmark(
...
-    #[command(about = "Download a key from S3")]
+    #[command(about = "Download keys from S3")]
     Real {
         #[arg(help = "Bucket name")]
         bucket: String,
-        #[arg(help = "Key name")]
-        key: String,
+        #[arg(
+            help = "Comma-separated list of key names",
+            value_delimiter = ',',
```
passaro, line 166: Is there a reason to set the delimiter? We could just allow this, as we do elsewhere:
```
--key ONE [--key TWO ...]
```

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -142,17 +157,22 @@ fn run_benchmark(
...
+            value_name = "KEYS"
+        )]
+        keys: Vec<String>,
         #[arg(long, help = "AWS region", default_value = "us-east-1")]
         region: String,
-        #[clap(
+        #[arg(
             long,
             help = "One or more network interfaces to use when accessing S3. Requires Linux 5.7+ or running as root.",
+            value_delimiter = ',',
```
passaro, line 175: Same here, now this is different from `mount-s3`.

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -196,8 +214,42 @@ struct CliArgs {
...
     initial_window_size: Option<usize>,
-    #[clap(long, help = "Output file to write the results to", value_name = "OUTPUT_FILE")]
-    pub output_file: Option<PathBuf>,
+    #[arg(long, help = "Output file to write the results to", value_name = "OUTPUT_FILE")]
+    output_file: Option<PathBuf>,
+    #[arg(
+        long,
+        help = "Maximum duration (in seconds) to run the benchmark",
+        value_name = "SECONDS",
+        value_parser = |arg: &str| -> Result<Duration, String> {
```
passaro: nit: make it function?

passaro (changes requested):

mountpoint-s3-client/examples/client_benchmark.rs
sahityadg: ~~Will send a separate PR~~

I didn't see the merge is blocked, will update

mountpoint-s3-client/examples/client_benchmark.rs
sahityadg: It is still supported while it also allows --bind eth0, eth1

mountpoint-s3-client/examples/client_benchmark.rs
sahityadg: `client_benchmark real [OPTIONS] <BUCKET> `[KEYS]...`

These are still positional arguments but allow comma separated list too. As I was scripting the commands, I found it useful. If there is a convention, we want to stick to, I can revert the change.

mountpoint-s3-client/examples/client_benchmark.rs
sahityadg: Waiting for the entire objects to download is not practical for overnight benchmarks with lots of objects, which is we we added the runtime option. So we want to capture the throughput without waiting for whole objects to be downloaded.

passaro: I see. I'm just slightly worried that a timeout mid-iteration may skew the results. Would we be able to see that from what we print here?

sahityadg: If a thread takes longer than the configured runtime, then total_bytes/total_duration would look worse (lower than actual throughput). However, we could look at total_bytes/configured_duration or discard the results when total_duration is more than the configured duration.

mountpoint-s3-client/examples/client_benchmark.rs
sahityadg: For now, cloning to keep the Vec<String> as we have to pass it to the CRT client.

passaro: The comment above still applies. Not a blocker though.

passaro (approved):

mountpoint-s3-client/examples/client_benchmark.rs
sahityadg: Sorry for the churn, addressed this in the latest commit.

passaro (approved):

muddyfish (approved):

PR #1511, opened 2025-07-11
Use a unified memory pool in Mountpoint

passaro:
Introduces a unified memory pool in Mountpoint. The pool adopts the CRT pool interface, so it can be used by the CRT client when requesting buffers. Ownership of the buffers is then passed to the prefetcher when they are returned from GetObject requests. The same memory pool is also used to serve reads from the local disk cache and for incremental uploads.

The main goal is to reduce overall memory usage and mitigate memory fragmentation issues. We may also observe performance gains in some scenarios since we can avoid copying the data received from GetObject.

### Does this change impact existing behavior?

No changes in file system behavior. It will publish new memory-related metrics.

### Does this change need a changelog entry? Does it require a version change?

Requires entries in the `fs` and `mount-s3` changelogs and new major versions.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

PR #1516, opened 2025-07-15
Add support for custom memory pools

passaro:
Introduces a `MemoryPool` trait in the client crate which allows users to provide their own memory pool implementation. This is part of the broader effort to use a unified memory pool in Mountpoint (see draft PR #1511).

This change introduces:
* The required code to bridge implementations of the new Rust trait to the CRT pool interface.
* A simple `MemoryPool` implementation to be used in tests.
* The `pool_tests` feature flags to use the above pool in the client tests, replacing the CRT default pool.
* A new CI workflow to run the client tests with the custom pool.

No.

Entry in the client changelog.

---

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -148,6 +148,64 @@ jobs:
       if: ${{ failure() && matrix.runner.name == 'Amazon Linux arm' }}
       run: ./.github/actions/scripts/save-coredump.sh
 
+
+  client-test:
+    name: Client Tests (${{ matrix.bucket-type.name }}, ${{ matrix.runner.name }}, ${{ matrix.pool.name }})
+    runs-on: ${{ matrix.runner.tags }}
+
+    environment: ${{ inputs.environment }}
+    env:
+      features: s3_tests,fips_tests,${{ matrix.pool.feature }},${{ matrix.bucket-type.feature }}
+      packages: --package mountpoint-s3-client --package mountpoint-s3-crt --package mountpoint-s3-crt-sys
+
+    strategy:
+      fail-fast: false
+      matrix:
+        bucket-type:
+        - name: S3
+          feature: 
+        - name: S3 Express One Zone
+          feature: s3express_tests
+        runner:
+        - name: Ubuntu x86
+          tags: [ubuntu-22.04] # GitHub-hosted
+        - name: Amazon Linux arm
+          tags: [self-hosted, linux, arm64]
+        pool:
+        - name: Test Pool
+          feature: pool_tests
+
+    steps:
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        role-to-assume: ${{ vars.ACTIONS_IAM_ROLE }}
+        aws-region: ${{ vars.S3_REGION }}
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        ref: ${{ inputs.ref }}
+        submodules: true
+        persist-credentials: false
+    - name: Set up Rust toolchain
+      uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        # setup-rust-toolchain sets "-D warnings" by default, and Rust treats any warning as compile error.
+        # We need to this currently because `mountpoint-s3-fuser` contains some warnings and it breaks the build.
+        rustflags: ""
+    - name: Install operating system dependencies
+      uses: ./.github/actions/install-dependencies
+      with:
+        # not required for client tests. TODO: make it optional.
+        fuseVersion: 2
+    - name: Build tests
+      run: cargo test ${{ env.packages }} --features '${{ env.features }}' --no-run
+    - name: Run tests
+      run: cargo test ${{ env.packages }} --features '${{ env.features }}'
+    - name: Save dump files
+      if: ${{ failure() && matrix.runner.name == 'Amazon Linux arm' }}
+      run: ./.github/actions/scripts/save-coredump.sh
+
   fstab:
     name: fstab tests (${{ matrix.runner.name }}
     runs-on: ${{ matrix.runner.tags }}
--- a/mountpoint-s3-client/src/lib.rs
+++ b/mountpoint-s3-client/src/lib.rs
@@ -40,6 +40,7 @@
 //!
 //! [awscrt]: https://docs.aws.amazon.com/sdkref/latest/guide/common-runtime.html
 
+#![deny(clippy::undocumented_unsafe_blocks)]
 // Make async trait docs not-ugly on docs.rs (https://github.com/dtolnay/async-trait/issues/213)
 #![cfg_attr(docsrs, feature(async_fn_in_trait))]
 
@@ -79,6 +80,8 @@ pub mod config {
     pub use mountpoint_s3_crt::io::io_library_init;
     #[doc(hidden)]
     pub use mountpoint_s3_crt::s3::s3_library_init;
+
+    pub use mountpoint_s3_crt::s3::pool::{MemoryPool, MemoryPoolFactory, MemoryPoolFactoryOptions};
 }
 
 /// Types used by all object clients
--- a/mountpoint-s3-client/src/lib.rs
+++ b/mountpoint-s3-client/src/lib.rs
@@ -63,7 +63,8 @@ pub mod error_metadata;
 pub use object_client::{ObjectClient, PutObjectRequest};
 
 pub use s3_crt_client::{
-    OnTelemetry, S3CrtClient, S3RequestError, get_object::S3GetObjectResponse, put_object::S3PutObjectRequest,
+    NewClientError, OnTelemetry, S3CrtClient, S3RequestError, get_object::S3GetObjectResponse,
+    put_object::S3PutObjectRequest,
 };
 
 /// Configuration for the S3 client
--- a/mountpoint-s3-crt/src/s3/pool.rs
+++ b/mountpoint-s3-crt/src/s3/pool.rs
@@ -1,0 +1,429 @@
+//! Bridge custom memory pool implementations to the CRT S3 Client interface.
+
+use std::marker::PhantomPinned;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::{fmt::Debug, ptr::NonNull};
+
+use mountpoint_s3_crt_sys::{
+    aws_allocator, aws_byte_buf, aws_byte_buf_from_empty_array, aws_byte_cursor, aws_future_s3_buffer_ticket,
+    aws_future_s3_buffer_ticket_acquire, aws_future_s3_buffer_ticket_new, aws_future_s3_buffer_ticket_release,
+    aws_future_s3_buffer_ticket_set_result_by_move, aws_ref_count_init, aws_s3_buffer_pool, aws_s3_buffer_pool_config,
+    aws_s3_buffer_pool_factory_fn, aws_s3_buffer_pool_reserve_meta, aws_s3_buffer_pool_vtable, aws_s3_buffer_ticket,
+    aws_s3_buffer_ticket_vtable,
+};
+
+use crate::ToAwsByteCursor as _;
+use crate::common::allocator::Allocator;
+
+/// A custom memory pool.
+pub trait MemoryPool: Clone + Send + Sync {
+    /// Associated buffer type.
+    type Buffer: AsMut<[u8]>;
+
+    /// Get a buffer of at least the requested size.
+    fn get_buffer(&self, size: usize) -> Self::Buffer;
+
+    /// Trim the pool.
+    ///
+    /// Return `true` if the pool freed any memory.
+    fn trim(&self) -> bool;
+}
+
+/// Factory for a custom memory pool.
+pub trait MemoryPoolFactory {
+    /// The [MemoryPool] implementation created by this factory.
+    type Pool: MemoryPool;
+
+    /// Create a memory pool instance.
+    fn create(&self, options: MemoryPoolFactoryOptions) -> Self::Pool;
+}
+
+impl<F, P> MemoryPoolFactory for F
+where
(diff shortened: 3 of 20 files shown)


.github/workflows/integration.yml
```diff
@@ -148,6 +148,64 @@ jobs:
...
+        - name: S3 Express One Zone
+          feature: s3express_tests
+        runner:
+        - name: Ubuntu x86
+          tags: [ubuntu-22.04] # GitHub-hosted
+        - name: Amazon Linux arm
+          tags: [self-hosted, linux, arm64]
+        pool:
+        - name: Test Pool
+          feature: pool_tests
```
muddyfish, line 176: And we're not testing without any features because they're already tested?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -0,0 +1,429 @@
...
+pub struct CrtBufferPoolFactory(Arc<CrtBufferPoolFactoryInner>);
+
+#[derive(Debug)]
+struct CrtBufferPoolFactoryInner {
+    factory_ptr: NonNull<libc::c_void>,
+    factory_fn: aws_s3_buffer_pool_factory_fn,
+    drop_fn: fn(*mut ::libc::c_void),
+}
+
+// SAFETY: `CrtBufferPoolFactoryInner` is safe to transfer across threads because it wraps a [MemoryPoolFactory] implementation that is [Send].
```
muddyfish, line 91: Is `MemoryPoolFactory` always `Send` and `Sync`?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -0,0 +1,429 @@
...
+    }
+}
+
+unsafe extern "C" fn buffer_pool_factory<PoolFactory: MemoryPoolFactory>(
+    allocator: *mut aws_allocator,
+    config: aws_s3_buffer_pool_config,
+    user_data: *mut libc::c_void,
+) -> *mut aws_s3_buffer_pool {
+    // SAFETY: `user_data` references a `Box` owned by the `CrtBufferPoolFactory` instance.
+    let pool_factory = unsafe { &*(user_data as *mut PoolFactory) };
```
muddyfish, line 130: The previous safety comment this is treated as pinned, but we're not using `Pin::new_unchecked` here

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -0,0 +1,429 @@
...
+            aws_future_s3_buffer_ticket_set_result_by_move(self.inner, &mut ticket);
+        }
+    }
+
+    /// Return the pointer to the inner `aws_future_s3_buffer_ticket` instance.
+    ///
+    /// # Safety
+    /// The returned pointer follows ref-counting rules and must be eventually released.
+    unsafe fn into_inner_ptr(mut self) -> *mut aws_future_s3_buffer_ticket {
+        std::mem::replace(&mut self.inner, std::ptr::null_mut())
```
muddyfish, line 415: How does interact with `Drop` on self?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -0,0 +1,429 @@
...
+        let ticket = self.make_ticket(buffer);
+        future.set(ticket);
+
+        future
+    }
+
+    fn make_ticket(&self, buffer: Pool::Buffer) -> Pin<Box<CrtTicket<Pool::Buffer>>> {
+        // Set up the vtable by pointing into the pool pinned address, but leave `impl_`
+        // and `ref_count` to be initialized after pinning the ticket.
+        let mut ticket = Box::pin(CrtTicket {
```
muddyfish, line 265: I must be missing something, but why not Pin after setting up the self-referential stuff rather than unsafe getting the pin?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -0,0 +1,429 @@
...
+    ///
+    /// # Safety
+    /// The raw pointer must have been obtained through [leak()].
+    unsafe fn from_raw(ticket: *mut aws_s3_buffer_ticket) -> Pin<Box<Self>> {
+        // SAFETY: `ticket` points to the `inner` field of a pinned instance.
+        unsafe { Pin::new_unchecked(Box::from_raw((*ticket).impl_ as *mut Self)) }
+    }
+}
+
+unsafe extern "C" fn ticket_claim<Buffer: AsMut<[u8]>>(mut ticket: *mut aws_s3_buffer_ticket) -> aws_byte_buf {
```
muddyfish, line 363: Could we have some docs on what this function is meant to do? Does it convert the buffer inside `aws_s3_buffer_ticket` into an `aws_byte_buf`?

.github/workflows/integration.yml
passaro: Yes, that's included in the existing tests. I plan to review those and likely move that case here, but as a separate change.

mountpoint-s3-crt/src/s3/pool.rs
passaro: It should be. Will fix.

mountpoint-s3-crt/src/s3/pool.rs
passaro: The key is that we need the address, i.e. the `Box`. Technically we could create a `Box<CrtTicket>`, use its address to set up the self-referential fields, **being careful not to move it**, and then pin it. It would be fine, but why not just use `Box::pin(CrtTicket {..})` to get a pinned box in the first place?

mountpoint-s3-crt/src/s3/pool.rs
passaro: The null makes it a no-op. Will add comment.

mountpoint-s3-crt/src/s3/pool.rs
passaro: We don't need (or want) to take ownership of the pinned box, just a reference to it. And with only the reference, we cannot move it anyway.

.github/workflows/integration.yml
vladem: what's the purpose of this dimension? will we have tests running for different pools in future?

mountpoint-s3-client/src/lib.rs
```diff
@@ -40,6 +40,7 @@
 //!
 //! [awscrt]: https://docs.aws.amazon.com/sdkref/latest/guide/common-runtime.html
 
+#![deny(clippy::undocumented_unsafe_blocks)]
```
vladem, line 43: does this attribute prohibits undocumented unsafe blocks? does it apply to crates which `mountpoint-s3-client` depend on, specifically to `mountpoint-s3-crt`?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -0,0 +1,432 @@
...
+impl Drop for CrtBufferPoolFactoryInner {
+    fn drop(&mut self) {
+        (self.drop_fn)(self.factory_ptr.as_ptr());
+    }
+}
+
+impl CrtBufferPoolFactory {
+    /// Builds a factory for the given pool.
+    pub fn new<PoolFactory: MemoryPoolFactory>(pool_factory: PoolFactory) -> Self {
+        let factory = Box::pin(pool_factory);
```
vladem, line 105: why pinning here? is it because the pool held by `pool_factory` is likely to be self-referential struct?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -0,0 +1,432 @@
...
+    /// Returns the factory callback and user_data pointer to pass to the CRT.
+    pub(crate) fn as_inner(&self) -> (aws_s3_buffer_pool_factory_fn, *mut ::libc::c_void) {
+        (self.0.factory_fn, self.0.factory_ptr.as_ptr())
+    }
+}
+
+unsafe extern "C" fn buffer_pool_factory<PoolFactory: MemoryPoolFactory>(
+    allocator: *mut aws_allocator,
+    config: aws_s3_buffer_pool_config,
+    user_data: *mut libc::c_void,
```
vladem, line 127: `user_data` is misleading, `pool_factory` would be a better name (or a comment for consistency with C signature?)

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -0,0 +1,432 @@
...
+        Self(Arc::new(CrtBufferPoolFactoryInner {
+            factory_ptr,
+            factory_fn: Some(buffer_pool_factory::<PoolFactory>),
+            drop_fn: drop_pool_factory::<PoolFactory>,
+        }))
+    }
+
+    /// Returns the factory callback and user_data pointer to pass to the CRT.
+    pub(crate) fn as_inner(&self) -> (aws_s3_buffer_pool_factory_fn, *mut ::libc::c_void) {
+        (self.0.factory_fn, self.0.factory_ptr.as_ptr())
```
vladem, line 120: is it possible to store typed `factory_ptr` in `CrtBufferPoolFactoryInner` and cast it to `libc::c_void` in this method? can we avoid having `drop_fn` in that case?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -0,0 +1,432 @@
...
+        // SAFETY: We're setting up the struct to be self-referential, and we're not moving out
+        // of the struct, so the unchecked deref of the pinned pointer is okay.
+        unsafe {
+            let pool_ref = Pin::get_unchecked_mut(Pin::as_mut(&mut crt_pool));
+            pool_ref.inner.vtable = &raw mut pool_ref.pool_vtable;
+            pool_ref.inner.impl_ = pool_ref as *mut CrtBufferPool<Pool> as *mut libc::c_void;
+            aws_ref_count_init(
+                &mut pool_ref.inner.ref_count,
+                &mut pool_ref.inner as *mut aws_s3_buffer_pool as *mut libc::c_void,
+                Some(pool_destroy::<Pool>),
```
vladem, line 208: does it mean that "outer" struct `CrtBufferPool` relies on refcounter of it's field `inner` to get destroyed eventually?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    fn reserve(&self, size: usize) -> CrtTicketFuture {
+        let future = CrtTicketFuture::new(&self.allocator);
+
+        // Get a buffer from the pool, build its ticket, and immediately fullfil the future.
+        // This will likely change later, when we make the method on the pool async.
+        let buffer = self.pool.get_buffer(size);
+        let ticket = self.make_ticket(buffer);
+        future.set(ticket);
```
vladem, line 258: do you foresee in future implementation which doesn't set the future immediately? will the `MemoryPool` trait require another async interface for that?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -0,0 +1,432 @@
...
+
+        future
+    }
+
+    fn make_ticket(&self, buffer: Pool::Buffer) -> Pin<Box<CrtTicket<Pool::Buffer>>> {
+        // Set up the vtable by pointing into the pool pinned address, but leave `impl_`
+        // and `ref_count` to be initialized after pinning the ticket.
+        let mut ticket = Box::pin(CrtTicket {
+            inner: aws_s3_buffer_ticket {
+                vtable: (&raw const self.ticket_vtable).cast_mut(),
```
vladem: may `self` (and `aws_s3_buffer_ticket->vtable`) be destroyed before the returned `aws_s3_buffer_ticket`?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -0,0 +1,432 @@
...
+
+    // Build the `aws_byte_buf` required by the CRT from the pointer and length of the buffer.
+    // SAFETY: `ptr` is a valid buffer with capacity >= `len`.
+    unsafe { aws_byte_buf_from_empty_array(ptr as *mut libc::c_void, len) }
+}
+
+unsafe extern "C" fn ticket_destroy<Buffer: AsMut<[u8]>>(data: *mut libc::c_void) {
+    let ticket = data as *mut aws_s3_buffer_ticket;
+    // SAFETY: `ticket` was obtained through `Ticket::leak`.
+    _ = unsafe { CrtTicket::<Buffer>::from_raw(ticket) };
```
vladem, line 378: how is the buffer (held by ticket) returned to the pool?

sahityadg (approved):
LGTM.

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -357,14 +359,16 @@ impl<Buffer: AsMut<[u8]>> CrtTicket<Buffer> {
...
     // SAFETY: `ticket` was obtained through `Ticket::leak`.
     let ticket = unsafe { CrtTicket::<Buffer>::ref_mut_from_raw(&mut ticket) };
 
     // SAFETY: the CRT guarantees to only use the returned buffer while holding the ticket.
     let aws_byte_cursor { len, ptr } = unsafe { ticket.buffer.as_mut().as_aws_byte_cursor() };
 
-    // SAFETY: `ptr` is a valid buffer with capacity >= `size`.
+    // Build the `aws_byte_buf` required by the CRT from the pointer and length of the buffer.
+    // SAFETY: `ptr` is a valid buffer with capacity >= `len`.
     unsafe { aws_byte_buf_from_empty_array(ptr as *mut libc::c_void, len) }
```
muddyfish, line 372: Can we add another comment here saying we're using `aws_byte_buf_from_empty_array` rather than `aws_byte_buf_from_array` because the memory we're returning is to be used in future operations rather than the current content being significant. It took me a few reads to realise

vladem (dismissed):
LGTM

muddyfish (approved):

PR #1517, opened 2025-07-16
Improve tracing for prefetcher read window increments

c-hagem:
Changes the BackpressureHandle to include the address of the backpressure handle so that we can easily match tracing about read window increments to the file handle.

Does not impact existing behaviour, no changelog entry or version change required.
---

### DIFF
--- a/mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
+++ b/mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
@@ -173,8 +173,12 @@ impl BackpressureController {
         let prev_window_end_offset = self.read_window_end_offset;
         let next_window_end_offset = prev_window_end_offset + len as u64;
         trace!(
+            controller_addr = format!("{:p}", self),
             next_read_offset = self.next_read_offset,
-            prev_window_end_offset, next_window_end_offset, len, "incrementing read window",
+            prev_window_end_offset,
+            next_window_end_offset,
+            len,
+            "incrementing read window",
         );
 
         // This should not block since the channel is unbounded


mountpoint-s3-fs/src/fs.rs
```diff
@@ -385,6 +385,7 @@ where
             InodeKind::Directory => return Err(InodeError::IsDirectory(lookup.inode_err()).into()),
             InodeKind::File => (),
         }
+        let fh = self.next_handle();
```
muddyfish: Putting this earlier means some handles are generated but not used - is this a problem?

c-hagem: I know - I do not think this has an effect (i.e. fuse does not expect read handle numbers to be in increasing order) but it would be nice to avoid it. Perhaps having an atomic counter for prefetch streams might be nicer, and then we#d just log that this Prefetchstream n is associated with file handle k?

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -173,8 +173,12 @@ impl BackpressureController {
         let prev_window_end_offset = self.read_window_end_offset;
         let next_window_end_offset = prev_window_end_offset + len as u64;
         trace!(
+            controller_addr = format!("{:p}", self),
```
passaro, line 176: Is this a different approach? At the very least, it deserve a comment. And the PR description should be updated.

c-hagem: Yes- I thought it was easier to print the address instead of passing the file handle number through so many places (and the potentially not usoing file handles also didn't seem nice).

muddyfish: We're planning on using shared memory pools, so the address won't be sufficient to identify the object as being the same if it gets de-allocated and then re-allocated.

c-hagem: But we are using the pool for buffers, not for Backpressure controllers (for now)?

muddyfish: Probably, though I'd still be cautious about using addresses as persistently unique identifiers throughout a Mountpoint client's life

PR #1524, opened 2025-07-18
Extract meta request type when reserving buffers

passaro:
Modify the new `MemoryPool` trait (and the CRT bridge) to propagate the type of the meta request which is reserving the buffer. Requires extending bindings to an additional private header from `aws-c-s3` in order to access the type of a `aws_s3_meta_request` pointer.

No, the new type information is not used yet.

No.

---


passaro:
This is an incremental step in preparation of #1511.

c-hagem (approved):

PR #1525, opened 2025-07-19
Decouple mem_limiter from client

passaro:
Minor refactor to decouple the memory limiter from the client implementation. The memory limiter only requires the client to retrieve information about its internal memory pool utilization. This change wraps that request in a type-erased closure and drop the generic parameter from the memory limiter and all related types.

No. Internal refactor only.

No.

---

### DIFF
--- a/mountpoint-s3-fs/examples/prefetch_benchmark.rs
+++ b/mountpoint-s3-fs/examples/prefetch_benchmark.rs
@@ -124,15 +124,37 @@ fn parse_duration(arg: &str) -> Result<Duration, String> {
         .map_err(|e| format!("Invalid duration: {e}"))
 }
 
-fn create_memory_limiter(args: &CliArgs, client: &S3CrtClient) -> Arc<MemoryLimiter<S3CrtClient>> {
-    let max_memory_target = if let Some(target) = args.max_memory_target {
-        target * 1024 * 1024
-    } else {
-        // Default to 95% of total system memory
-        let sys = System::new_with_specifics(RefreshKind::everything());
-        (sys.total_memory() as f64 * 0.95) as u64
-    };
-    Arc::new(MemoryLimiter::new(client.clone(), max_memory_target))
+impl CliArgs {
+    fn memory_target(&self) -> u64 {
+        if let Some(target) = self.max_memory_target {
+            target * 1024 * 1024
+        } else {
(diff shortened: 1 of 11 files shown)


mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -124,15 +124,37 @@ fn parse_duration(arg: &str) -> Result<Duration, String> {
...
-        target * 1024 * 1024
-    } else {
-        // Default to 95% of total system memory
-        let sys = System::new_with_specifics(RefreshKind::everything());
-        (sys.total_memory() as f64 * 0.95) as u64
-    };
-    Arc::new(MemoryLimiter::new(client.clone(), max_memory_target))
+impl CliArgs {
+    fn memory_target_in_bytes(&self) -> u64 {
+        if let Some(target) = self.max_memory_target {
```
renanmagagnin, line 129: np: it would be nice if the unit of `self.max_memory_target` was explicit. Maybe we could to that in the local `target` variable if we don't rename `self.max_memory_target`. But seems like it would be made consistent with `self.crt_memory_limit_gib` and `self.maximum_throughput_gbps`.

mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -124,15 +124,37 @@ fn parse_duration(arg: &str) -> Result<Duration, String> {
...
+        // Set up backpressure with the same initial window used in Mountpoint.
+        let mut client_config = S3ClientConfig::new()
+            .read_backpressure(true)
+            .initial_read_window(mountpoint_s3_fs::s3::config::INITIAL_READ_WINDOW_SIZE)
+            .endpoint_config(EndpointConfig::new(self.region.as_str()));
+        if let Some(throughput_target_gbps) = self.maximum_throughput_gbps {
+            client_config = client_config.throughput_target_gbps(throughput_target_gbps as f64);
+        }
+        if let Some(limit_gib) = self.crt_memory_limit_gib {
+            client_config = client_config.memory_limit_in_bytes(limit_gib * 1024 * 1024 * 1024);
```
renanmagagnin, line 148: np: as discussed offline, it would be ideal to improve how we do these unit conversations (applies to the whole code base)

PR #1526, opened 2025-07-19
Introduce UploaderConfig

passaro:
Group configuration parameters to initialize the `Uploader` component into a new struct. It will make easier to introduce new parameters in future changes.

No. Internal change only.

No.

---

### DIFF
--- a/mountpoint-s3-fs/src/fs.rs
+++ b/mountpoint-s3-fs/src/fs.rs
@@ -25,7 +25,7 @@ use crate::prefix::Prefix;
 use crate::superblock::{Superblock, SuperblockConfig};
 use crate::sync::atomic::{AtomicU64, Ordering};
 use crate::sync::{Arc, AsyncMutex, AsyncRwLock};
-use crate::upload::Uploader;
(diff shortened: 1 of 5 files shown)


passaro:
This is an incremental step in preparation of #1511.

mountpoint-s3-fs/src/fs.rs
```diff
@@ -166,10 +166,10 @@ where
             client.clone(),
             runtime,
             mem_limiter,
-            config.storage_class.to_owned(),
-            config.server_side_encryption.clone(),
-            client.write_part_size().unwrap(),
-            config.use_upload_checksums.then_some(ChecksumAlgorithm::Crc32c),
+            UploaderConfig::new(client.write_part_size().unwrap())
```
renanmagagnin, line 169: should we avoid the unwrap here?

passaro: We should probably make the part size getters always return a value (rather than `Option`). It's already the case that no implementation returns `None`. But out of scope for this change.

PR #1527, opened 2025-07-21
Automated Benchmark Result Upload

tadiwa-aizen:
### What changed and why?

This PR adds functionality to automatically upload benchmark results to an S3 bucket when benchmarks complete. The implementation includes:

1. A new `detect_result_folder()` function that determines the appropriate result folder path and source path based on Hydra's runtime configuration
2. A new `upload_results_to_s3()` function that uses AWS CLI to sync local benchmark results to the specified S3 bucket

These changes enable automated collection of benchmark results in a centralized S3 location, making it easier to analyze performance trends over time.

No.

No

---

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -72,13 +72,12 @@ def upload_results_to_s3(bucket_name: str, region: str) -> None:
     hydra_config = HydraConfig.get()
 
     if hydra_config.mode == RunMode.MULTIRUN:
-        parts = Path(hydra_config.runtime.output_dir).parent.parts
+        source_path = Path(hydra_config.runtime.output_dir).parent
 
-        date_part, time_part = parts[-2:]
+        assert len(source_path.parts) >= 2, "Source path must have at least 2 parts for date/time extraction"
+        date_part, time_part = source_path.parts[-2:]
 
         s3_target_path = f"s3://{bucket_name}/results/{date_part}/{time_part}"
-
-        source_path = Path(hydra_config.runtime.output_dir).parent
 
         aws_cmd = [
             "aws",
@@ -91,7 +90,7 @@ def upload_results_to_s3(bucket_name: str, region: str) -> None:
         ]
         result = subprocess.run(aws_cmd, capture_output=True, text=True)
         if result.returncode == 0:
-            log.info("Successfully uploaded multirun benchmark results to S3")
+            log.info("Successfully uploaded benchmark results to S3")
         else:
             log.error(f"S3 upload failed: {result.stderr.strip()}")
     else:
@@ -238,6 +237,17 @@ def run_experiment(cfg: DictConfig) -> None:
 
         # Mark success if we get here without exceptions
         metadata["success"] = True
+
+        result_bucket_name = common_config.get("s3_result_bucket")
+
+        # If region is not specified, default to 'us-east-1' as that is the only region we can be relavtively assued that tranium instances are available
+        region = common_config.get("region", "us-east-1")
+        if result_bucket_name:
+            log.info(f"Uploading benchmark results to S3 bucket '{result_bucket_name}'")
+            upload_results_to_s3(result_bucket_name, region)
+        else:
+            log.info("No results bucket specified (s3_result_bucket), skipping upload")
+
     except Exception:
         log.error("Benchmark execution failed:", exc_info=True)
         raise
@@ -249,17 +259,6 @@ def run_experiment(cfg: DictConfig) -> None:
         finally:
             write_metadata(metadata)
             metadata["end_time"] = datetime.now(tz=timezone.utc)
-
-            result_bucket_name = common_config.get("s3_result_bucket")
-
-            # If region is not specified, default to 'us-east-1' as that is the only region we can be relavtively assued that tranium instances are available
-            region = common_config.get("region", "us-east-1")
-
-            if result_bucket_name:
-                log.info(f"Uploading benchmark results to S3 bucket '{result_bucket_name}'")
-                upload_results_to_s3(result_bucket_name, region)
-            else:
-                log.info("No results bucket specified (s3_result_bucket), skipping upload")
 
 
 if __name__ == "__main__":


benchmark/benchmark.py
```diff
@@ -200,18 +256,27 @@ def run_experiment(cfg: DictConfig) -> None:
...
             benchmark.post_process()
-        except Exception:
-            log.error("Post-processing failed:", exc_info=True)
+        except Exception as e:
+            log.error(f"Post-processing failed: {str(e)}")
         finally:
             write_metadata(metadata)
             metadata["end_time"] = datetime.now(tz=timezone.utc)
 
+            bucket_name = common_config.get('s3_bucket')
```
c-hagem: Should we have a `s3_result_bucket` in the configuration to not conflate the bucket from which data is pulled with the bucket we are requesting data from?

sahityadg: +1, we should have a separate bucket for storing results and avoid reading/parsing of results from interfering with benchmark runs.

benchmark/benchmark.py
```diff
@@ -60,6 +62,64 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
         log.error("Failed to write metadata", exc_info=True)
 
 
+def detect_result_folder() -> Tuple[str, str]:
+    """
+    Detect the result folder path and source path for benchmark results using Hydra's internal API.
+    """
+    local_path = HydraConfig.get().runtime.output_dir
```
sahityadg: Do we need this? Should we just sync the Hydra sweep dir to S3 to start off? We could partition the results on S3 further by instance-id

benchmark/benchmark.py
```diff
@@ -60,6 +62,64 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
...
+        date_time_path = parts[1].lstrip("/\\")
+        
+        path_parts = date_time_path.split("/")
+        if path_parts and path_parts[-1].isdigit():
+            # use parent to include multirun.yaml
+            local_path = str(Path(local_path).parent)
+            date_time_path = "/".join(path_parts[:-1])
+        
+        s3_path = os.path.join("multirun", date_time_path)
+    elif "outputs" in str(path):
```
sahityadg: Does benchmark.py ever create outputs dir, isn't the default set to multirun? It is probably better to throw an exception if the output dir is not as expected?

benchmark/benchmark.py
tadiwa-aizen: Noted, I've added a result bucket in the args, and a fallback that if its not entered ..it'll assume you want the output in the same bucket as the one used for mountpoint.

benchmark/benchmark.py
tadiwa-aizen: Noted, I've updated it to  syncing the Hydra sweep directory for multirun, and the results are partitioned by instance-id on S3.

benchmark/benchmark.py
tadiwa-aizen: Yes it does, the benchmarks can be run as single direct experiments with the  `hydra.mode=RUN` flag. I added it incase someone prefers to run a quick experiment that way. eg

` uv run benchmark.py -- s3_bucket=tadiwaom-mountpoint-test region=eu-west-1 hydra.mode=RUN application_workers=4 benchmarks.fio.direct_io=true benchmarks.fio.size_gib=10 mountpoint.fuse_threads=8`

benchmark/benchmark.py
```diff
@@ -60,6 +62,46 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
         log.error("Failed to write metadata", exc_info=True)
 
 
+def upload_results_to_s3(bucket_name: str, region: str = "us-east-1") -> None:
```
renanmagagnin: np: is there a good reason to make the default value of region `us-east-1`? it seems we are not using it

benchmark/benchmark.py
```diff
@@ -60,6 +62,46 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
...
+    """
+    Upload benchmark results to S3 bucket using the AWS CLI.
+    """
+    try:
+        source_path = HydraConfig.get().runtime.output_dir
+        path = Path(source_path)
+        path_str = str(path)
```
muddyfish: Isn't `source_path` already a string? What is this actually doing?

benchmark/benchmark.py
```diff
@@ -60,6 +62,46 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
...
+
+        instance_id = get_ec2_instance_id() or "local"
+
+        if "multirun" in path_str:
```
muddyfish: This is not portable. What if the parent path happens to contain one of these strings?

benchmark/benchmark.py
```diff
@@ -60,6 +62,46 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
...
+            parts = path_str.split("multirun")[1].lstrip("/\\").split("/")
```
muddyfish: Why are we doing string magic here rather than using the `Path` object we created earlier?

benchmark/benchmark.py
```diff
@@ -60,6 +62,46 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
...
+            date_part, time_part = parts[0], parts[1]
+            source_path = os.path.dirname(source_path)
+            s3_target_path = os.path.join("results", instance_id, "multirun", date_part, time_part)
```
muddyfish: Do not use `os.path.join` for S3 paths, as the separator isn't consistent between OSes. Use `"/".join` instead

benchmark/benchmark.py
```diff
@@ -200,18 +242,27 @@ def run_experiment(cfg: DictConfig) -> None:
 
         # Mark success if we get here without exceptions
         metadata["success"] = True
-    except Exception:
-        log.error("Benchmark execution failed:", exc_info=True)
+    except Exception as e:
+        log.error(f"Benchmark execution failed: {str(e)}")
```
muddyfish: Why are we changing this?

benchmark/benchmark.py
```diff
@@ -200,18 +242,27 @@ def run_experiment(cfg: DictConfig) -> None:
...
             benchmark.post_process()
-        except Exception:
-            log.error("Post-processing failed:", exc_info=True)
+        except Exception as e:
+            log.error(f"Post-processing failed: {str(e)}")
         finally:
             write_metadata(metadata)
             metadata["end_time"] = datetime.now(tz=timezone.utc)
 
+            result_bucket_name = common_config.get('s3_result_bucket') or common_config.get('s3_bucket')
```
muddyfish: It feels odd to have a fallback here - why not make the result bucket required?

benchmark/benchmark.py
```diff
@@ -200,18 +242,27 @@ def run_experiment(cfg: DictConfig) -> None:
...
+            region = common_config.get('region', 'us-east-1')
```
muddyfish: Using single quotes where everything else uses double quotes.

benchmark/benchmark.py
tadiwa-aizen: My thinking was if the they dont add a result bucket then we can just use their current bucket for MP, and if they specify one..then we use that.

sahityadg: We could keep it optional but skip uploading if there is no results bucket.

benchmark/benchmark.py
tadiwa-aizen: the uvx ruff check said it was an undefined variable

muddyfish: In the original code, it wasn't using a variable, so I'm not sure why that might be

benchmark/benchmark.py
```diff
@@ -60,6 +62,50 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
...
+    try:
+        output_path = Path(HydraConfig.get().runtime.output_dir)
+        parts = output_path.parts
+        if "multirun" in parts:
+            multirun_idx = parts.index("multirun")
+            if len(parts) > multirun_idx + 2:
+                date_part = parts[multirun_idx + 1]
+                time_part = parts[multirun_idx + 2]
+                source_path = output_path.parent
+                s3_target_path = f"results/{date_part}/{time_part}"
```
muddyfish: I'd either inline this in the only use, or include the s3:// prefix here

benchmark/benchmark.py
tadiwa-aizen: Removed it from this part

benchmark/benchmark.py
tadiwa-aizen: Noted, i included the s3:// prefix in the path variable

benchmark/benchmark.py
```diff
@@ -60,6 +63,56 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
...
+
+            source_path = output_path.parent
+
+            parts = output_path.parts
+            date_idx = len(parts) - 3
+            time_idx = len(parts) - 2
+
+            if len(parts) >= 3:
+                date_part = parts[date_idx]
+                time_part = parts[time_idx]
```
muddyfish: `date_part, time_part = parts[-3:-1]`?

tadiwa-aizen: Noted

benchmark/benchmark.py
```diff
@@ -60,6 +63,53 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
         log.error("Failed to write metadata", exc_info=True)
 
 
+def upload_results_to_s3(bucket_name: str, region: str) -> None:
+    """
+    Upload benchmark results to S3 bucket using the AWS CLI.
+    Only uploads results from multirun directories.
+    """
+    try:
```
renanmagagnin: Remove the try catch and considering having during call

benchmark/benchmark.py
```diff
@@ -60,6 +63,53 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
...
+        hydra_config = HydraConfig.get()
+
+        if hydra_config.mode == RunMode.MULTIRUN:
+            output_path = Path(hydra_config.runtime.output_dir)
+
+            source_path = output_path.parent
```
renanmagagnin:
```suggestion
            output_path, source_path = Path(hydra_config.runtime.output_dir), output_path.parent
```

benchmark/benchmark.py
```diff
@@ -60,6 +63,53 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
...
+
+            parts = output_path.parts
+
+            if len(parts) >= 3:
+                date_part, time_part = parts[-3:-1]
```
renanmagagnin: We should make this part more readable. On idea that was discussed: break up the path into its parts, find the latest multirun occurence and get the next two parts. However, if we can assert that output_path.parts, we can just unpack it:
```
 date_part, time_part, _ = parts
```

renanmagagnin (changes requested):

benchmark/benchmark.py
tadiwa-aizen: Noted, thanks

benchmark/benchmark.py
```diff
@@ -206,12 +244,23 @@ def run_experiment(cfg: DictConfig) -> None:
     finally:
         try:
             benchmark.post_process()
-        except Exception:
-            log.error("Post-processing failed:", exc_info=True)
+        except Exception as e:
+            log.error(f"Post-processing failed: {str(e)}")
```
muddyfish: Why are we changing this?

benchmark/benchmark.py
```diff
@@ -60,6 +63,41 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
...
+def upload_results_to_s3(bucket_name: str, region: str) -> None:
+    """
+    Upload benchmark results to S3 bucket using the AWS CLI.
+    Only uploads results from multirun directories.
+    """
+
+    hydra_config = HydraConfig.get()
+
+    if hydra_config.mode == RunMode.MULTIRUN:
+        parts = Path(hydra_config.runtime.output_dir).parent.parts
```
renanmagagnin: Use `source_path` in this declaration.

benchmark/benchmark.py
```diff
@@ -60,6 +63,41 @@ def write_metadata(metadata: Dict[str, Any]) -> None:
...
+
+        date_part, time_part = parts[-2:]
```
renanmagagnin: We should assert that len of parts is at least two. Also make sure to explain the context for this to make the code more readable.

muddyfish (approved):

PR #1528, opened 2025-07-21
Support read sizes up to 1M for fio cached IO benchmarks. 

sahityadg:
This change allows configuring fio read sizes for benchmarks. For cached IO reads over 256K, the script overwrites the read_ahead_kb on the device itself. 

This also includes minor cleanup fixes from previous changes. 

No, benchmark scripts only

No, benchmark scripts only

---


benchmark/benchmarks/fio_benchmark.py
```diff
@@ -24,6 +25,24 @@ def __init__(self, cfg: DictConfig, metadata: Dict[str, Any]):
         self.fio_config = self.config_parser.get_fio_config()
         self.mount_dir = None
         self.target_pid = None
+        self.dev_id = None
+
+    def _get_dev_id(mount_dir):
```
c-hagem: When is this method called? Why does it not take `self` and updates the `self.dev_id`?

sahityadg: Sorry, missed including all the changes I had locally. Added them now.

PR #1529, opened 2025-07-21
Add custom memory pool implementation

passaro:
Introduce a custom implementation of a `MemoryPool` which can be used by the CRT S3 client. The new pool will eventually be adopted in Mountpoint, which will also use it to replace the allocations for disk cache blocks and incremental upload buffers.

This change extends the integration tests on the client crate to run with this pool implementation.

See docs in `memory/pool.rs` for more details on the new memory pool.

No, the new pool is only used in tests for now. 

No.

---

### DIFF
--- a/mountpoint-s3-client/tests/common/mod.rs
+++ b/mountpoint-s3-client/tests/common/mod.rs
@@ -78,6 +78,11 @@ pub fn set_up_client_config(config: S3ClientConfig) -> S3ClientConfig {
     #[cfg(feature = "pool_tests")]
     let config = config.memory_pool(memory_pool::new_for_tests());
 
+    #[cfg(feature = "fs_pool_tests")]
+    let config = config.memory_pool_factory(|options: mountpoint_s3_client::config::MemoryPoolFactoryOptions| {
+        mountpoint_s3_fs::memory::PagedPool::new([options.part_size()])
+    });
+
     config
 }
 
--- a/mountpoint-s3-fs/src/memory/buffers.rs
+++ b/mountpoint-s3-fs/src/memory/buffers.rs
@@ -1,0 +1,223 @@
+use bytes::Bytes;
+
+use crate::sync::Arc;
+
+use super::pages::PagedBufferPtr;
+use super::stats::{BufferKind, PoolStats};
+
+/// Pool buffer.
+#[derive(Debug)]
+pub struct PoolBuffer(PoolBufferInner);
+
+#[derive(Debug)]
+enum PoolBufferInner {
+    /// Buffer from the paged pool.
+    Primary { buffer_ptr: PagedBufferPtr, size: usize },
+    /// Buffer allocated independently.
+    Secondary(FreeBuffer),
+}
+
+impl PoolBuffer {
+    pub(super) fn new_primary(buffer_ptr: PagedBufferPtr, size: usize) -> Self {
+        assert!(size <= buffer_ptr.size());
+        Self(PoolBufferInner::Primary { buffer_ptr, size })
+    }
+
+    pub(super) fn new_secondary(size: usize, kind: BufferKind, stats: Arc<PoolStats>) -> Self {
+        Self(PoolBufferInner::Secondary(FreeBuffer::new(size, kind, stats)))
+    }
+
+    pub fn capacity(&self) -> usize {
+        match &self.0 {
+            PoolBufferInner::Primary { size, .. } => *size,
+            PoolBufferInner::Secondary(boxed) => boxed.data.len(),
+        }
+    }
+
+    pub fn into_bytes(self) -> Bytes {
+        Bytes::from_owner(self)
+    }
+}
+
+impl AsMut<[u8]> for PoolBuffer {
+    fn as_mut(&mut self) -> &mut [u8] {
+        match &mut self.0 {
+            PoolBufferInner::Primary { buffer_ptr, size } => {
+                // SAFETY: returned slice will be valid until this buffer is dropped.
+                unsafe { std::slice::from_raw_parts_mut(buffer_ptr.as_raw_ptr(), *size) }
+            }
+            PoolBufferInner::Secondary(boxed) => &mut boxed.data,
+        }
+    }
+}
+
+impl AsRef<[u8]> for PoolBuffer {
+    fn as_ref(&self) -> &[u8] {
+        match &self.0 {
+            PoolBufferInner::Primary { buffer_ptr, size } => {
+                // SAFETY: returned slice will be valid until this buffer is dropped.
+                unsafe { std::slice::from_raw_parts(buffer_ptr.as_raw_ptr(), *size) }
+            }
+            PoolBufferInner::Secondary(boxed) => &boxed.data,
+        }
+    }
+}
+
+/// A mutable buffer backed by the pool.
+#[derive(Debug)]
+pub struct PoolBufferMut {
+    buffer: PoolBuffer,
+    len: usize,
+}
+
+impl PoolBufferMut {
+    pub fn new(buffer: PoolBuffer) -> Self {
+        Self { buffer, len: 0 }
+    }
+
+    /// Append data from a slice to the end of this buffer. If reaching the buffer capacity,
+    /// split off the overflowing subslice.
+    ///
+    /// Returns the overflowing subslice.
+    pub fn append_from_slice<'a>(&mut self, slice: &mut &'a [u8]) -> &'a [u8] {
+        let available = self.buffer.capacity() - self.len;
+        let overflow = slice.split_off(available.min(slice.len())..).unwrap();
+        let new_len = self.len + slice.len();
+        self.buffer.as_mut()[self.len..new_len].copy_from_slice(slice);
+        self.len = new_len;
+        overflow
+    }
+
+    /// Set the length of this buffer, up to its capacity. When extending the
+    /// current length, exposes uninitialized data.
+    ///
+    /// Returns the resulting length of the buffer (<= capacity).
+    pub fn set_len_uninit(&mut self, length: usize) -> usize {
+        self.len = self.buffer.capacity().min(length);
+        self.len
+    }
+
+    pub fn is_full(&self) -> bool {
+        self.len == self.buffer.capacity()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.len == 0
+    }
+
+    pub fn len(&self) -> usize {
+        self.len
+    }
+
+    pub fn capacity(&self) -> usize {
+        self.buffer.capacity()
+    }
+
+    pub fn into_bytes(self) -> Bytes {
+        Bytes::from_owner(self)
+    }
+}
+
+impl AsRef<[u8]> for PoolBufferMut {
+    fn as_ref(&self) -> &[u8] {
+        &self.buffer.as_ref()[..self.len]
+    }
+}
+
+impl AsMut<[u8]> for PoolBufferMut {
+    fn as_mut(&mut self) -> &mut [u8] {
+        &mut self.buffer.as_mut()[..self.len]
+    }
+}
+
+#[derive(Debug)]
+struct FreeBuffer {
+    data: Box<[u8]>,
+    kind: BufferKind,
+    stats: Arc<PoolStats>,
+}
+
+impl FreeBuffer {
+    fn new(size: usize, kind: BufferKind, stats: Arc<PoolStats>) -> Self {
+        let data = vec![0u8; size].into_boxed_slice();
+        stats.reserve_bytes(data.len(), kind);
+        Self { data, kind, stats }
+    }
+}
+
+impl Drop for FreeBuffer {
+    fn drop(&mut self) {
+        self.stats.release_bytes(self.data.len(), self.kind);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::super::pages::tests::create_page;
+
+    use super::*;
+
+    use test_case::{test_case, test_matrix};
+
+    fn primary(buffer_size: usize) -> PoolBuffer {
+        let page = create_page(buffer_size);
+        let buffer_ptr = page
+            .try_reserve(BufferKind::Other)
+            .expect("should be able to reserve a buffer from a new page");
+
+        PoolBuffer::new_primary(buffer_ptr, buffer_size)
+    }
+
+    fn secondary(buffer_size: usize) -> PoolBuffer {
+        PoolBuffer::new_secondary(buffer_size, BufferKind::Other, Arc::new(PoolStats::default()))
+    }
+
+    #[test_case(primary)]
+    #[test_case(secondary)]
+    fn test_pool_buffer(create_fn: fn(usize) -> PoolBuffer) {
+        const BUFFER_SIZE: usize = 1024;
+        let mut pool_buffer = create_fn(BUFFER_SIZE);
+
+        assert_eq!(pool_buffer.capacity(), BUFFER_SIZE);
+
+        let data = &[42u8; BUFFER_SIZE];
+        pool_buffer.as_mut().copy_from_slice(data);
+
+        assert_eq!(pool_buffer.as_ref(), data);
+
+        let bytes = pool_buffer.into_bytes();
+        assert_eq!(bytes.as_ref(), data);
+    }
+
+    const BUFFER_SIZE: usize = 1024;
+    const SMALL_WRITE_SIZE: usize = 512;
+    const LARGE_WRITE_SIZE: usize = 1280;
+
+    #[test_matrix([primary, secondary], [SMALL_WRITE_SIZE, BUFFER_SIZE, LARGE_WRITE_SIZE])]
+    fn test_pool_buffer_mut(create_fn: fn(usize) -> PoolBuffer, write_size: usize) {
+        let mut pool_buffer = PoolBufferMut::new(create_fn(BUFFER_SIZE));
+
+        assert_eq!(pool_buffer.capacity(), BUFFER_SIZE);
+        assert_eq!(pool_buffer.len(), 0);
+        assert!(pool_buffer.is_empty());
+        assert!(!pool_buffer.is_full());
+
+        let slice = pool_buffer.as_ref();
+        assert!(slice.is_empty());
+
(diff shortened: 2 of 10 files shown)


mountpoint-s3-fs/src/memory/buffers.rs
```diff
@@ -0,0 +1,223 @@
+use bytes::Bytes;
+
+use crate::sync::Arc;
+
+use super::pages::PagedBufferPtr;
+use super::stats::{BufferKind, PoolStats};
+
+/// Pool buffer.
```
muddyfish: Documentation doesn't add anything - it's just the name of the struct.

mountpoint-s3-fs/src/memory/buffers.rs
```diff
@@ -0,0 +1,223 @@
...
+        self.buffer.as_mut()[self.len..new_len].copy_from_slice(slice);
+        self.len = new_len;
+        overflow
+    }
+
+    /// Set the length of this buffer, up to its capacity. When extending the
+    /// current length, exposes uninitialized data.
+    ///
+    /// Returns the resulting length of the buffer (<= capacity).
+    pub fn set_len_uninit(&mut self, length: usize) -> usize {
```
muddyfish: Is this something we want to expose? It seems unideal to expose uninitialised memory.

mountpoint-s3-fs/src/memory/buffers.rs
```diff
@@ -0,0 +1,223 @@
...
+        assert_eq!(pool_buffer.len(), 0);
+        assert!(pool_buffer.is_empty());
+        assert!(!pool_buffer.is_full());
+
+        let slice = pool_buffer.as_ref();
+        assert!(slice.is_empty());
+
+        let data = vec![42u8; write_size];
+        let mut slice = &data[..];
+        let remaining = pool_buffer.append_from_slice(&mut slice);
```
muddyfish, line 217: Given we're using the same buffer size, shouldn't the remaining always be empty here? If so, this isn't testing too much

mountpoint-s3-fs/src/memory/buffers.rs
```diff
@@ -0,0 +1,223 @@
...
+        let bytes = pool_buffer.into_bytes();
+        assert_eq!(bytes.as_ref(), data);
+    }
+
+    const BUFFER_SIZE: usize = 1024;
+    const SMALL_WRITE_SIZE: usize = 512;
+    const LARGE_WRITE_SIZE: usize = 1280;
+
+    #[test_matrix([primary, secondary], [SMALL_WRITE_SIZE, BUFFER_SIZE, LARGE_WRITE_SIZE])]
+    fn test_pool_buffer_mut(create_fn: fn(usize) -> PoolBuffer, write_size: usize) {
```
muddyfish: This could do with a larger matrix for write sizes and buffer sizes to check the 'remining' part

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -0,0 +1,246 @@
...
+/// buffers of a fixed size.
+#[derive(Debug, Clone)]
+pub struct Page {
+    inner: Arc<PageInner>,
+}
+
+#[derive(Debug)]
+struct PageInner {
+    bytes: *mut u8,
+    last_offset: *mut u8,
```
muddyfish: Can't this be computed from other fields in the structure? And this appears to only be used in asserts, so might not be used in release aside from possibly in `Debug`

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -0,0 +1,246 @@
...
+        }
+        // Prevent further use.
+        *bitmask = FULL_MASK;
+        true
+    }
+}
+
+const FULL_MASK: u16 = 0xFFFF;
+
+fn consume(bitmask: &mut u16) -> Option<(usize, PageStatus)> {
```
muddyfish, line 156: More comments here would be ideal - when is it meant to return empty vs non-empty (empty only if the page was previously never used?), and the number is the index in the underlying buffer?

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -0,0 +1,246 @@
...
+    ///
+    /// Returns [None] if the page is already full.
+    pub fn try_reserve(&self, kind: BufferKind) -> Option<PagedBufferPtr> {
+        let (index, page_status) = consume(&mut self.inner.free_bitmask.lock().unwrap())?;
+        let offset = index * self.inner.stats.buffer_size;
+        if let PageStatus::Empty = page_status {
+            self.inner.stats.remove_empty_page();
+        };
+
+        self.inner.stats.add_used_buffer(kind);
```
muddyfish: For metrics, would it also be useful to have a dimension on pages used by buffer? Or is that too much cardinality?

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -0,0 +1,246 @@
...
+        assert!(
+            ptr >= self.inner.bytes && ptr <= self.inner.last_offset,
+            "the pointer does not belong to this page"
+        );
+
+        // SAFETY: ptr points to the same allocated object as self.inner.bytes.
+        let offset = unsafe { ptr.offset_from(self.inner.bytes) };
+        let index = offset as usize / self.inner.stats.buffer_size;
+        let mask = !(1u16 << index);
+        self.inner.stats.remove_used_buffer(kind);
```
muddyfish: This stats thing should probably not be nestled in all the maths/logic

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -0,0 +1,246 @@
...
+    /// preventing any further use.
+    ///
+    /// Returns whether the page was invalidated. If `true` the page is ready to be
+    /// removed from the pool.
+    pub fn invalidate_if_empty(&self) -> bool {
+        let mut bitmask = self.inner.free_bitmask.lock().unwrap();
+        if *bitmask != 0 {
+            return false;
+        }
+        // Prevent further use.
```
muddyfish: Prevents further use by preventing new allocations by marking the pool as full. Releases can't happen as there are no active allocations.

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -0,0 +1,246 @@
...
+        *bitmask = FULL_MASK;
+        true
+    }
+}
+
+const FULL_MASK: u16 = 0xFFFF;
```
muddyfish: Can be written as `u16::MAX`, and then can the page be generic?

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -0,0 +1,246 @@
...
+    fn drop(&mut self) {
+        // SAFETY: `self.bytes` was allocated using `self.layout` in `Page::new`.
+        unsafe {
+            alloc::dealloc(self.bytes, self.layout);
+        }
+    }
+}
+
+impl Page {
+    const BUFFERS_PER_PAGE: usize = 16;
```
muddyfish: Can be written as `u16::BITS as usize`

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -0,0 +1,246 @@
...
+            self.inner.stats.remove_empty_page();
+        };
+
+        self.inner.stats.add_used_buffer(kind);
+
+        // SAFETY: ptr is in bounds of the allocated object, since offset < page_size.
+        let ptr = unsafe { self.inner.bytes.add(offset) };
+        Some(PagedBufferPtr {
+            ptr,
+            pool_page: self.clone(),
```
muddyfish, line 94: Should have a comment here - why is it safe to clone. This will prevent the child buffer from receiving updates to `free_bitmask`

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -0,0 +1,246 @@
...
+    use super::*;
+
+    pub fn create_page(buffer_size: usize) -> Page {
+        let pool_stats = PoolStats::default();
+        let stats = SizePoolStats::new(buffer_size, Arc::new(pool_stats));
+        Page::new(Arc::new(stats))
+    }
+
+    #[test]
+    fn test_reserve_buffers() {
```
muddyfish, line 212: Should also verify it becomes empty again after deallocating all buffers

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -0,0 +1,345 @@
...
+#[derive(Debug, Clone)]
+pub struct PagedPool {
+    inner: Arc<PagedPoolInner>,
+}
+
+impl PagedPool {
+    /// Creates a new pool with the given set of buffer sizes.
+    ///
+    /// Duplicate sizes are ignored.
+    pub fn new(buffer_sizes: impl Into<Vec<usize>>) -> Self {
```
muddyfish, line 51: A set might make slightly more semantic sense here, but might also make the interface harder to satisfy

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    fn trim(&self) -> bool {
+        self.trim()
+    }
+}
+
+#[derive(Debug)]
+struct PagedPoolInner {
+    size_pools: Vec<SizePool>,
```
muddyfish: Should have an assertion that the size pools are ordered by size

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -0,0 +1,345 @@
...
+        Some(&self.size_pools[index])
+    }
+
+    fn trim(&self) -> bool {
+        let mut removed = false;
+        for pool in &self.size_pools {
+            if pool.stats.empty_pages() == 0 {
+                continue;
+            }
+            let mut write = pool.pages.write().unwrap();
```
muddyfish, line 177: Bad variable name. Can't think of something immediately obviously better

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -0,0 +1,345 @@
...
+impl SizePool {
+    fn reserve(&self, kind: BufferKind) -> PagedBufferPtr {
+        {
+            let read_pages = self.pages.read().unwrap();
+            if let Some(buffer_ptr) = self.try_get_buffer_ptr(read_pages.iter(), kind) {
+                return buffer_ptr;
+            }
+        }
+
+        let mut write_pages = self.pages.write().unwrap();
```
muddyfish, line 220: Why do we need to do this twice?

mountpoint-s3-fs/src/memory/stats.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl<T> Index<BufferKind> for [T; BUFFER_KIND_COUNT] {
+    type Output = T;
+    fn index(&self, idx: BufferKind) -> &Self::Output {
+        &self[idx as usize]
+    }
+}
+
+impl BufferKind {
+    pub fn as_str(&self) -> &'static str {
```
muddyfish, line 98: I remember discussing using a library for this at some point when implementing https://github.com/awslabs/mountpoint-s3/blob/641f613c0339c1ba6bc9c53a55d37dc358f73b55/mountpoint-s3/src/data_cache.rs#L45-L56. Is this something we want to revisit?

mountpoint-s3-fs/src/memory/buffers.rs
passaro: It is unideal, but it allows to use a `PoolBufferMut` with [`Read`](https://doc.rust-lang.org/std/io/trait.Read.html) implementations. I'll mention it in the docs.

mountpoint-s3-fs/src/memory/buffers.rs
passaro: `LARGE_WRITE_SIZE` is larger than `BUFFER_SIZE` and will result in a non-empty `remaining` slice.

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -0,0 +1,246 @@
...
+            }
+        }
+    }
+    None
+}
+
+#[derive(Debug)]
+enum PageStatus {
+    Empty,
+    NonEmpty,
```
c-hagem, line 177: Should the page keep track of the largest unallocated chunk?

mountpoint-s3-fs/src/memory/buffers.rs
passaro: Will rework into a "fill" method taking in the `Read` implementation, so we can avoid exposing the uninitialized buffer.

mountpoint-s3-fs/src/memory/buffers.rs
```diff
@@ -220,4 +227,24 @@ mod tests {
         let bytes = pool_buffer.into_bytes();
         assert_eq!(bytes.as_ref(), &data[..appended_size]);
     }
+
+    #[test_matrix([primary, secondary], [SMALLER_THEN_BUFFER_SIZE, BUFFER_SIZE, LARGER_THAN_BUFFER_SIZE])]
+    fn test_pool_buffer_mut_fill(create_fn: fn(usize) -> PoolBuffer, read_size: usize) {
```
muddyfish, line 232: Should also check the behaviour when the buffer already has some content in it. Success cases should leave the buffer full, and failures should leave the length unchanged.

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -111,18 +122,37 @@ impl Page {
...
     /// removed from the pool.
     pub fn invalidate_if_empty(&self) -> bool {
-        let mut bitmask = self.inner.free_bitmask.lock().unwrap();
+        let mut bitmask = self.inner.reserved_bitmask.lock().unwrap();
         if *bitmask != 0 {
             return false;
         }
-        // Prevent further use.
+        self.inner.stats.remove_empty_page();
+        // Mark the page as full, even if no buffer is currently reserved. If a new request
```
muddyfish: *even though no buffer is currently reserved. We just checked this, and the bitmask is still locked, so we know it can't have changed

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -100,13 +103,15 @@ impl PagedPool {
         match self.inner.get_pool_for_size(size) {
             Some(pool) => {
                 let buffer_ptr = pool.reserve(kind);
-                metrics::histogram!("pool.reserve", "type" => "primary", "kind" => kind.as_str()).record(size as f64);
+                metrics::histogram!("pool.reserved_bytes", "type" => "primary", "kind" => kind.as_str())
+                    .record(size as f64);
                 let slack = pool.buffer_size() - size;
-                metrics::histogram!("pool.slack", "kind" => kind.as_str()).record(slack as f64);
+                metrics::histogram!("pool.slack_bytes", "kind" => kind.as_str()).record(slack as f64);
```
muddyfish: Is it worth adding `"type" => "primary"` here for future compatibility if we add a third type?

muddyfish: Also, do you think it's worth including the pool size here? That would help us deciding if we need to add new sizes

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -0,0 +1,352 @@
...
+/// maintain a [SizePool] instance, which will allocate "pages" of
+/// [BUFFERS_PER_PAGE](super::pages::Page::BUFFERS_PER_PAGE) (16) buffers on demand.
+///
+/// When a buffer of a given size is requested, the pool will return a buffer from one
+/// of 2 logical areas: **primary** and **secondary**:
+///
+/// - **primary**: the [SizePool] with the smallest buffer size large enough to contain
+///   the requested size is selected. If any of its existing pages has a free buffer,
+///   it is marked as reserved and returned, otherwise a new page is allocated.
+/// - **secondary**: if the requested size is larger than the maximum buffer size in
```
vladem, line 29: it's handy to have this implemented, but for the sake of understanding, will we use it in near future?

mountpoint-s3-client/tests/common/mod.rs
```diff
@@ -78,6 +78,11 @@ pub fn set_up_client_config(config: S3ClientConfig) -> S3ClientConfig {
     #[cfg(feature = "pool_tests")]
     let config = config.memory_pool(memory_pool::new_for_tests());
 
+    #[cfg(feature = "fs_pool_tests")]
+    let config = config.memory_pool_factory(|options: mountpoint_s3_client::config::MemoryPoolFactoryOptions| {
+        mountpoint_s3_fs::memory::PagedPool::new([options.part_size()])
```
vladem, line 83: looks that `MemoryPoolFactoryOptions` doesn't have all of the buffer sizes that will be required to create a pool in Mountpoint (e.g. it doesn't know anything about size of the cache block)

in Mountpoint, will this pool be constructed on the higher level and be just cloned in factory? `MemoryPoolFactoryOptions` will be ignored?

passaro: Yes. Mountpoint controls how the client is created, so already knows which part sizes will be used and can ignore the factory here. Also, it needs to share the pool across other components, like the disk cache. See #1511 for how we plan to integrate it.

mountpoint-s3-fs/src/memory/buffers.rs
```diff
@@ -247,4 +247,35 @@ mod tests {
...
+        if read_size + INITIAL_SIZE >= BUFFER_SIZE {
+            result.expect("fill from large enough slice should succeed");
+            assert_eq!(&pool_buffer.as_ref()[0..INITIAL_SIZE], &initial[..]);
+            assert_eq!(
+                &pool_buffer.as_ref()[INITIAL_SIZE..],
+                &data[..BUFFER_SIZE - INITIAL_SIZE]
+            );
+        } else {
+            result.expect_err("fill from small slice should fail");
+            assert_eq!(pool_buffer.len(), INITIAL_SIZE);
```
muddyfish: Should we also assert that the first part of the buffer is the initial data?

muddyfish (approved):

PR #1532, opened 2025-07-23
Unaligned

c-hagem:
Adds a metric to track and sets initial read window size to 2MB so that reads are aligned with the standard S3 part size and less reads cross part boundaries.

---

### DIFF
--- a/mountpoint-s3-fs/src/prefetch/part_stream.rs
+++ b/mountpoint-s3-fs/src/prefetch/part_stream.rs
@@ -290,6 +290,12 @@ where
         pin_mut!(request_stream);
         while let Some(next) = request_stream.next().await {
             let GetBodyPart { offset, data: mut body } = next?;
+            trace!("got a body part at offset {offset}");
+            const DEFAULT_PART_SIZE: u64 = 8 * 1024 * 1024;
+            if offset % DEFAULT_PART_SIZE != 0 {
+                metrics::counter!("s3.client.part_size_not_aligned", "part_size" => self.preferred_part_size.to_string())
+                    .increment(1);
+            }
             // pre-split the body into multiple parts as suggested by preferred part size
             // in order to avoid validating checksum on large parts at read.
             let mut curr_offset = offset;
--- a/mountpoint-s3-fs/src/s3/config.rs
+++ b/mountpoint-s3-fs/src/s3/config.rs
@@ -264,12 +264,12 @@ impl TargetThroughputSetting {
 // FUSE doesn't know the difference between regular reads and readahead reads, it will
 // send us a READ request for that 128k, so we'll have to block waiting for it even if
 // the application doesn't want it. This is all in the noise for sequential IO, but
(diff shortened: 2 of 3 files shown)


mountpoint-s3-fs/src/prefetch/part_stream.rs
```diff
@@ -290,6 +290,12 @@ where
         pin_mut!(request_stream);
         while let Some(next) = request_stream.next().await {
             let GetBodyPart { offset, data: mut body } = next?;
+            trace!("got a body part at offset {offset}");
+            const DEFAULT_PART_SIZE: u64 = 8 * 1024 * 1024;
+            if offset % DEFAULT_PART_SIZE != 0 {
+                metrics::counter!("s3.client.part_size_not_aligned", "part_size" => self.preferred_part_size.to_string())
```
muddyfish, line 296: We're not using the actual part size here, so this is guesswork rather than definitely accurate. Is this a problem?

c-hagem: It's the best we can do without implementing a bunch of logic

passaro: Not sure what we are trying to track here. The counter will be incremented on every part (except the first) when reading a file sequentially from the start.
There is also (pre-existing!) confusion between the S3 client part size and the `preferred_part_size` which applies to `PartQueue` parts and (badly) attempts to track read sizes.

mountpoint-s3-fs/src/s3/config.rs
```diff
@@ -264,12 +264,12 @@ impl TargetThroughputSetting {
...
 // the application doesn't want it. This is all in the noise for sequential IO, but
-// waiting for the readahead hurts random IO. So we add 128k to the first request size
-// to avoid the latency hit of the second request.
+// waiting for the readahead hurts random IO.
+// So we request 2MB to stay aligned with the part size (8MB).
 //
 // Note the CRT does not respect this value right now, they always return chunks of part size
 // but this is the first window size we prefer.
-pub const INITIAL_READ_WINDOW_SIZE: usize = 1024 * 1024 + 128 * 1024;
+pub const INITIAL_READ_WINDOW_SIZE: usize = 2 * 1024 * 1024;
```
mansi153, line 272: Is there some context here on why 2MB instead of the original or another value (say 1MB)? Also, is this really a problem if the read requests from Fuse come in with 128K sizes or does it only become a problem with larger read sizes? Or does it only on part_size of the check-summed parts and is immune to incoming read requests sizes?

mountpoint-s3-fs/src/prefetch/part_stream.rs
c-hagem: From my understanding, the counter is only incremented if a part we get from the crt is not aligned with the expected (CRT) part size. When changing to 2MB initial read size this metric is never emitted when using 256k reads.

PR #1538, opened 2025-07-25
Fix issue preventing incremental upload to handle very large write part sizes

passaro:
The append upload queue tries to limit the total memory used to buffer the data to write to 2 GiB. However, when setting  `--write-part-size` to values greater than 2 GiB, it would incorrectly set the queue capacity to 0 buffers and panic.

This change ensures that the queue allows for at least 1 buffer, even if that means exceeding the 2 GiB cap.

No.

Bug fix entry.

---

### DIFF
--- a/mountpoint-s3-fs/src/upload.rs
+++ b/mountpoint-s3-fs/src/upload.rs
@@ -140,6 +140,9 @@ where
         initial_offset: u64,
         initial_etag: Option<ETag>,
     ) -> AppendUploadRequest<Client> {
+        // Limit the queue capacity to hold buffers for a total of at most
+        // MAX_BYTES_IN_QUEUE, but ensure it allows at least 1 buffer.
+        let capacity = (MAX_BYTES_IN_QUEUE / self.buffer_size).max(1);
         let params = AppendUploadQueueParams {
             bucket,
             key,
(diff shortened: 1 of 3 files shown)


mountpoint-s3-fs/src/upload.rs
```diff
@@ -140,14 +140,17 @@ where
         initial_offset: u64,
         initial_etag: Option<ETag>,
     ) -> AppendUploadRequest<Client> {
+        // Limit the queue capacity to hold buffers for a total of at most
+        // MAX_BYTES_IN_QUEUE, but ensure it allows at least 1 buffer.
+        let capacity = (MAX_BYTES_IN_QUEUE / self.buffer_size).max(1);
```
mansi153, line 145: The change itself LGTM, but I just wondered if there is a `self.buffer_size > 0` check somewhere before we do this division?

passaro: No, there isn't. We should add one here, maybe with an assertion, and also a check on the part size (or write part size), which is where the value come from.

mansi153: I looked into the code and if I understand correctly, we have a check for read/write part-sizes being in the range [5MiB, 5GiB] in the CrtClientConfig. However, there are some tests and testing trait-implementations which don't do the same validation. So I'll add a >0 check for the `buffer_size` to avoid indirect exceptions here, but keep the write_part_size checks as it is.

PR #1540, opened 2025-07-25
Enforce valid buffer sizes for the memory pool

passaro:
The memory pool will only accept buffer sizes in the range (0, 64MiB] for the primary memory (i.e. allocated in pages of 16 buffers). For larger sizes, it will only use secondary memory (i.e. ad-hoc allocation for a single buffer).

The 64MiB cap reproduces the behavior of the internal CRT memory pool.

No.

No - part of the memory pool change.

---

### DIFF
--- a/mountpoint-s3-fs/src/memory/pool.rs
+++ b/mountpoint-s3-fs/src/memory/pool.rs
@@ -45,18 +45,35 @@ pub struct PagedPool {
 }
 
 impl PagedPool {
-    /// Creates a new pool with the given set of buffer sizes.
+    /// Maximum size for a primary buffer.
     ///
-    /// Duplicate sizes are ignored.
-    pub fn new(buffer_sizes: impl Into<Vec<usize>>) -> Self {
-        let mut buffer_sizes: Vec<_> = buffer_sizes.into();
-        buffer_sizes.sort();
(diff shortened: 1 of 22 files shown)


mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -45,18 +45,35 @@ pub struct PagedPool {
...
+
+    /// Create a new pool, configuring primary memory with the given set of
+    /// buffer sizes, if valid.
+    ///
+    /// Ignores invalid (0 or greater than [MAX_BUFFER_SIZE](Self::MAX_BUFFER_SIZE))
+    /// or duplicate values for buffer sizes. If no valid value is provided,
+    /// uses [DEFAULT_BUFFER_SIZE](Self::DEFAULT_BUFFER_SIZE).
+    pub fn new_with_candidate_sizes(buffer_sizes: impl Into<Vec<usize>>) -> Self {
+        let mut ordered_sizes: Vec<_> = buffer_sizes.into();
+        ordered_sizes.retain(|&size| size > 0 && size <= Self::MAX_BUFFER_SIZE);
```
muddyfish, line 66: NB: Later we should make this configurable in some way

PR #1543, opened 2025-07-28
Add non-zero check for incremental upload buffer sizes

mansi153:
Add non-zero check for upload buffer-sizes.

This is done so that we avoid runtime exceptions for invalid sized buffers and division by 0 during capacity calculations in current or future incremental uploads. The buffer_size currently is derived from `write-part-size`. We have a check for read/write part-sizes being in the range [5MiB, 5GiB] in the `CrtClientConfig`. However, there are some tests and testing trait-implementations which don't do the same validation and hence need explicit checks in the upload constructor.

No.

No, minor code enhancement entry.

---

### DIFF
--- a/mountpoint-s3-fs/src/upload.rs
+++ b/mountpoint-s3-fs/src/upload.rs
@@ -116,6 +116,7 @@ where
         mem_limiter: Arc<MemoryLimiter>,
         config: UploaderConfig,
     ) -> Self {
+        assert!(config.buffer_size > 0, "buffer_size should be greater than 0 for uploads");
         Self {
             client,
             runtime,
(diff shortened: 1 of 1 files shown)


mountpoint-s3-fs/src/upload.rs
```diff
@@ -152,6 +152,7 @@ where
         initial_offset: u64,
         initial_etag: Option<ETag>,
     ) -> AppendUploadRequest<Client> {
+        assert!(self.buffer_size > 0, "write-part-size should be greater than 0 for incremental uploads");
```
passaro: This seems the wrong place to check. Why not `UploaderConfig::new` (where buffer_size is passed)?

Also, the message here should be related to this type, rather than user-facing. `--write-part-size` is a CLI argument and should be validated elsewhere.

passaro (changes requested):

mountpoint-s3-fs/src/upload.rs
c-hagem: Maybe we should consider making the size (NonZero)[https://doc.rust-lang.org/std/num/type.NonZeroU32.html]?

passaro:
Closing: it's an internal defensive check and would need rebasing on months of changes.

PR #1545, opened 2025-07-28
Add versioning of the configuration format in mount_from_config example

vladem:
Example binary `mount_from_config` now accepts `config_version` parameter. This may be used to ensure that user is aware of updates to the configuration format and prevent from silent failures.

No.

No.

---

### DIFF
--- a/mountpoint-s3-fs/examples/mount_from_config.rs
+++ b/mountpoint-s3-fs/examples/mount_from_config.rs
@@ -24,6 +24,8 @@ use serde::Deserialize;
 use tempfile::tempdir_in;
 use tracing::info;
 
+const CONFIG_VERSION: &str = "0.0.1";
+
 #[derive(Debug, Deserialize)]
 #[serde(tag = "type")]
 enum ThroughputConfig {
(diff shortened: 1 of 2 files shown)


mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -248,6 +263,7 @@ fn main() -> Result<()> {
     let args = Args::parse();
     // Read the config
     let config = load_config(&args.config).context("Failed to load config")?;
+    config.validate_version()?;
```
muddyfish: We'd want to validate the version before loading the full json struct, so we can give the incorrect version message even if we add new fields.

mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -167,7 +172,29 @@ impl ConfigOptions {
     }
 }
 
+/// Reads the config_version field from a JSON config file and validates it against CONFIG_VERSION
+fn validate_config_version<P: AsRef<Path>>(path: P) -> Result<()> {
+    #[derive(Deserialize)]
```
muddyfish, line 177: Would using [`from_str`](https://docs.rs/serde_json/latest/serde_json/fn.from_str.html) multiple times avoid needing to re-open the file?

PR #1546, opened 2025-07-29
 Update prefetcher to increment client read window linearly with each read

mansi153:
_Rebased version of https://github.com/awslabs/mountpoint-s3/pull/1453, PRed to run CI benchmarks and analyse performance/behaviour_

Mountpoint's S3 client has a backpressure mechanism to controlling how much data to fetch. This change updates the way Mountpoint's prefetcher signals to Mountpoint's S3 client (using the AWS CRT internally) to fetch more data ahead of where a consuming application is reading to accelerate throughput.

Before this change, Mountpoint would wait for 50% of the existing window to be consumed. For example with a window of 2GiB, 1GiB must be read by the kernel before Mountpoint would inform the S3 client to fetch more data up to 2GiB ahead of the current position.

After this change, Mountpoint now sends this signal with every read by the kernel. For example, a 128KiB read by the Kernel to fill a page in the page cache will result in the CRT being updated to be 2GiB ahead of the end offset of the 128KiB read, where the readahead window has a size of 2GiB.

We observe improved throughput for single file handle sequential reading with this approach. For random reading and for multiple file handle reading, we don't see an observable change in throughput. We expect this may be a prerequisite for driving higher throughput with multiple file handles, with this potentially being one bottleneck among others.

This improves the way Mountpoint signals progress to its S3 client. We expect improvements in throughput, but the end-user behavior hasn't changed in a meaningful way.

A changelog entry has been added to note the change in algorithm, alongside ensuring a minor version bump. This is added to communicate the change, in case any issue is raised around the change in behavior. However, we do not expect any regressions given the benchmarking performed.

---

### DIFF
--- a/mountpoint-s3-fs/CHANGELOG.md
+++ b/mountpoint-s3-fs/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Adopt a unified memory pool to reduce overall memory usage. ([#1511](https://github.com/awslabs/mountpoint-s3/pull/1511))
 * Replace `S3Uri` with `S3Path` and consolidate related types like `Bucket` and `Prefix` into the `s3` module.
   ([#1535](https://github.com/awslabs/mountpoint-s3/pull/1535))
+* `PrefetchGetObject` now has an updated backpressure algorithm advancing the read window with each call to `PrefetchGetObject::read`, with the aim of higher sequential-read throughput. ([#1453](https://github.com/awslabs/mountpoint-s3/pull/1453))
 
 ## v0.6.0 (July 23, 2025)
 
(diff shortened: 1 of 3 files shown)


mountpoint-s3-fs/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Adopt a unified memory pool to reduce overall memory usage. ([#1511](https://github.com/awslabs/mountpoint-s3/pull/1511))
 * Replace `S3Uri` with `S3Path` and consolidate related types like `Bucket` and `Prefix` into the `s3` module.
   ([#1535](https://github.com/awslabs/mountpoint-s3/pull/1535))
+* `PrefetchGetObject` now has an updated backpressure algorithm advancing the read window with each call to `PrefetchGetObject::read`, with the aim of higher sequential-read throughput. ([#1453](https://github.com/awslabs/mountpoint-s3/pull/1453))
```
mansi153, line 8: I'll fix this and add the appropriate changelog entry and version updates

PR #1552, opened 2025-07-31
Add CLI option for publishing OTEL metrics to an endpoint 

15skumar:
This PR adds a CLI argument which enables users to run Mountpoint with the functionality of publishing metrics to a specified endpoint.

No impact on existing behavior. This CLI option is under a compile time flag.

Added functionality:
Run Mountpoint with --log-metrics-otlp http://localhost:4318 flag to enable publishing metrics to port 4318 (otlp port)
Optionally you can also specify the exporting interval with the --log-metrics-otlp-interval flag.

To verify the implementation I ran a docker container running the OpenTelemetry Collector at the default port, and ran Mountpoint with the new flag with endpoint specified.
I verified that the Mountpoint metrics were visible in the collector logs. Here is a screenshot of an example Mountpoint metric collected at the endpoint:

Screenshot 2025-07-31 at 17 39 57

---

### DIFF
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -280,6 +280,7 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
     #[clap(long, help = "Enable logging of summarized performance metrics", help_heading = LOGGING_OPTIONS_HEADER)]
     pub log_metrics: bool,
 
+    #[cfg(feature = "otlp_integration")]
     #[clap(
         long,
         help = "Enable OTLP metrics export to the specified endpoint",
@@ -288,6 +289,7 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
     )]
     pub log_metrics_otlp: Option<String>,
 
+    #[cfg(feature = "otlp_integration")]
     #[clap(
         long,
         help = "OTLP metrics export interval in seconds [default: 5]",
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -280,6 +280,24 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
     #[clap(long, help = "Enable logging of summarized performance metrics", help_heading = LOGGING_OPTIONS_HEADER)]
     pub log_metrics: bool,
 
+    #[clap(
+        long,
+        help = "Enable OTLP metrics export to the specified endpoint",
+        help_heading = LOGGING_OPTIONS_HEADER,
+        value_name = "ENDPOINT"
+    )]
+    pub log_metrics_otlp: Option<String>,
+
+    #[clap(
+        long,
+        help = "OTLP metrics export interval in seconds [default: 5]",
+        help_heading = LOGGING_OPTIONS_HEADER,
+        value_name = "SECONDS",
+        value_parser = value_parser!(u64).range(1..),
+        requires = "log_metrics_otlp"
+    )]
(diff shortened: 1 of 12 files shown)


mountpoint-s3/src/cli.rs
```diff
@@ -279,6 +279,26 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
...
+    )]
+    pub log_metrics_otlp: Option<String>,
+
+    #[cfg(feature = "otlp_integration")]
+    #[clap(
+        long,
+        help = "OTLP metrics export interval in seconds [default: 5]",
+        help_heading = LOGGING_OPTIONS_HEADER,
+        value_name = "SECONDS",
+        value_parser = value_parser!(u64).range(1..),
```
sahityadg, line 297: I think we should restrict this to a reasonable value

mountpoint-s3/src/cli.rs
```diff
@@ -279,6 +279,26 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
...
     pub log_metrics: bool,
 
+    #[cfg(feature = "otlp_integration")]
+    #[clap(
+        long,
+        help = "Enable OTLP metrics export to the specified endpoint",
+        help_heading = LOGGING_OPTIONS_HEADER,
+        value_name = "ENDPOINT"
+    )]
+    pub log_metrics_otlp: Option<String>,
```
sahityadg: Should we call this otlp_endpoint instead, as that's what we are expecting users to configure.

mountpoint-s3/src/cli.rs
```diff
@@ -279,6 +279,26 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
     #[clap(long, help = "Enable logging of summarized performance metrics", help_heading = LOGGING_OPTIONS_HEADER)]
     pub log_metrics: bool,
 
+    #[cfg(feature = "otlp_integration")]
+    #[clap(
+        long,
+        help = "Enable OTLP metrics export to the specified endpoint",
```
sahityadg: Should we instead say, OTLP endpoint url for publishing metrics. Does this also require log_metrics?

mountpoint-s3/src/cli.rs
```diff
@@ -279,6 +279,26 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
...
+    #[cfg(feature = "otlp_integration")]
+    #[clap(
+        long,
+        help = "OTLP metrics export interval in seconds [default: 5]",
+        help_heading = LOGGING_OPTIONS_HEADER,
+        value_name = "SECONDS",
+        value_parser = value_parser!(u64).range(1..),
+        requires = "log_metrics_otlp"
+    )]
+    pub log_metrics_otlp_interval: Option<u64>,
```
sahityadg: otlp_export_interval?

sahityadg (approved):

mountpoint-s3/src/cli.rs
15skumar: Yes, the otel meter provider has a default interval of 60 seconds, so that should be a reasonable default value (it is already implemented as 60 I have just forgotten to change the text here)

mountpoint-s3/src/cli.rs
15skumar: Yes that makes sense

mountpoint-s3/src/cli.rs
15skumar: It doesn't require log_metrics

sahityadg (approved):

sahityadg (approved):

PR #1554, opened 2025-08-04
Add benchmark sweeper configuration and auto-override for benchmark type

tadiwa-aizen:
### What changed and why?

Fixed an issue where benchmark parameters for unselected benchmark types were incorrectly included in multirun sweeps. Previously, all benchmark parameters were defined in the common sweeper config, causing irrelevant parameters to be swept even when running specific benchmark types.

__Changes:__

- Added auto-detection of `benchmark_type` parameter to automatically select appropriate sweeper config
- Split benchmark-specific sweep parameters into separate config files (`fio.yaml`, `prefetch.yaml`, `client-bp.yaml`, etc.)
- Moved common parameters to `base.yaml` sweeper config
- Now only relevant parameters are swept for each benchmark type
- Added tests that can be run via `uv add pytest --dev` and `uv run pytest tests/` to test the filtering and combination logic.

__Example__: Running `benchmark_type=fio` now only sweeps FIO-specific parameters instead of including unrelated prefetch or client-backpressure parameters.

No breaking changes. Existing commands work as before, but now sweep only relevant parameters for the specified benchmark type.

None needed

---

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -342,7 +342,7 @@ def run_experiment(cfg: DictConfig) -> None:
         benchmark = CrtBenchmark(cfg, metadata)
     elif benchmark_type == "client":
         benchmark = ClientBenchmark(cfg, metadata)
-    elif benchmark_type == "client-bp":
+    elif benchmark_type == "client_bp":
         benchmark = ClientBenchmark(cfg, metadata, backpressure=True)
     else:
         raise ValueError(f"Unsupported benchmark type: {benchmark_type}")
--- a/benchmark/conf/config.yaml
+++ b/benchmark/conf/config.yaml
@@ -5,20 +5,21 @@ defaults:
   - _self_
   - override hydra/job_logging: colorlog
   - override hydra/hydra_logging: colorlog
+  - override hydra/sweeper: multi
 
 # ===== Common parameters for all benchmarks =====
 s3_bucket: ???
 s3_result_bucket: !!null
 application_workers: 1
 iteration: 0
-iterations: 1
+iterations: 10
 run_time: 30  # Default run time in seconds
 read_size: 262144 # Defaults to 256KiB, can go up to 1MiB.
 read_part_size: !!null
 region: "us-east-1"
 write_part_size: 16777216  # 16 MiB, to allow for uploads of large files
 object_size_in_gib: 100  # Size of the object to benchmark
-benchmark_type: "fio" # fio, prefetch, client, client-bp, crt
+benchmark_type: "fio" # fio, prefetch, client, client_bp, crt
 s3_keys: !!null
 download_checksums: true
 crt_eventloop_threads: !!null # Number of ELG thread count
@@ -57,6 +58,7 @@ benchmarks:
       - sequential_read
     fio_benchmark: "${benchmarks.fio.fio_benchmarks[0]}"
     fio_io_engine: "psync"
+    fuse_threads: !!null
 
   prefetch:
     max_memory_target: !!null # memory upper-limit in MB
@@ -67,7 +69,7 @@ benchmarks:
   client:
     # None
 
-  client_backpressure:
+  client_bp:
     read_window_size: !!null #2147483648
 
 
@@ -77,14 +79,3 @@ hydra:
   mode: MULTIRUN
   job:
     chdir: true
-  sweeper:
-    # Global sweeper params - use this for common parameters across all benchmarks
-    # For specific parameter use sweep params under the specific benchmark type config
-    params:
-      'application_workers': 1, 4, 16, 64, 128
-      'iteration': "range(${iterations})"
-      'mountpoint.fuse_threads': 1, 16, 64
-      'benchmarks.fio.direct_io': false, true
-      # 'benchmarks.prefetch.max_memory_target': !!null, 1073741824, 2147483648  # null, 1GB, 2GB
-      #'benchmarks.client_backpressure.read_window_size': 8388608, 2147483648
-      'crt_eventloop_threads': !!null # 128 (default for Trn1), 256, 1024
--- a/benchmark/conf/hydra/sweeper/base.yaml
+++ b/benchmark/conf/hydra/sweeper/base.yaml
@@ -1,0 +1,8 @@
+# Base sweeper configuration with common parameters for all benchmark types
+_target_: hydra._internal.core_plugins.basic_sweeper.BasicSweeper
+max_batch_size: null
+params:
+  'application_workers': 1, 16 # 4, 64
+  'iteration': "range(${iterations})"
+  'network.maximum_throughput_gbps': 400
+  # 'crt_eventloop_threads': 128, 256, 1024
--- a/benchmark/conf/hydra/sweeper/fio.yaml
+++ b/benchmark/conf/hydra/sweeper/fio.yaml
@@ -1,0 +1,4 @@
+# @package hydra.sweeper
+params:
+  'benchmarks.fio.direct_io': false, true
+  'benchmarks.fio.fuse_threads': 16 #1, 64
--- a/benchmark/conf/hydra/sweeper/multi.yaml
+++ b/benchmark/conf/hydra/sweeper/multi.yaml
@@ -1,0 +1,10 @@
+# @package hydra.sweeper
+defaults:
+  - base
+  - fio
+  - prefetch
+  - client_bp
+  - client
+  - crt
+
+_target_: hydra_plugins.smart_sweeper.smart_benchmark_sweeper.SmartBenchmarkSweeper
--- a/benchmark/conf/hydra/sweeper/prefetch.yaml
+++ b/benchmark/conf/hydra/sweeper/prefetch.yaml
@@ -1,0 +1,3 @@
+# @package hydra.sweeper
+params: {}
+  #'benchmarks.prefetch.max_memory_target': null, 1073741824, 2147483648  # null, 1GiB, 2GiB
--- a/benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
+++ b/benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
@@ -1,0 +1,143 @@
+from dataclasses import dataclass
+import itertools
+import logging
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+import re
+from hydra.types import HydraContext
+from hydra.core.config_store import ConfigStore
+from hydra.core.override_parser.overrides_parser import OverridesParser
+from hydra.core.override_parser.types import Glob, Override
+from hydra.core.plugins import Plugins
+from hydra.plugins.launcher import Launcher
+from hydra.plugins.sweeper import Sweeper
+from hydra.types import TaskFunction
+from omegaconf import DictConfig, OmegaConf
+
+log = logging.getLogger(__name__)
+
+
+@dataclass
+class SmartBenchmarkSweeperConf:
+    _target_: str = "hydra_plugins.smart_sweeper.smart_benchmark_sweeper.SmartBenchmarkSweeper"
+    max_batch_size: Optional[int] = None
+    params: Optional[Dict[str, str]] = None
+
+
+ConfigStore.instance().store(group="hydra/sweeper", name="smart_benchmark", node=SmartBenchmarkSweeperConf)
+
+
+class SmartBenchmarkSweeper(Sweeper):
+    BENCHMARK_TYPE_REGEX = re.compile(r'^benchmarks\.([^.]+)\.')
+
+    def __init__(self, max_batch_size: Optional[int] = None, params: Optional[Dict[str, str]] = None):
+        self.max_batch_size = max_batch_size
+        self.params = params or {}
+        self.config: Optional[DictConfig] = None
+        self.launcher: Optional[Launcher] = None
+        self.hydra_context: Optional[HydraContext] = None
+
+    def _glob_match_parameter(
+        self, param_key: str, benchmark_type: str, exclude_patterns: Optional[List[str]] = None
+    ) -> bool:
+        """
+        Check if a parameter belongs to a specific benchmark type using Hydra's GLOB.
+        Only matches exact benchmark types: fio, prefetch, client, crt.
+        """
+        pattern = f"benchmarks.{benchmark_type}.*"
+        glob_filter = Glob(include=[pattern], exclude=exclude_patterns or [])
+        return len(glob_filter.filter([param_key])) > 0
+
+    def _process_overrides_to_sweep_lists(self, overrides: List[Override]) -> List[List[str]]:
+        lists = []
+        for override in overrides:
+            if override.is_sweep_override():
+                key = override.get_key_element()
+                sweep = [f"{key}={val}" for val in override.sweep_string_iterator()]
+                lists.append(sweep)
+            else:
+                key = override.get_key_element()
+                value = override.get_value_element_as_str()
+                lists.append([f"{key}={value}"])
+        return lists
+
+    def setup(self, *, hydra_context: HydraContext, task_function: TaskFunction, config: DictConfig) -> None:
+        self.config = config
+        self.launcher = Plugins.instance().instantiate_launcher(
+            hydra_context=hydra_context, task_function=task_function, config=config
+        )
+        self.hydra_context = hydra_context
+
+    def sweep(self, arguments: List[str]) -> Any:
+        benchmark_types = self._extract_benchmark_types(arguments)
+        log.info(f"Running benchmark types: {benchmark_types}")
+
+        # Save sweep config
+        sweep_dir = Path(self.config.hydra.sweep.dir)
+        sweep_dir.mkdir(parents=True, exist_ok=True)
+        OmegaConf.save(self.config, sweep_dir / "multirun.yaml")
+
+        params_conf = []
+        for k, v in self.params.items():
+            params_conf.append(f"{k}={v}")
+        params_conf.extend(arguments)
+
+        parser = OverridesParser.create()
+        parsed = parser.parse_overrides(params_conf)
+
+        all_combinations = []
+        for benchmark_type in benchmark_types:
+            type_combinations = self._generate_combinations_for_type(benchmark_type, parsed)
+            all_combinations.extend(type_combinations)
+
+        log.info(f"Generated {len(all_combinations)} total combinations")
+
+        returns = []
+        initial_job_idx = 0
+        if all_combinations:
+            self.validate_batch_is_legal(all_combinations)
+            results = self.launcher.launch(all_combinations, initial_job_idx=initial_job_idx)
+            returns.append(results)
+
+        return returns
+
+    def _extract_benchmark_types(self, arguments: List[str]) -> List[str]:
+        for arg in arguments:
+            if arg.startswith("benchmark_type="):
+                benchmark_type_str = arg.split("=", 1)[1]
+                return [bt.strip() for bt in benchmark_type_str.split(",")]
+        return ["fio"]
+
+    def _generate_combinations_for_type(self, benchmark_type: str, parsed_overrides: List[Override]) -> List[List[str]]:
+        common_overrides = []
+        type_specific_overrides = []
+        other_type_nulls = set()
+
+        for param_override in parsed_overrides:
+            param_key = param_override.get_key_element()
+            # Skip benchmark_type parameter - we handle it separately
+            if param_key == "benchmark_type":
+                continue
+
+            if param_key.startswith("benchmarks."):
+                if self._glob_match_parameter(param_key, benchmark_type):
+                    type_specific_overrides.append(param_override)
+                else:
+                    other_type_nulls.add(param_key)
+            else:
+                common_overrides.append(param_override)
+
+        param_lists = [[f"benchmark_type={benchmark_type}"]]
+
+        if common_overrides:
+            param_lists.extend(self._process_overrides_to_sweep_lists(common_overrides))
+
+        if type_specific_overrides:
+            param_lists.extend(self._process_overrides_to_sweep_lists(type_specific_overrides))
+
+        combinations = list(itertools.product(*param_lists))
+
+        null_assignments = [f"{param_key}=null" for param_key in sorted(other_type_nulls)]
+
+        result = [list(combination) + null_assignments for combination in combinations]
+        return result or [[f"benchmark_type={benchmark_type}"] + null_assignments]
--- a/benchmark/conf/hydra/sweeper/client.yaml
+++ b/benchmark/conf/hydra/sweeper/client.yaml
@@ -1,0 +1,2 @@
+# @package hydra.sweeper
+params: {}
--- a/benchmark/conf/hydra/sweeper/client_bp.yaml
+++ b/benchmark/conf/hydra/sweeper/client_bp.yaml
@@ -1,0 +1,3 @@
+# @package hydra.sweeper
+params: {}
+  #'benchmarks.client_bp.read_window_size': 2147483648, 8589934592, 17179869184  # 2GB, 8GB, 16GB
--- a/benchmark/conf/hydra/sweeper/crt.yaml
+++ b/benchmark/conf/hydra/sweeper/crt.yaml
@@ -1,0 +1,2 @@
+# @package hydra.sweeper
+params: {}
--- a/benchmark/hydra_plugins/__init__.py
+++ b/benchmark/hydra_plugins/__init__.py
@@ -1,0 +1,1 @@
+__path__ = __import__('pkgutil').extend_path(__path__, __name__)
--- a/benchmark/pyproject.toml
+++ b/benchmark/pyproject.toml
@@ -9,3 +9,7 @@ dependencies = [
     "hydra-core>=1.3.2",
     "psutil>=7.0.0",
 ]
+[dependency-groups]
+dev = [
+    "pytest",
+]
--- a/benchmark/tests/test_smart_sweeper_combinations.py
+++ b/benchmark/tests/test_smart_sweeper_combinations.py
@@ -1,0 +1,81 @@
+from hydra.core.override_parser.overrides_parser import OverridesParser
+from hydra_plugins.smart_sweeper.smart_benchmark_sweeper import SmartBenchmarkSweeper
+
+
+class TestSmartSweeperCombinations:
+    def setup_method(self):
+        self.sweeper = SmartBenchmarkSweeper()
+        self.parser = OverridesParser.create()
+
+    def test_glob_parameter_matching(self):
+        """Test GLOB-based parameter matching for different benchmark types."""
+
+        # Test exact matching
+        assert self.sweeper._glob_match_parameter("benchmarks.fio.direct_io", "fio")
+        assert self.sweeper._glob_match_parameter("benchmarks.prefetch.max_memory", "prefetch")
+        assert self.sweeper._glob_match_parameter("benchmarks.client.read_window_size", "client")
+
+        # Test non-matching
+        assert not self.sweeper._glob_match_parameter("benchmarks.fio.direct_io", "prefetch")
+        assert not self.sweeper._glob_match_parameter("benchmarks.prefetch.max_memory", "fio")
+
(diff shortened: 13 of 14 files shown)


benchmark/conf/hydra/sweeper/base.yaml
```diff
@@ -0,0 +1,7 @@
+# Base sweeper configuration with common parameters for all benchmark types
+_target_: hydra._internal.core_plugins.basic_sweeper.BasicSweeper
+max_batch_size: null
+params:
+  'application_workers': 1, 4, 16, 64, 128
```
mansi153: So we don't miss it later, can we pls remove 128 app workers from here as discussed?

benchmark/conf/config.yaml
```diff
@@ -3,6 +3,7 @@
```
mansi153: I could be misunderstanding something but is it possible to move all this default configuration (upto line 31) into `base.yml` too and refer to that as `@hydra.main(version_base=None, config_path="hydra/sweeper", config_name="base")` or so? 

We could also then move the benchmark specific static/default configs to their respective config files, for e.g. `mountpoint.mountpoint_max_background: !!null  # or say, 64` can move into the Fio yml file so it's easier to see what all defaults we're configuring and which all are then being overridden in the sweeper

benchmark/conf/hydra/sweeper/base.yaml
```diff
@@ -0,0 +1,7 @@
...
+  'iteration': "range(${iterations})"
+  'mountpoint.fuse_threads': 1, 16, 64
```
mansi153: I don't think this should be here. It should either be in `fio.yml` for now or a separate mountpoint.yml which the Fio one (and others in the future) can inherit this configuration from

benchmark/conf/hydra/benchmark_sweeper/prefetch.yaml
```diff
@@ -0,0 +1,3 @@
+# @package hydra.sweeper
+params:
+  'benchmarks.prefetch.max_memory_target': null, 1073741824, 2147483648  # null, 1GB, 2GB
```
mansi153: Should this mem_limiter config really be turned on by default?

benchmark/conf/hydra/benchmark_sweeper/client-bp.yaml
```diff
@@ -0,0 +1,3 @@
+# @package hydra.sweeper
+params:
+  'benchmarks.client_backpressure.read_window_size': 8388608, 2147483648  # 8MB, 2GB
```
mansi153: While we're at it, let's also think about what reasonable values we would want here. Let's not put 8MB as a test, as it's too small and might not be too meaningful. I think 2GB, 8GB and maybe 16GB are the values I'd suggest testing here.

benchmark/conf/hydra/sweeper/base.yaml
```diff
@@ -0,0 +1,7 @@
+# Base sweeper configuration with common parameters for all benchmark types
+_target_: hydra._internal.core_plugins.basic_sweeper.BasicSweeper
+max_batch_size: null
+params:
+  'application_workers': 1, 4, 16, 64, 128
+  'iteration': "range(${iterations})"
```
mansi153, line 6: Can we pls make this default to 10 (or something >1) here?

benchmark/benchmark.py
```diff
@@ -195,6 +196,35 @@ def managed(target_pid, with_bwm=False, with_perf_stat=False):
...
+def _auto_add_benchmark_sweeper() -> None:
+    """
+    Automatically add +hydra/benchmark_sweeper=<type> based on benchmark_type parameter.
+    This allows users to just specify benchmark_type instead of the full command.
+    """
+    benchmark_type = None
+    for arg in sys.argv[1:]:
+        if arg.startswith("benchmark_type="):
+            benchmark_type = arg.split("=", 1)[1]
+            break
```
mansi153: Do we stay backward compatible and default to `benchmark_type=fio` here as well so we don't miss sweeping the extra params, given we already silently default to fio benchmark with `benchmark_type: "fio"`?

benchmark/benchmark.py
```diff
@@ -195,6 +196,35 @@ def managed(target_pid, with_bwm=False, with_perf_stat=False):
...
+        )
+
+        if not has_sweeper_override:
+            # we add the sweeper override before the benchmark_type argument and find position of benchmark_type argument
+            for i, arg in enumerate(sys.argv):
+                if arg.startswith("benchmark_type="):
+                    sys.argv.insert(i, sweeper_override)
+                    break
+            else:
+                sys.argv.insert(1, sweeper_override)
```
mansi153: I'm not sure I understand. Do you mind elaborating how this works or a sample usage? Where is the `hydra/benchmark_sweeper=` value coming from? 

Can we do something like below, or maybe I'm missing something?

```
def get_config_name():
    env = os.getenv("benchmark_type", "fio")
    config_map = {
        "fio": "config_fio",
        "prefetch": "config_prefetch", 
        "client_bp": "config_client_bp",
        "client": "config_client",
        "crt": "config_crt"
    }
    return config_map.get(env, "config")

@hydra.main(version_base=None, config_path="conf/hydra/sweeper", config_name=get_config_name())
def run_experiment(cfg: DictConfig) -> None:
    blah blah
```

benchmark/conf/hydra/benchmark_sweeper/prefetch.yaml
tadiwa-aizen: I'd assume we'd want want to see what the data looks like? We can always ovveride it as null if we dont want it in a benchmark run

benchmark/conf/hydra/benchmark_sweeper/client-bp.yaml
tadiwa-aizen: Noted

benchmark/conf/hydra/sweeper/base.yaml
tadiwa-aizen: Noted, I think you meant the iterations ? i set it as 10 in the config file because this is jsut a range tracker i think

benchmark/conf/hydra/sweeper/base.yaml
tadiwa-aizen: Noted, Moved it to FIO

benchmark/conf/config.yaml
tadiwa-aizen: Hmm it seems Hydra's design expects sweeper configs to contain only sweep parameters, not application defaults. So what i think you're suggesting isnt working. 

Reason is Hydra instantiates the sweeper with config.hydra.sweeper BEFORE the main config is fully composed. When application defaults are in the sweeper config, interpolations like "${benchmarks.fio.fio_benchmarks[0]}" fail because the sweeper config is resolved in isolation. its a thing with the sweeper instantiation process from the parent sweeper class.

benchmark/benchmark.py
tadiwa-aizen: Noted, i had set it up that if you dont specify anything you have the fio benchmarks run with the default params , which would be direct io=0,but this is better.

benchmark/benchmark.py
tadiwa-aizen: The hydra/benchmark_sweeper= values come from config files in conf/hydra/benchmark_sweeper/ (like fio.yaml, multi.yaml, etc.). Our _auto_add_benchmark_sweeper() function automatically detects your benchmark_type and adds the right sweeper.


Its a bit odd but Hydra expects cfg.hydra.sweeper to contain the sweeper configuration.(Source - hydra.py lines 154-160) So i made conf/hydra/benchmark_sweeper/ as a custom config group.

benchmark/benchmark.py
```diff
@@ -196,6 +197,42 @@ def managed(target_pid, with_bwm=False, with_perf_stat=False):
...
 
 
+def _auto_add_benchmark_sweeper() -> None:
+    """
+    Automatically add +hydra/benchmark_sweeper=<type> based on benchmark_type parameter.
+    This allows users to just specify benchmark_type instead of the full command.
+    """
+    benchmark_type = None
+    for arg in sys.argv[1:]:
+        if arg.startswith("benchmark_type="):
```
c-hagem: I guess this is the only one we have to special case?

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -0,0 +1,131 @@
...
+        # Separate parameters by category
+        common_params = []
+        type_specific_params = []
+        other_type_nulls = []
+
+        for override in parsed_overrides:
+            key = override.get_key_element()
+
+            if key == "benchmark_type":
+                continue  # we ekip benchmark_type - we'll add it manually for each combination as you can see below
```
c-hagem: nit: typo "skip"

benchmark/benchmark.py
```diff
@@ -196,6 +197,42 @@ def managed(target_pid, with_bwm=False, with_perf_stat=False):
...
+            benchmark_type = arg.split("=", 1)[1]
```
muddyfish:
```
arg.removeprefix("benchmark_type=")
```

benchmark/benchmark.py
muddyfish: Why do we need to special case here? Why can't we use whatever we normally use for CLI argument parsing?

benchmark/benchmark.py
```diff
@@ -196,6 +197,42 @@ def managed(target_pid, with_bwm=False, with_perf_stat=False):
...
+    if benchmark_type:
+        config_name = "multi" if "," in benchmark_type else benchmark_type
+        sweeper_override = f"+hydra/benchmark_sweeper={config_name}"
+        has_sweeper_override = any(
+            arg.startswith("+hydra/benchmark_sweeper=") or arg.startswith("hydra/benchmark_sweeper=")
+            for arg in sys.argv[1:]
+        )
+
+        if not has_sweeper_override:
+            # we add the sweeper override before the benchmark_type argument and find position of benchmark_type argument
```
muddyfish: Is there any way we can avoid mutating `sys.argv` here?

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
tadiwa-aizen: good catch

benchmark/benchmark.py
tadiwa-aizen: Okay so , ive spent some time looking into a way to avoid this and their essentially 3 alternative options each with their own cons that i dont think we should bear.

So the idea is to **avoid mutating sys.argv** 

---

**Option 1 — Programmatic API (`initialize` / `compose`)**
Instead of `@hydra.main`, we manually init Hydra and compose the config:

```python
from hydra import initialize, compose

with initialize(config_path="conf"):
    cfg = compose(config_name="config", overrides=custom_overrides)
    run_experiment(cfg)
```

*Pros:* No `sys.argv` mutation
*Cons:* `compose()` skips multirun machinery (no sweeper/launcher/run dirs/cwd changes). We’d need to manually recreate Hydra’s execution pipeline, which is non-trivial and prone to drift from upstream. Doesn’t fit our current setup , and we would lose benchmark sweeper logic that generates multiple parameter combinations.

So...❌

---

**Option 2 — Call Hydra internals (`get_args`, `_run_hydra`)**
Bypass `@hydra.main` but still run Hydra with a custom arg list:

```python
from hydra._internal.utils import get_args_parser, _run_hydra

args_parser = get_args_parser()
args = args_parser.parse_args(custom_overrides)
_run_hydra(args=args, args_parser=args_parser,
           task_function=run_experiment,
           config_path="conf", config_name="config")
```

*Pros:* Full Hydra behaviour (sweeper, launcher, cwd handling) with no `sys.argv` mutation.
*Cons:* Uses `hydra._internal.*` private APIs, which are unstable and may break between Hydra versions. Higher maintenance risk as its internal and not stable.. 

So.. ❌ again...

---

**Option 3 — Hydra-native config group**
Move sweeper choice into config files instead of injecting via CLI:

```yaml
# conf/config.yaml
defaults:
  - benchmark_type: fio

# conf/benchmark_type/fio.yaml
defaults:
  - override /hydra/benchmark_sweeper: fio
benchmark_type: fio
```

Run with:

```bash
uv benchmark.py benchmark_type=fio
```

*Pros:* 100% public/idiomatic; no globals; works with multirun; trivial to maintain.
*Cons:* Breaks our current CLI UX — we can’t do `benchmark_type=fio,crt` anymore without redesign. Would require users to switch to predefined multi configs or lists. 

Running multiple like this `benchmark_type=fio,crt` is extremely important to our aims..so this wont work either:
So.. ❌..

---

**Conclusion:**
Option 1 requires rebuilding large parts of Hydra’s execution logic; Option 3 changes the CLI semantics in a way that doesn’t work with our benchmarks; Option 2 works but relies on unstable internals.

So, ... given these trade-offs, we’re keeping the current `sys.argv`-injection approach because it preserves behaviour with minimal complexity and maintenance cost.
---

benchmark/benchmark.py
tadiwa-aizen: So, The sweeper config (+hydra/benchmark_sweeper=...) must be injected before Hydra parses arguments, because Hydra decides its sweeper before @hydra.main runs. Since Hydra isn't running yet, the only available place to find benchmark_type is by manually reading sys.argv. .from my python understanding this method is run an import time , and thats before the hydra main function...

@muddyfish  Yes — benchmark_type is the only argument we have to special-case.Since hydra decides which sweeper to use before it parses the config, so at that stage the only way to read benchmark_type is from the raw CLI (sys.argv). We can’t rely on our usual CLI/config parsing because it happens after sweeper selection.

This also answers this comment: https://github.com/awslabs/mountpoint-s3/pull/1554#discussion_r2256794254

This "pre-Hydra scan" lets us detect benchmark_type early.and auto inject the correct +hydra/benchmark_sweeper=... override so the right sweeper config is loaded.


**We only do this for benchmark_type because it’s the only parameter that directly controls which sweeper config Hydra must load before composition.**

benchmark/benchmark.py
tadiwa-aizen: Noted, thanks

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -82,50 +82,36 @@ def _extract_benchmark_types(self, arguments: List[str]) -> List[str]:
...
-            key = override.get_key_element()
-
-            if key == "benchmark_type":
-                continue  # we ekip benchmark_type - we'll add it manually for each combination as you can see below
-            elif key.startswith("benchmarks."):
-                param_type = key.split(".")[1]  # Extract benchmark type from "benchmarks.fio.direct_io" -> "fio"
-                if param_type == benchmark_type:
-                    type_specific_params.append(override)  # This parameter belongs to our current benchmark type
+        common, type_specific, other_type_nulls = [], [], set()
+        type_re = re.compile(r'^benchmarks\.([^.]+)\.')
```
muddyfish: Should be a global constant

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -82,50 +82,36 @@ def _extract_benchmark_types(self, arguments: List[str]) -> List[str]:
...
+
+        for ov in parsed_overrides:
```
muddyfish: Variable name too short (not informative)

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -82,50 +82,36 @@ def _extract_benchmark_types(self, arguments: List[str]) -> List[str]:
...
+            k = ov.get_key_element()
+            if k == "benchmark_type":
+                continue
+
+            m = type_re.match(k)
```
muddyfish: Variable name not useful

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -82,50 +82,36 @@ def _extract_benchmark_types(self, arguments: List[str]) -> List[str]:
...
-        for override in parsed_overrides:
-            key = override.get_key_element()
-
-            if key == "benchmark_type":
-                continue  # we ekip benchmark_type - we'll add it manually for each combination as you can see below
-            elif key.startswith("benchmarks."):
-                param_type = key.split(".")[1]  # Extract benchmark type from "benchmarks.fio.direct_io" -> "fio"
-                if param_type == benchmark_type:
-                    type_specific_params.append(override)  # This parameter belongs to our current benchmark type
+        common, type_specific, other_type_nulls = [], [], set()
```
muddyfish: Split this onto 3 lines

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -82,50 +82,36 @@ def _extract_benchmark_types(self, arguments: List[str]) -> List[str]:
         return ["fio"]
 
     def _generate_combinations_for_type(self, benchmark_type: str, parsed_overrides) -> List[List[str]]:
```
muddyfish: What type is `parsed_overrides`?

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -82,50 +82,36 @@ def _extract_benchmark_types(self, arguments: List[str]) -> List[str]:
...
+            if m:
+                if m.group(1) == benchmark_type:
+                    type_specific.append(ov)
                 else:
-                    other_type_nulls.append(f"{key}=null")  # Null out parameters for other benchmark types
+                    other_type_nulls.add(k)
+            else:
+                common.append(ov)
+
+        def expand(ov):
```
muddyfish: Why is this a nested function?

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -82,50 +82,36 @@ def _extract_benchmark_types(self, arguments: List[str]) -> List[str]:
...
+                return [f"{key}={v}" for v in ov.sweep_string_iterator()]
             else:
-                common_params.append(
-                    override
-                )  # Parameters that apply to ALL benchmark types (e.g., s3_bucket, mountpoint.fuse_threads)
+                return [f"{key}={ov.get_value_element_as_str()}"]
 
-        param_lists = []
+        param_lists = [[f"benchmark_type={benchmark_type}"]]
+        param_lists += [expand(ov) for ov in common]
```
muddyfish: Consider using `map` and the `*` operator

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -82,50 +82,36 @@ def _extract_benchmark_types(self, arguments: List[str]) -> List[str]:
...
+        param_lists += [expand(ov) for ov in type_specific]
 
-        param_lists.append([f"benchmark_type={benchmark_type}"])
+        combos = itertools.product(*param_lists)
+        nulls = [f"{k}=null" for k in sorted(other_type_nulls)]
```
muddyfish: Why sorted?

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -82,50 +82,36 @@ def _extract_benchmark_types(self, arguments: List[str]) -> List[str]:
...
-            combinations = list(itertools.product(*param_lists))
-            result = []
-            for combo in combinations:
-                # Add the null values for other benchmark types
-                combo_list = list(combo) + other_type_nulls
-                result.append(combo_list)
-            return result
-        else:
-            return [[f"benchmark_type={benchmark_type}"] + other_type_nulls]
+        result = [list(c) + nulls for c in combos]
```
muddyfish: Why? Add comments

benchmark/conf/hydra/benchmark_sweeper/prefetch.yaml
renanmagagnin: these are GiB (not GB), right?

benchmark/conf/config.yaml
```diff
@@ -61,21 +63,11 @@ benchmarks:
     # None 
 
   client_backpressure:
-    read_window_size: !!null #2147483648 
+    read_window_size: !!null #2147483648
```
renanmagagnin, line 73: might be worth explaining this is 2Gib in the comment

benchmark/conf/hydra/benchmark_sweeper/prefetch.yaml
tadiwa-aizen: Yes , thanks

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
tadiwa-aizen: My thinking is that list comprehension is more 'Pythony', and easier to understand for people

benchmark/benchmarks/client_benchmark.py
```diff
@@ -16,7 +16,8 @@ def __init__(self, cfg: DictConfig, metadata: Dict[str, Any], backpressure=False
         self.backpressure = backpressure
         self.config_parser = BenchmarkConfigParser(cfg)
         self.common_config = self.config_parser.get_common_config()
-        self.client_config = self.config_parser.get_client_config()
+        config_section = 'client_backpressure' if backpressure else 'client'
+        self.client_config = self.config_parser.get_client_config(config_section)
```
sahityadg: Instead, should we just move client_backpressure config to client section in config? The assumption is there will be configuration that may be common to both client and client_backpressure.

benchmark/conf/config.yaml
```diff
@@ -50,7 +51,8 @@ benchmarks:
       - sequential_read
     fio_benchmark: "${benchmarks.fio.fio_benchmarks[0]}"
     fio_io_engine: "psync"
-  
+    fuse_threads: 64
```
sahityadg: Why are we changing the defaults? We could either not set it or set it to default of 16

benchmark/conf/hydra/sweeper/client-bp.yaml
```diff
@@ -0,0 +1,3 @@
+# @package hydra.sweeper
+params:
+  'benchmarks.client_backpressure.read_window_size': 2147483648, 8589934592, 17179869184  # 2GB, 8GB, 16GB
```
sahityadg: Do we want to sweep through all these values? Shouldn't this be 2G

benchmark/conf/hydra/sweeper/prefetch.yaml
```diff
@@ -0,0 +1,3 @@
+# @package hydra.sweeper
+params:
+  'benchmarks.prefetch.max_memory_target': null, 1073741824, 2147483648  # null, 1GiB, 2GiB
```
sahityadg: Should this be fixed to 2G?

mansi153: Should it not be just null? We're not setting a memory pressure for any other layer, why only here then?

Also @tadiwa-aizen the values need to be in MB and not bytes, AFAIR. So for setting a limit of 2GB, you pass 2048 here (not 2147483648)

benchmark/conf/hydra/sweeper/base.yaml
```diff
@@ -0,0 +1,6 @@
+# Base sweeper configuration with common parameters for all benchmark types
+_target_: hydra._internal.core_plugins.basic_sweeper.BasicSweeper
+max_batch_size: null
+params:
+  'application_workers': 1, 4, 16, 64
+  'iteration': "range(${iterations})"
```
mansi153: Why are we not overriding `network.maximum_throughput_gbps` to 400 here? Are we passing that as a command-line input instead? Also, same question for `network.interface_names` list

benchmark/conf/hydra/sweeper/multi.yaml
```diff
@@ -0,0 +1,11 @@
...
+defaults:
+  - base
+  - fio
+  - prefetch
+  - client-bp
+  - client
+  - crt
+
+_target_: hydra_plugins.smart_sweeper.smart_benchmark_sweeper.SmartBenchmarkSweeper
+max_batch_size: null
```
mansi153: Can you pls explain what this parameter is used for and how?

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -0,0 +1,154 @@
...
+                     ['benchmark_type=fio', 'network.interface_names=[ens32]', 'benchmarks.fio.direct_io=true', 'benchmarks.prefetch.max_memory_target=null']]
+        """
+        common = []
+        type_specific = []
+        other_type_nulls = set()
+
+        # This regex is essentially to extract benchmark type from parameter keys like "benchmarks.fio.direct_io"
+        # The pattern matches: "benchmarks." + (capture group: benchmark type) + "."
+
+        benchmark_type_mapping = {'client-bp': 'client_backpressure'}
```
mansi153: Maybe we just fix our config keys to be consistently named instead?

benchmark/conf/hydra/sweeper/multi.yaml
tadiwa-aizen: it was form the hydra documentation, it essentially controls how many jobs run simultaneously in each batch. I had thought it was mandatory to have it ,but it seems not, so ill remove it..

benchmark/conf/hydra/sweeper/fio.yaml
```diff
@@ -0,0 +1,4 @@
+# @package hydra.sweeper
+params:
+  'benchmarks.fio.direct_io': false, true
+  'benchmarks.fio.fuse_threads': 16 #1, 64
```
mansi153, line 4: I thought for 1 app-worker we also want to test with 1 fuse thread because we've seen it give better performance in the past, no?

benchmark/benchmarks/benchmark_config_parser.py
```diff
@@ -83,8 +83,9 @@ def get_crt_config(self) -> Dict[str, Any]:
             'crt_benchmarks_path': getattr(crt_cfg, 'crt_benchmarks_path', None),
         }
 
-    def get_client_config(self) -> Dict[str, Any]:
-        client_cfg = self.cfg.benchmarks.client
+    def get_client_config(self, config_section: str = 'client') -> Dict[str, Any]:
+        section_name = config_section.replace('-', '_')
```
mansi153: What is this doing and why?

benchmark/conf/config.yaml
```diff
@@ -57,6 +58,7 @@ benchmarks:
       - sequential_read
     fio_benchmark: "${benchmarks.fio.fio_benchmarks[0]}"
     fio_io_engine: "psync"
+    fuse_threads: 16
```
mansi153: Nit: I think this is not needed (and can be set to !!null) because we default to 16 in the MP code

benchmark/conf/hydra/sweeper/fio.yaml
tadiwa-aizen: Sure, i'll make it 1 and 16

mansi153 (dismissed):

benchmark/conf/config.yaml
```diff
@@ -57,6 +58,7 @@ benchmarks:
       - sequential_read
     fio_benchmark: "${benchmarks.fio.fio_benchmarks[0]}"
     fio_io_engine: "psync"
+    fuse_threads: !!null
```
mansi153, line 61: Not a blocker for this PR and sorry I missed it earlier, but not sure it's the best idea to have a MP.fuse_threads and a Fio.fuse_threads. If we can't merge the two config sets at runtime, let's keep them all under a single category (of "Fio" benchmark) to avoid confusion.

PR #1555, opened 2025-08-05
Add option to disable download checksums in performance tests

c-hagem:
Adds an option to our benchmarking code to disable verification of downloaded objects integrity.

Does not change existing behaviour, as it is only enabled when `EXPERIMENTAL_MOUNTPOINT_NO_DOWNLOAD_INTEGRITY_VALIDATION ` is set, and thus does not need a changelog entry.

---

### DIFF
--- a/benchmark/benchmarks/client_benchmark.py
+++ b/benchmark/benchmarks/client_benchmark.py
@@ -1,5 +1,6 @@
 import logging
 import subprocess
+import os
 from typing import Dict, Any
 
 from benchmarks.base_benchmark import BaseBenchmark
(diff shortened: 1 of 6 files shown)


benchmark/benchmarks/prefetch_benchmark.py
```diff
@@ -67,9 +68,16 @@ def run_benchmark(self) -> None:
             subprocess_args.extend(["--max-duration", str(run_time)])
 
         subprocess_args.extend(["--output-file", "prefetch-output.json"])
+        prefetch_env = {}
+        if not self.common_config['download_checksums']:
+            prefetch_env["EXPERIMENTAL_MOUNTPOINT_NO_DOWNLOAD_INTEGRITY_VALIDATION"] = "ON"
+
+        subprocess_env = os.environ.copy()
```
muddyfish: Nit: consider `os.environ | prefetch_env`

c-hagem: We use the same in the other scripts currently. Maybe let's fast-follow fix that?

benchmark/benchmarks/client_benchmark.py
```diff
@@ -70,8 +71,16 @@ def run_benchmark(self) -> Dict[str, Any]:
...
+        if not self.common_config['download_checksums']:
+            client_env["EXPERIMENTAL_MOUNTPOINT_NO_DOWNLOAD_INTEGRITY_VALIDATION"] = "ON"
+
+        subprocess_env = os.environ.copy()
+        subprocess_env.update(client_env)
+        log.debug("Subprocess env: %s", subprocess_env)
+
         log.info("Running client benchmark with args: %s", subprocess_args)
-        subprocess.run(subprocess_args, check=True, capture_output=True, text=True)
+        subprocess.run(subprocess_args, check=True, capture_output=True, text=True, env=client_env)
```
c-hagem: subprocess_env

benchmark/benchmarks/prefetch_benchmark.py
c-hagem: Fixed it.

PR #1557, opened 2025-08-05
Add helpful script for generating summary table from benchmark runs

c-hagem:
Adds a script that parses the benchmark output and autoamtically creates a table with only the parameters that changed between runs.

Does not need a changelog entry, as the script only parses hydra runs. 

---


benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,184 @@
...
+
+def parse_hydra_config(iteration_dir):
+    """Parse Hydra config and overrides for an iteration."""
+    config = {}
+    hydra_dir = os.path.join(iteration_dir, '.hydra')
+
+    # Read base config
+    config_path = os.path.join(hydra_dir, 'config.yaml')
+    if os.path.exists(config_path):
+        with open(config_path, 'r') as f:
```
muddyfish: Nit: `'r'` is default and not needed

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,184 @@
...
+            config.update(yaml.safe_load(f))
+
+    # Apply overrides
+    override_path = os.path.join(hydra_dir, 'overrides.yaml')
+    if os.path.exists(override_path):
+        with open(override_path, 'r') as f:
+            overrides = yaml.safe_load(f)
+            if overrides:
+                for override in overrides:
+                    if '=' in override:
```
muddyfish: This should be a function. Also, what if there's no `=`? Also, if the override is yaml, why not just assume it was given as a dictionary anyway and use the normal yaml key value syntax?

muddyfish: Also, why not just use normal nested yaml here?
```yaml
foo:
  bar:
    baz: 10
```

```py
config.update(yaml.safe_load(f))
```

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,184 @@
...
+                            current[parts[-1]] = value
+                        else:
+                            config[key] = value
+
+    return flatten_config(config)
+
+
+def flatten_config(config, parent_key='', sep='.'):
+    """Flatten nested configuration dictionary."""
+    items = []
```
muddyfish: Why is this a list rather than a dict?

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,184 @@
...
+
+def process_iteration(iteration_dir):
+    """Process a single iteration directory."""
+    config = parse_hydra_config(iteration_dir)
+
+    throughput = None
+    for file_pattern in ['crt_output.json', 'client-output.json', 'prefetch-output.json', 'fio.*.json']:
+        files = glob.glob(os.path.join(iteration_dir, file_pattern))
+        if files:
+            throughput = parse_benchmark_file(files[0])
```
muddyfish, line 120: The mechanism here for expected errors in parsing the file is not good. Is it intended? If so, I'd bring up the try/except into this function instead

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,184 @@
...
+import json
+import argparse
+import glob
+import yaml
+from tabulate import tabulate
+import numpy as np
+from collections import defaultdict
+
+
+def parse_hydra_config(iteration_dir):
```
muddyfish: None of the functions are using type hints, which makes this hard to read

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,184 @@
...
+    return dict(items)
+
+
+def parse_benchmark_file(file_path):
+    """Parse benchmark output file and return throughput."""
+    try:
+        with open(file_path, 'r') as f:
+            data = json.load(f)
+
+        if 'throughput_gbps' in data:  # CRT format
```
muddyfish: I think you could use a match statement here to destructure the type and make this better

benchmark/analysis-scripts/autogroup.py
c-hagem: The layout of the overrides file is i.e.
```
- application_workers=16
- iteration=24
- benchmark_type=prefetch
- mountpoint.fuse_threads=16
```

c-hagem: I think these files are automatically generated by hydra.

benchmark/analysis-scripts/autogroup.py
c-hagem: I think, if you run this script while a benchmark is still running, this way it will not fail. Will think about making it nicer.

benchmark/analysis-scripts/autogroup.py
c-hagem: Will replace with using omegconf module to have more robust parsing

benchmark/analysis-scripts/autogroup.py
c-hagem: Fixed that.

benchmark/analysis-scripts/autogroup.py
c-hagem: Added type hints

benchmark/analysis-scripts/autogroup.py
c-hagem: Added a match statement

benchmark/analysis-scripts/autogroup.py
c-hagem: Added warnings that are displayed if a file is not parseable

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,240 @@
...
+    config = OmegaConf.load(config_path)
+
+    # Apply overrides
+    override_path = os.path.join(hydra_dir, 'overrides.yaml')
+    if os.path.exists(override_path):
+        overrides = OmegaConf.load(override_path)
+        if overrides:
+            # Convert overrides list to OmegaConf and merge
+            override_list = OmegaConf.to_container(overrides, resolve=True)
+            if isinstance(override_list, list):
```
muddyfish, line 42: Nb: Do we want to warn if it's not a list??

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,240 @@
...
+            data = json.load(f)
+
+        match data:
+            # CRT format
+            case {'throughput_gbps': throughput}:
+                return throughput
+
+            # Client/Prefetch format
+            case {'summary': {'total_bytes': total_bytes, 'total_elapsed_seconds': total_seconds}}:
+                return (total_bytes * 8) / (1024 * 1024 * 1024 * total_seconds)
```
muddyfish: Should have a function for converting a bytes and seconds into a gigabits/second value. It's not completely obvious reading this:
```
return to_gigabits_per_second(byte_count=total_bytes, seconds=total_seconds)
```

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,240 @@
...
+            # FIO format with missing fields
+            case {'jobs': [job, *_]}:
+                read_data = job.get('read', {})
+                io_bytes = read_data.get('io_bytes', 0)
+                runtime_ms = read_data.get('runtime', 1000)
+                return (io_bytes * 8) / (1024 * 1024 * 1024 * (runtime_ms / 1000))
+
+            # Unknown format
+            case _:
+                print(f"{Fore.YELLOW}⚠️  Warning: Unknown format in {file_path}{Style.RESET_ALL}")
```
muddyfish: Printing errors to stdout isn't ideal - should be to stderr (`print(..., file=sys.stderr)`).

Alternatively (preferred), use the [`warnings.warn`](https://docs.python.org/3/library/warnings.html#warnings.warn) function

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,240 @@
...
+        return set()
+
+    varying = set()
+    # Get all keys from all configs
+    all_keys = set()
+    for config in all_configs:
+        all_keys.update(config.keys())
+
+    # Parameters to ignore
+    ignore_params = {'hydra.job.num', 'hydra.run.dir', 'hydra.job.id', 'hydra.job.name', 'iteration'}
```
muddyfish, line 141:
```py
all_keys -= ignore_params
```

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,240 @@
...
+
+    # Check which parameters vary
+    for key in all_keys:
+        if key not in ignore_params:
+            values = set()
```
muddyfish: Can be set comprehension here

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,240 @@
...
+
+    # Custom sorting function for benchmark types
+    def benchmark_type_sort_key(value: str) -> int:
+        benchmark_order = {'crt': 0, 'client': 1, 'client-bp': 2, 'prefetch': 3, 'fio': 4}
+        return benchmark_order.get(value, 999)  # Unknown types go to end
+
+    # Sort rows by all columns
+    def sort_key(row: List[str]) -> List[Union[int, float, str]]:
+        key_parts = []
+        for i, header in enumerate(aggregated_headers):
```
muddyfish:
```py
for value, header in zip(row, aggregated_headers):
```

benchmark/analysis-scripts/autogroup.py
c-hagem: Added a warning

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,249 @@
...
+                total_seconds = summary.get('total_elapsed_seconds', 1)
+                return to_gigabits_per_second(bytes=total_bytes, seconds=total_seconds)
+
+            # FIO format
+            case {'jobs': [{'read': {'io_bytes': io_bytes, 'runtime': runtime_ms}}, *_]}:
+                return to_gigabits_per_second(bytes=io_bytes, seconds=runtime_ms / 1000)
+
+            # FIO format with missing fields
+            case {'jobs': [job, *_]}:
+                read_data = job.get('read', {})
```
muddyfish: If the `read` field isn't present, can we really meaningfully calculate anything?

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,249 @@
...
+                runtime_ms = read_data.get('runtime', 1000)
+                return to_gigabits_per_second(bytes=io_bytes, seconds=runtime_ms / 1000)
+
+            # Unknown format
+            case _:
+                warnings.warn(f"Unknown format in {file_path}")
+                return None
+
+    except Exception as e:
+        warnings.warn(f" Warning: Error parsing {file_path}: {e}")
```
muddyfish: Nit: Preceeding space

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,249 @@
...
+    throughput = None
+    for file_pattern in ['crt_output.json', 'client-output.json', 'prefetch-output.json', 'fio.*.json']:
+        files = glob.glob(os.path.join(iteration_dir, file_pattern))
+        if files:
+            throughput = parse_benchmark_file(files[0])
+            if throughput is not None:
+                break
+
+    if throughput is None:
+        print(f"{Fore.YELLOW}⚠️  Warning: No valid throughput data found in {iteration_dir}{Style.RESET_ALL}")
```
muddyfish: Should also be `warning.warn`?

benchmark/analysis-scripts/autogroup.py
c-hagem: Thanks for catching that! I had accidentially used the 1024 based instead of 1000 based gibibits. Should be fixed now. @sahityadg Could you check that this logic is correct?

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,249 @@
...
+
+    return config, throughput
+
+
+def find_varying_parameters(all_configs: List[Dict[str, Any]]) -> Set[str]:
+    """Identify parameters that vary across configurations, ignoring run-specific parameters."""
+    if not all_configs:
+        return set()
+
+    varying = set()
```
muddyfish: Not used anymore

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,249 @@
...
+                iteration_path = os.path.join(root, dir_name)
+                config, throughput = process_iteration(iteration_path)
+                if throughput is not None:
+                    all_results.append((config, throughput, dir_name))
+
+    # Sort results by iteration number
+    all_results.sort(key=lambda x: int(x[2]))
+
+    # Find parameters that vary between iterations
+    varying_params = find_varying_parameters([config for config, _, _ in all_results])
```
muddyfish: Nit: This is only ever used with `sorted`. Why not do that here upfront?

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,249 @@
...
+    print(", ".join(sorted(varying_params)))
+
+    # Group by varying parameters
+    grouped_results = defaultdict(list)
+    for config, throughput, iter_num in all_results:
+        key = tuple((param, str(config.get(param, 'N/A'))) for param in sorted(varying_params))
+        grouped_results[key].append(throughput)
+
+    # Aggregated results table
+    aggregated_headers = list(sorted(varying_params)) + [
```
muddyfish: Nb: `sorted` always returns a list, so the list call is not needed

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,249 @@
...
+        "Min (Gbps)",
+        "Max (Gbps)",
+    ]
+    aggregated_rows = []
+    for config_key, throughputs in grouped_results.items():
+        row = []
+        for _, value in config_key:
+            row.append(value)
+        row.append(len(throughputs))
+        row.append(f"{np.mean(throughputs):.2f}")
```
muddyfish: Not blocking: you can use `statistics.mean`, `statistics.stdev`, `min`, and `max` here and drop the numpy requirement

benchmark/analysis-scripts/autogroup.py
c-hagem: Yes, sorting it once now.

benchmark/analysis-scripts/autogroup.py
c-hagem: Changed it to use statistics module.

benchmark/analysis-scripts/autogroup.py
c-hagem: No -- removed that match arm.

muddyfish (dismissed):

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,245 @@
...
+        warnings.warn(f"Warning: Error parsing {file_path}: {e}")
+        return None
+
+
+def process_iteration(iteration_dir: str) -> Tuple[Dict[str, Any], Optional[float]]:
+    """Process a single iteration directory."""
+    config = parse_hydra_config(iteration_dir)
+
+    throughput = None
+    for file_pattern in ['crt_output.json', 'client-output.json', 'prefetch-output.json', 'fio.*.json']:
```
sahityadg, line 117: fio.*.*json was a hack. Would be good to add a FIXME at least.

benchmark/analysis-scripts/autogroup.py
```diff
@@ -0,0 +1,245 @@
...
+    # Group by varying parameters
+    grouped_results = defaultdict(list)
+    for config, throughput, iter_num in all_results:
+        key = tuple((param, str(config.get(param, 'N/A'))) for param in sorted(varying_params))
+        grouped_results[key].append(throughput)
+
+    # Aggregated results table
+    aggregated_headers = varying_params + [
+        "Count",
+        "Avg (Gbps)",
```
sahityadg: Wondering if p50 is more useful

c-hagem: Changed to median

benchmark/analysis-scripts/autogroup.py
c-hagem: Added FIXME

muddyfish (approved):

PR #1563, opened 2025-08-07
Implement integrity checks for the metadata store

vladem:
Introduce checksum verification of:

1. entries provided in the input CSV manifest (and for the whole file)
2. entries stored in the metadata store

Slightly refactored `ManifestEntry` in order to avoid huge `match` blocks. This PR also adds an example of the manifest creation.

No, only `mount_from_config` example.

Yes, an entry for `mountpoint-s3-fs` crate, and a minor version bump of this crate. 

---

### DIFF
--- a/mountpoint-s3-crt/src/checksums/crc32c.rs
+++ b/mountpoint-s3-crt/src/checksums/crc32c.rs
@@ -42,6 +42,11 @@ impl Hasher {
         self.state = Crc32c(Self::crc32c(buf, self.state.0));
     }
 
+    /// Return the computed CRC32C checksum value without finalizing the hasher.
+    pub fn current_value(&self) -> Crc32c {
+        self.state
+    }
+
     /// Finalize the hash state and return the computed CRC32C checksum value.
     pub fn finalize(self) -> Crc32c {
         self.state
--- a/mountpoint-s3-fs/examples/create_manifest.rs
+++ b/mountpoint-s3-fs/examples/create_manifest.rs
@@ -1,0 +1,176 @@
+use std::{
+    fs::File,
+    hash::Hasher,
+    io::{BufWriter, Write},
+    path::{Path, PathBuf},
+};
+
+use anyhow::Result;
+use aws_config::BehaviorVersion;
+use aws_sdk_s3::{Client, config::Region, operation::list_objects_v2::ListObjectsV2Output};
+use clap::Parser;
+use crc32c::{self, Crc32cHasher, crc32c_combine};
+use csv::WriterBuilder;
+use serde::Serialize;
+
+#[derive(Debug, Parser)]
+#[clap(
+    name = "Mountpoint Manifest Creator",
+    about = "A helper to create a CSV manifest of a bucket using ListObjectsV2 requests."
+)]
+struct Opt {
+    /// The AWS Region.
+    #[clap(long)]
+    region: String,
+
+    /// The name of the bucket.
+    #[clap(long)]
+    bucket: String,
+
+    /// The prefix to list [must include '/' if not empty].
+    #[clap(long, default_value = "")]
+    prefix: String,
+
+    /// Whether to remove prefix from the keys in the output.
+    #[clap(long, default_value_t = false)]
+    remove_prefix: bool,
+
+    /// Where to write the manifest.
+    #[clap(long)]
+    output_file: PathBuf,
+}
+
+#[derive(Serialize)]
+struct CsvEntry {
+    /// Partial key of the S3 object.
+    ///
+    /// Must not contain S3 prefix, when the prefix is mounted.
+    /// May hold a directory marker (e.g. "dir1/dir2/"), which will be skipped
+    /// and won't affect the file system structure.
+    partial_key: String,
+    /// Etag of the S3 object.
+    etag: String,
+    /// Size of the S3 object.
+    size: usize,
+    /// CRC32C checksum of the fields.
+    checksum: u32,
+}
+
+impl CsvEntry {
+    fn new(partial_key: String, etag: String, size: usize) -> Self {
+        let checksum = Self::compute_checksum(&partial_key, &etag, size);
+        Self {
+            partial_key,
+            etag,
+            size,
+            checksum,
+        }
+    }
+
+    fn compute_checksum(partial_key: &str, etag: &str, size: usize) -> u32 {
+        let mut hasher = Crc32cHasher::new(0);
+        hasher.write(partial_key.as_bytes());
+        hasher.write(etag.as_bytes());
+        // we encode size with big endian byte order and with a fixed width of 8 bytes (rust: u64, java: long)
+        let size = size as u64;
+        hasher.write(size.to_be_bytes().as_ref());
+        hasher.finish() as u32
+    }
+
+    fn total_size(&self) -> usize {
+        const SIZE_LEN: usize = 8; // size is always 8 bytes
+        self.partial_key.len() + self.etag.len() + SIZE_LEN
+    }
+}
+
+fn write_to_manifest<W: Write>(
+    resp: &ListObjectsV2Output,
+    prefix: &str,
+    remove_prefix: bool,
+    writer: &mut csv::Writer<W>,
+    running_checksum: &mut u32,
+) -> Result<()> {
+    for object in resp.contents() {
+        let relative_key = if remove_prefix && !prefix.is_empty() {
+            &object.key().unwrap()[prefix.len() + 1..]
+        } else {
+            object.key().unwrap()
+        };
+
+        // Create a CsvEntry instance
+        let entry = CsvEntry::new(
+            relative_key.to_string(),
+            object.e_tag().unwrap().to_string().replace("\"", ""), // remove quotes (that's the default behavior in Java SDK)
+            object.size().unwrap() as usize,
+        );
+
+        // Update running_checksum using crc32c_combine (this may not be available in all libraries, alternatively re-compute checksum for each field)
+        *running_checksum = crc32c_combine(*running_checksum, entry.checksum, entry.total_size());
+
+        // Write the entry to the CSV writer
+        writer.serialize(entry)?;
+    }
+    Ok(())
+}
+
+async fn list_objects(
+    client: &Client,
+    bucket: &str,
+    prefix: &str,
+    remove_prefix: bool,
+    output_file: &Path,
+) -> Result<()> {
+    // Create a CSV writer
+    let file = File::options().write(true).create_new(true).open(output_file)?;
+    let writer = BufWriter::new(file);
+    let mut writer = WriterBuilder::new().has_headers(false).from_writer(writer);
+
+    // Running checksum of all entries in the manifest
+    let mut running_checksum: u32 = 0;
+
+    // Make ListObjectsV2 requests and write output to the CSV file
+    let mut resp = client.list_objects_v2().bucket(bucket).prefix(prefix).send().await?;
+    write_to_manifest(&resp, prefix, remove_prefix, &mut writer, &mut running_checksum)?;
+
+    while let Some(token) = resp.next_continuation_token {
+        resp = client
+            .list_objects_v2()
+            .bucket(bucket)
+            .continuation_token(token)
+            .prefix(prefix)
+            .send()
+            .await?;
+        write_to_manifest(&resp, prefix, remove_prefix, &mut writer, &mut running_checksum)?;
+    }
+
+    // Flush the writer to ensure all data is written
+    writer.flush()?;
+
+    // Print the final running checksum to stdout
+    println!("Running checksum of all entries: {running_checksum}");
+
+    Ok(())
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    tracing_subscriber::fmt::init();
+
+    let Opt {
+        region,
+        bucket,
+        prefix,
+        remove_prefix,
+        output_file,
+    } = Opt::parse();
+
+    let sdk_config = aws_config::defaults(BehaviorVersion::latest())
+        .region(Region::new(region))
+        .load()
+        .await;
+    let client = Client::new(&sdk_config);
+
+    list_objects(&client, &bucket, &prefix, remove_prefix, output_file.as_path()).await?;
+
+    Ok(())
+}
--- a/mountpoint-s3-fs/src/manifest/builder.rs
+++ b/mountpoint-s3-fs/src/manifest/builder.rs
@@ -3,11 +3,12 @@ use std::io::{self, BufReader};
 use std::path::{Path, PathBuf};
 use std::{collections::HashMap, fs::File};
 
+use mountpoint_s3_client::checksums::Crc32c;
 use serde::Deserialize;
 use thiserror::Error;
 
 use crate::fs::InodeKind;
-use crate::metablock::{ROOT_INODE_NO, ValidKey, ValidKeyError};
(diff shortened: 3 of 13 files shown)


mountpoint-s3-crt/src/checksums/crc32c.rs
```diff
@@ -42,6 +42,11 @@ impl Hasher {
         self.state = Crc32c(Self::crc32c(buf, self.state.0));
     }
 
+    /// Return the computed CRC32C checksum value without finalizing the hasher.
+    pub fn current_value(&self) -> Crc32c {
```
passaro: This requires a new client release. As an alternative, consider having the caller do `.clone().finalize()`.

mountpoint-s3-fs/examples/create_manifest.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    /// The name of the bucket.
+    #[clap(long)]
+    bucket: String,
+
+    /// The prefix to list [must include '/' if not empty].
+    #[clap(long, default_value = "")]
+    prefix: String,
+
+    /// Whether to remove prefix from the keys in the output.
```
passaro: What's the use case for this flag?

mountpoint-s3-fs/src/manifest/builder.rs
```diff
@@ -87,35 +99,74 @@ impl InputManifestEntry {
...
+            etag,
             size,
+            checksum: Crc32c::new(checksum),
         })
     }
 
-    fn into_db_entry(self, id: u64, parent_id: u64, channel_id: usize, parent_partial_key: ValidKey) -> DbEntry {
+    /// Creates an InputManifestEntry and computes the checksum.
+    ///
+    /// This is only useful in tests.
```
passaro, line 119: minor: then why does it need to be here? shouldn't it be a test helper?

mountpoint-s3-fs/src/manifest/core.rs
```diff
@@ -24,213 +25,219 @@ pub enum ManifestError {
...
+    /// Identifier of the S3 channel (bucket+prefix combination) this file belongs to.
+    channel_id: usize,
+    /// Partial key of the parent directory.
+    parent_partial_key: Option<ValidKey>,
+    /// Name of the file.
+    name: String,
+    /// Kind and associated metadata.
+    kind: ManifestEntryKind,
+    /// CRC32C checksum of this entry.
+    checksum: u32,
```
passaro: Why not use a `Crc32C` type?

mountpoint-s3-fs/src/manifest/csv_reader.rs
```diff
@@ -7,26 +8,65 @@ use super::builder::InputManifestEntry;
...
+            expected_checksum,
+        }
+    }
+
+    fn update_running_checksum(&mut self, csv_entry: &CsvEntry) {
+        // we re-compute checksum of each entry, because `Hasher` doesn't support combining pre-computed checksums
+        if let Some(hasher) = self.hasher.as_mut() {
+            hasher.update(csv_entry.partial_key.as_bytes());
+            hasher.update(csv_entry.etag.as_bytes());
+            // we encode size with big endian byte order and with a fixed width of 8 bytes (rust: u64, java: long)
```
passaro: Should we make `csv_entry.size` a `u64` then?

mountpoint-s3-fs/src/manifest/csv_reader.rs
```diff
@@ -7,26 +8,65 @@ use super::builder::InputManifestEntry;
...
+        Self {
+            reader,
+            file_path: file_path.to_string(),
+            hasher: Some(Hasher::new()),
+            expected_checksum,
+        }
+    }
+
+    fn update_running_checksum(&mut self, csv_entry: &CsvEntry) {
+        // we re-compute checksum of each entry, because `Hasher` doesn't support combining pre-computed checksums
```
passaro: Couldn't we use the `combine_checksums` function?

mountpoint-s3-fs/src/manifest/csv_reader.rs
```diff
@@ -7,26 +8,65 @@ use super::builder::InputManifestEntry;
...
+        }
     }
 }
 
 impl<R: Read> Iterator for CsvReader<R> {
     type Item = Result<InputManifestEntry, InputManifestError>;
 
+    /// Returns next validated [InputManifestEntry] or [InputManifestError].
+    ///
+    /// This method will compare the manifest checksum after the final CSV entry is processed.
```
passaro: This comment should be on the type rust doc.

mountpoint-s3-fs/src/manifest/csv_reader.rs
```diff
@@ -7,26 +8,65 @@ use super::builder::InputManifestEntry;
 
 pub struct CsvReader<R: Read> {
     reader: DeserializeRecordsIntoIter<R, CsvEntry>,
+    file_path: String,
+    hasher: Option<Hasher>,
```
passaro: Why an `Option`? The `new` constructor always sets it.

mountpoint-s3-fs/examples/create_manifest.rs
```diff
@@ -0,0 +1,176 @@
...
+        };
+
+        // Create a CsvEntry instance
+        let entry = CsvEntry::new(
+            relative_key.to_string(),
+            object.e_tag().unwrap().to_string().replace("\"", ""), // remove quotes (that's the default behavior in Java SDK)
+            object.size().unwrap() as usize,
+        );
+
+        // Update running_checksum using crc32c_combine (this may not be available in all libraries, alternatively re-compute checksum for each field)
```
passaro: What does that mean? Couldn't this use `combine_checksums`?

mountpoint-s3-fs/examples/create_manifest.rs
```diff
@@ -0,0 +1,176 @@
...
+    #[clap(long, default_value_t = false)]
+    remove_prefix: bool,
+
+    /// Where to write the manifest.
+    #[clap(long)]
+    output_file: PathBuf,
+}
+
+#[derive(Serialize)]
+struct CsvEntry {
```
passaro: Is this the same struct under the `manifest` module?

mountpoint-s3-fs/src/manifest/core.rs
```diff
@@ -24,213 +25,219 @@ pub enum ManifestError {
...
+        })
+    }
+}
+
+/// A helper to compute the checksum of the given set of fields.
+///
+/// Computes (partial_key, full_checksum, partial_checksum), with full_checksum including all of the fields
+/// and partial_checksum including only (partial_key, etag, size).
+///
+/// Note: this function will reconstruct the s3 key from parent_partial_key and name, this may mean unnecessary
```
passaro: This is a bit vague. When do we need the copies and when are they wasted?

passaro (commented):
I had a first pass and added a few comments.

One thing that may need some clarification is the difference between the "partial checksums" and the ones on the DB. And why we need both.

mountpoint-s3-fs/src/manifest/core.rs
vladem: Updated this comment to explain this and the difference between the "partial checksums" and the ones on the DB

mountpoint-s3-fs/src/manifest/csv_reader.rs
vladem: `mountpoint-s3-crt`'s `crc32c` module doesn't have a combine function, but the we can use the external `crc32c` crate (which has `crc32c_combine`) for this, I'll change that

mountpoint-s3-fs/src/manifest/csv_reader.rs
vladem: fixed

mountpoint-s3-fs/examples/create_manifest.rs
vladem: fixed above

mountpoint-s3-fs/examples/create_manifest.rs
vladem: if readers expect the bucket structure with the prefix, this may be set to `false`, it was helpful in some of my experiments

there was a mistake related to `/` in the end of prefix, I've fixed it

mountpoint-s3-fs/src/manifest/builder.rs
vladem: I agree, but it's also used in `builder.rs` unit tests, is there really a better place for it than `InputManifestEntry::new_without_checksum`?

mountpoint-s3-fs/src/manifest/csv_reader.rs
vladem: `InodeStat::size` is `usize`, I think the best we can do is to fail on CSV parsing on machines where `usize` is less than 8 bytes and larger size was found in the manifest (that's how it works now)

mountpoint-s3-fs/examples/create_manifest.rs
passaro: Which other readers are there besides our CsvReader? I'm inclined to remove.

mountpoint-s3-fs/src/manifest/builder.rs
passaro: consider marking with `cfg(test)` then.

mountpoint-s3-fs/src/manifest/csv_reader.rs
passaro: we already wrap that function in `mountpoint-s3-fs::checksums::combine_checksums`. It should be functionally the same, but you'll get to use the `Crc32c` type everywhere.

mountpoint-s3-fs/src/manifest/csv_reader.rs
```diff
@@ -35,25 +72,34 @@ impl<R: Read> Iterator for CsvReader<R> {
...
-    partial_key: String,
+    pub partial_key: String,
     /// Etag of the S3 object.
-    etag: String,
+    pub etag: String,
     /// Size of the S3 object.
-    size: usize,
+    pub size: usize,
+    /// CRC32C checksum of the fields.
+    pub checksum: u32,
```
passaro: Let's use `Crc32c` here. And also document how the checksum is calculated, i.e. order and endianess.

mountpoint-s3-fs/src/manifest/csv_reader.rs
```diff
@@ -35,25 +72,34 @@ impl<R: Read> Iterator for CsvReader<R> {
...
     /// May hold a directory marker (e.g. "dir1/dir2/"), which will be skipped
     /// and won't affect the file system structure.
-    partial_key: String,
+    pub partial_key: String,
     /// Etag of the S3 object.
-    etag: String,
+    pub etag: String,
     /// Size of the S3 object.
-    size: usize,
+    pub size: usize,
```
passaro: This should be `u64` to remove some of the ambiguity on the checksum calculation.

mountpoint-s3-fs/src/manifest/csv_reader.rs
passaro: I think we should use `u64` to match the checksum calculation and then validate when building the manifest (e.g. CsvEntry -> InputManifestEntry)

mountpoint-s3-fs/src/manifest/csv_reader.rs
vladem: we'll need to update `mountpoint_s3_crt::checksums::crc32c::Crc32c` to `#[derive(Deserialize, Serialize)]`, do we want that?

mountpoint-s3-fs/src/manifest/builder.rs
vladem: this method is used both in unit tests and integration tests, integration tests won't be able to access it if we put `cfg(test)`

I can duplicate this method to `tests::common::manifest` and have it annotated with `cfg(test)` here, does it sound ok?

mountpoint-s3-fs/examples/create_manifest.rs
```diff
@@ -118,27 +106,29 @@ async fn list_objects(
...
+    println!(
+        "Running checksum of all entries: {}",
+        crc32c_to_base64(&running_checksum)
+    );
 
     Ok(())
 }
 
-#[tokio::main]
+#[tokio::main(flavor = "current_thread")]
```
passaro, line 124: Why the change?

mountpoint-s3-fs/src/checksums.rs
```diff
@@ -255,6 +282,60 @@ pub enum IntegrityError {
     ChecksumMismatch(Crc32c, Crc32c),
 }
 
+// A Crc32c checksum which can be (de)serialized to a base64 encoded string.
```
passaro: nit: rustdoc comment

mountpoint-s3-fs/src/checksums.rs
```diff
@@ -255,6 +282,60 @@ pub enum IntegrityError {
...
+//
+// TODO: there should be a single Crc32c type implementing serialization.
```
passaro: We could also consider using a [serde attribute](https://serde.rs/field-attrs.html#with) instead. But happy to leave for now.

vladem: interesting option! can implement like this, but not sure if it's any better:

```rust
fn serialize_crc32c_base64<S>(crc: &Crc32c, serializer: S) -> Result<S::Ok, S::Error>
where
    S: Serializer,
{
    let encoded = crc32c_to_base64(crc);
    serializer.serialize_str(&encoded)
}

fn deserialize_crc32c_base64<'de, D>(deserializer: D) -> Result<Crc32c, D::Error>
where
    D: Deserializer<'de>,
{
    struct Crc32cVisitor;

    impl<'de> Visitor<'de> for Crc32cVisitor {
        type Value = Crc32c;

        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
            formatter.write_str("a base64-encoded CRC32C string")
        }

        fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
        where
            E: de::Error,
        {
            crc32c_from_base64(v).map_err(E::custom)
        }
    }

    deserializer.deserialize_str(Crc32cVisitor)
}

/// A Crc32c checksum which can be (de)serialized to a base64 encoded string.
///
/// TODO: there should be a single Crc32c type implementing serialization.
#[derive(Debug, Serialize, Deserialize)]
pub struct Crc32cBase64(
    #[serde(serialize_with = "serialize_crc32c_base64", deserialize_with = "deserialize_crc32c_base64")]
    Crc32c
);
```

mountpoint-s3-fs/examples/create_manifest.rs
vladem: I had this [failure](https://github.com/awslabs/mountpoint-s3/actions/runs/17950231016/job/51047224628):

> error: The default runtime flavor is `multi_thread`, but the `rt-multi-thread` feature is disabled.

As an alternative I can enable `rt-multi-thread` in `Cargo.toml`:

```
tokio = { version = "1.45.1", features = ["rt", "macros", "rt-multi-thread"] }
```

mountpoint-s3-fs/src/checksums.rs
passaro: Not quite: it would be an alternative to the `Crc32cBase64` type, with the attribute in the struct that contains the `Crc32c` directly.

But my comment was just a "FYI", let's discuss later.

passaro (approved):
LGTM

PR #1565, opened 2025-08-07
Bench client paged pool

mansi153:
Configure the MP client benchmark to use the new paged memory-pool instead of the default CRT memory pool. This is to make the client benchmark perform more comparably to prefetcher and other upper layers.

A second commit also fixes the throughput display units and max_target_throughput default value.

No, benchmark change only.

No, benchmark change only.

---

### DIFF
--- a/mountpoint-s3-client/examples/client_benchmark.rs
+++ b/mountpoint-s3-client/examples/client_benchmark.rs
@@ -13,6 +13,7 @@ use mountpoint_s3_client::mock_client::{MockClient, MockObject};
 use mountpoint_s3_client::types::{ClientBackpressureHandle, ETag, GetObjectParams, GetObjectResponse};
 use mountpoint_s3_client::{ObjectClient, S3CrtClient};
 use mountpoint_s3_crt::common::rust_log_adapter::RustLogAdapter;
+use mountpoint_s3_fs::memory::PagedPool;
 use serde_json::{json, to_writer};
 use tracing_subscriber::EnvFilter;
 use tracing_subscriber::fmt::Subscriber;
@@ -238,8 +239,10 @@ fn create_s3_client_config(region: &str, args: &CliArgs, nics: Vec<String>) -> S
     config = config.throughput_target_gbps(args.throughput_target_gbps);
     config = config.memory_limit_in_bytes(args.crt_memory_limit_gb * 1024 * 1024 * 1024);
     config = config.network_interface_names(nics);
-
     config = config.part_size(args.part_size);
+
+    let pool = PagedPool::new_with_candidate_sizes([args.part_size]);
+    config = config.memory_pool(pool.clone());
 
     if args.enable_backpressure {
         config = config.read_backpressure(true);
--- a/mountpoint-s3-client/examples/client_benchmark.rs
+++ b/mountpoint-s3-client/examples/client_benchmark.rs
@@ -117,11 +117,11 @@ fn run_benchmark(
         let received_size = received_size.load(Ordering::SeqCst);
         total_bytes += received_size;
         println!(
-            "{}: received {} bytes in {:.2}s: {:.2} Gib/s",
+            "{}: received {} bytes in {:.2}s: {:.2} Gb/s",
             iteration,
             received_size,
             elapsed.as_secs_f64(),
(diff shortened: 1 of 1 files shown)


mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -198,7 +199,7 @@ struct CliArgs {
     #[arg(
         long,
         help = "Desired throughput in Gbps",
-        default_value_t = 10.0,
+        default_value_t = 100.0,
```
c-hagem, line 202: Unrelated to the PR? Maybe we should just error out if users do not provide throughput for these benchmarks (at least when not mocked) -- silently defaulting / changing the default seems like something that could cause results to vary without being obvious.

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -238,9 +239,11 @@ fn create_s3_client_config(region: &str, args: &CliArgs, nics: Vec<String>) -> S
     config = config.throughput_target_gbps(args.throughput_target_gbps);
     config = config.memory_limit_in_bytes(args.crt_memory_limit_gb * 1024 * 1024 * 1024);
```
sahityadg: Do we still use this?

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -116,11 +117,11 @@ fn run_benchmark(
         let received_size = received_size.load(Ordering::SeqCst);
         total_bytes += received_size;
         println!(
-            "{}: received {} bytes in {:.2}s: {:.2} Gib/s",
+            "{}: received {} bytes in {:.2}s: {:.2} Gb/s",
             iteration,
             received_size,
             elapsed.as_secs_f64(),
-            (received_size as f64) / elapsed.as_secs_f64() / (1024 * 1024 * 1024 / 8) as f64
+            (received_size as f64) / elapsed.as_secs_f64() / (1000 * 1000 * 1000 / 8) as f64
```
muddyfish, line 124: Mentioned this in another Python PR - perhaps we should make this a function that can be reused to make sure the logic is clear and correct in all cases

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -238,9 +239,11 @@ fn create_s3_client_config(region: &str, args: &CliArgs, nics: Vec<String>) -> S
     config = config.throughput_target_gbps(args.throughput_target_gbps);
     config = config.memory_limit_in_bytes(args.crt_memory_limit_gb * 1024 * 1024 * 1024);
     config = config.network_interface_names(nics);
-
     config = config.part_size(args.part_size);
```
renanmagagnin: can these be chained together? not necessarily also chained with declaration but worth considering too. Also applies to the following `args.enable_backpressure` block

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: Thanks, valid suggestion! I tried to elaborate here: https://github.com/awslabs/mountpoint-s3/pull/1579#discussion_r2279256180 
Pls let me know if you have any suggestions/ideas!

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: Yes, will make it chained where I can

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: This is same as the default in benchmark_config_parser.py [see [here](https://github.com/awslabs/mountpoint-s3/blob/main/benchmark/benchmarks/benchmark_config_parser.py#L32)], where it silently defaults to 100; so this was only an attempt to make it consistent across different places of assignment. I believe 100 is a reasonable value to fallback to as this is what CRT recommends using for a single NIC.

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: You're right, with the paged pool we should not be configuring mem_limit explicitly. I'll remove this from both client and prefetch benchmarks.

renanmagagnin (approved):

PR #1566, opened 2025-08-08
Add `max_read_window_size` config to prefetcher benchmarks

mansi153:
Add `max_read_window_size` config to prefetcher benchmarks so we can set the backpressure window max value at runtime. 

The PR also fixes a few minor things - throughput calculation units from Gib/s to Gb/s, consistency of config names and improving some benchmarking defaults.

No, benchmarks change only.

No, benchmarks change only.

---

### DIFF
--- a/benchmark/benchmarks/prefetch_benchmark.py
+++ b/benchmark/benchmarks/prefetch_benchmark.py
@@ -72,6 +72,9 @@ class PrefetchBenchmark(BaseBenchmark):
         if not self.common_config['download_checksums']:
             prefetch_env["EXPERIMENTAL_MOUNTPOINT_NO_DOWNLOAD_INTEGRITY_VALIDATION"] = "ON"
 
+        if self.prefetch_config['max_read_window_size'] is not None:
+            prefetch_env["UNSTABLE_MOUNTPOINT_MAX_PREFETCH_WINDOW_SIZE"] = str(self.prefetch_config['max_read_window_size'])
+
         subprocess_env = os.environ.copy() | prefetch_env
         log.debug("Subprocess env: %s", subprocess_env)
 
--- a/benchmark/conf/config.yaml
+++ b/benchmark/conf/config.yaml
@@ -54,6 +54,7 @@ benchmarks:
   
   prefetch:
     max_memory_target: !!null
+    max_read_window_size: !!null
   
   crt:
     crt_benchmarks_path: !!null
@@ -62,7 +63,7 @@ benchmarks:
     # None 
 
   client_backpressure:
-    read_window_size: !!null #2147483648 
+    max_read_window_size: !!null #2147483648
 
 
 hydra:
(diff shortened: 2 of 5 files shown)


mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -218,7 +218,7 @@ struct CliArgs {
     #[arg(
         long,
         help = "Initial read window size in bytes, used to dictate how far ahead we request data from S3",
-        default_value = "0"
+        default_value = "2147483648"
```
sahityadg, line 221: My understanding is this is a fixed read window size in client benchmarks. So should we rename the parameter to read_window_size?

benchmark/conf/config.yaml
```diff
@@ -62,7 +63,7 @@ benchmarks:
     # None 
 
   client_backpressure:
-    read_window_size: !!null #2147483648 
+    max_read_window_size: !!null #2147483648
```
sahityadg, line 66: Is this a constant read window size or can it go up to this value?

benchmark/benchmarks/prefetch_benchmark.py
```diff
@@ -72,6 +72,9 @@ def run_benchmark(self) -> None:
         if not self.common_config['download_checksums']:
             prefetch_env["EXPERIMENTAL_MOUNTPOINT_NO_DOWNLOAD_INTEGRITY_VALIDATION"] = "ON"
 
+        if self.prefetch_config['max_read_window_size'] is not None:
+            prefetch_env["UNSTABLE_MOUNTPOINT_MAX_PREFETCH_WINDOW_SIZE"] = str(self.prefetch_config['max_read_window_size'])
```
sahityadg, line 76: For prefetcher, we could configure initial and max window size, do we want to configure both? Would it be a like for like comparision with client backpressure benchmark if we just configure max window size?

mountpoint-s3-client/examples/client_benchmark.rs
renanmagagnin, line 221: nit: can we also document here somehow that this number corresponds to 2GiB?

mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -233,9 +233,9 @@ fn main() -> anyhow::Result<()> {
         let received_size = received_bytes.load(Ordering::SeqCst);
         total_bytes += received_size;
         println!(
-            "{iteration}: received {received_size} bytes in {:.2}s: {:.2} Gib/s",
+            "{iteration}: received {received_size} bytes in {:.2}s: {:.2} Gb/s",
             elapsed.as_secs_f64(),
-            (received_size as f64) / elapsed.as_secs_f64() / (1024 * 1024 * 1024 / 8) as f64
+            (received_size as f64) / elapsed.as_secs_f64() / (1000 * 1000 * 1000 / 8) as f64
```
renanmagagnin, line 238: nit: can we make this value more readable? not sure if best to use a clearly named variable or a function to reuse in line 251 too

PR #1569, opened 2025-08-12
Clean up the cache sub-directory in a background thread

unexge:
Mountpoint tries to clean up the cache sub-directory during startup and exit in a blocking manner. This causes a cold start on Mountpoint if the cache sub-directory is not cleaned up in exit (e.g., if it's forcefully terminated) and if the cache sub-directory contains lots of files.

This change updates the clean up logic to do a rename first in the current thread to ensure Mountpoint starts with a fresh cache folder, and then spawns a detached background thread to remove the old cache directory. The cache sub-directory is still cleaned up in a blocking manner in the current thread during exit.

The cache will now be cleaned up in a background thread, this change may:

- Increase disk space usage temporarily during cleanup, as the new cache begins populating while the old cache still exists for a period of time
- Prevent cache cleanup errors from being propagated, since Mountpoint startup is no longer blocked by the cache cleanup process. Errors will now be just logged instead of propagated
   - In the worst-case scenario, this could cause Mountpoint to leave orphaned cache folders

Probably requires a changelog entry.

---

### DIFF
--- a/mountpoint-s3-fs/src/data_cache/cache_directory.rs
+++ b/mountpoint-s3-fs/src/data_cache/cache_directory.rs
@@ -40,8 +40,14 @@ pub enum ManagedCacheDirError {
 }
 
 impl ManagedCacheDir {
+    // This is the sub-directory where Mountpoint will store cache contents for the current Mountpoint instance.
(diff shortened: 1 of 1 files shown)


muddyfish:
Closing as we'll work on this in another PR

PR #1573, opened 2025-08-13
Enable resource monitoring for all benchmark types

c-hagem:
Adds the possibility to run resource monitoring for all benchmark types. This is achieved by introducing a `Command` abstraction that is returned by the benchmark. Then when the command has just started executing we start the monitoring with it, unless we already have a PID to monitor (used for FIO). (Thanks, Q )

Additionally changes the way we run most cargo commands by seperating the phase where replacing `cargo run` by instead doing `cargo build` and getting the executable path -- otherwise the compilation was part of the flamegraph. 

---

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -358,7 +358,6 @@ def run_experiment(cfg: DictConfig) -> None:
             process = subprocess.Popen(
                 command.args,
                 env=command.env,
-                cwd=command.cwd,
                 stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
                 text=True,
@@ -377,7 +376,6 @@ def run_experiment(cfg: DictConfig) -> None:
             cfg.monitoring.get("flamegraph_scripts_path"),
         ):
             stdout, stderr = process.communicate()
-            execution_time = time.time() - start_time
 
             if stdout is not None and isinstance(stdout, bytes):
                 stdout = stdout.decode('utf-8')
@@ -385,7 +383,7 @@ def run_experiment(cfg: DictConfig) -> None:
                 stderr = stderr.decode('utf-8')
 
             result = CommandResult(
-                returncode=process.returncode, stdout=stdout, stderr=stderr, execution_time=execution_time
+                returncode=process.returncode, stdout=stdout, stderr=stderr
             )
 
         metadata["success"] = True
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -352,16 +352,13 @@ def run_experiment(cfg: DictConfig) -> None:
         benchmark.setup()
         command = benchmark.get_command()
 
-        if command.capture_output:
-            process = subprocess.Popen(
-                command.args,
-                env=command.env,
-                stdout=subprocess.PIPE,
-                stderr=subprocess.PIPE,
-                text=True,
-            )
-        else:
-            process = subprocess.Popen(command.args, env=command.env, cwd=command.cwd)
+        process = subprocess.Popen(
+            command.args,
+            env=command.env,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+        )
 
         target_pid = metadata.get("target_pid", process.pid)
         metadata["target_pid"] = target_pid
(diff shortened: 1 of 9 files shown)


benchmark/benchmark.py
```diff
@@ -357,7 +379,23 @@ def run_experiment(cfg: DictConfig) -> None:
...
+            if stderr is not None and isinstance(stderr, bytes):
+                stderr = stderr.decode('utf-8')
+
+            # Create result and let command handle it
+            result = CommandResult(
+                returncode=process.returncode, stdout=stdout, stderr=stderr, execution_time=execution_time
+            )
+
+            # Process result using command's callback
+            result_metadata = command.handle_result(result)
```
sahityadg: Shouldn't we move this to post process and upload to s3 step after post process? Do we really need the Comand and CommandResult classes? Should the command execution just be a util?

benchmark/benchmarks/cargo_helper.py
```diff
@@ -0,0 +1,95 @@
+import json
+import logging
+import subprocess
+from typing import List, Optional, Dict
+import os
+
+log = logging.getLogger(__name__)
+
+
+def build_example(name: str, features: Optional[List[str]] = None, build_env: Optional[Dict[str, str]] = None) -> str:
```
sahityadg, line 10: Looks same as build_binary, we should switch to build binary everywhere and remove this.

sahityadg: Ignore this, missed that we need to pass --example/--bin param.

benchmark/benchmarks/crt_benchmark.py
```diff
@@ -101,14 +105,30 @@ def run_benchmark(self) -> Dict[str, Any]:
...
+        def handle_crt_result(result: CommandResult) -> Dict[str, Any]:
+            if result.returncode != 0:
+                log.error(f"CRT benchmark failed with exit code {result.returncode}")
+                if result.stderr:
+                    log.error(f"Error output: {result.stderr}")
+                raise subprocess.CalledProcessError(result.returncode, subprocess_args)
+
+            log.info("CRT benchmark completed successfully")
+
+            # Parse the output and extract metrics
```
sahityadg: nit: do we need these comments? The code is kind of self explanatory.

benchmark/benchmarks/client_benchmark.py
```diff
@@ -75,12 +80,19 @@ def run_benchmark(self) -> Dict[str, Any]:
         if not self.common_config['download_checksums']:
             client_env["EXPERIMENTAL_MOUNTPOINT_NO_DOWNLOAD_INTEGRITY_VALIDATION"] = "ON"
 
-        subprocess_env = os.environ.copy() | client_env
-        log.debug("Subprocess env: %s", subprocess_env)
+        def handle_client_result(result: CommandResult) -> Dict[str, Any]:
```
sahityadg: Like my previous comment, I'm not sure if we need this or move this logic to post processing. Even if we want a separate method for command output/error processing, wouldn't it be simpler to make it part of base_benchmark and call it from benchmark.py?

benchmark/benchmark.py
```diff
@@ -357,30 +376,39 @@ def run_experiment(cfg: DictConfig) -> None:
             cfg.monitoring.with_flamegraph,
             cfg.monitoring.get("flamegraph_scripts_path"),
         ):
-            benchmark.run_benchmark()
+            stdout, stderr = process.communicate()
+            execution_time = time.time() - start_time
```
sahityadg: are we using this anywhere? we should just remove as this should be in logs already.

benchmark/benchmarks/command.py
```diff
@@ -0,0 +1,29 @@
...
+    execution_time: float = 0.0
+
+
+@dataclass
+class Command:
+    """Represents a command to be executed with its environment."""
+
+    args: List[str]
+    env: Dict[str, str] = field(default_factory=dict)
+    cwd: Optional[str] = None
```
sahityadg: Is this used?

c-hagem: No -- removed for now

benchmark/benchmark.py
c-hagem: Removed it.

PR #1575, opened 2025-08-15
[Benchmarks] Ensure binaries are built with necessary flags for flamegraphing

c-hagem:
Ensures that frame pointers for C and Rust code are emitted when flamegraphing mountpoint.

---

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -300,6 +300,19 @@ class ResourceMonitoring:
     def _start_flamegraph(self):
         """Produces a flamegraph"""
 
+        try:
+            with open('/proc/sys/kernel/kptr_restrict', 'r') as f:
+                kptr_restrict_value = f.read().strip()
+            if kptr_restrict_value != '0':
+                log.warning(
+                    f"kernel.kptr_restrict is set to {kptr_restrict_value}, not 0. "
+                    f"For comprehensive flamegraphs, consider running: sudo sysctl kernel.kptr_restrict=0"
+                )
+            else:
+                log.info("kernel.kptr_restrict=0 - good for comprehensive flamegraphs")
+        except (OSError, IOError) as e:
+            log.warning(f"Could not check kernel.kptr_restrict: {e}")
+
         flamegraph_args = ["flamegraph", "--pid", str(self.target_pid), "-o", "flamegraph.svg"]
 
         log.info("Starting flamegraph with args %s", " ".join(flamegraph_args))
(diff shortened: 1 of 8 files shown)


benchmark/benchmark.py
```diff
@@ -300,6 +300,19 @@ def _start_perf_stat(self) -> any:
     def _start_flamegraph(self):
         """Produces a flamegraph"""
 
+        try:
+            with open('/proc/sys/kernel/kptr_restrict', 'r') as f:
+                kptr_restrict_value = f.read().strip()
+            if kptr_restrict_value != '0':
```
renanmagagnin, line 306: We should check the declaration of `kptr_restrict_value` to make sure it's available in the scope of the if statement

c-hagem: Apparently "A with statement does not create a scope (like if, for and while do not create a scope either)." (c.f. https://stackoverflow.com/questions/45100271/scope-of-variable-within-with-statement#:~:text=A%20with%20statement%20does%20not,(to%20the%20real%20scope).), so should be fine

benchmark/benchmark.py
```diff
@@ -300,6 +300,19 @@ def _start_perf_stat(self) -> any:
     def _start_flamegraph(self):
         """Produces a flamegraph"""
 
+        try:
+            with open('/proc/sys/kernel/kptr_restrict', 'r') as f:
```
c-hagem, line 304: Also check for paranoid

benchmark/benchmarks/cargo_helper.py
```diff
@@ -7,43 +7,61 @@
 log = logging.getLogger(__name__)
 
 
-def build_example(name: str, features: Optional[List[str]] = None, build_env: Optional[Dict[str, str]] = None) -> str:
+def build_example(
+    name: str,
+    features: Optional[List[str]] = None,
+    build_env: Optional[Dict[str, str]] = None,
+    flamegraph_enhancement: bool = False,
```
sahityadg: nit: more readable to call this also with_flamegraph I guess.

c-hagem: renamed

renanmagagnin (dismissed):

sahityadg (dismissed):

benchmark/benchmarks/mountpoint.py
```diff
@@ -52,7 +52,7 @@ def mount_mp(cfg: DictConfig, mount_dir: str) -> Dict[str, Any]:
             binary_name = "mount-s3"
 
         log.info(f"Compiling {binary_name} with features: {features}")
-        mountpoint_binary = build_binary(binary_name, features, build_env)
+        mountpoint_binary = build_binary(binary_name, features, build_env, flamegraph_enhancement)
```
sahityadg: rename this also with_flamegraph

PR #1579, opened 2025-08-15
Capability to set EventLoopGroup thread count as a configurable CRT client config at runtime

mansi153:
Add capability to set EventLoopGroup thread count as a configurable CRT client config at runtime so we can override the current default with different values (for e.g., during performance benchmarking) by setting an environment variable `UNSTABLE_CRT_EVENTLOOP_THREADS`. 

No, it only introduces an unstable environment variable based capability to _optionally_ configure the thread count for CRT's event loop group. It also introduces a change in the MP/client/prefetcher benchmarks to set that value using benchmark.py. The current default behaviours for Mountpoint and benchmarks stay the same.

No, as it's not a behavioural change.

---

### DIFF
--- a/mountpoint-s3-client/examples/client_benchmark.rs
+++ b/mountpoint-s3-client/examples/client_benchmark.rs
@@ -249,6 +249,14 @@ fn create_s3_client_config(region: &str, args: &CliArgs, nics: Vec<String>) -> S
         );
     }
 
+    const ENV_VAR_KEY_CRT_ELG_THREADS: &str = "UNSTABLE_CRT_EVENTLOOP_THREADS";
+    if let Some(crt_elg_threads) = std::env::var_os(ENV_VAR_KEY_CRT_ELG_THREADS) {
+        let crt_elg_threads = crt_elg_threads.to_string_lossy().parse::<u16>().unwrap_or_else(|_| {
+            panic!("Invalid value for environment variable {ENV_VAR_KEY_CRT_ELG_THREADS}. Must be positive integer.")
+        });
+        config = config.event_loop_threads(crt_elg_threads);
+    }
+
     config
 }
 
--- a/benchmark/benchmarks/benchmark_config_parser.py
+++ b/benchmark/benchmarks/benchmark_config_parser.py
@@ -44,6 +44,7 @@ class BenchmarkConfigParser:
             'with_perf_stat': getattr(self.cfg.monitoring, 'with_perf_stat', False),
             'with_flamegraph': getattr(self.cfg.monitoring, 'with_flamegraph', False),
             'download_checksums': getattr(self.cfg, 'download_checksums', True),
+            'crt_eventloop_threads': getattr(self.cfg, 'crt_eventloop_threads', None),
         }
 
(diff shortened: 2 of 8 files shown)


mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -249,6 +249,14 @@ fn create_s3_client_config(region: &str, args: &CliArgs, nics: Vec<String>) -> S
         );
     }
 
+    const ENV_VAR_KEY_CRT_ELG_THREADS: &str = "UNSTABLE_CRT_EVENTLOOP_THREADS";
+    if let Some(crt_elg_threads) = std::env::var_os(ENV_VAR_KEY_CRT_ELG_THREADS) {
+        let crt_elg_threads = crt_elg_threads.to_string_lossy().parse::<u16>().unwrap_or_else(|_| {
+            panic!("Invalid value for environment variable {ENV_VAR_KEY_CRT_ELG_THREADS}. Must be positive integer.")
```
c-hagem, line 255: I don't think that we should panic just because a user sets an unexpected value for an environment variable -- we are potentially stopping someone from using MP. For prefetch window, we just warn and use default -- should we do the same here (c.f. https://github.com/awslabs/mountpoint-s3/blob/73c9de1a2ab6e5130aac9cb6d269037601a67023/mountpoint-s3-fs/src/prefetch.rs#L144)?

c-hagem: I guess doesn't really apply to client -- but poerhaps to main Mountpoint binary?

mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -249,6 +249,14 @@ fn create_s3_client_config(region: &str, args: &CliArgs, nics: Vec<String>) -> S
         );
     }
 
+    const ENV_VAR_KEY_CRT_ELG_THREADS: &str = "UNSTABLE_CRT_EVENTLOOP_THREADS";
```
renanmagagnin, line 252: can we share one declaration among the three files instead of redeclaring it in each?

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: This is similar to the behaviour for `UNSTABLE_MOUNTPOINT_MAX_BACKGROUND` and `UNSTABLE_MOUNTPOINT_CONGESTION_THRESHOLD`. I am inclined to panic because it's an experimental setup so I think the user would want to know if they're using the capability wrongly instead of MP failing silently and falling back to default behaviour. WDYT?

mountpoint-s3-client/examples/client_benchmark.rs
mansi153: Where do you suggest putting the common config? I guess I also want to put the common helper method (for e.g the env var helper here) in a common place, but currently we don't have a module/crate to do that. I intend to move common benchmark/generic helper stuff in a common place in the future, but this was a bit urgent so I PRed the change without much refactor.

c-hagem (approved):

PR #1581, opened 2025-08-15
Add unstable flag to use FUSE_DEV_IOC_CLONE

adpeace:
When creating threads to communicate with the FUSE device, instead of sharing a file descriptor, allow the option to create a new file descriptor with a clone of the session using the `FUSE_DEV_IOC_CLONE` ioctl.

To enable this, set the `UNSTABLE_MOUNTPOINT_CLONE_FUSE_FD` environment variable to `true` for the mountpoint process.

This behavior is similar to libfuse which also has this as an option.  The benefit of enabling this is that some of the processing queues in the kernel data structures are no longer shared, which theoretically reduces lock contention.  Several tests on current mountpoint do not show a statistically significant performance benefit with this change so it's not the default, however, it could be that other bottlenecks are currently more important and this will be more relevant once they're addressed.

The change also updates the benchmark.py to offer this option when running benchmarks, so we can evaluate performance with and without the change.

No

No

---

### DIFF
--- a/benchmark/benchmarks/mountpoint.py
+++ b/benchmark/benchmarks/mountpoint.py
@@ -116,6 +116,9 @@ def mount_mp(cfg: DictConfig, mount_dir: str) -> Dict[str, Any]:
     if mp_config['mountpoint_congestion_threshold'] is not None:
         mp_env["UNSTABLE_MOUNTPOINT_CONGESTION_THRESHOLD"] = str(mp_config["mountpoint_congestion_threshold"])
 
+    if mp_config['mountpoint_clone_fuse_fd'] is not None:
+        mp_env["UNSTABLE_MOUNTPOINT_CLONE_FUSE_FD"] = str(mp_config["mountpoint_clone_fuse_fd"]).lower()
+
     mp_env["UNSTABLE_MOUNTPOINT_PID_FILE"] = f"{mount_dir}.pid"
     if not common_config['download_checksums']:
         mp_env["EXPERIMENTAL_MOUNTPOINT_NO_DOWNLOAD_INTEGRITY_VALIDATION"] = "ON"
--- a/mountpoint-s3-fuser/Cargo.toml
+++ b/mountpoint-s3-fuser/Cargo.toml
@@ -21,7 +21,7 @@ page_size = "0.6.0"
 serde = { version = "1.0.102", features = ["std", "derive"], optional = true }
 smallvec = "1.6.1"
 zerocopy = { version = "0.8", features = ["derive"] }
-nix = { version = "0.29.0", features = ["fs", "user"] }
+nix = { version = "0.29.0", features = ["fs", "user", "ioctl"] }
 
 [dev-dependencies]
 env_logger = "0.11.3"
--- a/mountpoint-s3-fuser/src/channel.rs
+++ b/mountpoint-s3-fuser/src/channel.rs
@@ -1,5 +1,8 @@
 use std::{
-    fs::File,
+    fs::{
+        OpenOptions,
+        File,
+    },
     io,
     os::{
         fd::{AsFd, BorrowedFd},
@@ -9,7 +12,8 @@ use std::{
 };
 
 use libc::{c_int, c_void, size_t};
-
+use nix::ioctl_read;
+use nix::Result as NixResult;
 use crate::reply::ReplySender;
 
 /// A raw communication channel to the FUSE kernel driver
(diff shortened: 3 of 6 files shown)


benchmark/benchmarks/mountpoint.py
```diff
@@ -116,6 +116,11 @@ def mount_mp(cfg: DictConfig, mount_dir: str) -> Dict[str, Any]:
     if mp_config['mountpoint_congestion_threshold'] is not None:
         mp_env["UNSTABLE_MOUNTPOINT_CONGESTION_THRESHOLD"] = str(mp_config["mountpoint_congestion_threshold"])
 
+    if mp_config['mountpoint_clone_fuse_fd'] is not None:
+        mp_env["UNSTABLE_MOUNTPOINT_CLONE_FUSE_FD"] = str(mp_config["mountpoint_clone_fuse_fd"]).lower()
+
+    mp_env['MOUNTPOINT_LOG'] = 'fuser=info'
```
c-hagem: This disables default mountpoint logging.

adpeace: true, will fix or remove.  We don't really need this but I wanted to see my log message confirming that the change had actually been enabled.

mountpoint-s3-fuser/Cargo.toml
passaro, line 1: `mountpoint-s3-fuser` is a copy from our internal fork of `fuser`, hosted on the `fuser/fork` [branch](https://github.com/awslabs/mountpoint-s3/tree/fuser/fork).

We should commit these changes to that branch and then use the [vendor-fuser.sh](https://github.com/awslabs/mountpoint-s3/blob/main/vendor-fuser.sh) script to bring them here.

mountpoint-s3-fuser/src/channel.rs
```diff
@@ -30,6 +34,29 @@ impl Channel {
         Self(device)
     }
 
+    /// Create a worker channel by opening a new /dev/fuse file descriptor and
+    /// associating it with the session using FUSE_DEV_IOC_CLONE ioctl.
+    pub fn clone_channel(&self) -> io::Result<Self> {
+        let worker_file = OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open("/dev/fuse")?;
```
passaro, line 43: It's not obvious to me why we are issuing a new open instead of duplicating the current file descriptor. Is this what libfuse does? It could do with some explanation.

passaro: Also, will this support the "fd mounts" used in the CSI Driver v2? I'd guess not.

mountpoint-s3-fuser/src/channel.rs
```diff
@@ -30,6 +34,29 @@ impl Channel {
...
+        let worker_fd = worker_file.as_raw_fd();
+        let mut session_fd = self.0.as_raw_fd();
+
+        // Associate the worker fd with the session fd using FUSE_DEV_IOC_CLONE
+        ioctl_read!(fuse_dev_ioc_clone, 229, 0, libc::c_int);
```
passaro, line 48: doc on the magic number

mountpoint-s3-fuser/src/session.rs
```diff
@@ -148,8 +148,35 @@ impl<FS: Filesystem> Session<FS> {
...
+            .unwrap_or(false);
+
+        info!("FUSE channel cloning: {}", if clone_fd { "enabled" } else { "disabled" });
+
+        // Create a worker channel for this thread if cloning is enabled
+        // Otherwise use the original channel directly
+        if clone_fd {
+            // Create a worker channel for this thread using FUSE_DEV_IOC_CLONE
+            // This allows multiple threads to read from the FUSE device without contention
+            let worker_channel = self.ch.clone_channel()?;
```
passaro, line 169: Should we think about falling back to the default on error?

mountpoint-s3-fuser/src/session.rs
```diff
@@ -148,8 +148,35 @@ impl<FS: Filesystem> Session<FS> {
...
     /// This version also notifies callers of kernel requests before and after they
     /// are dispatched to the filesystem.
-    pub fn run_with_callbacks<FA, FB>(&self, mut before_dispatch: FB, mut after_dispatch: FA) -> io::Result<()> 
-    where 
+    pub fn run_with_callbacks<FA, FB>(&self, before_dispatch: FB, after_dispatch: FA) -> io::Result<()>
+    where
+        FB: FnMut(&Request<'_>),
+        FA: FnMut(&Request<'_>),
+    {
+        // Check if UNSTABLE_MOUNTPOINT_CLONE_FUSE_FD environment variable is set to true
```
passaro, line 156: My preference would be to avoid these deeply nested environment overrides. Could we surface these as a `fuser` configuration option and have the env variable checked in `mount-s3`?

passaro (changes requested):

mountpoint-s3-fuser/Cargo.toml
adpeace: Oh right, of course.  Will do that, thanks.

mountpoint-s3-fuser/src/channel.rs
adpeace: This is the same way libfuse does it.  Here's their code: https://github.com/libfuse/libfuse/blob/8b8ce5a8242e78db4f1d22437d8e1f015d71765f/lib/fuse_loop_mt.c#L258-L293

It's true that this won't work for CSI.  There's not an obvious way to do so, apart from passing a number of FDs rather than a single one?  I wouldn't see that as a blocker for an unstable flag, but agreed it's something we need to consider.

mountpoint-s3-fuser/src/session.rs
adpeace: yeah; I nearly wrote that code and then decided not to.  My reasoning is that the customer opted into this behaviour by setting the flag, and it'd be better to fail rather than invisibly falling back.  Especially in e.g. our performance tests, we need to know that the configuration was actually active.

If this were default behaviour we'd definitely have to fall back – the flag has been around for a while, introduced in kernel 4.2 I think, but not sure if any of our supported distros have kernels with it, for example.

mountpoint-s3-fuser/src/session.rs
adpeace: yeah; agree that would be cleaner.

PR #1584, opened 2025-08-22
Adapt GetObject part size to the requested range

passaro:
Set the part size on GetObject requests to the requested range, if it is smaller than the client read part size. Since in this case the request will not be split in parts, the change only affects the way the CRT internally reserves the memory for the data to return.

Addresses an issue where the Prefetcher could cause higher peak memory usage with the unified memory pool compared to the previous approach (internal CRT pool + copy into a new buffer), since small requests would consume part-size buffers for longer. 

No functional changes.

**TODO: update client and fs**

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client/get_object.rs
+++ b/mountpoint-s3-client/src/s3_crt_client/get_object.rs
@@ -77,7 +77,16 @@ impl S3CrtClient {
                 .map_err(S3RequestError::construction_failure)?;
 
             let mut options = message.into_options(S3Operation::GetObject);
-            options.part_size(self.inner.read_part_size as u64);
+
+            // Use the client read part size, or the requested range if smaller.
+            let part_size = (self.inner.read_part_size as u64).min(
+                params
+                    .range
+                    .as_ref()
+                    .map(|range| range.end - range.start)
+                    .unwrap_or(u64::MAX),
+            );
+            options.part_size(part_size);
 
             let mut headers_sender = Some(event_sender.clone());
             let part_sender = event_sender.clone();
--- a/mountpoint-s3/src/run.rs
+++ b/mountpoint-s3/src/run.rs
@@ -211,6 +211,7 @@ fn mount(args: CliArgs, client_builder: impl ClientBuilder) -> anyhow::Result<Fu
     // Set up a paged memory pool
     let pool = PagedPool::new_with_candidate_sizes([
         args.cache_block_size_in_bytes() as usize,
+        mountpoint_s3_fs::s3::config::INITIAL_READ_WINDOW_SIZE,
         client_config.part_config.read_size_bytes,
         client_config.part_config.write_size_bytes,
     ]);
--- a/mountpoint-s3-client/tests/client.rs
+++ b/mountpoint-s3-client/tests/client.rs
@@ -2,9 +2,15 @@
 
 pub mod common;
 
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+
+use aws_sdk_s3::primitives::ByteStream;
 use common::*;
-use mountpoint_s3_client::S3CrtClient;
-use mountpoint_s3_client::config::S3ClientConfig;
+use futures::{StreamExt as _, pin_mut};
+use mountpoint_s3_client::config::{MemoryPool, MetaRequestType, S3ClientConfig};
+use mountpoint_s3_client::types::{GetBodyPart, GetObjectParams};
+use mountpoint_s3_client::{ObjectClient, S3CrtClient};
 
 #[tokio::test]
 async fn test_create_client_twice() {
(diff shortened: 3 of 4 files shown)


mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -77,7 +77,16 @@ impl S3CrtClient {
...
 
             let mut options = message.into_options(S3Operation::GetObject);
-            options.part_size(self.inner.read_part_size as u64);
+
+            // Use the client read part size, or the requested range if smaller.
+            let part_size = (self.inner.read_part_size as u64).min(
+                params
+                    .range
+                    .as_ref()
+                    .map(|range| range.end - range.start)
```
mansi153, line 86: This means the `first_request_stream` will always request the part size `min(INITIAL_READ_WINDOW_SIZE, object_size)`, right?

Few questions:

- Is it the right understanding that the new logic also makes the splitting of metarequests more sensible, given we always start the second metarequest with range [INITIAL_READ_WINDOW_SIZE, end of object] (and not [PART_SIZE, end of object]) even though the first request has caused CRT to return up to part size?
- I believe this should also reduce the ttfb for the first read? 
- Do we have this risk of blocked memory in writes too, and do we intend to make any changes there to reduce default part-size etc.?

mountpoint-s3-client/src/s3_crt_client/get_object.rs
```diff
@@ -77,7 +77,16 @@ impl S3CrtClient {
                 .map_err(S3RequestError::construction_failure)?;
 
             let mut options = message.into_options(S3Operation::GetObject);
-            options.part_size(self.inner.read_part_size as u64);
+
+            // Use the client read part size, or the requested range if smaller.
```
mansi153, line 81: Can we pls add a slightly more elaborate explanation of why we're doing this, for future context? The PR description is sufficient, however that sometimes gets lost when we refactor code for unrelated reasons :(

mountpoint-s3/src/run.rs
```diff
@@ -211,6 +211,7 @@ fn mount(args: CliArgs, client_builder: impl ClientBuilder) -> anyhow::Result<Fu
     // Set up a paged memory pool
     let pool = PagedPool::new_with_candidate_sizes([
         args.cache_block_size_in_bytes() as usize,
+        mountpoint_s3_fs::s3::config::INITIAL_READ_WINDOW_SIZE,
```
mansi153, line 214: For my understanding, this implies that for objects with [INITIAL_READ_WINDOW_SIZE < objects_size < read_part_size], we will still be using a peak memory of [INITIAL_READ_WINDOW_SIZE + part_size] right (because we advance the CRT backpressure window halfway through the first metarequest being read)?

passaro: Only if `2 * INITIAL_READ_WINDOW_SIZE <= object_size < read_part_size`. What matters is the range on the second request, which will be `object_size - INITIAL_READ_WINDOW_SIZE`. If that is greater than `INITIAL_READ_WINDOW_SIZE`, it will require a full `read_part_size` buffer.

mountpoint-s3-client/src/s3_crt_client/get_object.rs
passaro: > This means the `first_request_stream` will always request the part size `min(INITIAL_READ_WINDOW_SIZE, object_size)`, right?

Right.

> Few questions:
>
> - Is it the right understanding that the new logic also makes the splitting of metarequests more sensible, given we always start the second metarequest with range [INITIAL_READ_WINDOW_SIZE, end of object] (and not [PART_SIZE, end of object]) even though the first request has caused CRT to return up to part size?

That is not changed. We are still requesting `INITIAL_READ_WINDOW_SIZE` bytes as before, the only difference is that we also tell the CRT to use `INITIAL_READ_WINDOW_SIZE` as the part size.

> - I believe this should also reduce the ttfb for the first read? 

No changes either.

> - Do we have this risk of blocked memory in writes too, and do we intend to make any changes there to reduce default part-size etc.?

That's always been the case, but for writes we do not know the size in advance, so no trivial fix there.

mansi153: > That is not changed. We are still requesting INITIAL_READ_WINDOW_SIZE bytes as before, the only difference is that we also tell the CRT to use INITIAL_READ_WINDOW_SIZE as the part size.

Right! But won't this also change the S3 request range (and hence, size) now that we're requesting a 1.128M part instead of an 8M part? Previously we would receive data upto 8M offset but then re-request the 1.128M-8M range in the second meta request. Now we only receive 1.128M in the first go because it's the part size. So that is a change with this fix, no?

(^My comment about reduced ttfb was also based on this understanding, that smaller part GET => smaller S3 latency)

mountpoint-s3/src/run.rs
mansi153: Oh right, yes, makes sense

mountpoint-s3-client/src/s3_crt_client/get_object.rs
passaro: We were already only requesting `INITIAL_READ_WINDOW_SIZE` (1.125 MiB) before the change. `read_part_size` only comes into play for the buffer reservation, not for the data that is downloaded.

PR #1586, opened 2025-08-28
Add option to use latest directory and add run printing

c-hagem:
Adds two features to the autogroup.py script, namely
 - the possibility to print either the numbers of all runs of a category (ordered descendingly by throughput), when specifying `--runs=all`, or just to print max, median and min run numbers `--runs=rep` (for representative)
-  makes default order of throughputs consistent with sorting order if runs are specified
-  adds possibility to use print the latest run, which is used automatically when no directory is specified

Only changes behaviour of the benchmarking autogroup script, so no Changelog entries / version bumps needed.

---

### DIFF
--- a/benchmark/analysis-scripts/autogroup.py
+++ b/benchmark/analysis-scripts/autogroup.py
@@ -13,7 +13,7 @@ import glob
 import csv
 import warnings
 import statistics
-import sys
+
 
 from tabulate import tabulate
 from collections import defaultdict
@@ -157,8 +157,7 @@ def find_varying_parameters(all_configs: List[Dict[str, Any]]) -> Set[str]:
 def find_multirun_dir(index: int = 0) -> str:
     """Find the Nth latest directory in multirun (0=most recent, 1=previous, etc.)"""
     if not Path('multirun').exists():
-        warnings.warn("multirun directory not found")
-        sys.exit(1)
+        raise FileNotFoundError("multirun directory not found")
 
     # This ensures that alphabetical sorting will correctly find latest
     sorted_subdirs = sorted(
@@ -168,9 +167,12 @@ def find_multirun_dir(index: int = 0) -> str:
     )
 
     if not sorted_subdirs:
-        warnings.warn("No experiment directories found in multirun")
-        sys.exit(1)
-    return sorted_subdirs[index][1]
+        raise FileNotFoundError("No experiment directories found in multirun")
+
+    try:
+        return sorted_subdirs[index][1]
+    except IndexError:
+        raise IndexError(f"Index {index} is out of range. Only {len(sorted_subdirs)} experiment directories found.")
 
 
 def main() -> None:
@@ -181,7 +183,9 @@ def main() -> None:
 
     parser.add_argument('--csv-output', help='Optional CSV file to write the results to')
     parser.add_argument(
-        '--runs', choices=['tri', 'all'], help='Show run numbers in results (tri=min/median/max, all=all runs)'
+        '--runs',
+        choices=['rep', 'all'],
+        help='Show run numbers in results (rep(resentative)=min/median/max, all=all runs)',
     )
     args = parser.parse_args()
 
@@ -192,8 +196,8 @@ def main() -> None:
         try:
             base_dir = find_multirun_dir()
             print(f"Using inferred base directory: {base_dir}")
-        except IndexError:
-            print("Invalid argument, cannot find latest directory")
+        except (IndexError, FileNotFoundError) as e:
+            print(f"Error: {e}")
             return
 
     # List to store all results
@@ -233,8 +237,7 @@ def main() -> None:
 
     results_rows = []
     for config_key, throughput_data in grouped_results.items():
-        throughputs = [t for t, _ in throughput_data]
-        run_numbers = [r for _, r in throughput_data]
+        throughputs, run_numbers = zip(*throughput_data)
 
         row = []
         for _, value in config_key:
@@ -242,24 +245,21 @@ def main() -> None:
 
         # Add run numbers column if requested
         if args.runs:
-            if args.runs == "tri":
+            sorted_by_throughput = sorted(zip(throughputs, run_numbers), reverse=True)
+            if args.runs == "rep":
                 # Find min, max, and median run numbers based on throughput
-                sorted_by_throughput = sorted(zip(throughputs, run_numbers))
-                min_run = sorted_by_throughput[0][1]
-                max_run = sorted_by_throughput[-1][1]
+                min_run = sorted_by_throughput[-1][1]
+                max_run = sorted_by_throughput[0][1]
                 median_idx = len(sorted_by_throughput) // 2
                 median_run = sorted_by_throughput[median_idx][1]
 
                 selected_runs = [max_run, median_run, min_run]
-                # Remove duplicates while preserving order
-                unique_runs = []
-                for run in selected_runs:
-                    if run not in unique_runs:
-                        unique_runs.append(run)
+                # Remove duplicates while preserving order using dict.fromkeys()
+                # (works in python > 3.7)
+                unique_runs = list(dict.fromkeys(selected_runs))
 
                 row.append(",".join(unique_runs))
             else:
-                sorted_by_throughput = sorted(zip(throughputs, run_numbers), reverse=True)
                 all_runs = [r for _, r in sorted_by_throughput]
                 row.append(",".join(all_runs))
 
@@ -300,7 +300,7 @@ def main() -> None:
     # Display results
     if args.runs == "all":
         print("\nResults Summary (with all run numbers):")
-    elif args.runs == "tri":
+    elif args.runs == "rep":
         print("\nResults Summary (with representative run numbers):")
     else:
         print("\nResults Summary:")
--- a/benchmark/analysis-scripts/autogroup.py
+++ b/benchmark/analysis-scripts/autogroup.py
@@ -152,17 +152,102 @@ def find_varying_parameters(all_configs: List[Dict[str, Any]]) -> Set[str]:
     return varying
 
 
+def find_multirun_dir(index: int = 0, sort_by_date: bool = True) -> str:
+    """Find the Nth latest directory in multirun (0=most recent, 1=previous, etc.)"""
+    multirun_path = 'multirun'
+    if not os.path.exists(multirun_path):
+        raise FileNotFoundError("multirun directory not found")
+
+    # Get all subdirectories and their sort keys
+    all_subdirs = []
+    for item in os.listdir(multirun_path):
+        item_path = os.path.join(multirun_path, item)
+        if os.path.isdir(item_path):
+            # Find all subdirectories within this date directory
+            for subitem in os.listdir(item_path):
+                subitem_path = os.path.join(item_path, subitem)
+                if os.path.isdir(subitem_path):
+                    if sort_by_date:
+                        # Parse the date and time from the path structure
+                        # Expected format: multirun/YYYY-MM-DD/HH-MM-SS
+                        try:
+                            date_str = item  # e.g., "2025-08-27"
+                            time_str = subitem  # e.g., "16-49-08"
+                            datetime_str = f"{date_str} {time_str.replace('-', ':')}"
+                            all_subdirs.append((datetime_str, subitem_path))
+                        except Exception:
+                            # Fall back to mtime if parsing fails
+                            mtime = os.path.getmtime(subitem_path)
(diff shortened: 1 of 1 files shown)


benchmark/analysis-scripts/autogroup.py
```diff
@@ -152,17 +152,102 @@ def find_varying_parameters(all_configs: List[Dict[str, Any]]) -> Set[str]:
...
 def main() -> None:
     parser = argparse.ArgumentParser(description='Print benchmark throughput data automatically grouped')
-    parser.add_argument('--base-dir', required=True, help='Base directory containing benchmark results')
+
+    # Positional argument for base directory (optional)
+    parser.add_argument('base_dir', nargs='?', help='Base directory containing benchmark results')
+
+    # Options for automatic latest directory selection
+    parser.add_argument(
+        '--latest',
```
sahityadg: It is probably more easy to use to default to latest when no multirun folder is specified, rather than having to specify latest.

benchmark/analysis-scripts/autogroup.py
```diff
@@ -152,17 +152,102 @@ def find_varying_parameters(all_configs: List[Dict[str, Any]]) -> Set[str]:
...
+        nargs='?',
+        const=0,
+        metavar='N',
+        help='Use the Nth latest multirun directory (0=most recent, 1=previous, etc.)',
+    )
+    parser.add_argument(
+        '--latest-order',
+        choices=['date', 'mod_time'],
+        default='date',
+        help='Sort latest directories by date in path or modification time',
```
sahityadg: I'm not sure what is the usecase for sorting based on mtime. It is probably simpler to sort it based on timestamps. in multirun output path names.

benchmark/analysis-scripts/autogroup.py
```diff
@@ -152,17 +152,102 @@ def find_varying_parameters(all_configs: List[Dict[str, Any]]) -> Set[str]:
...
+
     parser.add_argument('--csv-output', help='Optional CSV file to write the results to')
+    parser.add_argument(
+        '--runs', choices=['tri', 'all'], help='Show run numbers in results (tri=min/median/max, all=all runs)'
+    )
     args = parser.parse_args()
 
+    # Validate conflicting arguments
+    if args.base_dir and args.latest is not None:
+        parser.error("Cannot specify both a base directory and --latest option")
```
sahityadg: Similar to previous comment: We can simplify this by removing --latest arg altogether. We should use the base dir when specified and if not default to latest.

benchmark/analysis-scripts/autogroup.py
```diff
@@ -184,30 +269,68 @@ def main() -> None:
...
-        "Min (Gbps)",
-        "Max (Gbps)",
-    ]
-    aggregated_rows = []
-    for config_key, throughputs in grouped_results.items():
+        grouped_results[key].append((throughput, iter_num))
+
+    # Generate aggregated results table with optional Run Numbers column
+    if args.runs:
+        results_headers = varying_params + [
```
sahityadg: To avoid updating headers in two places, we could extract the common results_headers and prepend with run numbers based on arg.run option.

benchmark/analysis-scripts/autogroup.py
```diff
@@ -152,17 +152,102 @@ def find_varying_parameters(all_configs: List[Dict[str, Any]]) -> Set[str]:
...
+    parser.add_argument(
+        '--latest-order',
+        choices=['date', 'mod_time'],
+        default='date',
+        help='Sort latest directories by date in path or modification time',
+    )
+
     parser.add_argument('--csv-output', help='Optional CSV file to write the results to')
+    parser.add_argument(
+        '--runs', choices=['tri', 'all'], help='Show run numbers in results (tri=min/median/max, all=all runs)'
```
sahityadg: naive question probably - what does tri stand for here?

benchmark/analysis-scripts/autogroup.py
```diff
@@ -152,17 +152,102 @@ def find_varying_parameters(all_configs: List[Dict[str, Any]]) -> Set[str]:
     return varying
 
 
+def find_multirun_dir(index: int = 0, sort_by_date: bool = True) -> str:
```
sahityadg: I think we should always sort by date and remove mtime based sorting. As I said elsewhere, I'm not sure when that'd diverge from the path name or the use case for mtime

Isn't explicitly specifying multirun folder simpler when we want a different run results than maintaining these conventions of prev etc.?

benchmark/analysis-scripts/autogroup.py
c-hagem: tripartite

benchmark/analysis-scripts/autogroup.py
c-hagem: Okay, will remove the choices for ordering

benchmark/analysis-scripts/autogroup.py
c-hagem: Changed the logic slightly to remove duplication

benchmark/analysis-scripts/autogroup.py
c-hagem: Removed mtime based sorting, I kept the option to specify an index

benchmark/analysis-scripts/autogroup.py
c-hagem: Yes, it now just uses the explicit dir if specified and otherwise defaults to latest

sahityadg (dismissed):

benchmark/analysis-scripts/autogroup.py
```diff
@@ -152,17 +154,53 @@ def find_varying_parameters(all_configs: List[Dict[str, Any]]) -> Set[str]:
     return varying
 
 
+def find_multirun_dir(index: int = 0) -> str:
+    """Find the Nth latest directory in multirun (0=most recent, 1=previous, etc.)"""
+    if not Path('multirun').exists():
+        warnings.warn("multirun directory not found")
```
muddyfish: Raise an exception here?

benchmark/analysis-scripts/autogroup.py
```diff
@@ -152,17 +154,53 @@ def find_varying_parameters(all_configs: List[Dict[str, Any]]) -> Set[str]:
...
+    sorted_subdirs = sorted(
+        [(f"{p.parent.name} {p.name.replace('-', ':')}", str(p)) for p in Path('multirun').glob("*/*/")],
+        key=lambda x: x[0],
+        reverse=True,
+    )
+
+    if not sorted_subdirs:
+        warnings.warn("No experiment directories found in multirun")
+        sys.exit(1)
+    return sorted_subdirs[index][1]
```
muddyfish: Above there are warnings, but if the index is out of range, this can raise an exception without a handy description.

benchmark/analysis-scripts/autogroup.py
muddyfish: I've also not heard of it. Wondering if there's a slightly less technical term which also works here

benchmark/analysis-scripts/autogroup.py
```diff
@@ -184,30 +222,56 @@ def main() -> None:
...
+    # Generate aggregated results table with optional Run Numbers column
+    results_headers = (
+        varying_params
+        + (["Run Numbers"] if args.runs else [])
+        + ["Count", "Max (Gbps)", "Median (Gbps)", "Min (Gbps)", "Std Dev (Gbps)"]
+    )
+
+    results_rows = []
+    for config_key, throughput_data in grouped_results.items():
+        throughputs = [t for t, _ in throughput_data]
```
muddyfish: Nit: `throughputs, run_numbers = zip(*throughput_data)`

benchmark/analysis-scripts/autogroup.py
```diff
@@ -184,30 +222,56 @@ def main() -> None:
...
         row = []
         for _, value in config_key:
             row.append(value)
+
+        # Add run numbers column if requested
+        if args.runs:
+            if args.runs == "tri":
+                # Find min, max, and median run numbers based on throughput
+                sorted_by_throughput = sorted(zip(throughputs, run_numbers))
+                min_run = sorted_by_throughput[0][1]
```
muddyfish: Unclear why we're zipping after just unzipping above

benchmark/analysis-scripts/autogroup.py
```diff
@@ -184,30 +222,56 @@ def main() -> None:
...
+                selected_runs = [max_run, median_run, min_run]
+                # Remove duplicates while preserving order
+                unique_runs = []
+                for run in selected_runs:
+                    if run not in unique_runs:
+                        unique_runs.append(run)
+
+                row.append(",".join(unique_runs))
+            else:
+                sorted_by_throughput = sorted(zip(throughputs, run_numbers), reverse=True)
```
muddyfish: Why are we reverse sorting here but not above?

benchmark/analysis-scripts/autogroup.py
```diff
@@ -184,30 +222,56 @@ def main() -> None:
...
+                # Find min, max, and median run numbers based on throughput
+                sorted_by_throughput = sorted(zip(throughputs, run_numbers))
+                min_run = sorted_by_throughput[0][1]
+                max_run = sorted_by_throughput[-1][1]
+                median_idx = len(sorted_by_throughput) // 2
+                median_run = sorted_by_throughput[median_idx][1]
+
+                selected_runs = [max_run, median_run, min_run]
+                # Remove duplicates while preserving order
+                unique_runs = []
```
muddyfish: How large is this list realistically getting? This approach is `O(n^2)`

c-hagem: The list only has.3 elements (i.e. selected_runs).

benchmark/analysis-scripts/autogroup.py
c-hagem: Above we pick min, p50 and max , I guess we could use reverse sorting in both

benchmark/analysis-scripts/autogroup.py
c-hagem: This will be caught as an IndexError, since we currently only use index 0 i think.

benchmark/analysis-scripts/autogroup.py
c-hagem: Adjusted to reverse sort once

benchmark/analysis-scripts/autogroup.py
c-hagem: Fixed to faster method

benchmark/analysis-scripts/autogroup.py
c-hagem: Fixed

benchmark/analysis-scripts/autogroup.py
```diff
@@ -184,30 +226,52 @@ def main() -> None:
...
+    for config_key, throughput_data in grouped_results.items():
+        throughputs, run_numbers = zip(*throughput_data)
+
         row = []
         for _, value in config_key:
             row.append(value)
+
+        # Add run numbers column if requested
+        if args.runs:
+            sorted_by_throughput = sorted(zip(throughputs, run_numbers), reverse=True)
```
muddyfish, line 248: Isn't `zip(throughputs, run_numbers)` equivalent to `throughput_data`?

benchmark/analysis-scripts/autogroup.py
```diff
@@ -184,30 +226,52 @@ def main() -> None:
...
+                # Find min, max, and median run numbers based on throughput
+                min_run = sorted_by_throughput[-1][1]
+                max_run = sorted_by_throughput[0][1]
+                median_idx = len(sorted_by_throughput) // 2
+                median_run = sorted_by_throughput[median_idx][1]
+
+                selected_runs = [max_run, median_run, min_run]
+                # Remove duplicates while preserving order using dict.fromkeys()
+                # (works in python > 3.7)
+                unique_runs = list(dict.fromkeys(selected_runs))
```
muddyfish, line 259: I think your previous approach was more readable 😅
Perhaps just move it to a function instead?

benchmark/analysis-scripts/autogroup.py
```diff
@@ -184,30 +226,52 @@ def main() -> None:
...
+
+                row.append(",".join(unique_runs))
+            else:
+                all_runs = [r for _, r in sorted_by_throughput]
```
muddyfish, line 263: Isn't this `run_numbers`?

tadiwa-aizen:
Closing this as we've implemented this functionality in the daily benchmark cdk @c-hagem

PR #1587, opened 2025-08-28
Add a fuser mount option to use FUSE_DEV_IOC_CLONE.

adpeace:
This mimics functionality from libfuse. When enabled, it will create a new file descriptor for each fuse thread by opening /dev/fuse again and using the FUSE_DEV_IOC_CLONE ioctl to associate the file handle with the original filesystem.  The benefit of enabling this is that some of the processing queues in the kernel data structures are no longer shared, which theoretically reduces lock contention.

No.

No.

---

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -1,5 +1,5 @@
 use std::{
-    fs::File,
+    fs::{File, OpenOptions},
     io,
     os::{
         fd::{AsFd, BorrowedFd},
@@ -8,9 +8,21 @@ use std::{
     sync::Arc,
 };
 
-use libc::{c_int, c_void, size_t};
-
 use crate::reply::ReplySender;
+use libc::{c_int, c_void, size_t};
+use nix::ioctl_read;
+use nix::Result as NixResult;
+
+/// FUSE_DEV_IOC_CLONE ioctl constant matching the constant defined by the FUSE kernel module in fuse.h
+const IOCTL_FUSE_DEV_IOC_CLONE: u8 = 229;
+
+ioctl_read!(
+    /// Ioctl to clone a fuse session onto a new file handle
+    fuse_dev_ioc_clone,
+    IOCTL_FUSE_DEV_IOC_CLONE,
+    0,
+    libc::c_int
+);
 
 /// A raw communication channel to the FUSE kernel driver
 #[derive(Debug)]
@@ -28,6 +40,31 @@ impl Channel {
     /// the given path to the channel.
     pub(crate) fn new(device: Arc<File>) -> Self {
         Self(device)
+    }
+
+    /// Create a worker channel by opening a new /dev/fuse file descriptor and
+    /// associating it with the session using FUSE_DEV_IOC_CLONE ioctl.
+    pub fn clone_channel(&self) -> io::Result<Self> {
+        let worker_file = OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open("/dev/fuse")?;
+        let worker_fd = worker_file.as_raw_fd();
+        let mut session_fd = self.0.as_raw_fd();
+
+        // Associate the worker fd with the session fd using FUSE_DEV_IOC_CLONE
+        // SAFETY: `session_fd` is a valid open file descriptor. The ioctl
+        // FUSE_DEV_IOC_CLONE expects a pointer to an int containing the fd
+        // to clone from. The pointer is valid for the duration of the call.
+        let result: NixResult<libc::c_int> = unsafe {
+            let val = &mut session_fd as *mut libc::c_int;
+            fuse_dev_ioc_clone(worker_fd, val)
+        };
+
+        if let Err(err) = result {
+            return Err(io::Error::new(io::ErrorKind::Other, err));
+        }
+        Ok(Self(Arc::new(worker_file)))
     }
 
     /// Receives data up to the capacity of the given buffer (can block).
--- a/src/session.rs
+++ b/src/session.rs
@@ -141,15 +141,47 @@ impl<FS: Filesystem> Session<FS> {
     /// Run the session loop that receives kernel requests and dispatches them to method
     /// calls into the filesystem.
     pub fn run(&self) -> io::Result<()> {
-        self.run_with_callbacks(|_| {}, |_| {})
+        self.run_with_callbacks(|_| {}, |_| {}, false)
     }
 
     /// Run the session loop that receives kernel requests and dispatches them to method
     /// calls into the filesystem.
     /// This version also notifies callers of kernel requests before and after they
     /// are dispatched to the filesystem.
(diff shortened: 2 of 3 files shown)


adpeace:
this built locally, need to look at what I did wrong, converted back to draft for now.

src/channel.rs
```diff
@@ -30,6 +37,29 @@ impl Channel {
...
+    pub fn clone_channel(&self) -> io::Result<Self> {
+        let worker_file = OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open("/dev/fuse")?;
+        let worker_fd = worker_file.as_raw_fd();
+        let mut session_fd = self.0.as_raw_fd();
+
+        // Associate the worker fd with the session fd using FUSE_DEV_IOC_CLONE
+        ioctl_read!(fuse_dev_ioc_clone, IOCTL_FUSE_DEV_IOC_CLONE, 0, libc::c_int);
```
passaro: This is defining a `fn`. We should make it clear in the comment and consider making it a rust-doc. Also, move outside `close_channel` or at the top of its scope.

src/channel.rs
```diff
@@ -30,6 +37,29 @@ impl Channel {
...
+        let result: NixResult<libc::c_int> = unsafe {
```
passaro, line 59: Add `// SAFETY: ` comment

src/mnt/mount_options.rs
```diff
@@ -55,6 +55,11 @@ pub enum MountOption {
     Sync,
     /// All I/O will be done asynchronously
     Async,
+    /// Use the FUSE_DEV_IOC_CLONE ioctl.
+    /// This will be used to create a fresh FD for each fuse worker thread.
+    /// This is also a libfuse option, we mimic the functionality in fuser.
+    CloneFd,
```
passaro: Is `MountOption` the right place for this? This is a property of the session, why should we send it to the kernel? 
Could we just make it an extra parameter (or a new variant) of `Session:run_with_callbacks` ?

src/session.rs
```diff
@@ -67,6 +67,8 @@ pub struct Session<FS: Filesystem> {
     pub(crate) initialized: AtomicBool,
     /// True if the filesystem was destroyed (destroy operation done)
     pub(crate) destroyed: AtomicBool,
+    /// Whether to use FUSE_DEV_IOC_CLONE ioctl for channel cloning
+    pub(crate) clone_fuse_fd: AtomicBool,
```
passaro: Consider whether we need it as a session field. If we do, why atomic? We don't expect to mutate it, right?

src/session.rs
```diff
@@ -161,10 +191,9 @@ impl<FS: Filesystem> Session<FS> {
             std::mem::align_of::<abi::fuse_in_header>(),
         );
         loop {
-            // Read the next request from the given channel to kernel driver
-            // The kernel driver makes sure that we get exactly one request per read
```
passaro: Can we keep (or rephrase/integrate) this comment?

src/mnt/mount_options.rs
passaro: Coming back to this: `Session:run_with_callbacks` is already an addition we made in this fork, so we could just add a parameter there.

adpeace: We could just add a parameter there.  I'm open to that approach but I thought using a `MountOption` was better because this is the public-facing interface from fuser to configure a mount, and ultimately this is a fuser feature that we're adding, then supporting in Mountpoint through an unstable flag.

Based on this comment in fuser:

```
/// Mount options accepted by the FUSE filesystem type
/// See 'man mount.fuse' for details
// TODO: add all options that 'man mount.fuse' documents and libfuse supports
```

and the libfuse implementation here:

https://github.com/libfuse/libfuse/blob/63579dffce8ab079a1a1eab047e63dc1211185d2/lib/helper.c#L51

I thought it was a fit.

The corresponding mountpoint change that matches this implementation is here: https://github.com/adpeace/mountpoint-s3/commit/78743e81a88839cf3f364444d5031530db8b9761. 

The alternative approach is that it has to be wired through as an option through to `mountpoint_s3_fs::fuse::session::FuseSession`, where it can be coped into the worker pool state and used when calling `run_with_callbacks`.

src/session.rs
adpeace: I think it needs to be a session field because that's where the clone actually happens.  You're right it doesn't need to be mutable, will change to regular boolean (I think I saw all the other fields were Atomic here and ran of time to properly understand why that was the case, thanks for pointing it out).

src/session.rs
adpeace: Yep.

src/session.rs
adpeace: Changed to immutable.  Let me know if you had something else in mind on the other aspect of this: whether it should be a session field, or resolve conversation please.

src/mnt/mount_options.rs
adpeace: Looking at MountOption some more.  Maybe I misunderstood its intent, as the build failure for fure_pure is interesting (I see why I missed it locally – it was a different build config in a feature I'd not enabled).

adpeace: OK, so you're right that in the current fuser implementation, all the mount options seem to get passed down to libfuse.  Libfuse's implementation has several different levels of argument/option processing, some of which are applied to userspace only, and others are passed to the kernel via `fusermount`.

My testing shows that:
(1) clone_fd is not passed to the kernel nor fusermount in regular libfuse;
(2) we do pass clonefd to the fuse session creation function in fuser, which is a different implementation depending on the version of libfuse being used.
(3) in libfuse2 (not sure about 3) it gets filtered somewhere, so when fusermount is finally `execve`'d it's no longer there.

However I don't feel comfortable relying on unknown filtering.  The option should not be passed to fusermount nor fuse session creation, as you mentioned.  So options are:
(a) add a different mount option type to fuser that it interprets within the userspace program but does not pass to the kernel (the libfuse implementation).
(b) copy what you'd already done for max_threads (which is the same kind of thing: it's another option in libfuse that's handled this way).

I'm inclined to go in the direction of the latter, given it's a smaller fuser change.

src/channel.rs
adpeace: ack.

src/mnt/mount_options.rs
passaro: I'd go with (b): just like `max_threads`, this is a configuration parameter for our multi-threaded session, `FuseSession`, which can then pass it to the `run_[..]` method on the underlying `Session`.

src/session.rs
passaro: See other comment: I'd rather make it a parameter to `run_with_callbacks` since it only really make sense for a multi-threaded session.

src/channel.rs
passaro: nit: `SAFETY` comments should go above the `unsafe` block

passaro (approved):

PR #1592, opened 2025-09-01
Adding more parameters to the sweeper combinations for FUSE threads and objects.

tadiwa-aizen:
# Adding more parameters to the sweeper combinations for FUSE threads and application workers

## What changed and why?

Expanded the parameter combinations in the benchmark sweeper configuration to include more values for testing:
- `application_workers`: Added values 4 and 64 to the existing 1, 16 combination
- `fuse_threads`: Added values 1 and 64 to the existing 16 value

This change supports an ongoing investigation that requires testing across a broader range of worker and thread configurations.

No breaking changes. This only adds additional parameter combinations for benchmarking - existing functionality remains unchanged.

## Does this change need a changelog entry? Does it require a version change?

No changelog entry or version change required. This is an internal benchmarking configuration change that doesn't affect the public API or user-facing functionality.

---

### DIFF
--- a/benchmark/conf/hydra/sweeper/fio.yaml
+++ b/benchmark/conf/hydra/sweeper/fio.yaml
@@ -1,4 +1,4 @@
 # @package hydra.sweeper
 params:
   'benchmarks.fio.direct_io': false, true
-  'benchmarks.fio.fuse_threads': 16 #1, 64
+  'benchmarks.fio.fuse_threads': 1, 16, 64
--- a/benchmark/conf/hydra/sweeper/base.yaml
+++ b/benchmark/conf/hydra/sweeper/base.yaml
@@ -2,7 +2,7 @@
 _target_: hydra._internal.core_plugins.basic_sweeper.BasicSweeper
 max_batch_size: null
 params:
-  'application_workers': 1, 16 # 4, 64
+  'application_workers': 1, 4, 16, 64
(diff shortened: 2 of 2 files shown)


benchmark/conf/hydra/sweeper/fio.yaml
```diff
@@ -1,4 +1,4 @@
 # @package hydra.sweeper
 params:
   'benchmarks.fio.direct_io': false, true
-  'benchmarks.fio.fuse_threads': 16 #1, 64
\ No newline at end of file
+  'benchmarks.fio.fuse_threads': 1, 16, 64
```
sahityadg, line 4: It'd be good to restrict the combinations of app workers and fuse_threads as we don't care about all combinations of these values. Can we specify them as a pair instead?

mansi153: I had the same suggestion at some point too, but it makes me wonder how feasible it is to maintain and extend a coupled setup like that. 
- Does it become more error-prone or something that requires context to update every time? 
- Do we really need an increased number of combinations and consequently a stricter sweeper (in mainline), given we only plot 1 object and 16 objects on day-to-day basis; or should this be something one can play around with during investigations on their machine? 
- Will our Jupyter notebooks continue to work with some values/combinations of threads/app_workers missing?

PR #1605, opened 2025-09-16
Change default logging level from WARN to INFO.

### LINKED ISSUE #1244: Update default logging level from `WARN` to `INFO`
Today, Mountpoint uses `WARN` logging by default (and also disabled CRT logging entirely as it is noisy for our use case). https://github.com/awslabs/mountpoint-s3/blob/76c36003006246811a584f648d902432bb70d7df/mountpoint-s3/src/cli.rs#L498-L526

Instead, we should have `INFO` logging by default so that that you can see messages such as when the mount occurred successfully and where, even where users may capture logs without `--debug` specified.

This isn't as simple as changing one line of code. For this issue, we should review where we perform `INFO` logging in Mountpoint, and downgrade or adjust if necessary. For example, we should ensure that metric logging is turned off or downgraded when the customer does not specify --log-metrics.

Priyankakarumuru1:
Fixes #1244
- Purpose: Improve visibility of important operational messages (mount success, location) without requiring --debug flag
- Users can now see essential mount information by default
### What changed and why?
1. Default Logging Level:
- Changed default logging level from WARN to INFO in cli.rs
- Added test_info_level_logging to verify the change
- Default INFO logs now show important operational messages:
INFO ThreadId(01) mountpoint_s3::cli: target network throughput 10 Gbps
INFO ThreadId(01) fuser::session: Mounting /tmp/test-mount-new
INFO ThreadId(01) mountpoint_s3::run: successfully mounted bucket 

2. Metrics Logging:
- Metrics now show when either --log-metrics is set OR debug level is enabled
- Explicitly turn off metrics when neither condition is met

3. Log Level Optimization:
- Changed setattr logging from INFO to DEBUG level as it's implementation detail is more appropriate for debugging rather than regular operation

- All existing log levels (--debug, --no-log) continue to work as before
- Only changes the default level to show more information
- Setattr logging moved to DEBUG level to reduce noise

- Changelog entries added:
  * mountpoint-s3: Change default logging level from WARN to INFO to improve visibility of important operational messages
  * mountpoint-s3-fs: Downgrade setattr logging level from INFO to DEBUG to reduce log noise
- Version changes:
  * mountpoint-s3: v1.20.0 -> v1.21.0 (for default logging level change)
  * mountpoint-s3-fs: v0.7.1 -> v0.7.2 (for setattr logging change)

---

### DIFF
--- a/mountpoint-s3-fs/Cargo.toml
+++ b/mountpoint-s3-fs/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "mountpoint-s3-fs"
 # See `/doc/PUBLISHING_CRATES.md` to read how to publish new versions.
-version = "0.7.1"
+version = "0.7.2"
 edition = "2024"
 license = "Apache-2.0"
 repository = "https://github.com/awslabs/mountpoint-s3"
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -673,7 +673,9 @@ impl CliArgs {
             };
             let crt_verbosity = if self.debug_crt { "debug" } else { "off" };
             filter.push_str(&format!(",{AWSCRT_LOG_TARGET}={crt_verbosity}"));
-            if !self.log_metrics {
+            // Only turn off metrics if neither --log-metrics nor --debug is set.
+            // In debug mode, we want to see all available information including metrics.
+            if !self.log_metrics && !self.debug {
                 filter.push_str(&format!(",{}=off", metrics::TARGET_NAME));
             }
             filter
--- a/mountpoint-s3/src/cli.rs
+++ b/mountpoint-s3/src/cli.rs
@@ -669,12 +669,12 @@ impl CliArgs {
             let mut filter = if self.debug {
                 String::from("debug")
             } else {
-                String::from("warn")
+                String::from("info")
             };
             let crt_verbosity = if self.debug_crt { "debug" } else { "off" };
             filter.push_str(&format!(",{AWSCRT_LOG_TARGET}={crt_verbosity}"));
-            if self.log_metrics {
-                filter.push_str(&format!(",{}=info", metrics::TARGET_NAME));
+            if !self.log_metrics {
+                filter.push_str(&format!(",{}=off", metrics::TARGET_NAME));
             }
             filter
         };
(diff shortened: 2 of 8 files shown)


passaro (commented):
A couple of recommendations:

* Link to the issue in the description (#1244).
* Can you provide an example of the new default log? What is included at INFO?
* We need to ensure that metrics are not logged by default. They should only be included if `--log-metrics` was explicitly set or the logging level was DEBUG or greater.

mountpoint-s3-fs/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "mountpoint-s3-fs"
 # See `/doc/PUBLISHING_CRATES.md` to read how to publish new versions.
-version = "0.7.1"
+version = "0.7.2"
```
passaro, line 4: `Cargo.lock` should also be included in the commit whenever there are changes to the `Cargo.toml` files.

mountpoint-s3/src/cli.rs
```diff
@@ -669,12 +669,14 @@ impl CliArgs {
...
             } else {
-                String::from("warn")
+                String::from("info")
             };
             let crt_verbosity = if self.debug_crt { "debug" } else { "off" };
             filter.push_str(&format!(",{AWSCRT_LOG_TARGET}={crt_verbosity}"));
-            if self.log_metrics {
-                filter.push_str(&format!(",{}=info", metrics::TARGET_NAME));
+            if self.log_metrics || self.debug {
+                filter.push_str(&format!(",{}", metrics::TARGET_NAME)); // Will use default INFO level
```
passaro: This should not be required. I think we only need to turn off the metrics when the user does not set `--log-metrics` explicitly.

passaro (changes requested):

mountpoint-s3/src/cli.rs
```diff
@@ -669,12 +669,12 @@ impl CliArgs {
...
                 String::from("debug")
             } else {
-                String::from("warn")
+                String::from("info")
             };
             let crt_verbosity = if self.debug_crt { "debug" } else { "off" };
             filter.push_str(&format!(",{AWSCRT_LOG_TARGET}={crt_verbosity}"));
-            if self.log_metrics {
-                filter.push_str(&format!(",{}=info", metrics::TARGET_NAME));
+            if !self.log_metrics {
```
passaro: I think we should also check `&& !self.debug` here. Sorry my previous comment missed it!

Also, we should document why. Something like the comment you had before would do.

PR #1616, opened 2025-09-24
Add an option to configure cache in `mount_from_config` example

vladem:
Add an option to configure cache in `mount_from_config` example.

No.

No.

---

### DIFF
--- a/mountpoint-s3-fs/examples/config.json.example
+++ b/mountpoint-s3-fs/examples/config.json.example
@@ -17,6 +17,8 @@
         "type": "IMDSAutoConfigure"
     },
     "memory_limit_bytes": 2147483648,
-    "cache": "./cache_dir",
-    "max_cache_size_mib": 10240
+    "disk_cache": {
+        "path": "./cache_dir",
+        "limit_bytes": 1073741824
+    }
 }
--- a/mountpoint-s3-fs/examples/mount_from_config.rs
+++ b/mountpoint-s3-fs/examples/mount_from_config.rs
@@ -34,6 +34,16 @@ enum ThroughputConfig {
     Explicit { throughput: f64 },
 }
 
+/// Configuration for disk cache
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct DiskCacheConfig {
+    /// Directory path for the cache
+    path: String,
+    /// Maximum size of the cache in bytes
+    limit_bytes: Option<u64>,
+}
+
 /// Configuration options for a Mountpoint instance
 #[derive(Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
@@ -73,11 +83,8 @@ struct ConfigOptions {
     uid: Option<u32>,
     gid: Option<u32>,
 
-    // Cache options
-    /// Directory to use for caching object content
-    cache: Option<String>,
-    /// Maximum size of the cache directory in MiB
-    max_cache_size_mib: Option<u64>,
+    /// Disk cache configuration
+    disk_cache: Option<DiskCacheConfig>,
 }
 
 impl ConfigOptions {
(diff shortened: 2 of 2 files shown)


mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -141,8 +147,38 @@ impl ConfigOptions {
...
+            limit: cache_limit,
+        };
+        Some(cache_config)
+    }
+
     fn build_data_cache_config(&self) -> DataCacheConfig {
-        DataCacheConfig::default()
+        let disk_cache_config = self.disk_data_cache_config();
+        match &disk_cache_config {
+            Some(_) => {
```
mansi153, line 179: Nit: I suppose this `match` should also check for the `self.max_cache_size_mib = 0` case, because we're not actually setting a limit in that case? Why does `disk_data_cache_config` return optional DiskDataCacheConfig anyway, given it actually always constructs the `cache_config`?

mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -141,8 +147,38 @@ impl ConfigOptions {
...
+                tracing::trace!("using local disk as a cache for object content");
+            }
+            None => {
+                tracing::trace!("using no cache");
+            }
+        }
+        DataCacheConfig {
+            disk_cache_config,
+            express_cache_config: None,
```
mansi153, line 188: Where is this config being used?

mountpoint-s3-fs/examples/mount_from_config.rs
vladem: the way we treat `max_size_in_mib == 0` is confusing (=> no cache), but it's aligned with how we do that in the main binary: https://github.com/vladem/mountpoint-s3/blob/mp-example-cache/mountpoint-s3/src/cli.rs#L610

`disk_data_cache_config` will return None if `self.cache == None` or if `self.max_cache_size_mib == Some(0)` (note that `?` on `None` returns `None` from  `disk_data_cache_config`)

so I think the logging is ok here?

mountpoint-s3-fs/examples/mount_from_config.rs
vladem: this configures shared cache: https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#shared-cache (but we want it disabled here)

mountpoint-s3-fs/examples/config.json.example
```diff
@@ -16,5 +16,7 @@
     "throughput_config": {
         "type": "IMDSAutoConfigure"
     },
-    "memory_limit_bytes": 2147483648
+    "memory_limit_bytes": 2147483648,
+    "cache": "./cache_dir",
+    "max_cache_size_mib": 10240
```
passaro: Should we wrap this in a nested object? e.g.
```json
"disk_cache": {
    "path": "./cache_dir",
    "limit_bytes": 100000000
},
```

I'd also use bytes like for the memory limit.

mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -141,8 +147,38 @@ impl ConfigOptions {
         }
     }
 
+    fn disk_data_cache_config(&self) -> Option<DiskDataCacheConfig> {
+        let path = self.cache.as_ref()?;
+        let cache_limit = match self.max_cache_size_mib {
+            // Fallback to no data cache.
+            Some(0) => return None,
```
passaro: Let's error out here. In `mount-s3` we kept it only for backward compatibility.

mountpoint-s3-fs/examples/mount_from_config.rs
passaro: (optional) consider adding a constructor from `DiskDataCacheConfig` to `DataCacheConfig`.

mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -229,19 +265,28 @@ fn setup_logging(config: &ConfigOptions) -> Result<(LoggingHandle, MetricsSinkHa
     Ok((logging, metrics))
 }
 
+fn setup_disk_cache_directory(cache_config: &mut DataCacheConfig) -> anyhow::Result<Option<ManagedCacheDir>> {
```
passaro, line 276: To avoid duplication, consider moving `mountpoint_s3::run::setup_disk_cache_directory` to `mountpoint_s3_fs::data_cache::cache_directory` (or maybe even make it a method on `DataCacheConfig`?) and adding a couple of parameters for customization.

But it could also be a separate change later.

vladem: good idea! I'd like to keep this and constructor from `DiskDataCacheConfig` to `DataCacheConfig` for next PR (it'd be good to have a version which doesn't require `mountpoint-s3-fs` release first)

vladem: this is for the next PR: https://github.com/vladem/mountpoint-s3/commit/97827d00c2f457add3111936bb15530ad23b204b

PR #1618, opened 2025-09-25
Align read window with part boundaries

vladem:
Memory limiter currently doesn’t account for the [“rounded up”](https://github.com/awslabs/mountpoint-s3/blob/d598453968b10b55c01371f0f594d586fcd8be43/mountpoint-s3-crt/src/s3/client.rs#L707) memory by CRT, which means up to `8MiB` of unaccounted memory per file handle. Based on this flaw in the memory model, limiter may choose larger windows, while there is actually no free memory for those. We should “round up” the read window to the next part boundary on our side. The metric `s3.client.read_window_excess_bytes` may be used to track the amount of “rounded up” memory, not accounted in the memory limiter.

To test the change, PR introduces `read_window_increment_failed` flag to `MockClient` and adds a `basic_read_test_mock_large` test.

Note that this fix doesn't apply to `CachingPartStream`: for this stream `BackpressureController` is not aware of `second_request_start` offset, which is required for the alignment.   

Patch version change and a change log to `mountpoint-s3-fs`, will add later.

---

### DIFF
--- a/mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
+++ b/mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
@@ -266,7 +266,7 @@ impl BackpressureController {
     ) -> u64 {
         if align_read_window && read_window_end > second_req_start {
             let relative_end_offset = read_window_end - second_req_start;
-            if relative_end_offset % part_size != 0 {
+            if !relative_end_offset.is_multiple_of(part_size) {
                 let aligned_relative_offset = part_size * (relative_end_offset / part_size + 1);
                 second_req_start + aligned_relative_offset
             } else {
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -741,7 +741,7 @@ impl ClientBackpressureHandle for MockBackpressureHandle {
         let prev_read_window_end_offset = self.read_window_end_offset.fetch_add(len as u64, Ordering::SeqCst);
         let read_window_end_offset = prev_read_window_end_offset + len as u64;
         let relative_read_window_end = read_window_end_offset - self.request_range.start;
-        if read_window_end_offset < self.request_range.end && (relative_read_window_end % self.part_size != 0) {
+        if read_window_end_offset < self.request_range.end && !relative_read_window_end.is_multiple_of(self.part_size) {
             tracing::warn!(
                 relative_read_window_end,
                 self.part_size,


mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -136,8 +145,17 @@ impl BackpressureController {
                 {
                     let new_read_window_end_offset = self
                         .next_read_offset
-                        .saturating_add(self.preferred_read_window_size as u64)
-                        .min(self.request_end_offset);
+                        .saturating_add(self.preferred_read_window_size as u64);
+                    let new_read_window_end_offset = Self::round_up_to_part_boundary(
+                        new_read_window_end_offset,
+                        self.second_request_start,
+                        self.min_read_window_size as u64,
```
passaro, line 152: We are assuming this is actually the read part size, right? We should make it explicit.

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -136,8 +145,17 @@ impl BackpressureController {
...
+                        self.align_read_window,
+                    );
+                    // NOTE: in the end of the object we still have up to "part_size" of unaccounted memory.
+                    // For more accurate memory limiting we could reserve in "full parts", but that would make
+                    // release more complicated. So accept this inaccuracy, and reserve only till `request_end_offset`.
+                    let new_read_window_end_offset = new_read_window_end_offset.min(self.request_end_offset);
```
passaro, line 158: Try to avoid shadowing here.

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -136,8 +145,17 @@ impl BackpressureController {
                 {
                     let new_read_window_end_offset = self
                         .next_read_offset
-                        .saturating_add(self.preferred_read_window_size as u64)
-                        .min(self.request_end_offset);
+                        .saturating_add(self.preferred_read_window_size as u64);
+                    let new_read_window_end_offset = Self::round_up_to_part_boundary(
```
passaro, line 149: Is this the right place? Or should we align in `scale_up` and `scale_down`, where we already modify `preferred_read_window_size`?

vladem: With the existing incrementing logic, I think, it is the right place. Above this line we "blindly" adding `preferred_read_window_size` to `new_read_window_end_offset`. Having a fixed `preferred_read_window_size` won't work for arbitrary `new_read_window_end_offset`, which is dictated by the reading side (e.g. we may need to increment the window at offset `4,194,305` or `4,194,306` depending on how much data was read).

vladem:
closing in favour of https://github.com/awslabs/mountpoint-s3/pull/1671

PR #1621, opened 2025-09-26
Add rpm specfile generator

tadiwa-aizen:
Added a new RPM spec file generator (`generate_spec.py`) that creates distribution-specific `.spec` files by extracting versions from `Cargo.toml`/`rust-toolchain.toml`, scanning git submodules for bundled library declarations.

Added  `generate_amzn2023.srpm.sh` to generate al2023 srpm via the spec file generator.

Also added the template file for Amazon Linux 2023 (`package/spec/templates/amzn2023.spec.template`) containing AL2023-specific BuildRequires, runtime dependencies, and build commands.

*Usage* 
`uv run --directory package/spec python generate_spec.py amzn2023`

**Please confirm there's no breaking change, or call out any that are made.**

No breaking change

**Does this change need a changelog entry? Does it require a version change?**

No

---


package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+Provides: bundled(aws-c-s3)
+Provides: bundled(aws-c-sdkutils)
+Provides: bundled(aws-lc)
+
+%global _description %{expand:
+Mountpoint for Amazon S3 is a high-performance file system for Amazon S3.
+It allows you to mount an S3 bucket as a local file system, providing
+high-throughput access to S3 objects through standard file system operations.
+
+This package includes bundled AWS Common Runtime (CRT) libraries and
```
muddyfish: Does building the srpm file always produce the exact same binary?

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+all Rust crate dependencies for reproducible builds.}
+
+%description %{_description}
+
+# This line extracts the main source tarball (Source0) into a directory named 
+# mountpoint-s3-%%{version} and applies any patches with -p1 strip level. The -a1
```
muddyfish: If you're explaining this, what does 'p1 strip level' actually mean?

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+cat > .cargo/config.toml <<'EOF'
+[source.crates-io]
+replace-with = "vendored-sources"
+
+[source.vendored-sources]
+directory = "vendor"
+EOF
+
+%build
+# FORTIFY_SOURCE throws warnings when optimization is missing, -Werror makes them fatal
```
muddyfish: Is comment still relevant? There's no `-Werror` anymore

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+# This suppresses those warnings and ensures CMake can detect pthread properly
+export CFLAGS="${CFLAGS:-%{optflags}} -O2 -Wno-error=cpp"
+export CMAKE_C_FLAGS="$CFLAGS"
+
+# This compiles the Rust project in optimized release mode (--release) while 
+# preventing any network access (--offline) and forbidding changes to the 
+# dependency lock file (--frozen), ensuring a reproducible build using only the 
+# pre-vendored dependencies.
+cargo build --release --frozen --offline
```
muddyfish: This is missing the normal rust flags we pass as part of the release

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+
+# Fedora packaging guidlines require a manifest that lists the names and versions of all
+# crates in the vendor tarball.  They also note that this file must be added as a %license file. Macros which do this automatically are not available on all platforms.
+grep -E '^(name|version) = ' Cargo.lock \
```
muddyfish: Why are we doing this if we're importing a pre-built license file?

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+/opt/aws/mountpoint-s3/NOTICE
+/opt/aws/mountpoint-s3/THIRD_PARTY_LICENSES
+/opt/aws/mountpoint-s3/VERSION
+/opt/aws/mountpoint-s3/bin
+/opt/aws/mountpoint-s3/bin/mount-s3
+/opt/aws/mountpoint-s3/cargo-vendor.txt
+/usr/bin/mount-s3
+/usr/sbin/mount.mount-s3
+
+%changelog
```
muddyfish: Is this changelog meant to be for this rpm or for Mountpoint itself?

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+/opt/aws/mountpoint-s3
+/opt/aws/mountpoint-s3/LICENSE
+/opt/aws/mountpoint-s3/NOTICE
+/opt/aws/mountpoint-s3/THIRD_PARTY_LICENSES
+/opt/aws/mountpoint-s3/VERSION
+/opt/aws/mountpoint-s3/bin
+/opt/aws/mountpoint-s3/bin/mount-s3
+/opt/aws/mountpoint-s3/cargo-vendor.txt
+/usr/bin/mount-s3
+/usr/sbin/mount.mount-s3
```
muddyfish: This file listing doesn't match what you state in the PR overview

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+# We reuse the same optimized flags as in the  %build section
+export CFLAGS="%{optflags} -O2 -Wno-error=cpp"
+export CMAKE_C_FLAGS="$CFLAGS"
+
+# ulimit -n 4096 increases the maximum number of open file descriptors to 4096 
+# because Mountpoint-S3's tests open many files simultaneously and would fail with
+# the default lower limit.
+ulimit -n 4096
+cargo test --release --frozen --offline -- \
+    --skip mnt::test::mount_unmount \
```
muddyfish: Why skip these tests?

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+BuildRequires:  glibc-headers
+BuildRequires:  glibc-static
+BuildRequires:  libstdc++-devel
+BuildRequires:  nasm
+BuildRequires:  make
+BuildRequires:  which
+
+Requires:       fuse
+
+# BUNDLED C/C++ LIBRARIES - Required virtual provides for security tracking
```
yerzhan7: Could you please add instructions to always add new CRT libraries to this file?

Maybe as a comment into https://github.com/awslabs/mountpoint-s3/blob/7854f020d9f31efaf0e392367cf844f059c30b47/mountpoint-s3-crt-sys/UPDATING_CRT.md or https://github.com/awslabs/mountpoint-s3/blob/7854f020d9f31efaf0e392367cf844f059c30b47/mountpoint-s3-crt-sys/build.rs#L14-L25 so that we won't forget.

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+Source1:        mountpoint-s3-%{version}-vendor.tar.gz
+Source2:        LICENSE
+Source3:        NOTICE
+Source4:        THIRD_PARTY_LICENSES
+
+# BuildRequires lists dependencies that must be installed on the build system before the package can be compiled.
+BuildRequires:  clang
+BuildRequires:  clang-devel
+BuildRequires:  llvm
+BuildRequires:  rust
```
yerzhan7: Which version of rust is this? Can we specify exact version to match rust version in https://github.com/awslabs/mountpoint-s3/blob/7854f020d9f31efaf0e392367cf844f059c30b47/rust-toolchain.toml#L2?

package/amazon-linux-packaging.spec
yerzhan7: Do we need to talk about CRT and Rust libraries in the package description? Sounds like just an implementation detail (which could change in the future).

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+BuildRequires:  fuse-devel
+BuildRequires:  glibc-devel
+BuildRequires:  glibc-headers
+BuildRequires:  glibc-static
+BuildRequires:  libstdc++-devel
+BuildRequires:  nasm
+BuildRequires:  make
+BuildRequires:  which
+
+Requires:       fuse
```
yerzhan7: Can we specify that we need fuse v2? E.g. `>= 2.9.0`?

package/amazon-linux-packaging.spec
yerzhan7: Did we test/validate that **fstab** feature works with this new Mountpoint AL2023 RPM? As I understand there are no automatic tests for fstab for this new RPM

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
+%bcond_without check
+
+Name:           mount-s3
+Version:        1.20.0
+Release:        1.amzn2023
```
yerzhan7: In the future there might be new Amazon Linux version and we might need to release RPM versions for both AL2023 and the new Amazon Linux. So, would we re-use the same spec file for both Amazon linux versions or will we create 2 separate spec files? 

I am just thinking that maybe we should rename this file to `package/amazon-linux-2023-packaging.spec` if we hardcode AL2023 on this line?

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+Summary:        High-performance file system for Amazon S3
+
```
yerzhan7: Should we also add `BuildArchitectures: x86_64 aarch64`?

package/amazon-linux-packaging.spec
yerzhan7: I think you can do `1.%{dist}` to that automatically expands to `1.amz2023` via macro?

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
...
+Provides: bundled(aws-c-auth)
+Provides: bundled(aws-c-cal)
+Provides: bundled(aws-c-common)
+Provides: bundled(aws-c-compression)
+Provides: bundled(aws-checksums)
+Provides: bundled(aws-c-http)
+Provides: bundled(aws-c-io)
+Provides: bundled(aws-c-s3)
+Provides: bundled(aws-c-sdkutils)
+Provides: bundled(aws-lc)
```
yerzhan7: I think we should specify exact version that we are bundling so that it's easier to track vulnerabilities.

E.g. `Provides: bundled(aws-lc) = 1.61.3`

package/amazon-linux-packaging.spec
```diff
@@ -0,0 +1,149 @@
+%bcond_without check
+
+Name:           mount-s3
+Version:        1.20.0
+Release:        1.amzn2023
+Summary:        High-performance file system for Amazon S3
```
yerzhan7: Should we make it consistent with our https://github.com/awslabs/mountpoint-s3/blob/main/package/mount-s3.spec#L4?

Same for `description`

package/amazon-linux-packaging.spec
tadiwa-aizen: Noted, I'll leave the 
BuildRequires:  rust >= 1.88
BuildRequires:  cargo >= 1.88

yerzhan7: we should match exact version as in our other builds? (if we want same/reproducable builds)

Let's do ` rust = 1.88` instead?

package/amazon-linux-packaging.spec
tadiwa-aizen: Noted, removed, it

package/amazon-linux-packaging.spec
tadiwa-aizen: So the manifest is different from the THIRD_PARTY_LICENSES, which is essentially legal stuff and not much detail on dependancies.

The manifest allows for security scanners to see everything we've bundled. It's like the provides section above but for the entire projects dependancies, and not just the CRT.

As seen here: https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/#_vendored_dependencies

package/amazon-linux-packaging.spec
tadiwa-aizen: K*j* doesnt allow you to mount in the builds run.

package/amazon-linux-packaging.spec
tadiwa-aizen: Yes, fstab works fine, i've tested it. We could add automated tests the the current AMI test CR if you'd like?

package/amazon-linux-packaging.spec
tadiwa-aizen: This is the changelog for the desert linux package mainly. We already have one here. For this will mainly track each release into AL, thus i assume its best to track each release from 1.20.0 onwards.

yerzhan7: Can we just provide link to our changelog on GitHub here instead? Do we really need to duplicate changelog here?

package/amazon-linux-2023-packaging.spec
```diff
@@ -0,0 +1,121 @@
...
+BuildRequires:  nasm
+BuildRequires:  make
+BuildRequires:  which
+BuildRequires:  rust-packaging
+BuildRequires:  rust-toolset
+
+ExclusiveArch: x86_64 aarch64
+
+# BUNDLED C/C++ LIBRARIES - Required virtual provides for security tracking
+Provides: bundled(aws-c-auth) = 0.9.0
```
passaro: Do we need to hard-code the version numbers here? Also, they do not look correct to me.

package/amazon-linux-2023-packaging.spec
```diff
@@ -0,0 +1,121 @@
...
+%autosetup -n mountpoint-s3
+%cargo_prep -v vendor
+
+%build
+# This suppresses optimization warnings
+export CFLAGS="${CFLAGS:-%{optflags}} -O2 -Wno-error=cpp"
+export CMAKE_C_FLAGS="$CFLAGS"
+
+export RUSTFLAGS="${RUSTFLAGS} -C debuginfo=line-tables-only"
+cargo build --release --frozen --offline
```
passaro: What about the version number?

package/amazon-linux-2023-packaging.spec
```diff
@@ -0,0 +1,121 @@
...
+
+# We reuse the same optimized flags as in the  %build section
+export CFLAGS="%{optflags} -O2 -Wno-error=cpp"
+export CMAKE_C_FLAGS="$CFLAGS"
+
+# ulimit -n 4096 increases the maximum number of open file descriptors to 4096
+# because Mountpoint-S3's tests open many files simultaneously and would fail with
+# the default lower limit.
+ulimit -n 4096
+cargo test --release --frozen --offline -- \
```
passaro: What is the purpose of these tests? When are they run?

package/amazon-linux-2023-packaging.spec
```diff
@@ -0,0 +1,121 @@
+%bcond_without check
+
+Name:           mount-s3
+Version:        1.20.0
```
passaro: Does the version number need to be hard-coded?

passaro (changes requested):
Could you simplify the PR description? In general, it should contain a summary and most importantly the **reason** for the change. In this case, a lot of the additional information seems more appropriate for docs to be committed instead. For example, should the "how to test" section be its own separate doc? or script?

package/generate_spec.py
```diff
@@ -0,0 +1,157 @@
...
+
+cargo build --release
+%cargo_vendor_manifest
+
+%install
+mkdir -p %{buildroot}/opt/aws/mountpoint-s3/bin
+mkdir -p %{buildroot}/%{_prefix}/sbin
+mkdir -p %{buildroot}/%{_bindir}
+mkdir -p %{buildroot}/%{_bindir}
+cp target/release/mount-s3 %{buildroot}/opt/aws/mountpoint-s3/bin/mount-s3
```
yerzhan7: Should we use `install -m 755 ` instead of `cp`? Install allows us to set permissions.

package/generate_spec.py
```diff
@@ -0,0 +1,157 @@
...
+cp NOTICE %{buildroot}/opt/aws/mountpoint-s3/
+cp LICENSE %{buildroot}/opt/aws/mountpoint-s3/
+cp THIRD_PARTY_LICENSES %{buildroot}/opt/aws/mountpoint-s3/
+cp cargo-vendor.txt %{buildroot}/opt/aws/mountpoint-s3/
+echo "%{version}" > %{buildroot}/opt/aws/mountpoint-s3/VERSION
+ln -sf /opt/aws/mountpoint-s3/bin/mount-s3 %{buildroot}/%{_bindir}/mount-s3
+ln -sf /opt/aws/mountpoint-s3/bin/mount-s3 %{buildroot}/%{_prefix}/sbin/mount.mount-s3
+
+%files
```
yerzhan7: Also, I think it's best practice to add `%defattr(-,root,root,-)`. EFS Utils has this in their spec file.

package/generate_spec.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+cargo build --release
+%cargo_vendor_manifest
+
+%install
+mkdir -p %{buildroot}/opt/aws/mountpoint-s3/bin
+mkdir -p %{buildroot}/%{_prefix}/sbin
+mkdir -p %{buildroot}/%{_bindir}
+mkdir -p %{buildroot}/%{_bindir}
```
yerzhan7: This is duplicated

package/generate_spec.py
```diff
@@ -0,0 +1,157 @@
...
+echo "%{version}" > %{buildroot}/opt/aws/mountpoint-s3/VERSION
+ln -sf /opt/aws/mountpoint-s3/bin/mount-s3 %{buildroot}/%{_bindir}/mount-s3
+ln -sf /opt/aws/mountpoint-s3/bin/mount-s3 %{buildroot}/%{_prefix}/sbin/mount.mount-s3
+
+%files
+%dir /opt/aws/mountpoint-s3
+%dir /opt/aws/mountpoint-s3/bin
+/opt/aws/mountpoint-s3/bin/mount-s3
+%doc /opt/aws/mountpoint-s3/NOTICE
+%license /opt/aws/mountpoint-s3/LICENSE
```
yerzhan7: Should be `%license %attr(644,root,root) /opt/aws/mountpoint-s3/LICENSE`?

package/generate_spec.py
```diff
@@ -0,0 +1,157 @@
...
+cp NOTICE %{buildroot}/opt/aws/mountpoint-s3/
+cp LICENSE %{buildroot}/opt/aws/mountpoint-s3/
+cp THIRD_PARTY_LICENSES %{buildroot}/opt/aws/mountpoint-s3/
+cp cargo-vendor.txt %{buildroot}/opt/aws/mountpoint-s3/
+echo "%{version}" > %{buildroot}/opt/aws/mountpoint-s3/VERSION
+ln -sf /opt/aws/mountpoint-s3/bin/mount-s3 %{buildroot}/%{_bindir}/mount-s3
+ln -sf /opt/aws/mountpoint-s3/bin/mount-s3 %{buildroot}/%{_prefix}/sbin/mount.mount-s3
+
+%files
+%dir /opt/aws/mountpoint-s3
```
yerzhan7: Should we do `%dir %attr(755,root,root) /opt/aws/mountpoint-s3`?

package/generate_spec.py
```diff
@@ -0,0 +1,157 @@
...
+%dir /opt/aws/mountpoint-s3/bin
+/opt/aws/mountpoint-s3/bin/mount-s3
+%doc /opt/aws/mountpoint-s3/NOTICE
+%license /opt/aws/mountpoint-s3/LICENSE
+%license /opt/aws/mountpoint-s3/cargo-vendor.txt
+/opt/aws/mountpoint-s3/THIRD_PARTY_LICENSES
+/opt/aws/mountpoint-s3/VERSION
+%{_bindir}/mount-s3
```
yerzhan7: Should be `%attr(755,root,root) %{_bindir}/mount-s3` ?

package/generate_spec.py
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env python3
```
yerzhan7: Should we commit the generated spec file to this git repo? Otherwise, it's hard for me as a reviewer to review the final spec file.

yerzhan7: Could you make this file as executable?

```
chmod +x package/generate_spec.py
```

tadiwa-aizen: Sure, i'll add it in the next revision

package/spec/templates/amzn2023.spec.template
```diff
@@ -0,0 +1,49 @@
+{# Amazon Linux 2023 Template - Distribution-specific sections #}
+
```
tadiwa-aizen, line 2: This is the spec file created:

```sh
%bcond_without check

Name:           mount-s3
Version:        1.21.0
Release:        amzn2023
Summary:        Mountpoint for Amazon S3

License:        Apache-2.0
URL:            https://github.com/awslabs/mountpoint-s3 
Source0:        mountpoint-s3-1.21.0.tar.gz
Source1:        LICENSE
Source2:        NOTICE
Source3:        THIRD_PARTY_LICENSES

BuildRequires:  clang
BuildRequires:  clang-devel
BuildRequires:  rust >= 1.88      
BuildRequires:  cargo >= 1.88    
BuildRequires:  cmake
BuildRequires:  gcc
BuildRequires:  gcc-c++
BuildRequires:  git
BuildRequires:  pkgconfig
BuildRequires:  fuse-devel
BuildRequires:  glibc-devel
BuildRequires:  glibc-headers
BuildRequires:  glibc-static
BuildRequires:  libstdc++-devel
BuildRequires:  nasm
BuildRequires:  make
BuildRequires:  which
BuildRequires:  rust-packaging
BuildRequires:  rust-toolset

ExclusiveArch: x86_64 aarch64

Provides: bundled(aws-c-auth) = 0.9.0
Provides: bundled(aws-c-cal) = 0.9.2
Provides: bundled(aws-c-common) = 0.12.4
Provides: bundled(aws-c-compression) = 0.3.1
Provides: bundled(aws-c-http) = 0.10.3
Provides: bundled(aws-c-io) = 0.21.1
Provides: bundled(aws-c-s3) = 0.8.5
Provides: bundled(aws-c-sdkutils) = 0.2.4
Provides: bundled(aws-checksums) = 0.2.6
Provides: bundled(aws-lc) = 1.53.1
Provides: bundled(s2n-tls) = 1.5.18

Requires:       ca-certificates
Requires:       fuse >= 2.9.0
Requires:       fuse-libs >= 2.9.0

%description
Mountpoint for Amazon S3 is a simple, high-throughput file client for
mounting an Amazon S3 bucket as a local file system. With Mountpoint for Amazon
S3, your applications can access objects stored in Amazon S3 through file
operations like open and read. Mountpoint for Amazon S3 automatically
translates these operations into S3 object API calls, giving your applications
access to the elastic storage and throughput of Amazon S3 through a file
interface.

%build
export CFLAGS="${CFLAGS:-%{optflags}} -O2 -Wno-error=cpp"
export CMAKE_C_FLAGS="$CFLAGS"

export MOUNTPOINT_S3_AWS_RELEASE="amzn2023"

cargo build --release

%cargo_vendor_manifest

%prep
%autosetup -n mountpoint-s3
%cargo_prep -v vendor

%install
mkdir -p %{buildroot}/opt/aws/mountpoint-s3/bin
mkdir -p %{buildroot}/%{_prefix}/sbin
mkdir -p %{buildroot}/%{_bindir}
install -m 755 target/release/mount-s3 %{buildroot}/opt/aws/mountpoint-s3/bin/mount-s3
install -m 644 NOTICE %{buildroot}/opt/aws/mountpoint-s3/
install -m 644 LICENSE %{buildroot}/opt/aws/mountpoint-s3/
install -m 644 THIRD_PARTY_LICENSES %{buildroot}/opt/aws/mountpoint-s3/
install -m 644 cargo-vendor.txt %{buildroot}/opt/aws/mountpoint-s3/
echo "1.21.0" > %{buildroot}/opt/aws/mountpoint-s3/VERSION
ln -sf /opt/aws/mountpoint-s3/bin/mount-s3 %{buildroot}/%{_bindir}/mount-s3
ln -sf /opt/aws/mountpoint-s3/bin/mount-s3 %{buildroot}/%{_prefix}/sbin/mount.mount-s3

%files
%defattr(-,root,root,-)
%dir %attr(755,root,root) /opt/aws/mountpoint-s3
%dir %attr(755,root,root) /opt/aws/mountpoint-s3/bin
%attr(755,root,root) /opt/aws/mountpoint-s3/bin/mount-s3
%doc %attr(644,root,root) /opt/aws/mountpoint-s3/NOTICE
%license %attr(644,root,root) /opt/aws/mountpoint-s3/LICENSE
%license %attr(644,root,root) /opt/aws/mountpoint-s3/cargo-vendor.txt
%attr(644,root,root) /opt/aws/mountpoint-s3/THIRD_PARTY_LICENSES
%attr(644,root,root) /opt/aws/mountpoint-s3/VERSION
%attr(755,root,root) %{_bindir}/mount-s3
%attr(755,root,root) %{_prefix}/sbin/mount.mount-s3

%changelog
* Thu Oct 09 2025 Mountpoint-S3 Team <s3-opensource@amazon.com> - 1.21.0+amzn2023
- Mountpoint-S3 1.21.0 amzn2023 release
- Refer to https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/CHANGELOG.md
```

yerzhan7: I still see `rust >= 1.88`.  I thought we agreed to change it to exact version, no? `rust = 1.88`

yerzhan7: Also, `Source0:        mountpoint-s3-%{version}.tar.gz` still didn't to resolve to exact version? `mountpoint-s3-1.21.0.tar.gz`

package/generate_spec.py
```diff
@@ -0,0 +1,161 @@
...
+loads the corresponding template, and outputs a complete RPM spec file ready for rpmbuild.
+"""
+
+import subprocess
+import os
+import sys
+from datetime import datetime
+from jinja2 import Environment, FileSystemLoader
+
+script_dir = os.path.dirname(__file__)
```
muddyfish: Consider using `pathlib.Path` here - this makes path manipulation more readable

package/generate_spec.py
```diff
@@ -0,0 +1,161 @@
...
+project_root = os.path.dirname(script_dir)
+templates_dir = os.path.join(script_dir, "templates")
+
+
+def get_version():
+    cargo_path = os.path.join(project_root, "mountpoint-s3", "Cargo.toml")
```
muddyfish: Use https://docs.python.org/3/library/tomllib.html

package/generate_spec.py
```diff
@@ -0,0 +1,161 @@
...
+        if line.strip():
+            parts = line.strip().split(' ', 1)
+            if len(parts) == 2:
+                name, version = parts
+                versions[name] = version.lstrip('v')
+    return versions
+
+
+def generate_bundled_provides(submodule_versions):
+    provides = []
```
muddyfish: Nb: does a list comprehension look better here?

package/generate_spec.py
```diff
@@ -0,0 +1,161 @@
...
+    for lib_name, lib_version in submodule_versions.items():
+        provides.append(f"Provides: bundled({lib_name}) = {lib_version}")
+    return "\n".join(provides)
+
+
+def main():
+    if len(sys.argv) != 2:
```
muddyfish: Consider using `argparse`

package/generate_spec.py
```diff
@@ -0,0 +1,161 @@
...
+        'submodule_versions': submodule_versions,
+    }
+
+    # Setup Jinja2 env
+    env = Environment(loader=FileSystemLoader(templates_dir), trim_blocks=True, lstrip_blocks=True)
+
+    template = env.get_template(template_file)
+    template_content = template.render(**template_data)
+
+    spec_content = f"""%bcond_without check
```
muddyfish: I think this should ideally be it's own file

package/generate_spec.py
```diff
@@ -0,0 +1,161 @@
...
+
+Name:           mount-s3
+Version:        {version}
```
muddyfish: Why are we mixing f-strings and jinja templating?

package/spec/templates/amzn2023.spec.template
yerzhan7: > I still see rust >= 1.88. I thought we agreed to change it to exact version, no? rust = 1.88

Discussed offline.

package/generate_spec.py
tadiwa-aizen: Cool, ill try something like this
```py
def generate_bundled_provides(submodule_versions):
    return "\n".join([f"Provides: bundled({lib_name}) = {lib_version}" 
                      for lib_name, lib_version in submodule_versions.items()])
```

package/generate_spec.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python3
+
+"""
+This RPM spec generator creates distribution-specific .spec files using template inheritance.
```
muddyfish: Is this accurate?

package/generate_spec.py
```diff
@@ -0,0 +1,100 @@
...
+        return data["toolchain"]["channel"]
+
+
+def get_submodule_versions():
+    result = subprocess.run(
+        'git submodule foreach -q \'echo $name `git describe --tags`\'', capture_output=True, text=True, shell=True
+    )
+    versions = {}
+    for line in result.stdout.strip().split('\n'):
+        if line.strip():
```
muddyfish: Does this `if` statement need to be here?

package/generate_spec.py
```diff
@@ -0,0 +1,100 @@
...
+            match line.strip().split(' ', 1):
+                case [name, version]:
+                    versions[name] = version.lstrip('v')
```
muddyfish: Nit: use `removeprefix`

package/generate_spec.py
```diff
@@ -0,0 +1,100 @@
...
+    return versions
+
+
+def generate_bundled_provides(submodule_versions):
+    return "\n".join(
+        [f"Provides: bundled({lib_name}) = {lib_version}" for lib_name, lib_version in submodule_versions.items()]
```
muddyfish: Using a generator comprehension suffices

package/generate_spec.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Generate RPM spec files for different distributions")
+    parser.add_argument("build_target", help="Target distribution (e.g., amzn2023)")
+
+    args = parser.parse_args()
+    build_target = args.build_target
+    template_file = f"{build_target}.spec.template"
```
muddyfish: Should this be another argument?

package/generate_spec.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # Checking if template exists
+    template_path = templates_dir / template_file
+    if not template_path.exists():
+        print(f"Error: Template file {template_path} not found")
```
muddyfish: Raise an exception

package/spec/generate_spec.py
```diff
@@ -0,0 +1,100 @@
...
+    # Checking if template exists
+    template_path = templates_dir / template_file
+    if not template_path.exists():
+        print(f"Error: Template file {template_path} not found")
+        sys.exit(1)
+
+    version = get_version()
+    rust_version = get_rust_version()
+    submodule_versions = get_submodule_versions()
+    current_date = datetime.now().strftime("%a %b %d %Y")
```
muddyfish, line 66: We've traditionally used this `April 9, 2025` as our date format - why change it?

package/generate_spec.py
```diff
@@ -0,0 +1,100 @@
...
+        'current_date': current_date,
+        'submodule_versions': submodule_versions,
+    }
+
+    # Setting up template env
+    env = Environment(loader=FileSystemLoader(templates_dir), trim_blocks=True, lstrip_blocks=True)
+
+    template = env.get_template(template_file)
+
+    spec_content = template.render(**template_data)
```
muddyfish: Why use kwarg splatting here when you can just pass in arguments normally?

package/generate_spec.py
```diff
@@ -0,0 +1,100 @@
...
+
+    version = get_version()
+    rust_version = get_rust_version()
+    submodule_versions = get_submodule_versions()
+    current_date = datetime.now().strftime("%a %b %d %Y")
+
+    template_data = {
+        'version': version,
+        'rust_version': rust_version,
+        'build_target': build_target,
```
muddyfish: This is always constant for a given spec file. Why do we need to pass this in as an argument?

package/spec/generate_spec.py
tadiwa-aizen: Its RPM changelog format, they include the day, you can double check by running `dnf changelog x` .

package/generate_spec.py
```diff
@@ -0,0 +1,91 @@
...
+
+    args = parser.parse_args()
+    build_target = args.build_target
+
+    template_file = args.template or f'{build_target}.spec.template'
+
+    template_path = templates_dir / template_file
+
+    if not template_path.exists():
+        raise Exception(f"Template file {template_path} not found")
```
muddyfish: Use [`FileNotFoundError`](https://docs.python.org/3/library/exceptions.html#FileNotFoundError)

package/generate_spec.py
```diff
@@ -0,0 +1,91 @@
...
+    )
+    versions = {}
+    for line in result.stdout.strip().split('\n'):
+        match line.split(' ', 1):
+            case [name, version]:
+                versions[name] = version.removeprefix('v')
+    return versions
+
+
+def generate_bundled_provides(submodule_versions):
```
muddyfish: You're not actually using this function

package/templates/amzn2023.spec.template
```diff
@@ -0,0 +1,109 @@
...
+
+cargo build --release
+
+%cargo_vendor_manifest
+
+%prep
+%autosetup -n mountpoint-s3
+%cargo_prep -v vendor
+
+{% raw %}
```
muddyfish: Do we need these raw blocks for something?

package/generate_spec.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
```
muddyfish: We should remove this shebang as we're not planning on executing this directly, and instead through `uv`

package/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[project]
```
passaro: Why do we need this file? Will this impact the pre-existing python script?

package/pyproject.toml
```diff
@@ -0,0 +1,10 @@
...
+name = "mountpoint-s3-packaging"
+version = "0.1.0"
+description = "Packaging tools for Mountpoint-S3"
+readme = "README.md"
```
passaro: It points to the pre-existing readme. Is that intentional?

package/spec/generate_spec.py
```diff
@@ -0,0 +1,86 @@
...
+        match line.split(' ', 1):
+            case [name, version]:
+                versions[name] = version.removeprefix('v')
+    return versions
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Generate RPM spec files for different distributions")
+    parser.add_argument("build_target", help="Target distribution (e.g., amzn2023)")
+    parser.add_argument("--template", help="Custom template file (default: {build_target}.spec.template)")
```
passaro, line 50: I'd apply some YAGNI here and start with something simpler for the specific case we are targeting now.

package/.python-version
```diff
@@ -0,0 +1 @@
+3.12
```
passaro: Do we really need to be tied to a specific version?

package/templates/amzn2023.spec.template
```diff
@@ -0,0 +1,103 @@
...
+Summary:        Mountpoint for Amazon S3
+
+License:        Apache-2.0
+URL:            https://github.com/awslabs/mountpoint-s3 
+Source0:        mountpoint-s3-{{ version }}.tar.gz
+Source1:        LICENSE
+Source2:        NOTICE
+Source3:        THIRD_PARTY_LICENSES
+
+{# Build dependencies #}
```
passaro: What is this line? A template comment? Is it required?

passaro: Same for all the other similar ones. Either remove or use spec comments.

package/templates/amzn2023.spec.template
```diff
@@ -0,0 +1,103 @@
...
+translates these operations into S3 object API calls, giving your applications
+access to the elastic storage and throughput of Amazon S3 through a file
+interface.
+
+{# Build commands #}
+%build
+export CFLAGS="${CFLAGS:-%{optflags}} -O2 -Wno-error=cpp"
+export CMAKE_C_FLAGS="$CFLAGS"
+
+export MOUNTPOINT_S3_AWS_RELEASE="amzn2023"
```
passaro: This looks wrong. The environment variable should be `MOUNTPOINT_S3_AWS_RELEASE_TARGET`.

package/templates/amzn2023.spec.template
```diff
@@ -0,0 +1,103 @@
...
+Mountpoint for Amazon S3 is a simple, high-throughput file client for
+mounting an Amazon S3 bucket as a local file system. With Mountpoint for Amazon
+S3, your applications can access objects stored in Amazon S3 through file
+operations like open and read. Mountpoint for Amazon S3 automatically
+translates these operations into S3 object API calls, giving your applications
+access to the elastic storage and throughput of Amazon S3 through a file
+interface.
+
+{# Build commands #}
+%build
```
passaro: nit: could we move build after prep?

package/templates/amzn2023.spec.template
```diff
@@ -0,0 +1,103 @@
...
+%doc %attr(644,root,root) /opt/aws/mountpoint-s3/NOTICE
+%license %attr(644,root,root) /opt/aws/mountpoint-s3/LICENSE
+%license %attr(644,root,root) /opt/aws/mountpoint-s3/cargo-vendor.txt
+%attr(644,root,root) /opt/aws/mountpoint-s3/THIRD_PARTY_LICENSES
+%attr(644,root,root) /opt/aws/mountpoint-s3/VERSION
+%attr(755,root,root) %{_bindir}/mount-s3
+%attr(755,root,root) %{_prefix}/sbin/mount.mount-s3
+
+%changelog
+* {{ current_date }} Mountpoint-S3 Team <s3-opensource@amazon.com> - {{ version }}+amzn2023
```
passaro: This does not look like a useful changelog, i.e. there is no history of changes.
Do we need to include it in this form? Or rather just include the date of first release (without the version) and the link to the changelog on github?

passaro: Also please double-check the email address.

package/templates/amzn2023.spec.template
```diff
@@ -0,0 +1,103 @@
...
+interface.
+
+{# Build commands #}
+%build
+export CFLAGS="${CFLAGS:-%{optflags}} -O2 -Wno-error=cpp"
+export CMAKE_C_FLAGS="$CFLAGS"
+
+export MOUNTPOINT_S3_AWS_RELEASE="amzn2023"
+
+cargo build --release
```
passaro: Do we need the full build? I'd think we could just use `cargo build --release --bin mount-s3`.

passaro (changes requested):

package/templates/amzn2023.spec.template
dannycjones: This is not our e-mail address.

```suggestion
* {{ current_date }} Mountpoint-S3 Team <mountpoint-s3@amazon.com> - {{ version }}+amzn2023
```

package/pyproject.toml
dannycjones, line 1: Nice addition!

Update the [readme](https://github.com/awslabs/mountpoint-s3/blob/main/package/README.md) to call out we should use `uv`?

dannycjones (commented):
Only a passing review, don't block on me.

Can you resolve comment threads that are now resolved? I think there's a few that are stale.

package/generate_amzn2023_srpm.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -euo pipefail
+
+# Setup RPM build environment
+rpmdev-setuptree
+
+# Generate spec file
+uv run --directory package/spec python generate_spec.py amzn2023 --output ~/rpmbuild/SPECS/amzn2023.spec
```
passaro, line 8: Not great that we hard code the output path, but it seems `rpmdev-setuptree` uses `~/rpmbuild` by default. Could it be configured? If so, I'd set it as a parameter for the script. If not, let's add a comment at the top of the file at least.

passaro: On second pass: It seems `rpmdev-setuptree` uses `~/rpmbuild` by default. Unless that is configurable, no point in changing it here.

I'd add a comment at the top of the file though.

package/generate_amzn2023_srpm.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
passaro, line 1: Can you make this script executable?

package/pyproject.toml
lucasqml: **Checked against the current code — this does not hold.**

WHAT I COULD NOT REFUTE: the ordering is exactly as described (`gmail_ingest_policy.py:213-216` scans, `:228-229` size-gates), and the first reviewer's pathological timings reproduce on this checkout (4 MiB body + 21 MiB attachment = 803 ms; 6 MiB plain = 1176 ms; 6 MiB HTML-only = 2130 ms, all returning `too_large`). Everything below is what the first reviewer did not measure. 1. WRONG INPUT SHAPE. `size_estimate > 8 MiB` in real mail means ATTACHMENTS, not body — the reviewer's own reachability argument says so ("size_estimate counts attachment bytes that the format=full response never carries"), then benchmarks a 4 MiB *text body* anyway. Measured on the shape that actually produces `too_large` (25 MiB message, PDF attachment, ordinary body): 2 KiB body = 0.354 ms, 20 KiB = 3.254 ms, 200 KiB = 32.777 ms. The loop this runs in takes `await self._sleep(self._pace_seconds)` after EVERY message — `ingest_gmail_messages.py:556`, `_DEFAULT_PACE_SECONDS = 0.25` (`:71`), an unconditional deliberate 250 ms tick. The proposed saving is 0.14%–13% of one pace tick the code already chooses to spend. Also `decide` returns in 0.001 ms on a SPAM-labelled 4 MiB body: the label group at …

<sub>Verified by re-reading the file (line numbers in the finding are stale), tracing the call path, and — where library behaviour was load-bearing — running it against the installed library rather than reasoning from memory. A second reader then tried to refute the first verdict; this one did not survive. Resolving.</sub>

lucasqml: **Checked against the current code — this does not hold.**

WHAT I COULD NOT REFUTE: the ordering is exactly as described (`gmail_ingest_policy.py:213-216` scans, `:228-229` size-gates), and the first reviewer's pathological timings reproduce on this checkout (4 MiB body + 21 MiB attachment = 803 ms; 6 MiB plain = 1176 ms; 6 MiB HTML-only = 2130 ms, all returning `too_large`). Everything below is what the first reviewer did not measure. 1. WRONG INPUT SHAPE. `size_estimate > 8 MiB` in real mail means ATTACHMENTS, not body — the reviewer's own reachability argument says so ("size_estimate counts attachment bytes that the format=full response never carries"), then benchmarks a 4 MiB *text body* anyway. Measured on the shape that actually produces `too_large` (25 MiB message, PDF attachment, ordinary body): 2 KiB body = 0.354 ms, 20 KiB = 3.254 ms, 200 KiB = 32.777 ms. The loop this runs in takes `await self._sleep(self._pace_seconds)` after EVERY message — `ingest_gmail_messages.py:556`, `_DEFAULT_PACE_SECONDS = 0.25` (`:71`), an unconditional deliberate 250 ms tick. The proposed saving is 0.14%–13% of one pace tick the code already chooses to spend. Also `decide` returns in 0.001 ms on a SPAM-labelled 4 MiB body: the label group at …

<sub>Verified by re-reading the file (line numbers in the finding are stale), tracing the call path, and — where library behaviour was load-bearing — running it against the installed library rather than reasoning from memory. A second reader then tried to refute the first verdict; this one did not survive. Resolving.</sub>

PR #1624, opened 2025-09-30
Use OTel's exponential histograms for histogram metrics.

sahityadg:
With explicit hitograms, OTel uses fixed bucket bounds, which don't work well for all Mountpoint metrics as they don't fall within the default range. So we need to manually configure the bucket bounds for different metrics with different boundaries. However, with exponential histograms, the bucket boundss are automatically scaled and provide more accurate metrics. However, this relies on OTel SDK's unstable feature. 

In case this isn't supported in the future, we need to switch to explicit Buckets with different bounds for different metrics.

No, changes are under a feature flag

No, changes are under a feature flag

---

### DIFF
--- a/mountpoint-s3-fs/src/metrics_otel.rs
+++ b/mountpoint-s3-fs/src/metrics_otel.rs
@@ -66,6 +66,22 @@ impl OtlpMetricsExporter {
                     .build(),
             )
             .with_resource(resource)
+            .with_view(|instrument: &opentelemetry_sdk::metrics::Instrument| {
+                if matches!(instrument.kind(), opentelemetry_sdk::metrics::InstrumentKind::Histogram) {
+                    Some(
+                        opentelemetry_sdk::metrics::Stream::builder()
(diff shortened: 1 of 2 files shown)


mountpoint-s3-fs/src/metrics_otel.rs
```diff
@@ -66,6 +66,22 @@ impl OtlpMetricsExporter {
                     .build(),
             )
             .with_resource(resource)
+            .with_view(|instrument: &opentelemetry_sdk::metrics::Instrument| {
```
passaro, line 69: nit: should we import these `opentelemetry_sdk::metrics::` types? Applies to existing code as well. Happy to fix later.

passaro (approved):
LGTM!

mountpoint-s3-fs/src/metrics_otel.rs
sahityadg: Agree, will fix it in the next change.

PR #1625, opened 2025-09-30
Use delta temporality as default for exporting OTLP metrics

sahityadg:
By default, we will use Delta temporality instead of Cumulative temporality to minimise the network payload size while exporting metrics. However, cutomers can switch to Cumulative temporality if their backends don't support Delta temporality

No, the changes are under a feature flag

No, the changes are under a feature flag

---

### DIFF
--- a/mountpoint-s3-fs/src/metrics_otel.rs
+++ b/mountpoint-s3-fs/src/metrics_otel.rs
@@ -1,6 +1,8 @@
-use opentelemetry::global;
-use opentelemetry_otlp::{Protocol, WithExportConfig};
-use opentelemetry_sdk::metrics::Temporality;
+use opentelemetry::{global, metrics};
+use opentelemetry_otlp::{MetricExporter, Protocol, WithExportConfig};
+use opentelemetry_sdk::metrics::{
+    Aggregation, Instrument, InstrumentKind, PeriodicReader, SdkMeterProvider, Stream, Temporality,
+};
 use std::time::Duration;
 
 /// Get temporality preference from environment variable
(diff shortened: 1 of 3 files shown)


mountpoint-s3-fs/src/metrics_otel.rs
```diff
@@ -1,7 +1,24 @@
...
+use opentelemetry_otlp::{MetricExporter, Protocol, WithExportConfig};
+use opentelemetry_sdk::metrics::{
+    Aggregation, Instrument, InstrumentKind, PeriodicReader, SdkMeterProvider, Stream, Temporality,
+};
 use std::time::Duration;
 
+/// Get temporality preference from environment variable
+/// By default, we will use delta.
+fn get_temporality_from_env() -> Temporality {
+    let prefer_delta = std::env::var("MOUNTPOINT_OTLP_METRICS_TEMPORALITY_PREFERENCE")
```
muddyfish: Why did we choose an environment variable over a CLI argument?

sahityadg: OTel SDK also supports [env](https://code.amazon.com/reviews/CR-91421674/revisions/1#/details) variable to configure the temporality in some SDKs. But from my quick search, they aren't implemented in the OTel Rust SDK yet. 

I think anyone wanting to switch to cumulative temporality should use the standard OTel env vars. So I'm inclined to making this an experimental env value for benchmarking purposes (which is why I added them. We could just leverage OTel SDK's env variables when they are implemented or decide to make it a cli arg.

PR #1626, opened 2025-10-01
Fix lifetime ellision warnings and other lints

dannycjones:
To be ready for newer Rust versions, fix some lints which will start failing builds if we upgrade.

No, this only addresses lints by removing ambiguous code or updating to use new methods that make code clearer.

No, no behavior changes.

---

### DIFF
--- a/mountpoint-s3-client/src/s3_crt_client/put_object.rs
+++ b/mountpoint-s3-client/src/s3_crt_client/put_object.rs
@@ -81,6 +81,7 @@ impl S3CrtClient {
                 span,
                 move |metrics| {
                     if metrics.request_type() == RequestType::CreateMultipartUpload && !metrics.error().is_err() {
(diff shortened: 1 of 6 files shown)


mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -81,6 +81,7 @@ impl S3CrtClient {
                 span,
                 move |metrics| {
                     if metrics.request_type() == RequestType::CreateMultipartUpload && !metrics.error().is_err() {
+                        // Send signal on a successful CreateMultipartUpload request
```
dannycjones, line 84: Added a comment here because I do not believe it is clearer to collapse these three conditions into one line. Linter won't warn if there's a comment.

muddyfish (approved):

dannycjones:
Merging now so we're ready to bump to Rust 1.89.0.

PR #1627, opened 2025-10-01
Mountpoint Build Tag

tadiwa-aizen:
Add support for platform-specific build tagging by using environment variable `MOUNTPOINT_S3_AWS_RELEASE_TARGET` that appends platform suffixes to version strings for specific platform releases.

**Example**
Given that MOUNTPOINT_S3_AWS_RELEASE_TARGET is "amzn2023":
  - Official build:` mount-s3 1.21.0`
  - Official build with target: `mount-s3 1.21.0+amzn2023`
  - Unofficial build: `mount-s3 1.21.0-unofficial+abc1234`

No

No

---

### DIFF
--- a/mountpoint-s3/src/build_info.rs
+++ b/mountpoint-s3/src/build_info.rs
@@ -32,7 +32,7 @@ pub const FULL_VERSION: &str = {
     }
 };
 
-/// Checks environment to see the official Mountpoint for Amazon S3 release target
+/// Checks the environment for the official Mountpoint for Amazon S3 release target.
 const fn official_aws_release_target() -> Option<&'static str> {
     option_env!("MOUNTPOINT_S3_AWS_RELEASE_TARGET")
 }


mountpoint-s3/src/build_info.rs
```diff
@@ -10,7 +10,13 @@ mod built {
...
 pub const FULL_VERSION: &str = {
     if is_official_aws_release() {
-        built::PKG_VERSION
+        if is_al2023_build() {
+            const_format::concatcp!(built::PKG_VERSION, "-al2023")
+        } else {
+            built::PKG_VERSION
+        }
+    } else if is_al2023_build() {
+        const_format::concatcp!(built::PKG_VERSION, "-unofficial", git_commit_suffix(), "-al2023")
```
yerzhan7: Can we simplify this like this?

```
    let mut version = built::PKG_VERSION;
    
    if is_al2023_build() {
        version = const_format::concatcp!(version, "-al2023");
    }

    if !is_official_aws_release() {
        version = const_format::concatcp!(version, "-unofficial", git_commit_suffix());
    }
    
    version
```

tadiwa-aizen: This wont work because we're in a  const function/block where everything must be determined at compile time. So the assigning of version wont compile.

mountpoint-s3/src/build_info.rs
```diff
@@ -9,18 +9,38 @@ mod built {
...
 };
-
-/// Checks environment to see if this build is for an official Mountpoint for Amazon S3 release.
 const fn is_official_aws_release() -> bool {
     option_env!("MOUNTPOINT_S3_AWS_RELEASE").is_some()
 }
 
+/// Gets build target from BUILD_TARGET environment variable if set
+const fn build_target_suffix() -> Option<&'static str> {
+    option_env!("BUILD_TARGET")
```
yerzhan7: np: How about `MOUNTPOINT_S3_DISTRIBUTION_METHOD` or `MOUNTPOINT_S3_PACKAGE_PLATFORM` to better distinguish between the generic existing builds (DEB/RPM/TAR.gz) and platform-specific packaging builds (AL2023 native packaging/distribution).

yerzhan7 (approved):

mountpoint-s3/src/build_info.rs
```diff
@@ -9,18 +9,39 @@ mod built {
 
 /// Valid SemVer version constructed using declared Cargo version and short commit hash if needed.
 pub const FULL_VERSION: &str = {
+    const BUILD_TARGET_STR: &str = match build_target_suffix() {
```
passaro: Why match again here? Couldn't the function just return a (sometime empty) string?

mountpoint-s3/src/build_info.rs
```diff
@@ -9,18 +9,39 @@ mod built {
...
+        Some(target) => target,
+        None => "",
+    };
+
     if is_official_aws_release() {
-        built::PKG_VERSION
+        if build_target_suffix().is_some() {
```
passaro: Why call the same function again?

mountpoint-s3/src/build_info.rs
```diff
@@ -9,18 +9,39 @@ mod built {
...
         const_format::concatcp!(built::PKG_VERSION, "-unofficial", git_commit_suffix())
     }
 };
-
-/// Checks environment to see if this build is for an official Mountpoint for Amazon S3 release.
 const fn is_official_aws_release() -> bool {
     option_env!("MOUNTPOINT_S3_AWS_RELEASE").is_some()
 }
 
+/// Gets build target from BUILD_TARGET environment variable if set
```
passaro: Outdated comment

mountpoint-s3/src/build_info.rs
```diff
@@ -9,18 +9,39 @@ mod built {
...
+const fn build_target_suffix() -> Option<&'static str> {
+    match option_env!("MOUNTPOINT_S3_AWS_RELEASE") {
```
passaro: Try to avoid duplicating the env variable name.

passaro: Also, are we keeping the original `MOUNTPOINT_S3_AWS_RELEASE`? Or should we rather use something like  `MOUNTPOINT_S3_AWS_RELEASE_TARGET`?

mountpoint-s3/src/build_info.rs
```diff
@@ -9,18 +9,39 @@ mod built {
...
+        Some(value) => {
+            if value.is_empty() {
+                None // No build target suffix
```
passaro: Both official build with no target and unofficial build return `None`. That seems very confusing - see also previous comments.

package/package.py
```diff
@@ -136,7 +136,7 @@ def build_mountpoint_binary(metadata: BuildMetadata, args: argparse.Namespace) -
     env["CARGO_TARGET_DIR"] = target_dir
     if args.official:
         # Remove the commit from the User-agent version number
-        env["MOUNTPOINT_S3_AWS_RELEASE"] = "true"
+        env["MOUNTPOINT_S3_AWS_RELEASE"] = ""
```
passaro: Also update the comment above.

passaro (changes requested):
Please update title and description.

mountpoint-s3/src/build_info.rs
```diff
@@ -7,18 +7,25 @@ mod built {
...
     } else {
-        const_format::concatcp!(built::PKG_VERSION, "-unofficial", git_commit_suffix())
+        const_format::concatcp!(built::PKG_VERSION, "-unofficial", git_commit_suffix()) // unofficial build
     }
 };
 
-/// Checks environment to see if this build is for an official Mountpoint for Amazon S3 release.
-const fn is_official_aws_release() -> bool {
-    option_env!("MOUNTPOINT_S3_AWS_RELEASE").is_some()
+const fn release_info() -> (bool, &'static str) {
```
muddyfish: This should return an `Option<&'static str>`

mountpoint-s3/src/build_info.rs
```diff
@@ -7,18 +7,41 @@ mod built {
...
+        // is an unofficial build
         const_format::concatcp!(built::PKG_VERSION, "-unofficial", git_commit_suffix())
     }
 };
 
-/// Checks environment to see if this build is for an official Mountpoint for Amazon S3 release.
-const fn is_official_aws_release() -> bool {
-    option_env!("MOUNTPOINT_S3_AWS_RELEASE").is_some()
+/// Returns release build tag information.
+/// Uses a tuple to work around Rust's const evaluation limitations regarding pattern matching in const scopes.
```
passaro: Why not adopt the same solution used in `git_commit_suffix`? Is there any meaningful difference here that warrants a second hacky approach in the same module?

passaro (changes requested):
Please consider using the approach discussed and implemented offline. Also, please fix the typos in the examples provided in the description.

mountpoint-s3/src/build_info.rs
```diff
@@ -7,18 +7,33 @@ mod built {
...
+        } else {
+            const_format::concatcp!(built::PKG_VERSION, "+", RELEASE_INFO_STR)
+        }
     }
 };
 
-/// Checks environment to see if this build is for an official Mountpoint for Amazon S3 release.
-const fn is_official_aws_release() -> bool {
-    option_env!("MOUNTPOINT_S3_AWS_RELEASE").is_some()
+const fn release_info() -> Option<&'static str> {
```
passaro: Rename (e.g. `official_aws_release_target`) and re-add docs.

mountpoint-s3/src/build_info.rs
```diff
@@ -7,18 +7,33 @@ mod built {
...
+    } else {
+        // A little hacky so we can pull out the target as a const
+        const RELEASE_INFO_STR: &str = match RELEASE_INFO {
+            Some(target) => target,
+            // Evaluated at compile time, but never used
+            None => "unreachable",
+        };
+        #[allow(
+            clippy::const_is_empty,
+            reason = "The RELEASE_INFO_STR can be empty or non-empty depending on the release target"
```
passaro: Mention that it depends on the environment at build time

PR #1631, opened 2025-10-03
Keep a constant memory reservation for backwards seek for each file handle

vladem:
Currently we reserve memory for backwards seek only when an actual seek occurs. The memory is used even if there is no such seek. Also we reserve too few memory, up to `1MiB`, while the whole extra buffer of size `part_size` may be kept in RAM.

With this change MP makes a memory reservation upon the creation of `PrefetchGetObject` and releases memory once it is dropped. This replaces the existing mechanism which reserves memory in `PartQueue::push_front`. 

Patch version change and a change log to `mountpoint-s3-fs`, will add later.

---

### DIFF
--- a/mountpoint-s3-fs/src/prefetch.rs
+++ b/mountpoint-s3-fs/src/prefetch.rs
@@ -42,7 +42,9 @@ use tracing::trace;
 use crate::checksums::{ChecksummedBytes, IntegrityError};
 use crate::data_cache::DataCache;
 use crate::fs::error_metadata::{ErrorMetadata, MOUNTPOINT_ERROR_CLIENT};
+use crate::mem_limiter::{BufferArea, MemoryLimiter};
 use crate::object::ObjectId;
+use crate::sync::Arc;
 
 mod backpressure_controller;
 mod builder;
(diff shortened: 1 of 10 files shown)


mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -474,13 +495,25 @@ where
         histogram!("prefetch.contiguous_read_len")
             .record((self.next_sequential_read_offset - self.sequential_read_start_offset) as f64);
     }
+
+    /// The amount of memory reserved for a backwards seek window.
+    ///
+    /// The seek window size is rounded up to the nearest multiple of part_size.
+    fn seek_window_reservation(part_size: usize, seek_window_size: usize) -> u64 {
```
muddyfish, line 505: Nit: `backwards_seek_reservation`?

mountpoint-s3-fs/src/prefetch/seek_window.rs
passaro, line 36: Couldn't we mark the memory as released when we drop the parts here, rather than after a read? Wouldn't that also solve the issue?

vladem: it won't address the main issue: `SeekWindow` logically only holds `1MiB` of data, while physically it will hold up to `8MiB` (part size) 

releasing here would work if `SeekWindow` was aware of the actual allocated memory

PR #1632, opened 2025-10-03
Update CI approvals to use single job with 'needs' relationship

dannycjones:
This change reduces noise from CI approvals by using one single job with its own approval, which all other jobs "need" before they start.

`needs` definition: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idneeds

Reference: https://github.com/awslabs/s3-connector-for-pytorch/pull/373/

CI change only. It now has all jobs requiring approval wait on a single job to succeed, which is itself gated by the GitHub environment.

No.

---

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -29,11 +29,19 @@ env:
   MAX_VALUES_IN_BENCH_CHART: 30
 
 jobs:
+  approval:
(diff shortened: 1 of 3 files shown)


passaro (approved):
Ship it!

PR #1642, opened 2025-10-08
Unstable flag to open a new FD on each fuse thread.

adpeace:
This mimics the libfuse behaviour.  This is currently behind a flag because the performance and compatibility concerns are still being considered.

No.

No.

---

### DIFF
--- a/mountpoint-s3-fs/src/fuse/session.rs
+++ b/mountpoint-s3-fs/src/fuse/session.rs
@@ -26,6 +26,14 @@ pub struct FuseSession {
 
 type OnClose = Box<dyn FnOnce() + Send>;
 
+struct SessionAndConfig<FS>
+where
+    FS: Filesystem + Send + Sync + 'static,
+{
+    session: Session<FS>,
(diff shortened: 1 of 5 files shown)


mountpoint-s3-fs/src/fuse/session.rs
```diff
@@ -100,7 +114,9 @@ impl FuseSession {
                 .context("failed to spawn waiter thread")?
         };
 
-        WorkerPool::start(session, workers_tx, max_worker_threads).context("failed to start worker thread pool")?;
+        let session_and_config = SessionAndConfig { session, clone_fuse_fd };
+        WorkerPool::start(session_and_config, workers_tx, max_worker_threads, clone_fuse_fd)
```
passaro: This should now be:
```suggestion
        WorkerPool::start(session_and_config, workers_tx, max_worker_threads)
```

adpeace: Ye, now fixed.

PR #1644, opened 2025-10-09
Fix instructions for updating CRT dependencies

passaro:
The command to fetch tags could fail in case of conflict and bail out before upgrading the remaining submodules.

No.

No.

---


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "aws-lc"]
 	path = mountpoint-s3-crt-sys/crt/aws-lc
-	url = https://github.com/awslabs/aws-lc.git
+	url = https://github.com/aws/aws-lc.git
```
passaro, line 3: Not really relevant, but `aws-lc` was moved some time ago (though `awslabs` would still redirect it).

PR #1647, opened 2025-10-10
Emit a MOUNTPOINT_EVENT_READY when error logging is enabled

vladem:
Emit an event which signals that mountpoint is ready.

No.

Yes, `mountpoint-s3-fs`.

---


mountpoint-s3-fs/src/logging/error_logger.rs
```diff
@@ -140,6 +154,23 @@ pub struct Event {
 }
 
 impl Event {
+    pub fn new(operation: &str, event_code: &str) -> Self {
```
passaro: [suggestion:] Could we move the `Event` initialization at line 77 into another constructor? And think about appropriate names for both?

Issue #1614, opened 2025-09-22
Warning and incorrect behavior on file listing with concurrent read loads


gaberudy:
### Mountpoint for Amazon S3 version

mount-s3 v1.20.0

### AWS Region

us-east-1

### Describe the running environment

AWS m6i.2xlarge running mount-s3, scheduled jobs from remote machines are connecting via a sshfs mount of the server that has access to the mount-s3 folder.

### Mountpoint options

```shell
/usr/bin/mount-s3 -f --uid 166535 --gid 166535 --write-part-size 16777216 --force-path-style --allow-other --allow-delete --allow-overwrite demo-prod-gh-bucket "/opt/ghdata/share/CloudStorage"
```

### What happened?

Using mount-s3 in a bioinformatic workflow with ~ 700 files and folders created and used in a directory.

I'm doing work in batches of 10 workers on a host. In the latest v1.20.0 release, I can reproduce a behavior where, in a queue of 100 workers, on the 8th or 9th batch, I get errors seeing the files in the work directory. A program reports that a file doesn't exist, but it does. Sometimes it's recovered by the 9th or 10th batch. I suspect it has something to do with the number of files being over a page size in the API or something.

Note that downgrading to 1.14.0 (last known good version).

I tested downgrading to 1.19.0 and v1.14.0 and was able to reproduce the issue, although the messages where different.

It seems like some cached or intermediate value for the directory listing is out of sync with the folder (as files are added). The problem "fixes" itself, but results in jailed tasks / jobs as they see a truncated or incomplete listing for a directory.

Any suggestions or work-arounds or coudl mount-s3 auto-resolve this state inconsistency internally?

### Relevant log output

```shell
Logs at around the time of the error (1.20 and 1.19)
---
[WARN] ThreadId(12) readdirplus{req=52422080 ino=2711 fh=5801 offset=113}: mountpoint_s3_fs::fuse: readdirplus failed with errno 9: inode error: out-of-order readdir, expected offset 118 but got 113 on dir handle 5801
2025-09-22T17:25:45.807592Z  WARN ThreadId(12) readdirplus{req=52422080 ino=2711 fh=5801 offset=113}: mountpoint_s3_fs::fuse: readdirplus failed with errno 9: inode error: out-of-order readdir, expected offset 118 but got 113 on dir handle 5801
[...]

PR #1648, opened 2025-10-10
Fix readdir race condition with concurrent deletions

Priyankakarumuru1:
Fix "File does not exist" errors in readdir operations ([#1614](https://github.com/awslabs/mountpoint-s3/pull/1614)) that could occur during concurrent directory listing.

Root Cause: update_from_remote fails when processing a local entry in ReaddirHandle::next() because the inode is removed from the parent’s writing_children. This was incorrectly treated as FileDoesNotExist.

Changes made:

- Altered inode update logic for readdir: when a ReaddirEntry is known to be local, the update is skipped.

- Updated directory listing to handle concurrent modifications gracefully.

- Ensures that files removed from writing_children are skipped without causing errors.

No breaking changes. 

Yes, this requires a changelog entry
No version change needed.

---

### DIFF
--- a/mountpoint-s3-fs/src/superblock/readdir.rs
+++ b/mountpoint-s3-fs/src/superblock/readdir.rs
@@ -651,9 +651,12 @@ mod tests {
     use crate::sync::Arc;
     use mountpoint_s3_client::mock_client::MockClient;
 
-    /// Verifies readdir handles gracefully skip deleted local inodes.
-    /// Creates a file, obtains a readdir handle, deletes the file, then uses the handle.
-    /// Should complete successfully when local inodes are deleted concurrently.
+    /// Verifies that `readdir` gracefully skips local inodes that are no longer tracked
+    /// in the parent’s `writing_children`.  
+    /// Creates a file, obtains a readdir handle, finishes writing the file (removing it
+    /// from `writing_children`), then uses the handle.  
+    /// The test passes if `readdir` completes successfully without panicking, even if the
+    /// entry is not returned.
     #[tokio::test]
     async fn test_readdir_race_condition() {
         let bucket = Bucket::new("test-bucket").unwrap();


mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -134,14 +134,12 @@ impl ReaddirHandle {
...
             if let Some(next) = next {
-                let Ok(name) = next.name().try_into() else {
-                    // Short-circuit the update if we know it'll fail because the name is invalid
-                    warn!("{} has an invalid name and will be unavailable", next.description());
-                    continue;
-                };
-                let remote_lookup = self.remote_lookup_from_entry(inner, &next);
-                let lookup = inner.update_from_remote(self.dir_ino, name, remote_lookup)?;
-                return Ok(Some(lookup));
+                let name: OsString = next.name().into();
```
passaro: Revert the conversion here

mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -638,3 +645,53 @@ impl DirHandle {
...
+            .new_readdir_handle(FUSE_ROOT_INODE)
+            .await
+            .expect("Failed to create readdir handle");
+
+        superblock
+            .finish_writing(lookup.ino(), None)
+            .await
+            .expect("Finish writing failed");
+
+        superblock
```
vladem, line 692: consider checking the expected result of the `readdir`: do we expect the file to be present in the response or we accept that response is incomplete in this racy case?

mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -638,3 +645,53 @@ impl DirHandle {
...
+mod tests {
+    use crate::fs::FUSE_ROOT_INODE;
+    use crate::metablock::{InodeKind, Metablock};
+    use crate::s3::S3Path;
+    use crate::superblock::Superblock;
+    use mountpoint_s3_client::mock_client::MockClient;
+    use std::sync::Arc;
+
+    /// Verifies readdir handles gracefully skip deleted local inodes.
+    /// Creates a file, obtains a readdir handle, deletes the file, then uses the handle.
```
vladem: consider clarifying this comment a bit: "deletes the file" is a bit misleading, since the test doesn't call `unlink`, probably you've meant that file is removed from `writing_children` of the parent as a result of `finish_writing`?

mountpoint-s3-fs/src/superblock/readdir.rs
Priyankakarumuru1: We don’t assert that the file is present because in this race condition scenario, the inode may have been removed from the parent’s writing_children. The test ensures that `readdir` completes successfully without panicking, regardless of whether the entry is returned.

mountpoint-s3-fs/src/superblock/readdir.rs
Priyankakarumuru1: Thanks for the suggestion! I’ve updated the comment to clarify.

mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -133,18 +133,20 @@ impl ReaddirHandle {
...
-                let remote_lookup = self.remote_lookup_from_entry(inner, &next);
-                let lookup = inner.update_from_remote(self.dir_ino, name, remote_lookup)?;
-                return Ok(Some(lookup));
-            } else {
+            let Some(next) = next else {
                 return Ok(None);
+            };
+
+            let Ok(name) = next.name().try_into() else {
+                //Short-circuit the update if we know it'll fail because the name is invalid
```
passaro: nit: spacing

mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -159,23 +161,24 @@ impl ReaddirHandle {
         self.parent_ino
     }
 
-    /// Create a [RemoteLookup] for the given ReaddirEntry if appropriate.
-    fn remote_lookup_from_entry<OC: ObjectClient + Send + Sync>(
+    /// Process a ReaddirEntry and return the final LookedUpInode result.
+    /// Returns None if the entry should be skipped.
```
passaro: It does not seem to ever return `None`.

mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -193,12 +196,16 @@ impl ReaddirHandle {
...
+                RemoteLookup {
                     stat,
                     kind: InodeKind::File,
-                })
+                }
             }
-        }
+        };
+
+        // Remote entry, use update_from_remote
```
passaro:
```suggestion
        // Remote entry, update the superblock.
```

mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -638,3 +645,56 @@ impl DirHandle {
...
+
+    /// Verifies that `readdir` gracefully skips local inodes that are no longer tracked
+    /// in the parent’s `writing_children`.  
+    /// Creates a file, obtains a readdir handle, finishes writing the file (removing it
+    /// from `writing_children`), then uses the handle.  
+    /// The test passes if `readdir` completes successfully without panicking, even if the
+    /// entry is not returned.
+    #[tokio::test]
+    async fn test_readdir_race_condition() {
+        let bucket = crate::s3::Bucket::new("test-bucket").unwrap();
```
passaro: nit: import `Bucket` (line 652).

mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -159,23 +161,23 @@ impl ReaddirHandle {
...
         inner: &SuperblockInner<OC>,
         entry: &ReaddirEntry,
-    ) -> Option<RemoteLookup> {
-        match entry {
-            // If we made it this far with a local inode, we know there's nothing on the remote with
-            // the same name, because [LocalInode] is last in the ordering and so otherwise would
-            // have been deduplicated by now.
-            ReaddirEntry::LocalInode { .. } => None,
+        name: ValidName,
+    ) -> Result<Option<LookedUpInode>, InodeError> {
```
passaro: Remove `Option`

mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -193,12 +195,16 @@ impl ReaddirHandle {
...
                     kind: InodeKind::File,
-                })
+                }
             }
-        }
+        };
+
+        // Remote entry, update the superblock.
+        let lookup = inner.update_from_remote(self.dir_ino, name, Some(remote_lookup))?;
+        Ok(Some(lookup))
```
passaro: Could be simplified. E.g.:
```suggestion
        inner.update_from_remote(self.dir_ino, name, Some(remote_lookup))
```

mountpoint-s3-fs/src/superblock/readdir.rs
```diff
@@ -133,18 +133,20 @@ impl ReaddirHandle {
...
+            let Ok(name) = next.name().try_into() else {
+                // Short-circuit the update if we know it'll fail because the name is invalid
+                warn!("{} has an invalid name and will be unavailable", next.description());
+                continue;
+            };
+
+            if let Some(result) = self.lookup_from_entry(inner, &next, name)? {
+                return Ok(Some(result));
             }
+            // Continue to next entry if None returned
```
passaro:
```suggestion
            let lookup = self.lookup_from_entry(inner, &next, name)?;
            return Ok(Some(lookup));
```

passaro (changes requested):

PR #1653, opened 2025-10-14
Update S3 and FUSE metrics for clarity 

sahityadg:
Rename S3 and FUSE metrics in preparation for OTLP export. Update FUSE idle threads from gauge to histogram for better visibility into threads utilisation, instead of point of time values. 

This change also adds integration tests for metrics verification, and updates documentation with new metric names.

Yes, it changes metrics names in logs

Yes, needs a changelog entry and version change for mountpoint-s3-fs crate.

---

### DIFF
--- a/mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
+++ b/mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
@@ -1,13 +1,14 @@
-use crate::common::fuse::{TestSessionConfig, mock_session, read_dir_to_entry_names};
+use crate::common::fuse::{TestSessionConfig, mock_session};
 use crate::common::test_recorder::{Metric, TestRecorder};
 use mountpoint_s3_fs::S3FilesystemConfig;
 use mountpoint_s3_fs::metrics::defs::{
     ATTR_FUSE_REQUEST, FUSE_IO_SIZE, FUSE_REQUEST_ERRORS, FUSE_REQUEST_LATENCY, PREFETCH_RESET_STATE,
 };
 use rusty_fork::rusty_fork_test;
-use std::fs::{File, read_dir};
+use std::fs::File;
 use std::io::{Read, Seek, Write};
-use std::sync::Arc;
+use std::thread::sleep;
+use std::time::Duration;
 
 fn setup_recorder() -> TestRecorder {
     let recorder = TestRecorder::default();
@@ -15,33 +16,40 @@ fn setup_recorder() -> TestRecorder {
     recorder
 }
 
-fn get_metric(recorder: &TestRecorder, name: &str, labels: &[(&str, &str)]) -> Arc<Metric> {
-    recorder
-        .get(name, labels)
-        .unwrap_or_else(|| panic!("Expected metric '{name}' with labels {labels:?} to exist"))
+macro_rules! get_metric {
+    ($recorder:expr, $name:expr, $labels:expr) => {{
+        match $recorder.get($name, $labels) {
+            Some(metric) => metric,
+            None => {
+                eprintln!("Available metrics:");
+                $recorder.print_metrics();
+                panic!("Expected metric '{}' with labels {:?} to exist", $name, $labels)
+            }
+        }
+    }};
 }
 
-fn get_histogram(recorder: &TestRecorder, name: &str, labels: &[(&str, &str)]) -> Vec<f64> {
-    let metric = get_metric(recorder, name, labels);
-    match metric.as_ref() {
-        Metric::Histogram(_) => metric.histogram(),
-        _ => panic!("Metric '{name}' is not a histogram"),
-    }
+macro_rules! get_histogram {
+    ($recorder:expr, $name:expr, $labels:expr) => {{
+        let metric = get_metric!($recorder, $name, $labels);
+        match metric.as_ref() {
+            Metric::Histogram(_) => metric.histogram(),
+            _ => panic!("Metric '{}' is not a histogram", $name),
+        }
+    }};
 }
 
 fn assert_metric_exists(recorder: &TestRecorder, name: &str, labels: &[(&str, &str)]) {
-    let _ = get_metric(recorder, name, labels);
+    let _ = get_metric!(recorder, name, labels);
 }
 
 fn verify_common_metrics(recorder: &TestRecorder) {
-    for request in ["open", "getattr", "fsync"] {
+    for request in ["open", "getattr"] {
         assert_metric_exists(recorder, FUSE_REQUEST_LATENCY, &[(ATTR_FUSE_REQUEST, request)]);
     }
     assert_metric_exists(recorder, "fs.inodes", &[]);
     assert_metric_exists(recorder, "fs.inode_kinds", &[("kind", "file")]);
     assert_metric_exists(recorder, "fs.inode_kinds", &[("kind", "directory")]);
-    assert_metric_exists(recorder, "fuse.op_unimplemented", &[("op", "getxattr")]);
-
     assert_metric_exists(recorder, "fuse.total_threads", &[]);
     assert_metric_exists(recorder, "fuse.idle_threads", &[]);
 }
@@ -80,35 +88,18 @@ rusty_fork_test! {
         file3.write_all(&content).unwrap();
         file3.sync_all().unwrap();
 
-        let write_handle = get_metric(&recorder, "fs.current_handles", &[("type", "write")]);
+        let write_handle = get_metric!(&recorder, "fs.current_handles", &[("type", "write")]);
         assert!(write_handle.gauge() >= 3.0, "should have at least 3 write handles");
 
         for f in [file, file2, file3] {
             drop(f);
         }
 
-        let write_io_size = get_histogram(&recorder, FUSE_IO_SIZE, &[(ATTR_FUSE_REQUEST, "write")]);
+        let write_io_size = get_histogram!(&recorder, FUSE_IO_SIZE, &[(ATTR_FUSE_REQUEST, "write")]);
         assert_eq!(write_io_size.len(), 3, "should have 3 write operations");
 
-        let write_latency = get_histogram(&recorder, FUSE_REQUEST_LATENCY, &[(ATTR_FUSE_REQUEST, "write")]);
+        let write_latency = get_histogram!(&recorder, FUSE_REQUEST_LATENCY, &[(ATTR_FUSE_REQUEST, "write")]);
         assert_eq!(write_latency.len(), 3, "should have 3 write operations");
-
-        // List files
-        let read_dir_iter = read_dir(test_session.mount_path()).unwrap();
-        let dir_entries = read_dir_to_entry_names(read_dir_iter);
-        assert_eq!(dir_entries, vec!["test.txt", "test2.txt", "test3.txt"]);
-
-        let read_dir_latency = get_histogram(
-            &recorder,
-            FUSE_REQUEST_LATENCY,
-            &[(ATTR_FUSE_REQUEST, "readdirplus")],
-        );
-        assert!(!read_dir_latency.is_empty(), "should have 1 readdirplus operation");
-        assert!(recorder
-                .get(FUSE_IO_SIZE, &[(ATTR_FUSE_REQUEST, "readdirplus")])
-                .is_none(),
-            "io size should not be recorded for non read/write operations"
-        );
 
         assert_metric_exists(&recorder, "fuse.total_bytes", &[("type", "write")]);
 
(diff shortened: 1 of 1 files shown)


mountpoint-s3-fs/src/metrics.rs
```diff
@@ -412,13 +413,14 @@ mod tests {
...
     use crate::metrics::data::Metric;
-    use crate::metrics::defs::{ATTR_HTTP_STATUS, ATTR_S3_REQUEST, S3_REQUEST_FAILURE};
+    use crate::metrics::defs::{ATTR_HTTP_STATUS, ATTR_S3_REQUEST, S3_REQUEST_ERRORS};
     use crate::metrics_otel::{OtlpConfig, OtlpMetricsExporter};
     use metrics::{Key, Unit};
     use opentelemetry::metrics::MeterProvider as _;
     use opentelemetry_sdk::metrics::data::{AggregatedMetrics, MetricData, ResourceMetrics};
     use opentelemetry_sdk::metrics::in_memory_exporter::InMemoryMetricExporter;
     use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
+    use std::sync::Arc;
```
muddyfish: Nit: No newline

mountpoint-s3-fs/tests/common/metrics.rs
```diff
@@ -0,0 +1,170 @@
...
+
+                let actual_labels: Vec<_> = key.labels().map(|l| (l.key(), l.value())).collect();
+
+                // Must have exact same number of labels
+                if actual_labels.len() != labels.len() {
+                    return false;
+                }
+
+                // Every expected label must be in the actual labels
+                labels.iter().all(|(k, v)| actual_labels.contains(&(*k, *v)))
```
muddyfish: Nit: `O(n^2)` here though shouldn't matter due to it being only a test

mountpoint-s3-fs/tests/common/metrics.rs
```diff
@@ -0,0 +1,170 @@
...
+        };
+        histogram.lock().unwrap().push(value);
+    }
+}
+
+/// Helper functions for test assertions
+pub fn get_metric(recorder: &TestRecorder, name: &str, labels: &[(&str, &str)]) -> Arc<Metric> {
+    recorder
+        .get(name, labels)
+        .unwrap_or_else(|| panic!("Expected metric '{name}' with labels {labels:?} to exist"))
```
muddyfish: Use `expect`

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -0,0 +1,268 @@
...
+    fn setup() -> TestContext {
+        let context = TestContext::new();
+        context.setup();
+        context
+    }
+
+    fn get_metric(recorder: &TestRecorder, name: &str, labels: &[(&str, &str)]) -> Arc<Metric> {
+        recorder
+            .get(name, labels)
+            .unwrap_or_else(|| panic!("Expected metric '{name}' with labels {labels:?} to exist"))
```
muddyfish: Use `expect`

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -0,0 +1,268 @@
...
+
+        let read_handle = get_metric(&context.recorder, "fs.current_handles", &[("type", "read")]);
+        assert!(read_handle.gauge() >= 1.0, "should have at least 1 read handle");
+
+        drop(read_file);
+
+        std::thread::sleep(Duration::from_millis(100));
+
+        let read_io_size = get_histogram(&context.recorder, FUSE_IO_SIZE, &[(ATTR_FUSE_REQUEST, "read")]);
+        assert!(read_io_size.contains(&1024.0));
```
muddyfish, line 144: Is this assertion really what we want to verify?

mountpoint-s3-client/src/metrics.rs
dannycjones, line 1: We already said that logged metrics aren't stable in docs, but I think it would be worth adding a note to the changelog to cover how we expect the metrics to change. (Any log line format change, metric name changes, etc..)

mountpoint-s3-fs/src/metrics.rs
```diff
@@ -98,7 +99,7 @@ fn poll_process_metrics(sys: &mut System) {
         if let Some(process) = sys.process(pid) {
             // update the metrics only when there is some change, otherwise it will be too spammy.
             if last_mem != process.memory() {
-                metrics::gauge!("process.memory_usage").set(process.memory() as f64);
+                metrics::gauge!(PROCESS_MEMORY_USAGE).set(process.memory() as f64);
                 metrics::gauge!("system.available_memory").set(sys.available_memory() as f64);
```
dannycjones, line 103: ~Why not also for system memory now?~

_Update: Discussed offline, trying to avoid a broad change of all metrics being updated at this point. This will be done in a later PR._

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -0,0 +1,268 @@
...
+    ATTR_FUSE_REQUEST, FUSE_IO_SIZE, FUSE_REQUEST_ERRORS, FUSE_REQUEST_LATENCY, PREFETCH_RESET_STATE,
+};
+use once_cell::sync::OnceCell;
+use std::fs::{File, read_dir};
+use std::io::{Read, Seek, Write};
+use std::sync::Arc;
+use std::sync::Once;
+use std::time::Duration;
+
+static RECORDER: OnceCell<TestRecorder> = OnceCell::new();
```
dannycjones: I think we can replace all of the logic for RECORDER with std's LazyCell, or I'm assuming it'll actually need std's LazyLock.

https://doc.rust-lang.org/std/sync/struct.LazyLock.html

This should avoid taking a dependency on once_cell crate.

mountpoint-s3-fs/src/metrics.rs
passaro: Also: `std::sync::Arc` should be removed.

mountpoint-s3-fs/tests/common/metrics.rs
```diff
@@ -0,0 +1,170 @@
+//! Test utilities for metrics validation
```
passaro: Is this a duplicate of `test_recorder.rs`?

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -0,0 +1,268 @@
...
+
+#[cfg(feature = "s3_tests")]
```
passaro: The tests seem to be using `mock_session`, why do we need S3?

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -0,0 +1,268 @@
...
+#[cfg_attr(test, serial_test::serial)]
+mod tests {
```
passaro: No need for the `tests` module here (in "integration" tests)

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -0,0 +1,268 @@
...
+        let filesystem_config = S3FilesystemConfig {
+            allow_overwrite: true,
+            ..Default::default()
+        };
+        let config = TestSessionConfig {
+            filesystem_config,
+            ..Default::default()
+        };
+
+        let test_session = mock_session::new("write_metrics", config);
```
passaro: The convention is to use the test name, so it would be "test_fuse_write_metrics" here. Same for the other tests.

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -0,0 +1,268 @@
...
+        file3.sync_all().unwrap();
+
+        let write_handle = get_metric(&context.recorder, "fs.current_handles", &[("type", "write")]);
+        assert!(write_handle.gauge() >= 3.0, "should have at least 3 write handles");
+
+        for f in [file, file2, file3] {
+            drop(f);
+        }
+
+        std::thread::sleep(Duration::from_millis(100));
```
passaro: Why is the sleep needed? We record the writes before the `write` requests return, right?

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        assert_metric_exists(&context.recorder, "fs.inodes", &[]);
+        assert_metric_exists(&context.recorder, "fs.inode_kinds", &[("kind", "file")]);
+        assert_metric_exists(&context.recorder, "fs.inode_kinds", &[("kind", "directory")]);
+        assert_metric_exists(&context.recorder, "fuse.op_unimplemented", &[("op", "getxattr")]);
+    }
+
+    #[test]
+    fn test_random_access_triggers_out_of_order() {
```
passaro: You mean "reset_state"? :)

passaro (changes requested):

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
sahityadg: We don't need them, I missed cleaning up as I switched to using mock sessions.

mountpoint-s3-fs/tests/common/metrics.rs
sahityadg: Yes, included it by mistake. Sorry!

mountpoint-s3-fs/tests/common/metrics.rs
sahityadg: Sorry, I included this file by mistake.

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
sahityadg: My understanding is also that these are recorded synchronously. But on CI, the tests were flaky without this, especially the common metric checks with gauges.

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
sahityadg: Yeah, we want to review the io size (bytes transferred) in a read request and there is at least one data point from the read.

mountpoint-s3-client/src/metrics.rs
sahityadg: yes, that is the plan.

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
sahityadg: Realised this was due to other tests in the same process updating metrics. So updated the tests to use rusty_fork.

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -0,0 +1,226 @@
...
+    assert_metric_exists(recorder, "fs.inodes", &[]);
+    assert_metric_exists(recorder, "fs.inode_kinds", &[("kind", "file")]);
+    assert_metric_exists(recorder, "fs.inode_kinds", &[("kind", "directory")]);
+    assert_metric_exists(recorder, "fuse.op_unimplemented", &[("op", "getxattr")]);
+
+    assert_metric_exists(recorder, "fuse.total_threads", &[]);
+    assert_metric_exists(recorder, "fuse.idle_threads", &[]);
+}
+
+rusty_fork::rusty_fork_test! {
```
passaro: nit: `rusty_fork_test` is already imported, so `rusty_fork::` should not be necessary

mountpoint-s3-fs/src/metrics/tracing_span.rs
```diff
@@ -1,5 +1,6 @@
 use std::time::Instant;
 
+use crate::metrics::defs::{ATTR_FUSE_REQUEST, FUSE_REQUEST_LATENCY};
```
passaro: tiny nit: prefer `super::defs:...` since it's a "sibling" module.

mountpoint-s3/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased (v1.21.0)
 
 * Change default logging level from WARN to INFO to improve visibility of important operational messages. ([#1605](https://github.com/awslabs/mountpoint-s3/pull/1605))
+* Change FUSE and S3 request metric names in logs (([#1630](https://github.com/awslabs/mountpoint-s3/pull/1630), [#1653](https://github.com/awslabs/mountpoint-s3/pull/1653))
```
passaro: Thanks. Could you add the same entry to the `fs` crate and increment its version?

mountpoint-s3-fs/CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
-## Unreleased
+## Unreleased (v0.8.2)
```
passaro, line 1: The version should be increased in `Cargo.toml` (as well as in the dependency in `../mountpoint-s3/Cargo.toml`)

mountpoint-s3-fs/CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
...
+* Change FUSE and S3 request metric names in logs (([#1630](https://github.com/awslabs/mountpoint-s3/pull/1630), [#1653](https://github.com/awslabs/mountpoint-s3/pull/1653))
```
passaro: nit: newline

passaro (dismissed):
LGTM!

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -0,0 +1,226 @@
...
+    let metric = get_metric(recorder, name, labels);
+    match metric.as_ref() {
+        Metric::Histogram(_) => metric.histogram(),
+        _ => panic!("Metric '{name}' is not a histogram"),
+    }
+}
+
+fn assert_metric_exists(recorder: &TestRecorder, name: &str, labels: &[(&str, &str)]) {
+    let _ = get_metric(recorder, name, labels);
+}
```
dannycjones: Not blocking but it would be good to update at some point. These should be macros in order to preserve the line numbers when an assertion failed. Otherwise, we only get the line numbers here (21 and 28) for each test failure.

As an example, here's what we do for permission assertions elsewhere: https://github.com/awslabs/mountpoint-s3/blob/5502a861ee11eaa6dc61aa8e711262b2d4388d8c/mountpoint-s3-fs/src/data_cache/cache_directory.rs#L170-L182

It would help for instance with the stack trace here: https://github.com/awslabs/mountpoint-s3/pull/1653#pullrequestreview-3356705006

passaro (approved):

PR #1656, opened 2025-10-15
Add integration test for OTel integration

sahityadg:
This change tests Mountpoint metrics integration with OTel collector

No, adds an integration test

No, adds an integration test

---

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -261,3 +261,42 @@ jobs:
     - name: Save dump files
       if: ${{ failure() }}
       run: ./.github/actions/scripts/save-coredump.sh
+
+  metrics:
+    name: Metrics Tests (${{ matrix.runner.name }})
+    runs-on: ${{ matrix.runner.tags }}
+
+    needs: approval
+
+    strategy:
+      fail-fast: false
+      matrix:
+        runner:
+        - name: Ubuntu x86
+          tags: [ubuntu-22.04]
+        - name: AL2 arm
+          tags: [self-hosted, linux, arm64]
+
+    steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v5
+        with:
+          role-to-assume: ${{ vars.ACTIONS_IAM_ROLE }}
+          aws-region: ${{ vars.S3_REGION }}
+      - name: Checkout code
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.ref }}
+          submodules: true
+          persist-credentials: false
+      - name: Set up Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          rustflags: ""
+      - name: Install operating system dependencies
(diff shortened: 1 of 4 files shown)


.github/workflows/integration.yml
```diff
@@ -261,3 +261,60 @@ jobs:
...
+          fi
+          VERSION=$(curl -s https://api.github.com/repos/open-telemetry/opentelemetry-collector-releases/releases/latest | jq -r '.tag_name' | sed 's/^v//')
+          echo "Downloading OTel collector version: $VERSION for $OS/$ARCH"
+          URL="https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v${VERSION}/otelcol_${VERSION}_${OS}_${ARCH}.tar.gz"
+          echo "URL: $URL"
+          curl -L "$URL" -o /tmp/otelcol.tar.gz
+          echo "Extracting..."
+          tar -xzf /tmp/otelcol.tar.gz -C /tmp
+          chmod +x /tmp/otelcol
+          /tmp/otelcol --version
```
muddyfish: Perhaps we should add `otelcol` to PATH so we can just assume it's 'installed' in the script rather than being in a specific place

.github/workflows/integration.yml
```diff
@@ -261,3 +261,60 @@ jobs:
...
+          persist-credentials: false
+      - name: Set up Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          rustflags: ""
+      - name: Install operating system dependencies
+        uses: ./.github/actions/install-dependencies
+        with:
+          fuseVersion: 2
+      - name: Install OpenTelemetry Collector
```
passaro: (non-blocking) consider making this a script, or an action as the one above.

.github/workflows/integration.yml
sahityadg: Is the concern that the script needs to know and pass the absolute path for otelcol binary? 

I can change. ~~But for testing purposes, isn't this fine?~~

mountpoint-s3/tests/metrics/otel_export.sh
```diff
@@ -0,0 +1,201 @@
...
+      features="--features otlp_integration"
+      ;;
+    "with_both")
+      args+=(--log-metrics --otlp-endpoint="${OTLP_ENDPOINT}" --otlp-export-interval=5)
+      features="--features otlp_integration"
+
+      ;;
+  esac
+  
+  cargo run --quiet --release ${features} -- "${args[@]}"
```
passaro: I think we should always build with `--features otlp_integration`. We want to test that `--log-metrics` without  `--otlp-endpoint` works correctly when the features is enabled, i.e. with the code that will eventually be released.

sahityadg: I added this mainly to test the current state without the feature. I was planning to make this change as part of removing the feature flag PR. But what you suggested may be better as that will give us the confidence to remove the feature flag

passaro (approved):

PR #1660, opened 2025-10-16
benchmark: Refactor resource monitoring tools into separate classes

adpeace:
Extract individual monitoring tools (mpstat, bwm-ng, perf-stat, flamegraph) from benchmark ResourceMonitoring class into separate tool classes that implement a common MonitoringTool interface. This improves benchmark code maintainability and makes adding new monitoring tools easier.

- Add benchmark/monitoring package with base MonitoringTool ABC
- Extract MpstatTool, BwmNgTool, PerfStatTool, FlamegraphTool classes
- Refactor ResourceMonitoring to manage list of tool instances
- Maintain backward compatibility with existing managed() API
- Add unit tests

No.

No.

---

### DIFF
--- a/benchmark/monitoring/tools.py
+++ b/benchmark/monitoring/tools.py
@@ -1,0 +1,217 @@
+import logging
+import os
+import signal
+import subprocess
+from typing import Optional
+
+import psutil
+
+from .base import MonitoringTool
+
+log = logging.getLogger(__name__)
+
+
+class MpstatTool(MonitoringTool):
+    def __init__(self):
+        self.process = None
+        self.output_file = None
+
+    def start(self) -> None:
+        self.output_file = open('mpstat.json', 'w')
+        log.info("Starting mpstat monitoring")
+        # fmt: off
+        self.process = subprocess.Popen([
+            "/usr/bin/mpstat",
+            "-P", "ALL",
+            "-o", "JSON",
+            "1"
+        ], stdout=self.output_file)
+        # fmt: on
+
+    def stop(self) -> None:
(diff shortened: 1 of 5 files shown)


benchmark/monitoring/tools.py
```diff
@@ -0,0 +1,217 @@
...
+        self.output_file = open('mpstat.json', 'w')
+        log.info("Starting mpstat monitoring")
+        # fmt: off
+        self.process = subprocess.Popen([
+            "/usr/bin/mpstat",
+            "-P", "ALL",
+            "-o", "JSON",
+            "1"
+        ], stdout=self.output_file)
+        # fmt: on
```
muddyfish, line 29: What do these bits do?

adpeace: The `fmt`? commands?

The code was moved from its original location in benchmark.py, where these were already there.  They stop ruff's formatting check from complaining about how the elements in the list are laid out.  The layout is intentional: the arguments to options go next to the option.

dannycjones: This PR is just moving code around so this is not at all a recommendation unless we go back to make changes again.

This might be neater for future. It applies to the previous statement rather than line.


```python
self.process = subprocess.Popen([
    "/usr/bin/mpstat",
    "-P", "ALL",
    "-o", "JSON",
    "1"
], stdout=self.output_file)  # fmt: skip
```

dannycjones (approved):

dannycjones (approved):
Thank you for fixing the mess I led us on a path to creating!

PR #1664, opened 2025-10-21
Add info logging around FUSE session join

dannycjones:
This change adds some new `INFO` logging around exits. In the past, we've seen tickets where Mountpoint "spontaneously" unmounts. It's not clear what's going on in those tickets and has not been possible to reproduce given no access to those systems. This change adds a little bit of extra logging to try and give better visibility into what Mountpoint thinks is happening.

No change to end-user behavior. Only new logs are added at `INFO` level, which is shown to customers.

No, just a simple logging addition.

---

### DIFF
--- a/mountpoint-s3-fs/src/fuse/session.rs
+++ b/mountpoint-s3-fs/src/fuse/session.rs
@@ -156,7 +156,7 @@ impl FuseSession {
             handler();
         }
 
-        trace!("unmounting filesystem");
+        info!("attempting unmount");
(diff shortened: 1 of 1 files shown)


mountpoint-s3-fs/src/fuse/session.rs
passaro: Should we also upgrade this? And log something like "Attempting unmount"?

passaro (approved):

PR #1671, opened 2025-10-23
[DRAFT] Align read window with part boundaries

vladem:
Memory limiter currently doesn’t account for the [“rounded up”](https://github.com/awslabs/mountpoint-s3/blob/d598453968b10b55c01371f0f594d586fcd8be43/mountpoint-s3-crt/src/s3/client.rs#L707) memory by CRT, which means up to 8MiB of unaccounted memory per file handle. Based on this flaw in the memory model, limiter may choose larger windows, while there is actually no free memory for those. We should “round up” the read window to the next part boundary on our side.

While working on https://github.com/awslabs/mountpoint-s3/pull/1618 we've faced the issue that `request_start_offset` is not always available to the code picking the new `read_window_end_offset`. Specifically when cache is used `BackpressureController` has no information about the S3 request start offset. 

This PR attempts to move the code picking `read_window_end_offset` close to the client and thus make `request_start_offset` available to it.

No.

No.

---

### DIFF
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -5,9 +5,8 @@
 use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::fmt::Write;
-use std::ops::Range;
 use std::pin::Pin;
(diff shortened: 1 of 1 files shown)


vladem:
closing in favour of https://github.com/awslabs/mountpoint-s3/pull/1707

PR #1695, opened 2025-11-03
Add explicit permissions to workflows

passaro:
Some of the workflows did not explicitly specify permissions and just relied on the default setting, triggering [code scanning alerts](https://github.com/awslabs/mountpoint-s3/security/code-scanning).

No.

No.

---

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -9,7 +9,6 @@ on:
     types: [ "checks_requested" ]
 
 permissions:
-  id-token: write
   contents: read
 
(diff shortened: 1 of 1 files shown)


.github/workflows/package.yml
```diff
@@ -8,6 +8,10 @@ on:
   merge_group:
     types: [ "checks_requested" ]
 
+permissions:
+  id-token: write
+  contents: read
```
dannycjones, line 12: Why does this workflow need `id-token: write`?

PR #1700, opened 2025-11-07
Use explicit result type in readdir reply callback

passaro:
The readdir reply callback in the `metablock` module was misleading and its documentation was wrong: `TryAddDirEntry` would return `false` on success (contrary to the rustdoc and common sense). Note though that all existing invocations were using it correctly.

This change introduces an explicit return type to avoid any confusion, at the cost of some increase in verbosity.

No, code tidy only.

No.

---

### DIFF
--- a/mountpoint-s3-fs/src/fs.rs
+++ b/mountpoint-s3-fs/src/fs.rs
@@ -18,9 +18,8 @@ use crate::async_util::Runtime;
 use crate::logging;
 use crate::mem_limiter::MemoryLimiter;
 use crate::memory::PagedPool;
-use crate::metablock::Metablock;
+use crate::metablock::{AddDirEntry, AddDirEntryResult, InodeInformation, Metablock};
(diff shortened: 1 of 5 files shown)


mountpoint-s3-fs/src/fs.rs
```diff
@@ -579,15 +578,19 @@ where
         is_readdirplus: bool,
         mut reply: R,
     ) -> Result<(), Error> {
-        let adder: TryAddDirEntry = Box::new(move |information, name, offset, generation| {
-            reply.add(DirectoryEntry {
+        let adder: AddDirEntry = Box::new(move |information, name, offset, generation| {
+            if reply.add(DirectoryEntry {
```
vladem, line 582: just to acknowledge: we could also change the signature of `DirectoryReplier::add`, but it is only used here and its documentation is OK, so keeping as is

PR #1702, opened 2025-11-11
Add 'workflow complete' GHA targets to codify CI requirements

dannycjones:
Mountpoint's GitHub repository today blocks PR approvals based on a list of jobs in the branch protection rules. This is not visible to contributors and also not possible to update unless you have additional privileges on the repository. When new workflows are added, it is frequently missed to update the branch protection rules.

This change instead defines targets within the workflow YAML files, such that it is clear what the requirements are. Changes in requirements can go through version control and code review.

I've only covered the tests and integration test workflows, as most other workflows have only one job. These workflows are also where most changes occur.

Following this commit, the GitHub branch protection rules should be updated to require only these new targets for these two workflows.

It changes CI only. There will be no actual change in behavior, just a change in how we maintain the rules.

No.

---

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Approval Gate
         run: echo "Approved!"
 
-  test:
+  fs-test:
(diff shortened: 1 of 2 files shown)


.github/workflows/integration.yml
dannycjones, line 1: We probably need a linter for whitespace... Another time.

PR #1704, opened 2025-11-11
Update Mountpoint's semantics to enable opening a new file handle on an inode if all the existing open handles have been flushed

mansi153:
Update Mountpoint's semantics to enable opening a new file handle on an inode if **all** the existing open handles have been flushed.

This allows for the filesystem to not depend on a `release` request to complete pending (delayed) upload and cleanup the internal state of the inode within Mountpoint; alleviating situations where an `open` request made immediately after a close(`flush`) request leads to a race condition between the `release` following the close and the new `open`, which is sometimes an issue when the upload to S3 has been delayed until `release`.

The semantics continue to allow **only one writer OR one/many readers** concurrently active for a file; however, now a new handle can be opened which can override the current active handles (despite not being officially released) if all the active handles are marked "flushed".
As part of opening the new handle, Mountpoint will also attempt uploading any pending data written for the previous file handle. All the requests to the overridden handle(s) will then start to fail or be no-op.
Multiple concurrent writers or concurrent readers and writers are still not allowed.

A handle is marked "flushed" when a close/`flush` is called on a file descriptor mapped to that handle. A following `read`/`write` request will revert that flushed state and signify that the handle is actively in use and can not be overridden. We maintain this information at the individual handle level and also in an inode-locked map of handles.

Breaking changes:
- Requests made to a duplicate file descriptor for a flushed file handle will start to fail (for e.g. a `read`/`write` would fail with `EBADF: file handle has been invalidated by a newer handle opened`) or be no-op (for e.g. `flush`, `release`) if a new `open` has overridden the flushed handle(s).
- A race condition can occur between a `read`/`write` request for a duplicate file descriptor on an existing (flushed) handle and multiple concurrent `open` requests, and any of them might succeed due to parallel processing of FUSE requests within Mountpoint. However, only one of them will ever succeed and there cannot be two concurrent writers or reader+writer for the inode at any point in time.
- An `open` request for an inode might fail if the pending upload to S3 fails. This is independent of whether the file has been truncated in the second `open`.

Yes, it needs a changelog entry and update to the semantics.md. Yes it also requires a version change.

---

### DIFF
--- a/mountpoint-s3-fs/src/superblock.rs
+++ b/mountpoint-s3-fs/src/superblock.rs
@@ -2015,19 +2015,20 @@ pub enum InodeMapError {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
-    use crate::fs::{FUSE_ROOT_INODE, TimeToLive, ToErrno};
-    use crate::metablock::AddDirEntryResult;
-    use crate::s3::{Bucket, Prefix};
-    use mountpoint_s3_client::mock_client::Operation;
     use mountpoint_s3_client::{
-        mock_client::{MockClient, MockObject},
+        mock_client::{MockClient, MockObject, Operation},
         types::ETag,
     };
     use std::ffi::OsString;
     use std::str::FromStr;
     use test_case::test_case;
     use time::{Duration, OffsetDateTime};
+
+    use crate::fs::{FUSE_ROOT_INODE, TimeToLive, ToErrno};
+    use crate::metablock::AddDirEntryResult;
+    use crate::s3::{Bucket, Prefix};
+
+    use super::*;
 
     /// Check an Inode's stat matches a series of fields.
     macro_rules! assert_inode_stat {
@@ -2293,11 +2294,11 @@ mod tests {
 
     #[tokio::test]
     async fn test_getattr_with_inode_local_invalid_stat_force_revalidate() {
-        let (metablock, client) = setup_test_metablock();
+        let (superblock, client) = setup_test_superblock();
         let head_object_counter = client.new_counter(Operation::HeadObject);
-        let inode = create_test_inode(&metablock, WriteStatus::LocalUnopened, true, None, true).unwrap();
+        let inode = create_test_inode(&superblock, WriteStatus::LocalUnopened, true, None, true).unwrap();
 
-        let result = metablock.getattr_with_inode(inode.ino(), true).await;
+        let result = superblock.getattr_with_inode(inode.ino(), true).await;
         assert!(result.is_ok());
         let looked_up = result.unwrap();
         assert_eq!(looked_up.write_status, WriteStatus::LocalUnopened);
@@ -2308,12 +2309,12 @@ mod tests {
 
     #[tokio::test]
     async fn test_getattr_with_inode_remote_valid_stat_no_force_revalidate() {
-        let (metablock, client) = setup_test_metablock();
+        let (superblock, client) = setup_test_superblock();
         let test_etag = ETag::for_tests();
         let head_object_counter = client.new_counter(Operation::HeadObject);
-        let inode = create_test_inode(&metablock, WriteStatus::Remote, false, Some(test_etag.clone()), false).unwrap();
+        let inode = create_test_inode(&superblock, WriteStatus::Remote, false, Some(test_etag.clone()), false).unwrap();
 
-        let result = metablock.getattr_with_inode(inode.ino(), false).await;
+        let result = superblock.getattr_with_inode(inode.ino(), false).await;
         assert!(result.is_ok());
         let looked_up = result.unwrap();
         assert_eq!(looked_up.write_status, WriteStatus::Remote);
@@ -2324,13 +2325,13 @@ mod tests {
 
     #[tokio::test]
     async fn test_getattr_with_inode_remote_force_revalidate() {
-        let (metablock, client) = setup_test_metablock();
+        let (superblock, client) = setup_test_superblock();
         let test_etag = ETag::for_tests();
         let head_object_counter = client.new_counter(Operation::HeadObject);
         client.add_object("prefix/test inode", MockObject::constant(0xaa, 1024, test_etag.clone()));
-        let inode = create_test_inode(&metablock, WriteStatus::Remote, false, Some(test_etag.clone()), false).unwrap();
+        let inode = create_test_inode(&superblock, WriteStatus::Remote, false, Some(test_etag.clone()), false).unwrap();
 
-        let result = metablock.getattr_with_inode(inode.ino(), true).await;
+        let result = superblock.getattr_with_inode(inode.ino(), true).await;
         assert!(result.is_ok());
         let looked_up = result.unwrap();
         assert_eq!(looked_up.write_status, WriteStatus::Remote);
@@ -2341,13 +2342,14 @@ mod tests {
 
     #[tokio::test]
     async fn test_getattr_with_stale_inode_remote_no_force_revalidate() {
-        let (metablock, client) = setup_test_metablock();
+        let (superblock, client) = setup_test_superblock();
         let head_object_counter = client.new_counter(Operation::HeadObject);
+        // Create the object with a different e-tag than which the inode is created with
         client.add_object("prefix/test inode", MockObject::constant(0xaa, 1024, ETag::for_tests()));
         let local_etag = ETag::from_str("Stale local etag").unwrap();
-        let inode = create_test_inode(&metablock, WriteStatus::Remote, false, Some(local_etag), true).unwrap();
+        let inode = create_test_inode(&superblock, WriteStatus::Remote, false, Some(local_etag), true).unwrap();
 
-        let result = metablock.getattr_with_inode(inode.ino(), false).await;
+        let result = superblock.getattr_with_inode(inode.ino(), false).await;
         if let Err(InodeError::StaleInode {
             remote_key,
             old_inode,
@@ -2362,12 +2364,12 @@ mod tests {
         }
     }
 
-    fn setup_test_metablock() -> (Superblock<Arc<MockClient>>, Arc<MockClient>) {
+    fn setup_test_superblock() -> (Superblock<Arc<MockClient>>, Arc<MockClient>) {
         let bucket = Bucket::new("test_bucket").unwrap();
         let prefix = "prefix/";
         let client = Arc::new(MockClient::config().bucket(bucket.to_string()).part_size(32).build());
         let prefix = Prefix::new(prefix).expect("valid prefix");
-        let metablock = Superblock::new(
+        let superblock = Superblock::new(
             client.clone(),
             S3Path::new(bucket, prefix.clone()),
             SuperblockConfig {
@@ -2375,17 +2377,17 @@ mod tests {
                 s3_personality: S3Personality::Standard,
             },
         );
-        (metablock, client)
+        (superblock, client)
     }
 
     fn create_test_inode(
-        metablock: &Superblock<Arc<MockClient>>,
+        superblock: &Superblock<Arc<MockClient>>,
         write_status: WriteStatus,
         is_new_file: bool,
         etag: Option<ETag>,
         invalidate_stat: bool,
     ) -> Result<Inode, InodeError> {
-        let parent_inode = metablock.inner.get(FUSE_ROOT_INODE).unwrap();
+        let parent_inode = superblock.inner.get(FUSE_ROOT_INODE).unwrap();
         let etag = etag.map(|etag| etag.into_inner().into_boxed_str());
 
         let mut stat = InodeStat::for_file(
@@ -2397,13 +2399,14 @@ mod tests {
             std::time::Duration::from_secs(24 * 60 * 60),
         );
         if invalidate_stat {
+            // For testing the expired stat cases
             stat.update_validity(std::time::Duration::from_secs(0));
         }
         let state = InodeState::new(&stat, InodeKind::File, write_status);
 
         let mut parent_state = parent_inode.get_mut_inode_state()?;
 
-        let inode = metablock.inner.create_inode_locked(
+        let inode = superblock.inner.create_inode_locked(
             &parent_inode,
             &mut parent_state,
             ValidName::parse_str("test inode")?,
@@ -2411,7 +2414,8 @@ mod tests {
             state,
             is_new_file,
         )?;
-        metablock.inner.remember(&inode);
+        // Manually add the inode to the Superblock's inode map, done by lookup/create in reality
+        superblock.inner.remember(&inode);
         Ok(inode)
     }
 


mountpoint-s3-fs/src/superblock.rs
```diff
@@ -321,6 +323,97 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
...
+        let completion_hook = {
+            let mut state = looked_up_inode.inode.get_mut_inode_state()?;
+            //debug!("Open: got inode state lock for writing");
+            match state.write_status {
+                WriteStatus::LocalUnopened => {
+                    state.write_status = WriteStatus::LocalOpenForWriting;
+                    state.stat.size = 0;
+                    self.inner
+                        .inode_handles
+                        .set_writer(&state, handle_id)
```
vladem: is it actually possible that `set_writer` fails when `WriteStatus::LocalUnopened`? what is the scenario?

we also modify the `state` above, should we rollback on an error?

mountpoint-s3-fs/src/superblock/inode.rs
```diff
@@ -240,6 +244,7 @@ pub struct InodeState {
     pub stat: InodeStat,
     pub write_status: WriteStatus,
     pub kind_data: InodeKindData,
+    pub completion_hook: Option<CompletionHook>,
```
vladem: Let's document what this hook does, when it is created and when it is triggered.

mountpoint-s3-fs/src/fs/handles.rs
```diff
@@ -39,77 +40,84 @@ where
     Client: ObjectClient + Clone + Send + Sync + 'static,
 {
     /// The file handle has been assigned as a read handle
-    Read(PrefetchGetObject<Client>),
+    Read {
+        request: PrefetchGetObject<Client>,
+        flushed: bool,
```
vladem, line 46: "flushed" seems to be recorded in multiple places: here and in the superblock (`HandleState::Inactive`), I remember that duplication optimises something, could you remind me which specific scenario please?

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -321,6 +323,97 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
...
+        if let Some(hook) = completion_hook
+            && let Some(lookup) = hook.trigger().await?
+        {
+            Ok(lookup)
+        } else {
+            Ok(looked_up_inode.into())
+        }
+    }
+
+    async fn start_writing(
```
vladem: Let's document what does this method allow. E.g. opening for writing if (and only if) all of the other file handles for this inode have been flushed.

mountpoint-s3-fs/src/fs/handles.rs
mansi153: Yeah that's right, and it's done to avoid querying the inode open_handles map (which requires locking the inode) every time a read/write is received on a file handle and only if the handle is currently in the "flushed" state.

mountpoint-s3-fs/src/fs/handles.rs
```diff
@@ -39,77 +40,84 @@ where
...
+                        stat.etag.as_ref().map(|e| e.into())
+                    };
+                    let current_offset = if is_truncate { 0 } else { stat.size as u64 };
+                    let request = fs.uploader.start_incremental_upload(
+                        bucket.to_string(),
+                        full_key.into(),
+                        current_offset,
+                        initial_etag.clone(),
+                    );
+                    FileHandleState::Write {
```
passaro: nit: could we return the `UploadState` here and on the mpu branch? And then build the `FileHandleState` once for both cases?

mountpoint-s3-fs/src/fs/handles.rs
```diff
@@ -170,19 +179,22 @@ where
...
                     UploadState::Failed(_) | UploadState::Completed => unreachable!("checked above"),
-                };
+                }
                 Err(e)
             }
         }
     }
 
-    pub async fn commit(&mut self, fs: &S3Filesystem<Client>, handle: &FileHandle<Client>) -> Result<(), Error> {
+    /// Commit and return whether the handle should be set to "flushed".
```
passaro: comment out of date? (returns void!)

mountpoint-s3-fs/src/fs/handles.rs
```diff
@@ -272,84 +280,93 @@ where
...
-        fs: &S3Filesystem<Client>,
+        &mut self,
+        metablock: Arc<dyn Metablock>,
         ino: InodeNo,
         key: &S3Location,
-    ) -> Result<bool, Error> {
-        match self {
+        fh: u64,
+    ) -> Result<Option<Lookup>, InodeError> {
+        match std::mem::replace(self, UploadState::Completed) {
```
passaro, line 334: This will change to `Completed` even if it was previously `Failed`. Is that intended?

mountpoint-s3-fs/src/metablock.rs
```diff
@@ -62,13 +72,31 @@ pub trait Metablock: Send + Sync {
...
-
-    /// Prepare an inode (referenced by `ino`) to start reading.
-    async fn start_reading(&self, ino: InodeNo) -> Result<(), InodeError>;
+    async fn finish_writing(&self, ino: InodeNo, etag: Option<ETag>, fh: u64) -> Result<Lookup, InodeError>;
 
     /// Finish reading from the inode (referenced by `ino`)
-    async fn finish_reading(&self, ino: InodeNo) -> Result<(), InodeError>;
+    async fn finish_reading(&self, ino: InodeNo, fh: u64) -> Result<(), InodeError>;
+
+    /// Updates status of the inode and of containing "local" directories.
```
passaro: Rustdocs of all the new methods need updating.

mountpoint-s3-fs/src/metablock.rs
```diff
@@ -163,3 +191,31 @@ impl WriteMode {
         }
     }
 }
+
+#[derive(Debug, Clone, Copy)]
+pub enum ReadWriteMode {
```
passaro, line 216: rustdocs - for `NewHandle` as well

mountpoint-s3-fs/src/superblock/handles_map.rs
```diff
@@ -0,0 +1,160 @@
...
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+
+use super::InodeLockedForWriting;
+
+/// Tracks currently open readers and writers (file handles) for Inodes.
+/// Ensures that the Inodes are locked for writing while performing any operation.
+#[derive(Debug, Default)]
+pub struct InodeHandleMap {
+    handles: RwLock<HashMap<InodeNo, InodeHandles>>,
```
passaro: Do we need the `RwLock` here? It seems we only ever use `.write()`.

mountpoint-s3-fs/src/superblock/handles_map.rs
```diff
@@ -0,0 +1,160 @@
...
+            entry.get_mut().writer = None;
+            if entry.get().is_empty() {
+                entry.remove();
+            }
+            return true;
+        }
+        false
+    }
+
+    pub fn remove_inode(&self, ino: u64) -> bool {
```
passaro, line 162: nit: rustdoc

mountpoint-s3-fs/src/superblock/inode.rs
passaro: Consider a more descriptive name, e.g. `pending_upload`.

mountpoint-s3-fs/src/fs.rs
```diff
@@ -866,4 +867,196 @@ mod tests {
...
+        // Lookup inode of the dir1 directory
+        let entry = fs.lookup(FUSE_ROOT_INODE, "dir1".as_ref()).await.unwrap();
+        assert_eq!(entry.attr.kind, FileType::Directory);
+        let dir_ino = entry.attr.ino;
+
+        // Open a file for write in "dir1"
+        let dentry = fs
+            .mknod(
+                dir_ino,
+                format!("{}2.txt", test_name).as_ref(),
```
passaro, line 1065: Why do we need to include the test name? This is only used against a mock client.

mountpoint-s3-fs/src/fs.rs
```diff
@@ -866,4 +867,196 @@ mod tests {
...
+
+    fn setup_mock_fs(test_name: &str, allow_overwrite: bool, incremental_upload: bool) -> S3Filesystem<MockClient> {
+        let bucket = Bucket::new("bucket").unwrap();
+        let client = MockClient::config()
+            .bucket(bucket.to_string())
+            .enable_backpressure(true)
+            .initial_read_window_size(1024 * 1024)
+            .part_size(1024 * 1024)
+            .build();
+        // Create "dir1" in the client to avoid creating it locally
```
passaro, line 1084: Do we need "dir1" at all? Any drawback in working at the root level?

mountpoint-s3-fs/src/metablock/completion.rs
```diff
@@ -0,0 +1,61 @@
+use mountpoint_s3_client::ObjectClient;
+use std::pin::Pin;
+
+use crate::fs::{FileHandle, FileHandleState};
+use crate::sync::{Arc, AsyncMutex};
+
+use super::{InodeError, Lookup, Metablock};
+
+#[derive(Clone, Debug)]
+pub struct CompletionHook {
```
passaro: As mentioned in other comments, this could have a more descriptive name. And needs rustdocs.

mountpoint-s3-fs/src/metablock/completion.rs
```diff
@@ -0,0 +1,61 @@
...
+                };
+                state.complete_if_in_progress(metablock, ino, &location, fh).await
+            }),
+            result: None,
+        }));
+
+        Self { state }
+    }
+
+    pub async fn trigger(&self) -> CompletionResult {
```
passaro: Add rustdoc & rename. Suggestions: `wait_for_completion`? `ensure_completed`?

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -811,35 +926,82 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
...
-        self.inner.reader_counts.write().unwrap().remove_reader(&state);
+        self.inner.open_handles.remove_reader(&state, fh);
+        Ok(())
+    }
+
+    async fn release_writer(
+        &self,
+        ino: InodeNo,
+        fh: u64,
+        completion_handle: CompletionHook,
```
passaro: rename

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -811,35 +926,82 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
...
+        location: &S3Location,
+    ) -> Result<(), InodeError> {
+        let completion_hook = self.flush_writer(ino, fh, completion_handle.clone()).await;
```
passaro: why all the `clone`s? and the explicit `drop`s at the end?

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1084,6 +1246,9 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
         match inodes.decrease_lookup_count(ino, n) {
             Ok(Some(removed_inode)) => {
                 trace!(ino, "removing inode from superblock");
+                if self.inner.open_handles.remove_inode(ino) {
+                    debug!("File handles found for forgotten inode")
```
passaro: typo: handles -> handle

Also, we should include the `ino` in the log

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1124,6 +1289,26 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
             }
         }
     }
+
+    async fn validate_handle(&self, ino: InodeNo, fh: u64, mode: ReadWriteMode) -> Result<bool, InodeError> {
```
passaro: nit: consider renaming to `try_validate_handle` or `try_activate_handle`.

Alternatively, we could refactor to `activate_handle` and return a new `InodeError` variant when we are not able to (instead of `false`). I think we error out in `fs` in that case anyway.

mountpoint-s3-fs/src/fs.rs
mansi153: I found it makes it easier to debug compared to having generic and same (e.g. "hello.txt") names across all tests. Do you see a potential issue with using distinct names in new tests?

mountpoint-s3-fs/src/fs.rs
mansi153: No drawbacks as far as I'm aware. I just kept it consistent with the other existing FS test `test_open_with_corrupted_sse`, but I think it can be gotten rid of

mountpoint-s3-fs/src/fs/handles.rs
mansi153: Yeah, could do that, thanks for pointing it out

mountpoint-s3-fs/src/metablock.rs
```diff
@@ -163,3 +191,31 @@ impl WriteMode {
...
+    Read,
+    Write,
+}
+
+#[derive(Debug)]
+pub struct NewHandle {
```
vladem, line 226: what is `NewHandle`? how is it different from `FileHandle`?

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -321,6 +323,91 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
...
+            }
+            locked_inode.pending_upload_hook.clone()
+        };
+
+        if let Some(upload_hook) = pending_upload_hook
+            && let Some(lookup) = upload_hook.wait_for_completion().await?
+        {
+            Ok(lookup)
+        } else {
+            Ok(looked_up_inode.into())
```
vladem: why do we return a lookup from the completion hook? may it be different from `looked_up_inode`?

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -321,6 +323,91 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
...
+    async fn start_reading(&self, looked_up_inode: LookedUpInode, fh: u64) -> Result<Lookup, InodeError> {
+        let pending_upload_hook = {
+            let mut locked_inode = looked_up_inode.inode.get_mut_inode_state()?;
+            if matches!(locked_inode.write_status, WriteStatus::LocalUnopened)
+                || !self.inner.open_handles.try_add_reader(&locked_inode, fh)
+            {
+                return Err(InodeError::InodeNotReadableWhileWriting(looked_up_inode.inode.err()));
+            }
+            if matches!(locked_inode.write_status, WriteStatus::LocalOpenForWriting) {
+                locked_inode.write_status = WriteStatus::Remote;
```
vladem: nit: comment / debug log about what's happening here: open for read "overtook" the write handle so switching status to `WriteStatus::Remote`

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -321,6 +323,91 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
...
+            }
+            locked_inode.pending_upload_hook.clone()
+        };
+
```
vladem: can we add `debug_assert!(matches!(locked_inode.write_status, WriteStatus::Remote));` here? may the inode be in any other status? what if completion hook failed?

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -321,6 +323,91 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
         trace!("Added handle with id: {}", handle_id);
         Ok(handle_id)
     }
+
+    async fn start_reading(&self, looked_up_inode: LookedUpInode, fh: u64) -> Result<Lookup, InodeError> {
```
vladem: nit: make it explicit that opening for reads is allowed if inode is in one of the following statuses by either a comment, or allowlist (currently you have a denylist here):

- WriteStatus::Remote
- WriteStatus::PendingRename
- WriteStatus::LocalOpenForWriting (if all FHs are flushed)

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -716,43 +803,62 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
...
+        flags: OpenFlags,
+    ) -> Result<NewHandle, InodeError> {
+        let force_revalidate = !self.inner.config.cache_config.serve_lookup_from_cache || flags.direct_io();
+        let looked_up_inode = self.getattr_with_inode(ino, force_revalidate).await?;
+        match looked_up_inode.inode.kind() {
+            InodeKind::Directory => return Err(InodeError::IsDirectory(looked_up_inode.inode.err())),
+            InodeKind::File => (),
+        }
+
+        let mode = if flags.contains(OpenFlags::O_RDWR) {
```
vladem, line 825: is `Superblock` the right place to choose the `ReadWriteMode`?

mountpoint-s3-fs/src/superblock/handles_map.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    /// Set an existing reader to active.
+    ///
+    /// Return `false` if the reader did not exist.
+    pub fn try_activate_reader(&self, locked_inode: &InodeLockedForWriting<'_>, fh: u64) -> bool {
+        let mut handles = self.handles.lock().unwrap();
+        if let Some(entry) = handles.get_mut(&locked_inode.ino)
+            && let Some(reader) = entry.readers.get_mut(&fh)
```
vladem, line 72: if `reader` is not found it's unexpected, right? should we report it as an error?

I am thinking of a case when `fs::read` happens and `flushed == true` there

mountpoint-s3-fs/src/metablock.rs
```diff
@@ -52,23 +56,47 @@ pub trait Metablock: Send + Sync {
...
     async fn inc_file_size(&self, ino: InodeNo, len: usize) -> Result<usize, InodeError>;
 
-    /// Called when the filesystem has finished writing to the inode refernced by `ino`.
+    /// Called when the filesystem has finished writing to the inode referenced by `ino`.
     /// Allows the implementor to make necessary adjustments / update its internal structure.
-    async fn finish_writing(&self, ino: InodeNo, etag: Option<ETag>) -> Result<(), InodeError>;
-
-    /// Prepare an inode (referenced by `ino`) to start reading.
-    async fn start_reading(&self, ino: InodeNo) -> Result<(), InodeError>;
+    async fn finish_writing(&self, ino: InodeNo, etag: Option<ETag>, fh: u64) -> Result<Lookup, InodeError>;
```
vladem, line 77: consider merging `finish_*` methods into one with an argument `mode: ReadWriteMode` specifying the type (the same for `flush_*`)

this would be consistent with `try_activate_handle` signature

mountpoint-s3-fs/src/manifest/metablock.rs
```diff
@@ -278,4 +282,31 @@ impl Metablock for ManifestMetablock {
...
+
+    async fn try_activate_handle(&self, _ino: InodeNo, _fh: u64, _mode: ReadWriteMode) -> Result<bool, InodeError> {
+        Ok(true)
+    }
+
+    async fn flush_reader(&self, _ino: InodeNo, _fh: u64) -> Result<bool, InodeError> {
+        Ok(true)
+    }
+
+    async fn flush_writer(
```
vladem, line 295: this method and the one below should probably return an error

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1084,6 +1244,9 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
         match inodes.decrease_lookup_count(ino, n) {
             Ok(Some(removed_inode)) => {
                 trace!(ino, "removing inode from superblock");
+                if self.inner.open_handles.remove_inode(ino) {
+                    debug!("Open file handle(s) found for forgotten inode {}", ino);
```
vladem: should never happen? I'd add `debug_assert!(!self.inner.open_handles.remove_inode(ino))` then

mountpoint-s3-fs/src/superblock/handles_map.rs
mansi153: Yes, in that case we return `false` to `fs::read` and it errors out with "EBADF: file handle has been invalidated by a newer handle opened"

It is expected because another open could have overridden this read handle in the time when it was flushed and when this read request came through. Think of a duplicate FD case as an example

mountpoint-s3-fs/src/metablock.rs
mansi153: I would prefer to keep them separate for now given they don't do very similar things code-wise, and we already had separate finish-reading/writing methods. `try_activate_handle` is a much simpler method, I'm apprehensive of combining the `flush-*` and `finish-*` ones into one big confusing method. WDYT?

mountpoint-s3-fs/src/superblock.rs
mansi153: The completion of the pending upload results in the object `etag` being updated, and hence we want to convey the latest inode stat values (with updated validity) and LookedUpInode to the caller awaiting the hook completion before it proceeds with the rest of the request. 
A failed hook completion returns the stale information it has (same as the existing lookup) but invalidates the inode's stats instead so that we refresh them from S3 when next queried.

mountpoint-s3-fs/src/superblock.rs
mansi153: At this point, the inode is remote/pendinRename and has at least one active reader in the handles_map. Which means it can't be opened for writing while we complete this buffered state of things (the hook completion). It can however be opened for reading/be renamed, but we allow all that [with the caveat that the state in S3 we know of is latest from "when we last checked"], but anything trying to read would have first awaited this pendingUploadHook within `start_reading`, so I think we should be fine with the current state of the inode.

can we add debug_assert!(matches!(locked_inode.write_status, WriteStatus::Remote)); - No, because we can be in pendingRename and we need to support that :) 

may the inode be in any other status? - yes, pendingRename. But you're right in that the check should be more restrictive to avoid future write_status checks incomplete, I'll fix that.

what if completion hook failed? - We will convey that failure to the `open`(s) and error out the `open` for read to tell the user the pending upload failed.

mountpoint-s3-fs/src/superblock.rs
mansi153: The superblock `open_handle` now encapsulates the majority of the open logic so it makes sense to keep it here. The motivation to move this from FS into superblock is also because we need to await the pending upload hook before we do `start_reading/writing` within the new `open`, and that needed some refactoring of method responsibilities.

mountpoint-s3-fs/src/superblock.rs
mansi153: Yes, in theory it shoudn't. But I'm not sure we want this to be an internal invariant which we track and fail the `forget` because of it - given we don't do any kind of checks in our `fs` (say in file_handles map) to verify we're not forgetting an inode with open handles. 

But we can discuss if we want to add it as a breaking change, in which case I'd incline towards just `assert!` instead of a special debug-mode feature, WDYT?

mountpoint-s3-fs/src/superblock.rs
vladem: Please add a comment with the note about updated etag, maybe to `wait_for_completion` method.

mountpoint-s3-fs/src/superblock.rs
vladem: I think this if-block is redundant as the status change will happen in `upload_hook.wait_for_completion().await?`. I would still add a debug log about "overriding" the handle and also a `debug_assert!` in the end of this method on the final `write_status` (should be `WriteStatus::Remote`).

mountpoint-s3-fs/src/superblock.rs
vladem: > can we add debug_assert!(matches!(locked_inode.write_status, WriteStatus::Remote)); - No, because we can be in pendingRename and we need to support that :)

Let's add a `debug_assert!`, that we're in one of those states. In my opinion the state machine must be explicit (with asserts on transitions), otherwise it is very hard to follow.

mountpoint-s3-fs/src/superblock.rs
vladem: What is the downside of adding a `debug_assert!`? The benefit is that, if our assumptions are incorrect, we will see test failures and have a chance to address. It's quite important that the `pending_upload_hook` for the inode with a  writer is completed, before we forget the inode.

mountpoint-s3-fs/src/superblock.rs
mansi153: It can't, but we should keep the state change after the set_writer check. Updated the code.

mountpoint-s3-fs/src/superblock.rs
mansi153: I retained the boolean return logic for simplicity of division of responsibilities, but have renamed it to `try_reactivate_handle`

mountpoint-s3-fs/src/superblock.rs
mansi153: Added it to `finish_writing`

mountpoint-s3-fs/src/superblock.rs
mansi153: I haven't written this exactly, but added some detail about the method's working. I think adding semantic detail here might be too much detail for a Rustdoc.

mountpoint-s3-fs/src/fs/handles.rs
```diff
@@ -42,10 +42,13 @@ where
     /// The file handle has been assigned as a read handle
     Read {
         request: PrefetchGetObject<Client>,
-        flushed: bool,
+        flushed: bool, // Set to true when `flush` called on the handle, and unset on a `read`
```
passaro: tiny nit: comment could be a rustdoc

mountpoint-s3-fs/src/superblock/handles_map.rs
```diff
@@ -48,8 +52,8 @@ impl InodeHandleMap {
     }
 
     /// Set an existing reader to inactive.
-    pub fn deactivate_reader(&self, locked_inode: &InodeLockedForWriting<'_>, fh: u64) {
-        let mut handles = self.handles.write().unwrap();
+    pub fn try_deactivate_reader(&self, locked_inode: &InodeLockedForWriting<'_>, fh: u64) {
```
passaro: `try_` is confusing here without returning a `bool`. If the failure case is not really interesting for callers, revert the name and document that it's a no op if the reader is not there (or is already deactivated).

mountpoint-s3-fs/src/metablock.rs
```diff
@@ -70,33 +70,40 @@ pub trait Metablock: Send + Sync {
...
     async fn finish_reading(&self, ino: InodeNo, fh: u64) -> Result<(), InodeError>;
 
-    /// Updates status of the inode and of containing "local" directories.
+    /// Called when the filesystem has called `flush` on a read handle for the inode referenced by `ino`.
+    /// Allows the implementor to make necessary adjustments / update its internal structure.
     async fn flush_reader(&self, ino: InodeNo, fh: u64) -> Result<bool, InodeError>;
 
-    /// Updates status of the inode and of containing "local" directories.
+    /// Called when the filesystem has called `flush` on a write handle for the inode referenced by `ino`.
+    /// Allows the implementor to make necessary adjustments / update its internal structure.
```
passaro: "Allows the implementor to make necessary adjustments / update its internal structure" is not too useful.

I'd rather explain why/when it returns a `PendingUploadHook`.

passaro: Similarly for some of the other methods.

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -866,7 +884,11 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
         Ok(state.stat.size)
     }
 
-    /// Updates status of the inode and of containing "local" directories.
+    /// Concludes a writing operation for a file handle and marks the inode closed for writing anymore.
+    ///
+    /// Transitions the inode and all it's ancestor directories from local writing state to remote
```
passaro: typo: its

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1278,6 +1319,15 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
                     metrics::counter!("metadata_cache.inode_forgotten_before_expiry")
                         .increment(state.stat.is_valid().into());
                 };
+                if self.inner.open_handles.remove_inode(ino) {
+                    // This should never happen, but it is good to have this visibility to detect any
+                    // discrepancies in our inode handles' tracking logic or tests involving `forget`
+                    debug_assert!(
```
passaro: Can we guarantee that we would always complete `release` before `forget`? If not, this will panic in tests even if the kernel issued the requests in the expected order.

passaro (commented):
Left a few additional suggestions. For the next (final?) round of changes, we should add changelog entries and increment version numbers if needed.

mountpoint-s3-fs/src/superblock/handles_map.rs
vladem: Do we also need to check that there is no writer: `&& entry.writer.is_none()`?

Otherwise, I think in this case we will allow the read:
    
```
fd1 = open("a.txt", "r")
fd2 = dup(fd1)
close(fd1) // flush, but no release yet (deactivated, but still in the `entry.readers` map)
fd3 = open("a.txt", "w") // override (all readers are inactive)
read(fd2) // `try_reactivate_handle` succeeds and the read is allowed (while there is an unflushed writer!)
```

Or am I missing something and something in our code prohibits such scenario? Note that some data may be written to `fd3` before the `read(fd2)`.

mountpoint-s3-fs/src/metablock.rs
mansi153: I will add some more detail about return types etc. 
But this is just the metablock documentation to keep the details generic. The superblock methods have their own rustdocs so I'm inclined to not have too much information in the type abstraction

mountpoint-s3-fs/src/superblock/handles_map.rs
mansi153: Ah, the `entry.readers.clear();` in `set_writer` should have removed the reader fh (mapping to fd1/fd2) as part of `start_writing` for the fd3 `open`, so `try_activate_reader` will return `false` for fd2 and hence the `read(fd2)` will fail with `libc::EBADF, "file handle has been invalidated by a newer handle opened"`. Does that make sense or am I misinterpreting your question?

mountpoint-s3-fs/src/superblock.rs
mansi153: Ah, no, the state change in `finish_writing` (called as part of `upload_hook.wait_for_completion().await?`) is only if it's releasing a currently open writer. If it's been triggered by a new `open`, it'll respect the state change (to `LocalOpenForWriting` or `Remote` based on what the new open is for) done in the new open. 

So we need this `locked_inode.write_status = WriteStatus::Remote;`. Maybe we don't need the `if`, but I didn't want to be too loose and update state in case it can be `PendingRename` in some case and we don't want to move to `Remote` there.

mountpoint-s3-fs/src/superblock.rs
vladem: I can see the point, given that requests may be processed out of order, but can't such condition lead to data loss? If so, I'd rather have this assert in tests, for us to know how frequent it is. 

If I understand the case with `forget` before `release` correctly than we can have something like this:

1. empty file created (`open`, `flush`), write to S3 delayed until `release`;
2. `forget` processed first, inode removed from `InodeMap`;
3. `release` processed next, `metablock.release_writer` errors out on inexistent inode. Empty file never gets written to S3 (no object data but we lost the key itself).

mountpoint-s3-fs/tests/fuse_tests/write_test.rs
```diff
@@ -1668,3 +1663,160 @@ fn append_fails_on_object_replaced_s3() {
...
+#[test_matrix(MOCK, [Open::ForReading, Open::ForWriting], [ATOMIC_UPLOAD, INCREMENTAL_UPLOAD])]
+fn open_after_closing_empty_file_test(
+    creator_fn: impl TestSessionCreator,
+    second_open_type: Open,
+    upload_mode: UploadMode,
+) {
+    const KEY: &str = "empty.txt";
+
+    let config = TestSessionConfig {
+        filesystem_config: S3FilesystemConfig {
```
vladem, line 1687: Some of the tests in this matrix are flaky, probably worth increasing `metadata_ttl` here:

```rust
filesystem_config: S3FilesystemConfig {
    allow_overwrite: true,
    cache_config: CacheConfig {
        serve_lookup_from_cache: true,
        dir_ttl: Duration::from_secs(600),
        file_ttl: Duration::from_secs(600),
        ..Default::default()
    },
    ..Default::default()
}
```

I've observed the failure:

```
thread 'fuse_tests::write_test::open_after_closing_empty_file_test::s3_open_forreading_atomic_upload_expects' panicked at mountpoint-s3-fs/tests/fuse_tests/write_test.rs:1719:6:
second open should succeed: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

And another one (after increased `metadata_ttl`):
```
2025-12-18T17:55:02.378395Z  WARN open{req=24 ino=2 pid=4503 name="empty.txt"}: mountpoint_s3_fs::fuse: open failed with errno 9: no E-Tag for inode 2
...
thread 'fuse_tests::write_test::open_after_closing_empty_file_test::mock_open_forreading_incremental_upload_expects' panicked at mountpoint-s3-fs/tests/fuse_tests/write_test.rs:1724:6:
second open should succeed: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }
```

mountpoint-s3-fs/src/superblock.rs
mansi153: I agree with the issue, but think now I agree with @passaro and feel that `debug_assert!` is not the best solution to the problem. Given it's a race condition, we should acknowledge that it might not happen in tests frequently enough, or be ignored as unnecessarily flaky or unrelated.

- Since MP should be the one processing the `release` _after_ the `forget`, I think it makes sense for the `release` to then error out in the test/production scenario.
- The `forget` logic needs to be more robust than it is today. MP should do this check _before_ forgetting the inode and error out the `forget` earlier in the request path (or not, if we want to abide by the Kernel completely?) to protect its internal state. Querying `open_handles` is potentially a good way to do that, but I think it makes sense to do it in a separate PR (I will create a bug ticket).

mountpoint-s3-fs/tests/fuse_tests/write_test.rs
mansi153: As discussed, this is due to the race condition between a `lookup` during an `Open` and a `Release` being processed concurrently. Working on it as a separate ticket

mountpoint-s3-fs/src/metablock.rs
passaro: Agree on the distinction between the trait and the implementation in superblock. I would just remove the "Allows the implementor ..." bit completely: it does not add anything.

Rather, let's try to capture the contract between caller and implementers here. Specifically for `flush_writer`: the caller passes in the pending upload, why? (so it can be completed automatically in some cases depending by the implementation). Why is the pending upload (optionally) returned? Why would a caller need it? (I suspect this is only useful in superblock and we should change the trait method).

mountpoint-s3-fs/src/superblock.rs
passaro: If this is a pre-existing issue (that we can now detect by looking at `open_handles`, but not trivially fix), I'd add a TODO in this PR -- whether we log the issue or `debug_assert` is secondary, as long as we are addressing it (in a separate PR).

mountpoint-s3-fs/src/metablock.rs
mansi153: No, the callers pass in the hook because creating it requires using the generic `Client` trait (part of the FileHandle) which the Metablock should not depend on. I'll add some more clarification on the metablock, but I think the superblock doc and implementation is more helpful -- primarily because the method definitions are as such to re-use code and make signatures generic enough for all metablock implementations, more than any other logical reason.

mountpoint-s3-fs/src/fs/flags.rs
```diff
@@ -69,6 +69,7 @@ pub struct OpenFlags(i32);
 libc_flags! {
     OpenFlags : i32 {
         O_WRONLY,
+        O_RDONLY,
```
passaro: This is now inconsistent with the rustdoc above. Either clarify the doc or revert.

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -264,21 +266,39 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
...
             let sync = inode.get_inode_state()?;
-            if sync.stat.is_valid() {
-                let stat = sync.stat.clone();
-                let is_remote = sync.is_remote();
-                drop(sync);
+            (sync.is_remote(), sync.stat.clone(), sync.is_new())
+        };
+
+        if !is_remote || !force_revalidate_if_remote {
+            // If the inode is local (open/unopened), extend its stat's validity before returning.
```
passaro: This seems wrong: we are only updating the copy of the stats the we return here, not the original in the inode state.

mountpoint-s3-fs/src/superblock/inode.rs
```diff
@@ -248,12 +253,18 @@ impl InodeState {
             stat: stat.clone(),
             kind_data: InodeKindData::default_for(kind),
             write_status,
+            pending_upload_hook: None,
         }
     }
 
     pub fn is_remote(&self) -> bool {
+        // TODO: should this check also include PendingRename?
```
passaro: Any reason not to? Where is `is_remote()` used now? Let's also consider inverting it to `is_local()` if it makes sense.

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -705,63 +793,118 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
...
+        {
+            lookup_after_upload
+        } else {
+            inode_lookup.into()
+        };
+
+        match mode {
+            ReadWriteMode::Read => Ok(NewHandle::read(lookup)),
+            ReadWriteMode::Write => Ok(NewHandle::write(lookup)),
         }
```
passaro: nit: do we even need the special constructors?

```suggestion
        Ok(NewHandle { lookup, mode })
```

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1634,6 +1902,7 @@ pub struct LookedUpInode {
     pub stat: InodeStat,
     pub path: Arc<S3Path>,
     pub is_remote: bool,
+    pub is_new: bool,
```
passaro: Related to the previous comment on `InodeState::remote()`: another option could be to simply copy `write_status` here rather than multiple flags. It would force to think about the specific status at each use point.

passaro (commented):
Here are a few notes on the latest changes. We should add specific tests around the new lookup on open behavior (local vs remote).

mountpoint-s3-fs/src/superblock/inode.rs
mansi153: I've removed this now , and for usages of the method converted the checks to `write_status = Remote`. I would incline to not changing that behaviour as part of this PR, and keeping it limited to how Lookup uses the inode state information.

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -264,23 +266,39 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
...
+                let validity = match inode.kind() {
+                    InodeKind::File => self.inner.config.cache_config.file_ttl,
+                    InodeKind::Directory => self.inner.config.cache_config.dir_ttl,
+                };
+                sync.stat.update_validity(validity);
             }
+            (is_remote, sync.stat.clone(), sync.write_status)
+        };
+
+        if (!is_remote || !force_revalidate_if_remote) && stat.is_valid() {
```
passaro: minor: could we move this inside the block above? it would be clearer (avoid taking our 3 variables) and save a state.clone() when revalidation is needed

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -321,6 +339,77 @@ impl<OC: ObjectClient + Send + Sync> Superblock<OC> {
...
+                self.setup_inode_for_writing(locked_inode, inode, handle_id)?;
+                if is_truncate {
+                    locked_inode.stat.size = 0;
+                }
+                Ok(locked_inode.pending_upload_hook.clone())
+            }
+        }
+    }
+
+    fn setup_inode_for_writing(
```
passaro: nit: consider making this a local function in `start_writing`

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -811,35 +972,93 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
...
+                let write_status = locked_inode.write_status;
+                drop(locked_inode);
+
+                Ok(LookedUpInode {
+                    inode,
+                    stat,
+                    path: self.inner.s3_path.clone(),
+                    write_status,
+                }
+                .into())
```
passaro: nit: here and at least in another place, we build a `LookedUpInode` only to immediately turn it into a `Lookup`. Would we save anything if we avoided it?

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1061,14 +1280,14 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
...
-                .create_inode_locked(&parent_inode, &mut parent_state, name, kind, state, true)?;
+            let inode =
+                self.inner
+                    .create_inode_locked(&parent_inode, &mut parent_state, name, kind, state.clone(), true)?;
             LookedUpInode {
                 inode,
                 stat,
                 path: self.inner.s3_path.clone(),
-                is_remote: false,
+                write_status: state.write_status,
```
passaro: Use `WriteStatus::LocalUnopened` directly here, or save it into a variable before line 1282. But avoid cloning `InodeState` because of this.

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1491,12 +1746,12 @@ impl<OC: ObjectClient + Send + Sync> SuperblockInner<OC> {
             }
             (Some(remote), None) => {
                 let state = InodeState::new(&remote.stat, remote.kind, WriteStatus::Remote);
-                self.create_inode_locked(&parent, &mut parent_state, name, remote.kind, state, false)
+                self.create_inode_locked(&parent, &mut parent_state, name, remote.kind, state.clone(), false)
```
passaro: Avoid the clone

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1562,12 +1817,12 @@ impl<OC: ObjectClient + Send + Sync> SuperblockInner<OC> {
                 );
                 let state = InodeState::new(&remote.stat, remote.kind, WriteStatus::Remote);
                 let new_inode =
-                    self.create_inode_locked(&parent, &mut parent_state, name, remote.kind, state, false)?;
+                    self.create_inode_locked(&parent, &mut parent_state, name, remote.kind, state.clone(), false)?;
```
passaro: And here

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -2069,6 +2292,241 @@ mod tests {
         }
     }
 
+    /*#[tokio::test]
```
passaro: Is this still a work in progress?

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -2015,6 +2015,11 @@ pub enum InodeMapError {
 
 #[cfg(test)]
 mod tests {
+    use super::*;
```
passaro: nit: revert/fix imports

mountpoint-s3-fs/src/superblock.rs
```diff
@@ -2292,240 +2291,129 @@ mod tests {
...
-            SuperblockConfig {
-                cache_config: CacheConfig::new(TimeToLive::Duration(ttl)),
-                s3_personality: S3Personality::Standard,
-            },
-        );
-        let inode = create_test_inode(WriteStatus::LocalUnopened, true); // valid stat
-
-        let result = metablock.getattr_with_inode(inode.ino(), false).await;
+    async fn test_getattr_with_inode_local_invalid_stat_force_revalidate() {
+        let (metablock, client) = setup_test_metablock();
```
passaro: nit: rename metablock to superblock in fn and variable names

PR #1705, opened 2025-11-17
Add AL2023 to installation documentation

tadiwa-aizen:
Updating Installation documentation to account for AL2023.

No

No. It is a documentation update

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ On other RPM-based distributions (Amazon Linux 2, Fedora, CentOS, RHEL; excludin
 
     wget https://s3.amazonaws.com/mountpoint-s3-release/latest/x86_64/mount-s3.rpm
     sudo yum install -y ./mount-s3.rpm
-    
+
 On Ubuntu, use these commands instead (for Graviton instances, replace `x86_64` with `arm64` in the URL):
 
     wget https://s3.amazonaws.com/mountpoint-s3-release/latest/x86_64/mount-s3.deb


doc/INSTALL.md
```diff
@@ -10,6 +10,23 @@ Mountpoint for Amazon S3 is only available for Linux operating systems.
...
+
+Verify that Mountpoint for Amazon S3 is successfully installed by entering the following command:
+   ```
+   mount-s3 --version
+   ```
+
+You should see output similar to the following:
+   ```
+   mount-s3 1.21.0+1.amzn2023
+   ```
```
dannycjones, line 28: Are these steps or not? The indentation is off.

I'd update to be 1,2,3 or I'd drop entirely the `--version`, etc..

doc/INSTALL.md
```diff
@@ -10,6 +10,23 @@ Mountpoint for Amazon S3 is only available for Linux operating systems.
...
+
 ### RPM-based distributions (Amazon Linux, Fedora, CentOS, RHEL; excluding SUSE)
```
dannycjones: We should address AL in this section now.

Ultimately, this is for AL2. What about AL2023, is this a way to get a version earlier or not?

README.md
```diff
@@ -28,7 +28,11 @@ Mountpoint for Amazon S3 is generally available! We're tracking future feature d
 
 ## Getting started
 
-Run these two commands to install Mountpoint for Amazon S3 on your Amazon Linux EC2 instance (for Graviton instances, replace `x86_64` with `arm64` in the URL):
+On Amazon Linux 2023, you can install Mountpoint for Amazon S3 directly from the repository:
+
+    sudo dnf install mount-s3
+
+On other Amazon Linux versions, run these two commands (for Graviton instances, replace `x86_64` with `arm64` in the URL):
```
dannycjones: Drop AL2? We point to `INSTALL.md` anyway.

doc/INSTALL.md
dannycjones: If we change this, please can we be mindful about the anchor link. Take a look at it now, and then what it becomes. We can use a HTML anchor with no body to preserve old links.

tadiwa-aizen: You mean address the fact that they can install an RPM with a version of MP thats not available on the AL2023 distribution?

dannycjones: Yeah, exactly. If we don't want to recommend that, we should update to "Amazon Linux 2" or drop.

README.md
```diff
@@ -28,10 +28,9 @@ Mountpoint for Amazon S3 is generally available! We're tracking future feature d
 
 ## Getting started
 
-Run these two commands to install Mountpoint for Amazon S3 on your Amazon Linux EC2 instance (for Graviton instances, replace `x86_64` with `arm64` in the URL):
+On Amazon Linux 2023, you can install Mountpoint for Amazon S3 directly from the repository:
```
muddyfish, line 31: Why are we removing the previous instructions for RPM based instances? AL is not the only RPM based OS out there

doc/INSTALL.md
```diff
@@ -12,22 +12,22 @@ The instructions for downloading and installing Mountpoint for Amazon S3 depend
 
 ### Amazon Linux 2023 (AL2023)
 
-Mountpoint is now available directly in the Amazon Linux 2023 repository, so you can install it directly using the following command:
+Mountpoint is available directly in the Amazon Linux 2023 repository.
```
dannycjones: I think we should probably call out the version since it was added for now.


```suggestion
Mountpoint is available directly in the Amazon Linux 2023 repository since AL2023 version 2023.9.20251110.
```

README.md
dannycjones: My bad, I suggested that we simplify but I think you're right. https://github.com/awslabs/mountpoint-s3/pull/1705#discussion_r2534841230

I'd suggest that we make it clear that the instructions after AL2023 are for "other RPM-based distributions" dropping the phrase "other Amazon Linux" and keep the rest.

PR #1707, opened 2025-11-24
Align read window with part boundaries + configurable `initial_request_size`

vladem:
Re-created https://github.com/awslabs/mountpoint-s3/pull/1618, in this PR:

- we align read window end to the part boundary for the second request (see `round_up_to_part_boundary` method);
- we update mock client to allow testing of this change;
- we add `PrefetcherConfig::initial_request_size` field and use it in `mount_from_config.rs` example.

Minor version change and a change log to `mountpoint-s3-fs`, will add later. Patch version change to the `mountpoint-s3-fs-client`.

---

### DIFF
--- a/mountpoint-s3-crt/CHANGELOG.md
+++ b/mountpoint-s3-crt/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Unreleased
+## Unreleased (v0.13.5)
 
 ## v0.13.4 (October 30, 2025)
 
--- a/mountpoint-s3-fs/Cargo.toml
+++ b/mountpoint-s3-fs/Cargo.toml
@@ -9,7 +9,7 @@ description = "Mountpoint S3 main library"
 
 [dependencies]
 mountpoint-s3-fuser = { path = "../mountpoint-s3-fuser", version = "0.1.1", features = ["abi-7-28", "libfuse"] }
-mountpoint-s3-client = { path = "../mountpoint-s3-client", version = "0.19.4" }
+mountpoint-s3-client = { path = "../mountpoint-s3-client", version = "0.19.5" }
 
 anyhow = { version = "1.0.100", features = ["backtrace"] }
 async-channel = "2.5.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2517,7 +2517,7 @@ dependencies = [
 
 [[package]]
 name = "mountpoint-s3-client"
-version = "0.19.4"
+version = "0.19.5"
 dependencies = [
  "async-io",
  "async-lock",
@@ -2562,7 +2562,7 @@ dependencies = [
 
 [[package]]
 name = "mountpoint-s3-crt"
-version = "0.13.4"
+version = "0.13.5"
 dependencies = [
  "anyhow",
  "clap",
--- a/mountpoint-s3-client/CHANGELOG.md
+++ b/mountpoint-s3-client/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Unreleased
+## Unreleased (v0.19.5)
 
 ## v0.19.4 (October 30, 2025)
 
--- a/mountpoint-s3-client/Cargo.toml
+++ b/mountpoint-s3-client/Cargo.toml
@@ -1,14 +1,14 @@
 [package]
 name = "mountpoint-s3-client"
 # See `/doc/PUBLISHING_CRATES.md` to read how to publish new versions.
-version = "0.19.4"
+version = "0.19.5"
 edition = "2024"
 license = "Apache-2.0"
 repository = "https://github.com/awslabs/mountpoint-s3"
 description = "High-performance Amazon S3 client for Mountpoint for Amazon S3."
 
 [dependencies]
-mountpoint-s3-crt = { path = "../mountpoint-s3-crt", version = "0.13.4" }
+mountpoint-s3-crt = { path = "../mountpoint-s3-crt", version = "0.13.5" }
 
 async-trait = "0.1.89"
 auto_impl = "1.3.0"
--- a/mountpoint-s3-crt/Cargo.toml
+++ b/mountpoint-s3-crt/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "mountpoint-s3-crt"
 # See `/doc/PUBLISHING_CRATES.md` to read how to publish new versions.
-version = "0.13.4"
+version = "0.13.5"
 edition = "2024"
 license = "Apache-2.0"
 repository = "https://github.com/awslabs/mountpoint-s3"
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -4,7 +4,6 @@
 * Remove redundant cache merics in logs. ([#1716](https://github.com/awslabs/mountpoint-s3/pull/1716), [#1721](https://github.com/awslabs/mountpoint-s3/pull/1721))
 * Rename cache metrics for consistency. ([#1721](https://github.com/awslabs/mountpoint-s3/pull/1721))
 * Add cache metrics for OTLP export. ([#1724](https://github.com/awslabs/mountpoint-s3/pull/1724))
-* Improve memory limiter accuracy when no disk cache is used. ([#1707](https://github.com/awslabs/mountpoint-s3/pull/1707))
 
 ## v1.21.0 (Oct 30, 2025)
 
--- a/mountpoint-s3/Cargo.toml
+++ b/mountpoint-s3/Cargo.toml
@@ -8,7 +8,7 @@ default-run = "mount-s3"
 
 [dependencies]
 mountpoint-s3-fs = { path = "../mountpoint-s3-fs", version = "0.8.3" }
-mountpoint-s3-client = { path = "../mountpoint-s3-client", version = "0.19.4" }
+mountpoint-s3-client = { path = "../mountpoint-s3-client", version = "0.19.5" }
 
 anyhow = { version = "1.0.100", features = ["backtrace"] }
 clap = { version = "4.5.48", features = ["derive"] }


mountpoint-s3-client/src/mock_client.rs
```diff
@@ -120,6 +128,7 @@ pub struct MockClient {
     objects: Arc<RwLock<BTreeMap<String, MockObject>>>,
     in_progress_uploads: Arc<RwLock<BTreeSet<String>>>,
     operation_counts: Arc<RwLock<HashMap<Operation, u64>>>,
+    read_window_increment_failed: Arc<AtomicBool>,
```
passaro, line 133: Why do we need a shared flag on the client, rather than independent ones on the response (and backpressure handle)?

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -720,11 +731,25 @@ fn validate_checksum(
...
 }
 
 impl ClientBackpressureHandle for MockBackpressureHandle {
     fn increment_read_window(&mut self, len: usize) {
-        self.read_window_end_offset.fetch_add(len as u64, Ordering::SeqCst);
+        let prev_read_window_end_offset = self.read_window_end_offset.fetch_add(len as u64, Ordering::SeqCst);
+        let read_window_end_offset = prev_read_window_end_offset + len as u64;
+        let relative_read_window_end = read_window_end_offset - self.request_range.start;
+        if read_window_end_offset < self.request_range.end && !relative_read_window_end.is_multiple_of(self.part_size) {
+            tracing::warn!(
```
passaro, line 751: Are we always logging this warning? I.e. regardless of the `fail_on_non_aligned_read_window` setting?

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -783,6 +810,15 @@ impl Stream for MockGetObjectResponse {
     type Item = ObjectClientResult<GetBodyPart, GetObjectError, MockClientError>;
 
     fn poll_next(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        if self.fail_on_non_aligned_read_window && self.read_window_increment_failed.load(Ordering::SeqCst) {
```
passaro: Wouldn't it be simpler to peek inside the `MockBackpressureHandle`, if present, rather than duplicating the failed flag?

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -114,6 +115,9 @@ pub struct PrefetcherConfig {
     /// The maximum distance the prefetcher will seek backwards before resetting and starting a new
     /// S3 request. We keep this much data in memory in addition to any inflight requests.
     pub max_backward_seek_distance: u64,
+    /// Size of the initial request. This request, of size possibly smaller than part_size,
+    /// is made to lower the latency in small-random-reads usage pattern.
+    pub initial_request_size: usize,
```
passaro, line 130: Should we allow to skip the first request? When `initial_request_size == 0` maybe? Or make it optional?

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -400,7 +404,7 @@ where
             range,
             read_part_size,
             preferred_part_size: self.preferred_part_size,
-            initial_read_window_size,
+            initial_read_window_size: self.config.initial_request_size,
```
passaro: This seems confusing now: is this the size of the initial request, or the initial value of the read window for the second request?

mountpoint-s3-client/src/mock_client.rs
vladem: That is unfortunate, yes. This "invalidates" the whole client after a failed check. If we "invalidate" a single request, the errors are just not observable in tests. My suspicion is, that the failure most often occurs in `readahead` reads and kernel just ignores errors on those.

mountpoint-s3-fs/src/prefetch/part_stream.rs
```diff
@@ -224,6 +224,7 @@ impl<Client: ObjectClient + Clone + Send + Sync + 'static> ObjectPartStream<Clie
             max_read_window_size: config.max_read_window_size,
             read_window_size_multiplier: config.read_window_size_multiplier,
             request_range: range.into(),
+            align_read_window: true,
```
sahityadg: Why do we want to make this optional? While the initial read window makes sense, shouldn't we align the read window always?

vladem: We don't align for the caching stream. Alignment is relative to the S3 request start and for the caching stream we don't know it: after X MB were read from the cache the actual S3 request starts. This X is unknown to the `BackpressureController`.

mountpoint-s3-fs/src/prefetch.rs
vladem: It is the size of the initial request and of the initial read window (to pass to `BackpressureLimiter`). If we want to have an option to disable initial request (i.e. set its size to 0), then I think we need 2 options (in `PrefetcherConfig` and `RequestTaskConfig`):

- initial_request_size: usize
- initial_read_window_size: usize

Or alternatively:

- initial_read_window_size: usize
- disable_random_read_optimisation: bool

Do you think it is worth implementing skip of the first request? Which way of configuration seems more clear?

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -17,6 +17,15 @@ pub enum BackpressureFeedbackEvent {
     PartQueueStall,
 }
 
+/// Read window alignment config.
+///
+/// Align only offsets starting from `align_from_offset` (and relative to it).
+#[derive(Copy, Clone, Debug)]
+pub enum ReadWidnowAlignmentConfig {
```
passaro: Typo

```suggestion
pub enum ReadWindowAlignmentConfig {
```

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -102,8 +115,11 @@ pub fn new_backpressure_controller(
         next_read_offset: config.request_range.start,
         request_end_offset: config.request_range.end,
         mem_limiter,
+        read_window_alignment_config: config.read_window_alignment_config,
     };
 
+    tracing::debug!(?controller, "initialising backpressure controller");
```
passaro: This and other new logging entries should be at `trace` level

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -136,8 +152,13 @@ impl BackpressureController {
                 {
                     let new_read_window_end_offset = self
                         .next_read_offset
-                        .saturating_add(self.preferred_read_window_size as u64)
-                        .min(self.request_end_offset);
+                        .saturating_add(self.preferred_read_window_size as u64);
+                    let new_read_window_end_offset =
```
passaro: tiny nit: I find name shadowing a bit confusing here, could we use descriptive names for the intermediate values? e.g. `preferred_*`, `aligned_*`

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -231,6 +254,29 @@ impl BackpressureController {
...
         }
     }
+
+    /// Round up `read_window_end` to next part boundary (relative to request start).
+    ///
+    /// This function ensures that read window end offsets are aligned to part boundaries for the second request,
+    /// which helps optimize memory usage.
+    ///
+    /// Note: window excludes the last byte, denoted by `read_window_end`.
+    fn round_up_to_part_boundary(read_window_end: u64, config: ReadWidnowAlignmentConfig) -> u64 {
```
passaro: nit: this could be a method on `ReadWindowAlignmentConfig`

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -17,6 +17,15 @@ pub enum BackpressureFeedbackEvent {
     PartQueueStall,
 }
 
+/// Read window alignment config.
+///
+/// Align only offsets starting from `align_from_offset` (and relative to it).
+#[derive(Copy, Clone, Debug)]
+pub enum ReadWidnowAlignmentConfig {
+    On { align_from_offset: u64, part_size: u64 },
+    Off,
```
passaro: Suggestion:

```suggestion
    AlignToPartSize { from_offset: u64, part_size: u64 },
    Disable,
```

mountpoint-s3-fs/src/s3/config.rs
passaro: Rename to `INITIAL_REQUEST_SIZE` and possibly move to the prefetcher module. Also consider where else it is used and whether it still make sense. For example, in benchmarks and tests, we should now set `initial_read_window` on clients to read part size.

mountpoint-s3-fs/src/prefetch/part_stream.rs
```diff
@@ -417,6 +438,8 @@ fn read_from_request<'a, Client: ObjectClient + 'a>(
         let mut client_backpressure_handle = request.backpressure_handle()
             .expect("S3 client backpressure should always be enabled in Mountpoint")
             .clone();
+        // If `initial_read_window` configured on the client is smaller than what `backpressure_limiter` expects, immediately increase the window.
+        // Among other cases, this ensures progress when `initial_read_window` on the client is lower than `part_size` or `initial_request_size`.
```
passaro: I don't understand this comment: isn't progress guaranteed in any cases except when `initial_read_window` is set to `0` (which the client should capture anyway)?

vladem: I've tried to clarify, hope it makes sense.

```rust
// If `initial_read_window` configured on the client is smaller than what
// `backpressure_limiter` expects, immediately increase the window to guarantee progress.
//
// Without this line we could get stuck in `request.next()` when `initial_read_window`
// on the client is lower than the one on the `backpressure_limiter`.
//
// Example: This could happen if `initial_request_size` is 3x larger than `part_size` and
// `initial_read_window` on the client equals to `part_size`.
// In that case, `wait_for_read_window_increment` will return `None`, assuming that the
// window end (set in the `backpressure_limiter`) is also known to the client. However,
// the client will only know about a lower value configured by `initial_read_window`.
```

As an example, without the line below this comment, the following test case fails (with `empty read window` error):

https://github.com/vladem/mountpoint-s3/blob/1fdee96746a4ef13078aa28dec4b096e695f8642/mountpoint-s3-fs/src/prefetch.rs#L1276

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -62,6 +62,19 @@ use part_stream::{PartStream, RequestRange, RequestTaskConfig};
...
+// which is a common IO size. But Linux's readahead will try to read an extra 128k on on
+// top of a 1MiB read, which we'd have to wait for a second request to service. Because
+// FUSE doesn't know the difference between regular reads and readahead reads, it will
+// send us a READ request for that 128k, so we'll have to block waiting for it even if
+// the application doesn't want it. This is all in the noise for sequential IO, but
+// waiting for the readahead hurts random IO. So we add 128k to the first request size
+// to avoid the latency hit of the second request.
+//
+// Note the CRT does not respect this value right now, they always return chunks of part size
+// but this is the first window size we prefer.
```
passaro: We are not using this as initial read window size anymore, so the last paragraph does not apply.

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -106,6 +108,12 @@ impl MockClientConfig {
         self
     }
 
+    /// Enable a check to ensure all read window increments are aligned with part boundaries
```
passaro, line 111: Let's document here that this is client-wide

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -320,6 +344,28 @@ impl ReadWindowIncrementQueue {
...
 }
 
+impl ReadWindowAlignmentConfig {
+    /// Round up `read_window_end` to next part boundary (relative to request start).
+    ///
+    /// This function ensures that read window end offsets are aligned to part boundaries for the second request,
+    /// which helps optimize memory usage.
+    ///
+    /// Note: window excludes the last byte, denoted by `read_window_end`.
+    fn round_up_to_part_boundary(&self, read_window_end: u64) -> u64 {
```
passaro: name and description only apply to the `AlignToPartSize` case. I'd rename to something more generig (e.g. `align`) and possible expand the description in the case definition.

Also, move the impl closer to the type definition.

mountpoint-s3-fs/src/prefetch/caching_stream.rs
```diff
@@ -478,7 +480,7 @@ mod tests {
                 range,
                 read_part_size: client_part_size,
                 preferred_part_size: 256 * KB,
-                initial_read_window_size,
+                initial_request_size: initial_read_window_size,
```
passaro: nit: rename `initial_read_window_size` to `initial_request_size`

mountpoint-s3-client/src/mock_client.rs
vladem: Adding this:

```rust
/// Enable a check to ensure all read window increments are aligned with part boundaries
///
/// Warning: A failed check on one request will result in all other requests from the client to fail.
```

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
vladem: adding this:

```rust
/// Align `read_window_end` to next part boundary relative to `from_offset`. If alignment is disabled, returns
/// the `read_window_end` unchanged.
///
/// Whend enabled, this function ensures that read window end offsets are aligned to part boundaries, which
/// improves accuracy of tracking reserved memory.
///
/// Note: window excludes the last byte, denoted by `read_window_end`.
fn align(&self, read_window_end: u64) -> u64 {
```

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -26,6 +26,29 @@ pub enum ReadWindowAlignmentConfig {
     Disable,
 }
 
+impl ReadWindowAlignmentConfig {
+    /// Align `read_window_end` to next part boundary relative to `AlignToPartSize::from_offset`.
+    /// If alignment is disabled, returns the `read_window_end` unchanged.
+    ///
+    /// Whend enabled, this function ensures that read window end offsets are aligned to part boundaries,
```
passaro: typo: "Whend"

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -491,8 +492,7 @@ mod tests {
     #[test_case(1512, 1000, 512, 1512; "offset exactly at part boundary")]
     #[test_case(1513, 1000, 512, 2024; "offset just past part boundary")]
     fn test_round_up_to_part_boundary(offset: u64, from_offset: u64, part_size: u64, expected: u64) {
```
passaro: nit: rename according to method

You could also have `ReadWindowAlignmentConfig` as an argument instead of `from_offset` and `part_size` (just a suggestion).

passaro (commented):
Looks good. Only minor comments. Shall we include changelog/version changes?

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
vladem: renamed, but with argument it becomes it bit bulky, so inclined to leave as is:

```rust
    #[test_case(ReadWindowAlignmentConfig::AlignToPartSize { from_offset: 1000, part_size: 100 }, 500, 500; "offset before second request start")]
    #[test_case(ReadWindowAlignmentConfig::AlignToPartSize { from_offset: 1000, part_size: 512 }, 1000, 1000; "offset at second request start")]
    #[test_case(ReadWindowAlignmentConfig::AlignToPartSize { from_offset: 1000, part_size: 512 }, 1500, 1512; "offset after second request start, needs alignment")]
    #[test_case(ReadWindowAlignmentConfig::AlignToPartSize { from_offset: 1000, part_size: 512 }, 2024, 2024; "offset after second request start, already aligned")]
    #[test_case(ReadWindowAlignmentConfig::AlignToPartSize { from_offset: 1000, part_size: 512 }, 1001, 1512; "offset just after second request start, needs alignment")]
    #[test_case(ReadWindowAlignmentConfig::AlignToPartSize { from_offset: 1000, part_size: 512 }, 1512, 1512; "offset exactly at part boundary")]
    #[test_case(ReadWindowAlignmentConfig::AlignToPartSize { from_offset: 1000, part_size: 512 }, 1513, 2024; "offset just past part boundary")]
    fn test_read_window_alignment(alignment_config: ReadWindowAlignmentConfig, offset: u64, expected: u64) {
        let result = alignment_config.align(offset);
        assert_eq!(result, expected);
    }
```

mountpoint-s3-crt/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## Unreleased
+## Unreleased (v0.13.5)
```
passaro, line 1: This is because we missed a previous change, right? Could you add it here?

> Add new CRT request metrics to telemetry data (#1701)

mountpoint-s3-fs/Cargo.toml
passaro, line 4: This is the wrong version, it should have been updated to `0.8.4` already.

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
passaro: That's okay. But, fyi, one trick to simplify it could be to add `use ReadWindowAlignmentConfig::AlignToPartSize`.

mountpoint-s3-crt/CHANGELOG.md
vladem: Ack, will add in a follow up PR for crates release.

PR #1714, opened 2025-12-03
Extend autogroup.py to present benchmark output in json format

Priyankakarumuru1:
### What changed and why?
Added --json-output option to export benchmark results in JSON format with separate parameter keys

No breaking changes. Only adds new optional --json-output functionality.

No

---

### DIFF
--- a/benchmark/analysis-scripts/autogroup.py
+++ b/benchmark/analysis-scripts/autogroup.py
@@ -152,13 +152,31 @@ def find_varying_parameters(all_configs: List[Dict[str, Any]]) -> Set[str]:
     return varying
 
 
+def combine_raw_values(all_results: List[Tuple[Dict[str, Any], float, str]]) -> List[Dict[str, Any]]:
+    varying_params = sorted(find_varying_parameters([config for config, _, _ in all_results]))
+
+    grouped_results = defaultdict(list)
+    for config, throughput, iter_num in all_results:
+        key = tuple((param, str(config.get(param, 'N/A'))) for param in sorted(varying_params) if param != 'iteration')
+        grouped_results[key].append(throughput)
+
+    combined_results = []
+    for config_key, throughputs in grouped_results.items():
+        result = {}
+        for param, value in config_key:
+            result[param] = value
+        result['throughputs'] = [round(t, 2) for t in throughputs]
+        result['count'] = len(throughputs)
+        combined_results.append(result)
+
+    return combined_results
+
+
 def main() -> None:
     parser = argparse.ArgumentParser(description='Print benchmark throughput data automatically grouped')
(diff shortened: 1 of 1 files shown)


benchmark/analysis-scripts/autogroup.py
```diff
@@ -209,6 +211,27 @@ def main() -> None:
         row.append(f"{max(throughputs):.2f}")
         aggregated_rows.append(row)
 
+        # Add percentiles
```
sahityadg: The current option already gives us p0 (min), p50 (median) and p100 (max).

What is missing is an option that prints all raw throughput values, not percentiles from all iterations. I'm not sure these percentiles are necessarily useful with the default of 10 iterations.

benchmark/analysis-scripts/autogroup.py
```diff
@@ -156,6 +156,10 @@ def main() -> None:
     parser = argparse.ArgumentParser(description='Print benchmark throughput data automatically grouped')
     parser.add_argument('--base-dir', required=True, help='Base directory containing benchmark results')
     parser.add_argument('--csv-output', help='Optional CSV file to write the results to')
+    parser.add_argument(
+        '--raw-values-csv', help='Optional CSV file to write raw values per iteration for CloudWatch metrics'
```
sahityadg: Can we add an option to write the output in json  and add --json-output or even replace the csv option with json if we think json format is more useful than csv. 

By default, we are presenting the stats like Median, Min, Max etc, in csv. My suggestion is to present the raw values in json,  so that we we can parse whatever stats we want.

benchmark/analysis-scripts/autogroup.py
```diff
@@ -244,6 +248,29 @@ def sort_key(row: List[str]) -> List[Union[int, float, str]]:
...
+    if args.raw_values_csv:
+        raw_data = []
+        for config, throughput, iter_num in all_results:
+            # Create dimension string in a:b:c format
+            dimension_values = []
+            for param in sorted(varying_params):
+                dimension_values.append(str(config.get(param, 'N/A')))
+
+            raw_entry = {
+                "dimensions": ":".join(dimension_values),
```
sahityadg: We should probably use each test parameter as a separate key to make parsing easier.

benchmark/analysis-scripts/autogroup.py
```diff
@@ -244,6 +248,23 @@ def sort_key(row: List[str]) -> List[Union[int, float, str]]:
             writer.writerows(aggregated_rows)
         print(f"\nResults written to CSV: {args.csv_output}")
 
+    # Write raw values JSON if requested
```
sahityadg: Not a blocker for this change. But we should refactor to split the main method and cleanup the script. As this is currently a utility for parsing benchmarks, we can follow up.

PR #1716, opened 2025-12-04
Add a log based metric to capture cache hits for read requests

sahityadg:
The current cache metrics indicate how many from prefetcher were served from the cache. While they are useful for providing prefetcher's cache utilisation, we can't directly map them to see how many FUSE/FS requests were served from cache. As Prefetcher aligns the request size, the metric will capture the cache hits for the adjusted length of the request.

This PR also removes existing cache metrics from the prefetcher as they can be misleading or redundant. 

Yes, removes existing metrics from logs 

Yes

---

### DIFF
--- a/mountpoint-s3-fs/CHANGELOG.md
+++ b/mountpoint-s3-fs/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Unreleased (v0.8.4)
 
-* Add metric to track cache hit rate in logs ([#1716](https://github.com/awslabs/mountpoint-s3/pull/1716))
+* Add metric to track cache hit rate in logs. ([#1716](https://github.com/awslabs/mountpoint-s3/pull/1716))
 * Remove redundant cache merics in logs. ([#1716](https://github.com/awslabs/mountpoint-s3/pull/1716))
 
 ## v0.8.3 (October 30, 2025)
--- a/mountpoint-s3-fs/src/prefetch.rs
+++ b/mountpoint-s3-fs/src/prefetch.rs
@@ -294,10 +294,16 @@ where
 
         match self.try_read(offset, length).await {
             Ok((data, cache_hit)) => {
-                // These map to file system read requests, so record it as fuse metric. We are only
-                // recording when cache_hit is true, as otlp counter doesn't preserve the number of
-                // data points for using avg stat. FIXME: Revisit if this needs to be a histogram.
+                // Record cache hit metric for FUSE layer. We only record a cache hit when ALL parts
+                // for this read request were served from cache storage (disk/express), not from S3.
+                // Partial cache hits (some parts from cache, some from S3) are counted as cache misses
+                // to provide a clear binary metric. Parts served from in-memory prefetch buffers
+                // (from previous S3 requests) don't count as cache hits.
+
                 if !data.is_empty() && cache_hit {
+                    // We only increment the counter on cache_hit=true because OTLP counters don't preserve
+                    // data point counts needed for meaningful averages.
+                    // FIXME: Consider using histogram to capture partial hit ratios.
                     metrics::counter!("fuse.cache_hit").increment(1);
                 }
                 Ok(data)
--- a/mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
+++ b/mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
@@ -224,9 +224,9 @@ rusty_fork_test! {
     #[test]
     fn test_cache_metrics() {
         let recorder = setup_recorder();
+        let cache_dir = tempfile::tempdir().unwrap();
 
         let test_session = mock_session::new_with_cache(|block_size, pool| {
-            let cache_dir = tempfile::tempdir().unwrap();
             let cache_config = DiskDataCacheConfig {
                 cache_directory: cache_dir.path().to_path_buf(),
                 block_size,
@@ -239,8 +239,9 @@ rusty_fork_test! {
         test_session.client().put_object("file.txt", &content).unwrap();
         let path = test_session.mount_path().join("file.txt");
 
-        // First read - populate cache
+        // First read - populate cache and wait for asynchronous cache updates
         File::open(&path).unwrap().read_to_end(&mut Vec::new()).unwrap();
+        sleep(Duration::from_millis(100));
 
         // No cache hits but cache should be updated
         assert!(recorder.get("fuse.cache_hit", &[]).is_none(), "Cache hit metric should not exist after cache miss");
@@ -261,6 +262,7 @@ rusty_fork_test! {
         test_session.client().put_object("large_file.txt", &large_content).unwrap();
         let large_path = test_session.mount_path().join("large_file.txt");
         File::open(&large_path).unwrap().read_to_end(&mut Vec::new()).unwrap();
+        sleep(Duration::from_millis(100));
 
         // Verify eviction metrics
         assert_metric_exists(&recorder, "disk_data_cache.eviction_duration_us", &[]);
@@ -269,9 +271,9 @@ rusty_fork_test! {
     #[test]
     fn test_multi_level_cache_metrics() {
         let recorder = setup_recorder();
+        let cache_dir = tempfile::tempdir().unwrap();
 
         let test_session = mock_session::new_with_cache(|block_size, pool| {
-            let cache_dir = tempfile::tempdir().unwrap();
             let disk_config = DiskDataCacheConfig {
                 cache_directory: cache_dir.path().to_path_buf(),
                 block_size,
@@ -279,13 +281,15 @@ rusty_fork_test! {
             };
             let disk_cache = Arc::new(DiskDataCache::new(disk_config, pool.clone()));
 
-            let express_client = MockClient::config()
-                .bucket("test_bucket".to_string())
-                .part_size(1024 * 1024)
-                .build();
-            let express_config = ExpressDataCacheConfig::new("test_bucket", "test_source")
-                .block_size(block_size);
-            let express_cache = ExpressDataCache::new(express_client, express_config);
+            let express_cache = {
+                let express_client = MockClient::config()
+                    .bucket("test_bucket")
+                    .part_size(1024 * 1024)
+                    .build();
+                let express_config = ExpressDataCacheConfig::new("test_bucket", "test_source")
+                    .block_size(block_size);
+                ExpressDataCache::new(express_client, express_config)
+            };
 
             let runtime = Runtime::new(ThreadPool::builder().pool_size(1).create().unwrap());
             MultilevelDataCache::new(disk_cache, express_cache, runtime)
@@ -300,6 +304,8 @@ rusty_fork_test! {
         sleep(Duration::from_millis(100));
 
         // Verify cache population metrics
+        // Currently, disk_data_cache.read_duration_us is only recorded on cache hits, not misses.
+        // But they are recorded for express.
         assert!(recorder.get("fuse.cache_hit", &[]).is_none(), "Cache hit metric should not exist after cache miss");
         assert_metric_exists(&recorder, "disk_data_cache.write_duration_us", &[]);
         assert_metric_exists(&recorder, "disk_data_cache.total_bytes", &[("type", "write")]);
--- a/mountpoint-s3/CHANGELOG.md
+++ b/mountpoint-s3/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Unreleased (v1.22.0)
 
-* Add metric to track cache hit rate in logs ([#1716](https://github.com/awslabs/mountpoint-s3/pull/1716))
+* Add metric to track cache hit rate in logs. ([#1716](https://github.com/awslabs/mountpoint-s3/pull/1716))
 * Remove redundant cache merics in logs. ([#1716](https://github.com/awslabs/mountpoint-s3/pull/1716))
 
 ## v1.21.0 (Oct 30, 2025)


mountpoint-s3-fs/src/prefetch/caching_stream.rs
```diff
@@ -378,7 +375,13 @@ where
 
 /// Creates a Part that can be streamed to the prefetcher if the given bytes
 /// are in the request range, otherwise return None.
-fn try_make_part(bytes: &ChecksummedBytes, offset: u64, object_id: &ObjectId, range: &RequestRange) -> Option<Part> {
+fn try_make_part(
+    bytes: &ChecksummedBytes,
+    offset: u64,
+    object_id: &ObjectId,
+    range: &RequestRange,
+    from_cache: bool,
```
dannycjones: An enum would be preferable here so its easy to read. Something like `PartSource::Cache`, `PartSource::S3`.

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -353,6 +364,7 @@ where
             debug_assert!(current_task.remaining() > 0);
 
             let part = current_task.read(to_read as usize).await?;
+            cache_hit &= part.is_from_cache();
```
dannycjones: just assign?


```suggestion
            cache_hit = part.is_from_cache();
```

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -291,18 +291,28 @@ where
...
         }
-        result
     }
 
     async fn try_read(
         &mut self,
         offset: u64,
         length: usize,
-    ) -> Result<ChecksummedBytes, PrefetchReadError<Client::ClientError>> {
+    ) -> Result<(ChecksummedBytes, bool), PrefetchReadError<Client::ClientError>> {
```
dannycjones, line 322: Propagate the source here with the same enum we define inside part?

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -311,11 +321,12 @@ where
         // and it can also be used as a lower bound in case the read size is too small.
         // The upper bound is 1MiB since it should be a common IO size.
         let max_preferred_part_size = 1024 * 1024;
+        let mut cache_hit = true;
```
dannycjones: nit: I think it would be more idiomatic to just return the value on line 329 than setup a mutable variable here, and mutate it later. later on, we can just use `let cache_hit = part.is_from_cache();`

mountpoint-s3/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased (v1.22.0)
 
+* Add metric to track cache hit rate and remove redundant cache merics in logs.
```
dannycjones: I would update to this (fine for both this crate and fs crate, since the change is the same).


```suggestion
* Add log-only metric to track cache hit rate.
* Remove redundant log-only cache metrics: `prefetch.blocks_served_from_cache`, `prefetch.blocks_requested_to_client`.
```

- Make it really clear we're only talking about the logged metrics and not OTEL metrics.
- Explicitly name the metrics we're removing. (don't care too much about those internal metrics we're adding)

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -291,18 +291,28 @@ where
...
-        if result.is_err() {
-            self.reset_prefetch_to_offset(offset);
+
+        match self.try_read(offset, length).await {
+            Ok((data, cache_hit)) => {
+                // Record cache hits only if we actually returned data. These map to file system
+                // read requests, so record it as fs metric.
+                if !data.is_empty() && cache_hit {
+                    metrics::counter!("fs.cache_hit").increment(1);
+                }
```
dannycjones: Does it make sense to omit the cache hit in this case? Customers are still going to see a `read` FS request, just with a returned bytes length of 0. Thus if they are doing metric maths like `sum(fs.cache_hit) / count(fs.read_latency_us)` (ignore wrong metric name) then it'll show an artifically high cache hit, even above 100%?

mountpoint-s3-fs/src/prefetch/caching_stream.rs
sahityadg: That was my initial idea. With multi level cache, it turned out to be a bigger change. I intend to follow up on it, but happy to discuss if we want to bring that forward.

mountpoint-s3-fs/src/prefetch.rs
sahityadg: If a request was served from multiple parts, we want to set the cache hit to true only if all parts were from cache.

So we will count requests with partial cache hits as misses and it's a gap with this approach.

mountpoint-s3/CHANGELOG.md
sahityadg: Sure, will fix.

mountpoint-s3-fs/src/prefetch/caching_stream.rs
dannycjones: I'm only suggesting to capture Cache or S3, not which cache. Is it still complex?

mountpoint-s3-fs/src/prefetch.rs
dannycjones: Ah yes, makes sense.

The bit maniupulation threw me off. Maybe this? But I'm not too fussed.


```suggestion
            cache_hit = cache_hit || part.is_from_cache();
```

mountpoint-s3-fs/src/prefetch.rs
dannycjones: This does need to be mutable given we're updating it in a loop later.

I am tempted to suggest we define this closer to the loop and return the literal on line 329, but its up to you.

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -311,11 +321,12 @@ where
         // and it can also be used as a lower bound in case the read size is too small.
         // The upper bound is 1MiB since it should be a common IO size.
         let max_preferred_part_size = 1024 * 1024;
+        let mut cache_hit = true;
         self.preferred_part_size = self.preferred_part_size.max(length).min(max_preferred_part_size);
 
         let remaining = self.size.saturating_sub(offset);
         if remaining == 0 {
-            return Ok(ChecksummedBytes::default());
+            return Ok((ChecksummedBytes::default(), cache_hit));
```
vladem: nit: make it explicitly not a cache hit (code above handles it, but may be a bit misleading)

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -40,7 +40,7 @@ macro_rules! get_histogram {
 }
 
 fn assert_metric_exists(recorder: &TestRecorder, name: &str, labels: &[(&str, &str)]) {
-    let _ = get_metric!(recorder, name, labels);
+    let _: std::sync::Arc<Metric> = get_metric!(recorder, name, labels);
```
vladem, line 50: is type really required here?

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -213,4 +213,90 @@ rusty_fork_test! {
...
+        let content = vec![b'x'; 1024 * 1024];
+        test_session.client().put_object("file.txt", &content).unwrap();
+        let path = test_session.mount_path().join("file.txt");
+
+        // First read - populate cache
+        File::open(&path).unwrap().read_to_end(&mut Vec::new()).unwrap();
+
+        // No cache hits but cache should be updated
+        assert!(recorder.get("fs.cache_hit", &[]).is_none(), "Cache hit metric should not exist after cache miss");
+        assert_metric_exists(&recorder, "disk_data_cache.write_duration_us", &[]);
```
vladem, line 248: may it be flaky? (cache writes are async)

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -213,4 +213,90 @@ rusty_fork_test! {
...
+
+        let test_session = mock_session::new_with_cache(|block_size, pool| {
+            use mountpoint_s3_fs::data_cache::{DiskDataCache, DiskDataCacheConfig};
+            let cache_dir = tempfile::tempdir().unwrap();
+            let cache_config = DiskDataCacheConfig {
+                cache_directory: cache_dir.path().to_path_buf(),
+                block_size,
+                limit: mountpoint_s3_fs::data_cache::CacheLimit::TotalSize { max_size: 2 * 1024 * 1024 },
+            };
+            DiskDataCache::new(cache_config, pool)
```
vladem: is it a multilevel cache? I think you have to use `MultilevelDataCache::new` to create it

mountpoint-s3-fs/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
-## Unreleased
+## Unreleased (v0.8.4)
+
+* Add metric to track cache hit rate and remove redundant cache merics in logs.
```
vladem: nit: add link to the PR

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -291,18 +291,28 @@ where
             next_seq_offset = self.next_sequential_read_offset,
             "read"
         );
-        let result = self.try_read(offset, length).await;
-        if result.is_err() {
-            self.reset_prefetch_to_offset(offset);
+
+        match self.try_read(offset, length).await {
```
passaro, line 295: I'm unconvinced this is the right layer to measure cache hits. All data read here was already prefetched in memory, regardless of where it originally came from. For example. multiple reads at the same offset will count as multiple cache misses even if we only made a single request to S3 and kept the data in RAM.

mountpoint-s3/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## Unreleased (v1.22.0)
 
+* Add metric to track cache hit rate in logs ([#1716](https://github.com/awslabs/mountpoint-s3/pull/1716))
```
dannycjones: nitpick

```suggestion
* Add metric to track cache hit rate in logs. ([#1716](https://github.com/awslabs/mountpoint-s3/pull/1716))
```

same on fs crate

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
dannycjones: I have a similar concern, but see that later tests have sleeps to address this.

Is there anything more reliable we can do (that isn't really complex)? And why do we not think we'll run into it here?

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -213,4 +220,100 @@ rusty_fork_test! {
             "Random access should increment reset_state metric"
         );
     }
+
+    #[test]
+    fn test_cache_metrics() {
+        let recorder = setup_recorder();
+
+        let test_session = mock_session::new_with_cache(|block_size, pool| {
+            let cache_dir = tempfile::tempdir().unwrap();
```
dannycjones: Is this going to behave correctly?

We created the temp directory, use the path in the disk data cache config, but then I expect this to be dropped and the temporary directory to be deleted and no longer registered for cleanup.

I think the temporary cache directory should be setup outside of the `mock_session` creation function.

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -213,4 +220,100 @@ rusty_fork_test! {
...
+            };
+            let disk_cache = Arc::new(DiskDataCache::new(disk_config, pool.clone()));
+
+            let express_client = MockClient::config()
+                .bucket("test_bucket".to_string())
+                .part_size(1024 * 1024)
+                .build();
+            let express_config = ExpressDataCacheConfig::new("test_bucket", "test_source")
+                .block_size(block_size);
+            let express_cache = ExpressDataCache::new(express_client, express_config);
```
dannycjones: nit

idiomatically it would be nice to just assign all to `express_cache` at the end within a scope


```suggestion
            let express_cache = {
                let express_client = MockClient::config()
                    .bucket("test_bucket".to_string())
                    .part_size(1024 * 1024)
                    .build();
                let express_config = ExpressDataCacheConfig::new("test_bucket", "test_source")
                    .block_size(block_size);
                ExpressDataCache::new(express_client, express_config)
            };
```

mountpoint-s3-fs/src/prefetch.rs
dannycjones: Given I believe we're OK with this edge cache, can we add this concern to the comment for this metric below?

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
sahityadg: From testing multiple runs, the disk cache metrics seem to be updated immediately. With multi layer one, I had to add delay as the tests were flaky. There is some delay which I'm not sure is to do with express cache or multilayer stuff, will have to look into it separately.

I don't want to add delay when not needed or overcomplicate the setup for the purpose of metrics validation.

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
sahityadg: No, it works just because the cache operations are succeeding before temp dir is cleaned up. Will fix it.

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
dannycjones: We agreed to add it anyway, 100ms. Now addressed.

dannycjones (approved):
LGTM!

I think we missed actually bumping the versions in this PR, but we can do it in the next one. https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONTRIBUTING.md#version-numbers

PR #1718, opened 2025-12-04
Add requester process' PID to file system logs

mansi153:
Add requester's PID to Filesystem logs.
It also adds an entry point "New request" log to all the (supported) filesystem methods as a proxy for tracking incoming requests at FUSER.

This makes it easier to trace requests dispatched to Mountpoint, especially during workflows using multiple customer processes to make requests concurrently for the same inode(s). 

Additionally, the commit adds/re-orders some other fields in the logs (for a few FS methods) to reattain a consistent order of logging request parameters.

Sample log:
```
2025-12-04T14:56:23.330127Z DEBUG ThreadId(11) lookup{req=3 ino=1 name="._." pid=1860}:head_object{id=3 bucket="multinictesting-iad-benchmarksetupbucket07d0221d-jc1kskgzz2gx" key="._."}: mountpoint_s3_client::s3_crt_client::head_object: new request
```

The commit also does some minor refactoring to name unused method parameters more consistent and adhering to Rust guidelines.

No, only (warn-level and higher) logging change.
No breaking changes.

No, and no.
Logging change only.

---

### DIFF
--- a/mountpoint-s3-fs/CHANGELOG.md
+++ b/mountpoint-s3-fs/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * Fix memory limiter ignoring container cgroup memory limits, which could cause out-of-memory issues in memory-constrained containers. ([#1806](https://github.com/awslabs/mountpoint-s3/pull/1806))
 * Add `S3FilesystemConfig::content_type_detection` option to configure automatic content type inference for new uploads. When set to `ContentTypeDetection::Auto`, Mountpoint will infer the `Content-Type` of new objects based on their file extension. ([#1790](https://github.com/awslabs/mountpoint-s3/pull/1790))
-* Add additional debug information to FUSE operation logs including the ID of the process triggering the file system operation. Some FUSE operations have a new log printed where it didn't already have one. ([#1718](https://github.com/awslabs/mountpoint-s3/pull/1718))
+* Add additional debug information to FUSE operation logs including the ID of the process triggering the file system operation. Some FUSE operations have a new log printed where they didn't already have one. ([#1718](https://github.com/awslabs/mountpoint-s3/pull/1718))
 
 ## v0.9.3 (April 28, 2026)
 
--- a/mountpoint-s3-fs/CHANGELOG.md
+++ b/mountpoint-s3-fs/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Fix memory limiter ignoring container cgroup memory limits, which could cause out-of-memory issues in memory-constrained containers. ([#1806](https://github.com/awslabs/mountpoint-s3/pull/1806))
 * Add `S3FilesystemConfig::content_type_detection` option to configure automatic content type inference for new uploads. When set to `ContentTypeDetection::Auto`, Mountpoint will infer the `Content-Type` of new objects based on their file extension. ([#1790](https://github.com/awslabs/mountpoint-s3/pull/1790))
+* Add additional debug information to FUSE operation logs including the ID of the process triggering the file system operation. Some FUSE operations have a new log printed where it didn't already have one. ([#1718](https://github.com/awslabs/mountpoint-s3/pull/1718))
 
 ## v0.9.3 (April 28, 2026)
 
--- a/mountpoint-s3-fs/src/fuse.rs
+++ b/mountpoint-s3-fs/src/fuse.rs
@@ -101,15 +101,15 @@ impl<Client> Filesystem for S3FuseFilesystem<Client>
 where
     Client: ObjectClient + Clone + Send + Sync + 'static,
 {
-    #[instrument(level="warn", skip_all, fields(req=req.unique()))]
-    fn init(&self, req: &Request<'_>, config: &mut KernelConfig) -> Result<(), libc::c_int> {
+    #[instrument(level="warn", skip_all, fields(req=_req.unique(), pid=_req.pid()))]
+    fn init(&self, _req: &Request<'_>, config: &mut KernelConfig) -> Result<(), libc::c_int> {
         if let Some(error_logger) = self.error_logger.as_ref() {
             error_logger.event("mount", MOUNTPOINT_EVENT_READY);
         }
         block_on(self.fs.init(config).in_current_span())
     }
 
-    #[instrument(level="warn", skip_all, fields(req=req.unique(), ino=parent, name=?name))]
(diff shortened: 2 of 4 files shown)


mountpoint-s3-fuser/src/ll/request.rs
dannycjones, line 1: I wouldn't touch `-fuser` unless we are really confident its required. The goal is to minimize the divergence from the upstream crate.

mountpoint-s3-fs/src/fuse.rs
```diff
@@ -101,44 +101,44 @@ impl<Client> Filesystem for S3FuseFilesystem<Client>
 where
     Client: ObjectClient + Clone + Send + Sync + 'static,
 {
-    #[instrument(level="warn", skip_all, fields(req=req.unique()))]
-    fn init(&self, req: &Request<'_>, config: &mut KernelConfig) -> Result<(), libc::c_int> {
+    #[instrument(level="warn", skip_all, fields(req=_req.unique(), pid=_req.pid()))]
+    fn init(&self, _req: &Request<'_>, config: &mut KernelConfig) -> Result<(), libc::c_int> {
```
dannycjones, line 105: Why do we use `_req` here? And elsewhere in the file we're adding the `_` prefix.

Is there something that's highlighting this as a warning/error?

mountpoint-s3-fs/src/fuse.rs
dannycjones: Why add name here, we don't ever emit it?

What's the consequence of adding name when not set?

mountpoint-s3-fuser/src/ll/request.rs
mansi153: Ah, should I PR to the upstream then? I feel it is important to track and make sense of incoming requests especially during situations with multiple (potentially) conflicting operations happening on an inode at the same time.

mountpoint-s3-fs/src/fuse.rs
mansi153: Yes, prefixing unused variables with `_` is standard Rust practice and recommended by the compiler.
The file was an inconsistent state where some methods had this in place and others didn't, and my IDE notified me of it; so it seemed like a good idea to just clean it up as part of this PR.

mountpoint-s3-fs/src/fuse.rs
mansi153: My understanding is we emit the name in a similar format across other FS methods too, so I tried to make it consistent across the board. It would default to a blank string if not set, but I haven't looked into where it gets set. I'll check and remove the `name` from the log if it's not.

mountpoint-s3-fs/src/fuse.rs
dannycjones: Fair enough, IDE is warning. Sounds good then.

Is that RustRover?

mountpoint-s3-fuser/src/ll/request.rs
dannycjones: Do we really need it? We are already instrumenting all the methods in `mountpoint-s3-fs::fuse::S3FuseFilesystem` with PID?

mansi153: Ok, I have undone the changes to `-fuser`. 
Instead, added a "new request" debug log at the point when the Fuse request reaches the `-fs` implementations so we (hopefully) don't have to wait for a log later in the request path to deduce the order of executions.

mountpoint-s3-fs/src/fuse.rs
mansi153: Removed the name field in `readdir`/`readidrplus` since it seems a bit unnecessary to scope in fixing logging in specific methods as part of this PR - although I feel it would be nice to reach a state of consistent logging format for our FS requests.

mountpoint-s3-fs/src/fuse.rs
```diff
@@ -101,44 +101,48 @@ impl<Client> Filesystem for S3FuseFilesystem<Client>
...
         if let Some(error_logger) = self.error_logger.as_ref() {
             error_logger.event("mount", MOUNTPOINT_EVENT_READY);
         }
         block_on(self.fs.init(config).in_current_span())
     }
 
-    #[instrument(level="warn", skip_all, fields(req=req.unique(), ino=parent, name=?name))]
+    #[instrument(level="warn", skip_all, fields(req=req.unique(), ino=parent, name=?name, pid=req.pid()))]
     fn lookup(&self, req: &Request<'_>, parent: InodeNo, name: &OsStr, reply: ReplyEntry) {
+        debug!("New request");
```
passaro, line 114: Have you considered using something like this instead? https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.with_span_events

passaro (commented):
The change looks good. It does need a new version and a short entry in the changelog.

mountpoint-s3-fs/src/fuse.rs
dannycjones: Given that this would impact S3 client spans and potentially libraries like OTLP, I'm inclined not to use it. It's a shame there's not a way to opt-in in the `instrument` macro.

dannycjones (dismissed):
Rebase, and add changelog and version bump. Then I think we can merge.

dannycjones:
Added changelog entries, ensured crate versions are bumped accordingly.

mountpoint-s3-fs/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * Fix memory limiter ignoring container cgroup memory limits, which could cause out-of-memory issues in memory-constrained containers. ([#1806](https://github.com/awslabs/mountpoint-s3/pull/1806))
 * Add `S3FilesystemConfig::content_type_detection` option to configure automatic content type inference for new uploads. When set to `ContentTypeDetection::Auto`, Mountpoint will infer the `Content-Type` of new objects based on their file extension. ([#1790](https://github.com/awslabs/mountpoint-s3/pull/1790))
+* Add additional debug information to FUSE operation logs including the ID of the process triggering the file system operation. Some FUSE operations have a new log printed where it didn't already have one. ([#1718](https://github.com/awslabs/mountpoint-s3/pull/1718))
```
passaro: Shouldn't it be:
```suggestion
* Add additional debug information to FUSE operation logs including the ID of the process triggering the file system operation. Some FUSE operations have a new log printed where they didn't already have one. ([#1718](https://github.com/awslabs/mountpoint-s3/pull/1718))
```

passaro (approved):

PR #1721, opened 2025-12-10
Refactor cache metrics for consistency and completeness

sahityadg:
This PR streamlines cache metrics collection across disk and express cache implementations.

This change renames caching metrics for consistency with other OTLP metrics. This change also captures latency, bytes_transferred from/to cache and errors consistently across both disk and express cache implementations. 

Yes, updates cache metrics in logs

Yes

---

### DIFF
--- a/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
+++ b/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
@@ -21,14 +21,14 @@ use tracing::{trace, warn};
 use crate::checksums::IntegrityError;
 use crate::data_cache::DataCacheError;
 use crate::memory::{BufferKind, PagedPool};
+use crate::metrics::defs::{
+    ATTR_CACHE, CACHE_DISK, CACHE_EVICT_LATENCY, CACHE_GET_ERRORS, CACHE_GET_IO_SIZE, CACHE_GET_LATENCY,
+    CACHE_PUT_ERRORS, CACHE_PUT_IO_SIZE, CACHE_PUT_LATENCY, CACHE_TOTAL_SIZE,
+};
 use crate::object::ObjectId;
 use crate::sync::Mutex;
 
 use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheResult};
-use crate::metrics::defs::{
-    ATTR_CACHE, ATTR_ERROR, CACHE_DISK, CACHE_EVICT_LATENCY, CACHE_GET_ERRORS, CACHE_GET_IO_SIZE, CACHE_GET_LATENCY,
-    CACHE_PUT_ERRORS, CACHE_PUT_IO_SIZE, CACHE_PUT_LATENCY, CACHE_TOTAL_SIZE,
-};
 
 /// Disk and file-layout versioning.
 const CACHE_VERSION: &str = "V2";
@@ -465,8 +465,7 @@ impl DataCache for DiskDataCache {
         let start = Instant::now();
         let block_key = DiskBlockKey::new(cache_key, block_idx);
         let path = self.get_path_for_block_key(&block_key);
-        let read_result = self.read_block(&path, cache_key, block_idx, block_offset);
-        let result = match read_result {
+        let result = match self.read_block(&path, cache_key, block_idx, block_offset) {
             Ok(None) => {
                 // Cache miss.
                 Ok(None)
@@ -481,7 +480,7 @@ impl DataCache for DiskDataCache {
             }
             Err(err) => {
                 // Invalid block.
-                metrics::counter!(CACHE_GET_ERRORS, ATTR_CACHE => CACHE_DISK, ATTR_ERROR => err.reason()).increment(1);
+                metrics::counter!(CACHE_GET_ERRORS, ATTR_CACHE => CACHE_DISK).increment(1);
                 Err(err)
             }
         };
@@ -506,6 +505,8 @@ impl DataCache for DiskDataCache {
         let path = self.get_path_for_block_key(&block_key);
         trace!(?cache_key, ?path, "new block will be created in disk cache");
 
+        // Capture the put operation result separately from metrics recording
+        // to ensure we can record both success and error metrics consistently
         let put_result = (|| -> DataCacheResult<()> {
             let block = DiskBlock::new(cache_key, block_idx, block_offset, bytes).map_err(|err| match err {
                 DiskBlockCreationError::IntegrityError(_e) => DataCacheError::InvalidBlockContent,
@@ -533,18 +534,13 @@ impl DataCache for DiskDataCache {
             Ok(())
         })();
 
-        let result = match put_result {
-            Ok(()) => {
-                metrics::counter!(CACHE_PUT_IO_SIZE, ATTR_CACHE => CACHE_DISK).increment(bytes_len as u64);
-                Ok(())
-            }
-            Err(err) => {
-                metrics::counter!(CACHE_PUT_ERRORS, ATTR_CACHE => CACHE_DISK, ATTR_ERROR => err.reason()).increment(1);
-                Err(err)
-            }
-        };
+        if put_result.is_ok() {
+            metrics::counter!(CACHE_PUT_IO_SIZE, ATTR_CACHE => CACHE_DISK).increment(bytes_len as u64);
+        } else {
+            metrics::counter!(CACHE_PUT_ERRORS, ATTR_CACHE => CACHE_DISK).increment(1);
+        }
         metrics::histogram!(CACHE_PUT_LATENCY, ATTR_CACHE => CACHE_DISK).record(start.elapsed().as_micros() as f64);
-        result
+        put_result
     }
 
     fn block_size(&self) -> u64 {
--- a/mountpoint-s3-fs/src/metrics/defs.rs
+++ b/mountpoint-s3-fs/src/metrics/defs.rs
@@ -44,7 +44,6 @@ pub const CACHE_OVERSIZED_OBJECTS: &str = "cache.oversized_objects";
 // Attribute constants
 pub const ATTR_FUSE_REQUEST: &str = "fuse_request";
 pub const ATTR_CACHE: &str = "cache";
-pub const ATTR_ERROR: &str = "error";
 
 // Cache type constants
 pub const CACHE_DISK: &str = "disk";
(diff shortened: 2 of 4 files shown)


mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -461,29 +465,28 @@ impl DataCache for DiskDataCache {
         let start = Instant::now();
         let block_key = DiskBlockKey::new(cache_key, block_idx);
         let path = self.get_path_for_block_key(&block_key);
-        match self.read_block(&path, cache_key, block_idx, block_offset) {
+        let read_result = self.read_block(&path, cache_key, block_idx, block_offset);
+        let result = match read_result {
```
vladem: `read_result` vs `result` -- a bit confusing naming, may be keep self.read_block inlined in match?

```suggestion
        let result = match self.read_block(&path, cache_key, block_idx, block_offset) {
```

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -461,29 +465,28 @@ impl DataCache for DiskDataCache {
...
                     usage.lock().unwrap().refresh(&block_key);
                 }
                 Ok(Some(bytes))
             }
             Err(err) => {
-                // Invalid block. Count as cache miss.
-                metrics::counter!("disk_data_cache.block_hit").increment(0);
-                metrics::counter!("disk_data_cache.block_err").increment(1);
+                // Invalid block.
+                metrics::counter!(CACHE_GET_ERRORS, ATTR_CACHE => CACHE_DISK, ATTR_ERROR => err.reason()).increment(1);
```
vladem: what is the example value of `err.reason()`? may it result in high cardinality, e.g. may it contain S3 key? what about error in express cache case?

mountpoint-s3-fs/src/metrics/defs.rs
```diff
@@ -29,8 +29,26 @@ pub const PROCESS_MEMORY_USAGE: &str = "process.memory_usage";
 
 pub const PREFETCH_RESET_STATE: &str = "prefetch.reset_state";
 
+// Cache metric constants
```
vladem, line 32: nit: add `fuse.cache_hit` to the list of constants

mountpoint-s3-fs/src/metrics/defs.rs
```diff
@@ -29,8 +29,26 @@ pub const PROCESS_MEMORY_USAGE: &str = "process.memory_usage";
...
+pub const CACHE_GET_IO_SIZE: &str = "cache.get_io_size";
```
vladem, line 33: `io_size` is somewhat misleading, here it means total data read or written, while in other place `fuse.io_size` means the size of the request, we should help users distinguish these two. I would suggest `cache.get_total`.

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -497,37 +500,51 @@ impl DataCache for DiskDataCache {
...
+            Ok(()) => {
+                metrics::counter!(CACHE_PUT_IO_SIZE, ATTR_CACHE => CACHE_DISK).increment(bytes_len as u64);
+                Ok(())
+            }
+            Err(err) => {
+                metrics::counter!(CACHE_PUT_ERRORS, ATTR_CACHE => CACHE_DISK, ATTR_ERROR => err.reason()).increment(1);
+                Err(err)
+            }
+        };
+        metrics::histogram!(CACHE_PUT_LATENCY, ATTR_CACHE => CACHE_DISK).record(start.elapsed().as_micros() as f64);
```
vladem, line 542: acknowledging that now latency will include `temp_file.persist(path).map_err(|e| e.error)?;` operation and will be recorded for failed writes, nice!

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
```diff
@@ -304,22 +308,21 @@ rusty_fork_test! {
...
+        assert_metric_exists(&recorder, CACHE_GET_LATENCY, &[(ATTR_CACHE, CACHE_DISK)]);
+        assert_metric_exists(&recorder, CACHE_PUT_LATENCY, &[(ATTR_CACHE, CACHE_EXPRESS)]);
+        assert_metric_exists(&recorder, CACHE_PUT_IO_SIZE, &[(ATTR_CACHE, CACHE_EXPRESS)]);
+        assert_metric_exists(&recorder, CACHE_GET_LATENCY, &[(ATTR_CACHE, CACHE_EXPRESS)]);
 
         // Second read should hit disk cache after async write completes
         File::open(&path).unwrap().read_to_end(&mut Vec::new()).unwrap();
         sleep(Duration::from_millis(100));
 
         assert_metric_exists(&recorder, "fuse.cache_hit", &[]);
```
vladem, line 324: acknowledging that we should look at `CACHE_GET_IO_SIZE` metric if we wanted to check, which cache is utilised (where "hits" come from)

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -25,6 +25,10 @@ use crate::object::ObjectId;
 use crate::sync::Mutex;
 
 use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheResult};
+use crate::metrics::defs::{
+    ATTR_CACHE, ATTR_ERROR, CACHE_DISK, CACHE_EVICT_LATENCY, CACHE_GET_ERRORS, CACHE_GET_IO_SIZE, CACHE_GET_LATENCY,
+    CACHE_PUT_ERRORS, CACHE_PUT_IO_SIZE, CACHE_PUT_LATENCY, CACHE_TOTAL_SIZE,
+};
```
dannycjones: nit: this should go in the imports above along with the other `use crate::`.

mountpoint-s3-fs/src/metrics/defs.rs
sahityadg: Even for fuse.io_size, it captures bytes transferred in read or write requests. We use this to indicate bytes transferred in either put and get block, so they are kind of the same in my opinion. But as it is internal, we can revisit it.

mountpoint-s3-fs/src/metrics/defs.rs
sahityadg: yeah, in the next PR

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
sahityadg: Good point,  my understanding is we should see these reasons: https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-fs/src/data_cache.rs#L30-L42

For now, we can remove this and follow up separately.

mountpoint-s3-fs/tests/fuse_tests/metrics_test.rs
sahityadg: or the sample count of latency metrics if you want to look at all attempts.

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -497,37 +500,51 @@ impl DataCache for DiskDataCache {
...
-        metrics::histogram!("disk_data_cache.write_duration_us").record(write_start.elapsed().as_micros() as f64);
-        metrics::counter!("disk_data_cache.total_bytes", "type" => "write").increment(bytes_len as u64);
-        if let Some(usage) = &self.usage {
-            let mut usage = usage.lock().unwrap();
-            _ = temp_file.persist(path).map_err(|e| e.error)?;
-            usage.add(block_key, size);
-        } else {
-            _ = temp_file.persist(path).map_err(|e| e.error)?;
-        }
+        let put_result = (|| -> DataCacheResult<()> {
```
dannycjones, line 510: I don't think it's obvious why we've put this into closure and then invoke it. Can you add a comment?

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -497,37 +500,51 @@ impl DataCache for DiskDataCache {
...
+                metrics::counter!(CACHE_PUT_IO_SIZE, ATTR_CACHE => CACHE_DISK).increment(bytes_len as u64);
+                Ok(())
+            }
+            Err(err) => {
+                metrics::counter!(CACHE_PUT_ERRORS, ATTR_CACHE => CACHE_DISK, ATTR_ERROR => err.reason()).increment(1);
+                Err(err)
+            }
+        };
+        metrics::histogram!(CACHE_PUT_LATENCY, ATTR_CACHE => CACHE_DISK).record(start.elapsed().as_micros() as f64);
+        result
```
dannycjones: simplify?

```suggestion
        match &put_result {
            Ok(()) => {
                metrics::counter!(CACHE_PUT_IO_SIZE, ATTR_CACHE => CACHE_DISK).increment(bytes_len as u64);
            }
            Err(ref err) => {
                metrics::counter!(CACHE_PUT_ERRORS, ATTR_CACHE => CACHE_DISK, ATTR_ERROR => err.reason()).increment(1);
            }
        };
        metrics::histogram!(CACHE_PUT_LATENCY, ATTR_CACHE => CACHE_DISK).record(start.elapsed().as_micros() as f64);
        put_result
```

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
dannycjones: It's a set of strings. Really these should live in metrics `defs.rs` but if we are happier dropping it, that sounds good too.

https://github.com/awslabs/mountpoint-s3/blob/1c38582c03599351257ef8bd9b26dc374760b3fe/mountpoint-s3-fs/src/data_cache.rs#L46-L55

I think we should drop in both disk and Express if we do drop it though.

mountpoint-s3-fs/src/metrics/defs.rs
dannycjones: We'll go ahead and merge this, @vladem. If you don't feel its resolved, let me know.

passaro: Agree with @vladem, the naming is not consistent: `fuse.io_size` is a histogram tracking size of requests, this is a counter for the total get size.

sahityadg: My intent was to make this a histogram but missed in this PR. We should be able to sum if we want to look at the overall bytes transferred.

This is still an internal metric in logs. If we are not convinced with the name, happy to revisit.

PR #1728, opened 2025-12-16
Fix internal race condition in the incremental upload queue

passaro:
This change addresses an issue that caused the `upload::incremental::tests::test_append_failure_on_object_replaced` to fail [occasionally](https://github.com/awslabs/mountpoint-s3/actions/runs/20229391488/job/58068496627#step:8:576). The root cause was a race condition in `AppendUploadQueue` when attempting to write more data after a failure to retrieve the checksum algorithm, which would occasionally result in a panic instead of returning an `UploadAlreadyTerminated` error.

The issue could be fixed with a simple change (see first commit), but I opted to refactor the queue to use a single channel to return both the checksum algorithm and the PutObject responses, making error handling more uniform. 

**Note that there is no user impact**, since further writes after an error are not allowed by the "file system" layer.

No.

No.

---

### DIFF
--- a/mountpoint-s3-fs/src/upload/incremental.rs
+++ b/mountpoint-s3-fs/src/upload/incremental.rs
@@ -503,6 +503,7 @@ impl AsRef<[u8]> for ReservedBuffer {
 #[cfg(test)]
 mod tests {
     use std::collections::HashMap;
+    use std::time::Duration;
 
     use crate::mem_limiter::MINIMUM_MEM_LIMIT;
     use crate::memory::PagedPool;
@@ -515,7 +516,8 @@ mod tests {
     use mountpoint_s3_client::failure_client::{CountdownFailureConfig, countdown_failure_client};
     use mountpoint_s3_client::mock_client::{MockClient, MockObject};
     use mountpoint_s3_client::types::{ChecksumAlgorithm, ETag, GetObjectParams, GetObjectResponse};
-    use test_case::test_case;
+    use test_case::{test_case, test_matrix};
+    use tokio::time::sleep;
 
     fn new_uploader_for_test<Client>(
         client: Client,
@@ -998,7 +1000,8 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_append_failure_on_object_replaced() {
+    #[test_matrix([true, false], [Duration::ZERO, Duration::from_secs(1)])]
+    async fn test_append_failure_on_object_replaced(replace_before_start: bool, sleep_before_write: Duration) {
         let bucket = "bucket";
         let key = "hello";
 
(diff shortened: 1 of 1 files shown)


mountpoint-s3-fs/src/upload/incremental.rs
```diff
@@ -169,41 +181,42 @@ where
...
+            checksum_algorithm: None,
             pool,
             mem_limiter,
-            checksum_algorithm,
             _task_handle: task_handle,
         }
     }
 
     // Push given bytes with its checksum to the upload queue
     pub async fn push(&mut self, buffer: UploadBuffer) -> Result<(), UploadError<Client::ClientError>> {
```
vladem, line 218: do methods of `AppendUploadQueue` need to be `pub`?

mountpoint-s3-fs/src/upload/incremental.rs
```diff
@@ -306,43 +347,39 @@ where
...
-    response_sender: Sender<Result<PutObjectResult, UploadError<Client::ClientError>>>,
-) where
+    buffer_receiver: &Receiver<UploadBuffer>,
+    event_sender: &Sender<AppendUploadEvent<Client::ClientError>>,
+) -> Result<(), UploadError<Client::ClientError>>
+where
     Client: ObjectClient + Send + Sync + 'static,
 {
+    // Always send the checksum algorithm first.
+    let checksum_algorithm = get_checksum_algorithm(&client, &params).await?;
```
vladem, line 362: > The root cause was a race condition in AppendUploadQueue when attempting to write more data after a failure to retrieve the checksum algorithm, which would occasionally result in a panic instead of returning an UploadAlreadyTerminated error.

Why failure to retrieve the checksum algorithm happens? Is it because we sometimes detect etag mismatch on `HeadObject` (instead of `PutObject`)?

mountpoint-s3-fs/src/upload/incremental.rs
```diff
@@ -133,6 +132,19 @@ where
     }
 }
 
+/// Events emitted by the background queue.
+#[derive(Debug)]
+enum AppendUploadEvent<E> {
+    /// The algorithm used to compute checksums.
+    /// Only emitted once, before any [AppendUploadEvent::PutResponse] event.
+    ChecksumAlgorithm(Option<ChecksumAlgorithm>),
```
vladem, line 140: Couldn't we receive checksum "synchronously" in `AppendUploadQueue::new`? Probably that would mean a delay on `open`, is that the reason for making this request in `run_append_loop`?

mountpoint-s3-fs/src/upload/incremental.rs
```diff
@@ -234,8 +247,24 @@ where
...
+                    self.checksum_algorithm.insert(checksum_algorithm).clone()
+                }
+                AppendUploadEvent::PutResponse(_) => {
+                    unreachable!("cannot receive a response event before the checksum algorithm");
+                }
+                AppendUploadEvent::Error(upload_error) => return Err(upload_error),
+            }
+        };
         while self.requests_in_queue > 0 {
             match UploadBuffer::try_new(capacity, &checksum_algorithm, &self.pool, self.mem_limiter.clone())? {
```
vladem, line 269: not related: is there a limit for how much a single "append" file handle can reserve for its buffers? the total amount of free RAM is the only limit?

mountpoint-s3-fs/src/upload/incremental.rs
passaro: Yes, we do not want to add latency to `open` (not changed in this PR).

mountpoint-s3-fs/src/upload/incremental.rs
passaro: Not really. All the methods called by `AppendUploadRequest` are marked `pub` as opposed to `consume_next_response` which is internal, but the visibility would only be enforced if we moved it to a separate module.

mountpoint-s3-fs/src/upload/incremental.rs
passaro: Yes, there is a limit of 2GiB (`MAX_BYTES_IN_QUEUE`) enforced by the bounded buffer channel. Writes will block if the channel is full until a buffer is popped from the queue and its data uploaded.

mountpoint-s3-fs/src/upload/incremental.rs
passaro: Yes. And the error was handled and propagated correctly on the first write attempt. The issue prior to this change was that a subsequent write (on the `AppendUploadRequest`) would still try to get the algorithm and panic.

vladem (approved):

PR #1733, opened 2025-12-19
Fix internal failure on atomic upload

passaro:
Improve handling of errors on `CreateMultiPartUpload` in the atomic upload code path. Similarly to the change in #1728, the issue only manifests when attempting to further write or complete an upload after an error and it does not affect Mountpoint file system users, since that's already prevented at that level.

No, user-visible behavior not impacted.

No.

---

### DIFF
--- a/mountpoint-s3-client/src/failure_client.rs
+++ b/mountpoint-s3-client/src/failure_client.rs
@@ -435,8 +435,10 @@ pub fn countdown_failure_client<Client: ObjectClient>(
                 },
                 result_fn: |state| {
                     state.count += 1;
-                    if state.count >= state.fail_count {
-                        Err(state.error.take().unwrap())
+                    if state.count >= state.fail_count
+                        && let Some(error) = state.error.take()
+                    {
+                        Err(error)
                     } else {
                         Ok(())
                     }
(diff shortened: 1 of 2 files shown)


mountpoint-s3-client/src/failure_client.rs
```diff
@@ -435,8 +435,10 @@ pub fn countdown_failure_client<Client: ObjectClient>(
                 },
                 result_fn: |state| {
                     state.count += 1;
-                    if state.count >= state.fail_count {
-                        Err(state.error.take().unwrap())
+                    if state.count >= state.fail_count
+                        && let Some(error) = state.error.take()
```
vladem, line 439: so we will succeed on all next steps after taking the error? nit: document in `CountdownFailureConfig` (if not yet)

passaro: That is indeed what we currently document:

```
    /// For PUT, map entries are interpreted as follows (operations are numbered starting at 1):
    ///   (k -> Err(E) means return error E on the k'th PUT
    ///   (k -> Ok((n, E))) means return a put request object on the k'th put that
    ///       returns error E on the n'th write, otherwise writes to the underlying request.
```

We should improve it, maybe with something similar to `GetObjectFailureMode`. But that's a separate change.

PR #1735, opened 2026-01-05
Propose allowing unlink for LocalUnopened files

Maxbey:
**What changed and why?**
We encountered an issue when files became permanently stuck and could not be deleted. This happens when a process crashes (or is killed) between the `mknod` and `open` FUSE calls. 

The file is created in `LocalUnopened` state, but since the process never opens it for writing, there's no way to complete or cancel the write, and unlink becomes blocked if we try to delete the stale file.

Previously, unlinking any local file (LocalUnopened, LocalOpen, or PendingRename) was blocked with `unlink on local file not allowed until write is complete`. However, files in LocalUnopened state have never been opened for writing, so no S3 upload has started. These files exist only in the local inode tree and can be removed without S3 cleanup.

This change separates the handling of LocalUnopened from LocalOpen/PendingRename, allowing unlink to succeed for files that were created but never opened.

Yes. Previously, attempting to unlink a LocalUnopened file returned an error. Now it succeeds.

Yes.

---

### DIFF
--- a/mountpoint-s3-fs/src/superblock/inode.rs
+++ b/mountpoint-s3-fs/src/superblock/inode.rs
@@ -286,6 +286,36 @@ impl InodeKindData {
             },
         }
     }
+
+    pub fn remove_child(&mut self, name: &str, expected_ino: InodeNo) {
+        match self {
+            InodeKindData::File {} => {
+                debug_assert!(false, "remove_child called on a file, not a directory");
+            }
+            InodeKindData::Directory {
+                children,
+                writing_children,
+                ..
+            } => {
+                // We want to remove the original child.
+                // If there is a mismatch in inode number between the existing inode and the deleted inode, we invalidate the existing inode's stat.
+                // If for some reason the child with the specified name was already removed, we do nothing as this is the desired state.
+                if let Some(existing_inode) = children.get(name) {
+                    if existing_inode.ino() == expected_ino {
+                        children.remove(name);
+                        writing_children.remove(&expected_ino);
+                    } else {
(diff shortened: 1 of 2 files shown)


mountpoint-s3-fs/src/superblock/inode.rs
```diff
@@ -286,6 +286,36 @@ impl InodeKindData {
...
+                // We want to remove the original child.
+                // If there is a mismatch in inode number between the existing inode and the deleted inode, we invalidate the existing inode's stat.
+                // If for some reason the child with the specified name was already removed, we do nothing as this is the desired state.
+                if let Some(existing_inode) = children.get(name) {
+                    if existing_inode.ino() == expected_ino {
+                        children.remove(name);
+                        writing_children.remove(&expected_ino);
+                    } else {
+                        // Mismatch in inode number, thus the existing inode might not be the same one we deleted
+                        let mut state = existing_inode.get_mut_inode_state_no_check();
```
vladem: this branch will deadlock? we may have already acquired write lock on child in `unlink`

mountpoint-s3-fs/src/superblock/inode.rs
```diff
@@ -286,6 +286,36 @@ impl InodeKindData {
             },
         }
     }
+
+    pub fn remove_child(&mut self, name: &str, expected_ino: InodeNo) {
+        match self {
+            InodeKindData::File {} => {
+                debug_assert!(false, "remove_child called on a file, not a directory");
```
vladem: should return an error in the release build

PR #1737, opened 2026-01-07
Fix dependencies for analyzing results from benchmarks

Priyankakarumuru1:
**What changed and why?**

Added analysis dependency group to pyproject.toml containing omegaconf and tabulate packages required by the autogroup.py script for processing benchmark results.

No 

No 

---

### DIFF
--- a/benchmark/pyproject.toml
+++ b/benchmark/pyproject.toml
@@ -12,8 +12,6 @@ dependencies = [
 [dependency-groups]
 dev = [
     "pytest",
-]
-analysis = [
     "omegaconf",
     "tabulate",
 ]
--- a/benchmark/pyproject.toml
+++ b/benchmark/pyproject.toml
@@ -13,3 +13,7 @@ dependencies = [
 dev = [
     "pytest",
 ]
+analysis = [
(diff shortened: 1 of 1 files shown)


benchmark/pyproject.toml
```diff
@@ -13,3 +13,7 @@ dependencies = [
 dev = [
     "pytest",
 ]
+analysis = [
```
sahityadg: Have you tried adding these to dev dependency-group like pytest? 

That may be better as it won't need --group flag.

Priyankakarumuru1: That is actually much clear. I'll move omegaconf and tabulate to the dev group instead of creating a separate analysis group. Thanks for the suggestion!

sahityadg: These dependencies are already defined here: https://github.com/awslabs/mountpoint-s3/blob/main/benchmark/analysis-scripts/autogroup.py#L3-L6. So uv should just install them anyway. We don't need them in pyproject.toml unless we want to move all dependencies here.

dannycjones (approved):

PR #1746, opened 2026-01-15
Add `max_background` setting to `S3FilesystemConfig`

vladem:
Add `max_background` setting to `S3FilesystemConfig`.

No.

Yes, new version of `mountpoint-s3-fs`.

---

### DIFF
--- a/mountpoint-s3-fs/examples/mount_from_config.rs
+++ b/mountpoint-s3-fs/examples/mount_from_config.rs
@@ -106,6 +106,7 @@ impl ConfigOptions {
     fn build_filesystem_config(&self) -> Result<S3FilesystemConfig> {
         let mut fs_config = S3FilesystemConfig {
             cache_config: CacheConfig::new(mountpoint_s3_fs::fs::TimeToLive::Indefinite),
+            max_background_fuse_requests: self.max_background_fuse_requests,
             ..Default::default()
         };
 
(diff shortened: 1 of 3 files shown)


mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -85,6 +85,9 @@ struct ConfigOptions {
 
     /// Disk cache configuration
     disk_cache: Option<DiskCacheConfig>,
+
+    /// Limits the number of concurrent FUSE requests that the kernel may send, default: 64
+    max_background: Option<u16>,
```
muddyfish: `max_background_fuse_requests`? current name is not easy to understand.

mountpoint-s3-fs/src/fs.rs
```diff
@@ -202,12 +203,17 @@ where
                 old
             );
         } else {
-            const DEFAULT_MAX_BACKGROUND: u16 = 64;
-            let max_background_result = config.set_max_background(DEFAULT_MAX_BACKGROUND);
-            if max_background_result.is_err() {
+            let max_background_result = config.set_max_background(self.config.max_background_fuse_requests);
```
passaro: With this approach, the env variable will always override the value in configuration, even when explicitly set by the user. Shouldn't it be the other way around?

mountpoint-s3-fs/src/fs.rs
```diff
@@ -202,12 +203,17 @@ where
...
+            if let Ok(old) = max_background_result {
+                tracing::debug!(
+                    "Successfully overridden FUSE max_background configuration to {} (was {}) from config.",
```
passaro: "overridden" here may be confusing, e.g. when applying the default value

PR #1747, opened 2026-01-15
Prefetch Benchmark: Add simple random read

yerzhan7:
**What changed and why?** 

Need to extend prefetcher benchmark to test random read access pattern in addition to existing sequential read.

### Manual Testing

Successfully run benchmark: `uv run benchmark.py -- benchmark_type=prefetch`

```
Results Summary:
+-----------------------+------------------+--------------------------------------+---------+-----------------+------------------+--------------+--------------+
|   application_workers | benchmark_type   | benchmarks.prefetch.access_pattern   |   Count |   Median (Gbps) |   Std Dev (Gbps) |   Min (Gbps) |   Max (Gbps) |
+=======================+==================+======================================+=========+=================+==================+==============+==============+
|                     1 | prefetch         | random                               |      10 |            0.02 |             0    |         0.02 |         0.02 |
+-----------------------+------------------+--------------------------------------+---------+-----------------+------------------+--------------+--------------+
|                     1 | prefetch         | sequential                           |      10 |            8.38 |             0.25 |         8.16 |         8.85 |
+-----------------------+------------------+--------------------------------------+---------+-----------------+------------------+--------------+--------------+
|                    16 | prefetch         | random                               |      10 |            0.3  |             0.03 |         0.23 |         0.33 |
+-----------------------+------------------+--------------------------------------+---------+-----------------+------------------+--------------+--------------+
|                    16 | prefetch         | sequential                           |      10 |            4.1  |             0.6  |         3.01 |         4.82 |
+-----------------------+------------------+--------------------------------------+---------+-----------------+------------------+--------------+--------------+
```

No. Changing Benchmark scripts only.

No. Changing Benchmark scripts only.

---

### DIFF
--- a/mountpoint-s3-fs/examples/prefetch_benchmark.rs
+++ b/mountpoint-s3-fs/examples/prefetch_benchmark.rs
@@ -308,6 +308,7 @@ fn main() -> anyhow::Result<()> {
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)] // Acceptable for benchmark utility: some parameters only used by random access pattern
 async fn wait_for_download(
     mut request: PrefetchGetObject<S3CrtClient>,
     object_id: ObjectId,


mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -93,6 +101,22 @@ pub struct CliArgs {
...
+    )]
+    access_pattern: AccessPattern,
+
+    #[arg(
+        long,
+        help = "Random seed for pseudorandom read offsets (used with random access pattern)",
+        default_value_t = 1,
+        value_name = "SEED"
+    )]
+    randseed: u64,
```
passaro, line 120: We should define the random behavior in more details:
* Do we apply the same random access sequence to all objects and all iterations? (I'd say no)
* Should the sequence(s) be reproducible across platforms?
* What's the behavior when no seed is specified?

mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -275,15 +308,39 @@ async fn wait_for_download(
...
+                let bytes_read = bytes.len() as u64;
+                offset += bytes_read;
+                total_bytes_read += bytes_read;
+            }
+            Ok(total_bytes_read)
+        }
+        AccessPattern::Random => {
+            let mut rng = StdRng::seed_from_u64(randseed);
+            let mut total_bytes_read = 0;
+            let max_offset = size.saturating_sub(read_size);
```
passaro: Why restrict the offset? vs allowing smaller reads at the end of the file?

mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -275,15 +308,39 @@ async fn wait_for_download(
...
+
+            while Instant::now() < timeout {
```
passaro: We are making the random read mode time-based, but using the timeout value. That's not the original intent of the timeout parameter and does not work well with the iterations parameter. We need to either define a separate parameter, or consider a different approach (e.g. no. of reads or total size).

mountpoint-s3-fs/examples/prefetch_benchmark.rs
```diff
@@ -275,15 +308,39 @@ async fn wait_for_download(
...
+            while offset < size && Instant::now() < timeout {
+                let bytes = request.read(offset, read_size as usize).await?;
+                let bytes_read = bytes.len() as u64;
+                offset += bytes_read;
+                total_bytes_read += bytes_read;
+            }
+            Ok(total_bytes_read)
+        }
+        AccessPattern::Random => {
+            let mut rng = StdRng::seed_from_u64(randseed);
```
passaro: `StdRng` is not portable, I think. We probably want to use a different RNG: https://rust-random.github.io/book/guide-rngs.html. See also other comment about the random seed.

mountpoint-s3-fs/examples/prefetch_benchmark.rs
yerzhan7: > Do we apply the same random access sequence to all objects and all iterations? (I'd say no)

Good point. Will add object ID and iteration number to the seed.

> Should the sequence(s) be reproducible across platforms?

Will switch to [PCG64](https://docs.rs/rand_pcg/latest/rand_pcg/type.Pcg64.html) which should be portable and deterministic.

> What's the behavior when no seed is specified?

Currently, the default value for seed is 1. I.e. still deterministic.

mountpoint-s3-fs/examples/prefetch_benchmark.rs
yerzhan7: Will switch to PCG64, thanks!

mountpoint-s3-fs/examples/prefetch_benchmark.rs
yerzhan7: Will fix, thanks

mountpoint-s3-fs/examples/prefetch_benchmark.rs
yerzhan7: I've added `total_bytes_read < size` in addition to timeout.

PR #1749, opened 2026-01-18
Add support for S3-compatible providers

legraml:
This commit enhances Mountpoint for Amazon S3 to better support S3-compatible storage providers such as MinIO, Wasabi, DigitalOcean Spaces, and Backblaze B2.

Changes:
- Add auth_scheme_with_fallback() method to handle custom endpoints that may not provide full AWS endpoint resolution properties
- Update S3 CRT client to use fallback auth scheme for better compatibility with S3-compatible providers
- Add comprehensive documentation section in CONFIGURATION.md with examples for MinIO, Wasabi, DigitalOcean Spaces, and Backblaze B2
- Add note in README.md about S3-compatible provider support
- Add unit test for auth scheme fallback mechanism

The fallback mechanism ensures that when custom endpoints (like MinIO or Wasabi) don't provide full endpoint resolution metadata, Mountpoint falls back to using standard SigV4 signing with the s3 service name and the configured region.

While S3-compatible providers are not officially supported, these changes improve compatibility on a best-effort basis.

**TODO: Remove all TODO messages in desc. For PR description guidance, review [Contributing via Pull Requests](https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONTRIBUTING.md#contributing-via-pull-requests)**.

**TODO: What changed and why?**

**TODO: Please confirm there's no breaking change, or call out any that are made.**

**TODO: Confirm with justification if not required.**

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,9 @@ When you're finished accessing your bucket, you can unmount it (you might need `
 
 Mountpoint for Amazon S3 is designed for high-performance access to the Amazon S3 service. While it may be functional against other storage services that use S3-like APIs, we aren't able to provide support for those use cases, and they may inadvertently break when we make changes to better support Amazon S3. We welcome contributions of minor compatibility fixes or performance improvements for these services if the changes can be tested against Amazon S3.
 
+> [!NOTE]
+> Mountpoint can be used with S3-compatible storage providers such as MinIO, Wasabi, DigitalOcean Spaces, and Backblaze B2. See [Using S3-compatible storage providers](doc/CONFIGURATION.md#using-s3-compatible-storage-providers) in the configuration documentation for examples and setup instructions. Note that these providers are not officially supported and compatibility is provided on a best-effort basis.
+
 ## Contributing
 
 We welcome contributions to Mountpoint for Amazon S3! Please see [CONTRIBUTING.md](doc/CONTRIBUTING.md) for more information on how to report bugs or submit pull requests. We especially welcome contributions to issues tagged as [good first issues to work on](https://github.com/awslabs/mountpoint-s3/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22).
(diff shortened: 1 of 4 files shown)


dannycjones:
hey @legraml, thank you for your contribution.

Can you elaborate a bit more on why the code changes are needed? In particular, what `mount-s3` invocation/arguments are you using and what error do you see when using the third-party object storage?

I ran the following test based on the test added in this PR, but without adding fallbacks, and it still passes.

```rust
#[test]
fn test_auth_scheme_fallback_for_custom_endpoint() {
    // Test that fallback works when using custom endpoints that may not
    // provide full endpoint resolution properties (e.g., MinIO, Wasabi)
    let endpoint_config = EndpointConfig::new("us-east-1")
        .endpoint(mountpoint_s3_client::config::Uri::new_from_str(&mountpoint_s3_client::config::Allocator::default(), "https://minio.example.com").unwrap())
        .addressing_style(AddressingStyle::Path);

    let resolved_endpoint = endpoint_config.resolve_for_bucket("test-bucket").unwrap();

    // Use fallback method which should always succeed
    let auth_scheme = resolved_endpoint.auth_scheme()
        .expect("no error when using MinIO example endpoint");

    // Verify fallback defaults for S3-compatible providers
    assert_eq!(auth_scheme.signing_region(), "us-east-1");
    assert_eq!(auth_scheme.signing_name(), "s3");
    assert_eq!(auth_scheme.scheme_name(), SigningAlgorithm::SigV4);
    assert_eq!(auth_scheme.disable_double_encoding(), true);
}
```

If there is an issue we can resolve with a small fix, we'd be happy to explore adding it but I don't currently understand what is required here.

Thank you also for the addition to the configuration docs. While we won't be able to accept an enumeration of other third-party object storage out there, a section added to those docs would be welcomed - in particular, knowing which flags may be required for compatibility. In the meantime, I'd like to understand what problem you've faced here.

PR #1756, opened 2026-01-25
Add SHA256 checksum support for uploads

srijanshukla18:
## Description

This change extends the `--upload-checksums` CLI option to support SHA256 in addition to the existing CRC32C algorithm.

### Usage
Users can now specify:
- `--upload-checksums=crc32c` (existing default)
- `--upload-checksums=sha256` (new)
- `--upload-checksums=off`

### Changes

**Core changes:**
- Extended `UploadChecksums` enum in CLI to include `Sha256` variant
- Changed `S3FilesystemConfig` to store `ChecksumAlgorithm` instead of boolean
- Added `ChecksumConfig::trailing_sha256()` and `upload_review_sha256()` methods
- Modified `PutObjectTrailingChecksums` enum to carry algorithm information

**Implementation:**
- Updated `put_object.rs` to dispatch to appropriate checksum config based on algorithm
- Updated `atomic.rs` to handle both CRC32C and SHA256 for multi-part uploads
- Fixed mock_client to compute correct checksums for SHA256

### Backward Compatibility
The implementation maintains backward compatibility by defaulting to CRC32C when checksums are enabled without specifying an algorithm.

### Testing
- All existing tests updated to work with new enum structure
- Mock client updated to properly compute SHA256 checksums for parts

---

Related to extending checksum algorithm support in mountpoint-s3.

### DIFF
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -1206,11 +1206,19 @@ impl MockPutObjectRequest {
             .chunks(self.part_size)
             .map(|part| {
                 let size = part.len();
-                let checksum = if !matches!(self.params.trailing_checksums, PutObjectTrailingChecksums::Disabled) {
-                    let checksum = crc32c::checksum(part);
-                    Some(crc32c_to_base64(&checksum))
-                } else {
-                    None
+                let checksum = match self.params.trailing_checksums {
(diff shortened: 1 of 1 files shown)


mountpoint-s3-fs/src/upload/atomic.rs
passaro, line 27: This is still using CRC32C, regardless of the algorithm set in `trailing_checksums`. When trying to use `ChecksumAlgorithm::Sha256`, `verify_checksums` will fail.

PR #1758, opened 2026-01-28
Require initial-window-size when backpressure is enabled in client_benchmark

tadiwa-aizen:
Require read-window-size when backpressure is enabled in client_benchmark.

The client_benchmark was silently defaulting to 0-byte initial read window when `--initial-window-size` was not provided, causing backpressure tests to fail with minimal throughput (0-32 MiB instead of expected GBs).   Merged `--enable-backpressure` and `--initial-window-size` into a single `--backpressure-window-size` argument whose presence triggers backpressure mode.

No, it fixes it.

No

---

### DIFF
--- a/mountpoint-s3-client/examples/client_benchmark.rs
+++ b/mountpoint-s3-client/examples/client_benchmark.rs
@@ -66,10 +66,10 @@ fn run_benchmark(
                             .await
                             .expect("couldn't create get request");
                         let mut backpressure_handle = request.backpressure_handle().cloned();
-                        if let Some(window_size) = backpressure_window_size {
-                            if let Some(backpressure_handle) = backpressure_handle.as_mut() {
-                                backpressure_handle.ensure_read_window(window_size as u64);
-                            }
+                        if let Some(window_size) = backpressure_window_size
+                            && let Some(backpressure_handle) = backpressure_handle.as_mut()
+                        {
+                            backpressure_handle.ensure_read_window(window_size as u64);
                         }
 
                         let mut request = pin!(request);


mountpoint-s3-client/examples/client_benchmark.rs
```diff
@@ -209,8 +209,8 @@ struct CliArgs {
     enable_backpressure: bool,
     #[arg(
         long,
-        help = "Initial read window size in bytes, used to dictate how far ahead we request data from S3",
-        default_value = "0"
+        help = "Initial read window size in bytes, used to dictate how far ahead we request data from S3. Required when --enable-backpressure is set.",
+        required_if_eq("enable_backpressure", "true")
```
passaro: 1. Since it's required when backpressure is enabled, consider merging into a single argument.
2. "Initial" read window size is misleading here: we are actually using a sliding window.

We could have a `--backpressure-window-size` with `Option<usize>`.

benchmark/conf/hydra/sweeper/client_bp.yaml
```diff
@@ -1,3 +1,3 @@
 # @package hydra.sweeper
 params: {}
-  #'benchmarks.client_bp.read_window_size': 2147483648, 8589934592, 17179869184  # 2GB, 8GB, 16GB
\ No newline at end of file
+  'benchmarks.client_bp.read_window_size': 2147483648 # 8589934592, 17179869184  2GB, 8GB, 16GB
```
passaro: Clean up the comments. E.g.
```suggestion
  'benchmarks.client_bp.read_window_size': 2147483648 # (2GiB)
```

passaro (approved):
Looks good!

Issue #44, opened 2023-01-13
Support for mounting using fstab


dannycjones:
Some users of file connectors may rely on fstab to manage mounting file systems in their Linux operating system.

More info on fstab: https://www.redhat.com/sysadmin/etc-fstab

StarlightSyndrome:
This is somehow an additional info/request for implementing this feature.

I just have setup mount-s3 with systemd automount with a light wrapper script that does the actual mounting. A line in fstab would look like this

`my-s3-bucket /s3/my-s3-bucket fuse.mount-s3.sh noauto,x-systemd.automount,--read-only,--allow-other`

where the /usr/local/bin/mount-s3.sh script extracts the source target and the double dashed options and calls mount-s3 with it.

While this approach works well with autofs it does not with systemd automount since the latter already creates a mount on the target:

```
$ cat /proc/mounts
...
systemd-1 on /s3/my-s3-bucket type autofs (rw,relatime,fd=55,pgrp=1,timeout=0,minproto=5,maxproto=5,direct)
...
```

Subsequently mount-s3 fails on the actual mounting during triggered by accessing the mount point with: 
`Error: mount point /s3/my-s3-bucket is already mounted.`

So I think we need a switch to turn of this check.

Also I suggest that either we could filter the (comma separated) options passed by -o for those options with double dashes. Or we introduce our own parameter that can hold all our options like so:

`my-s3-bucket /s3/my-s3-bucket fuse.mount-s3.sh noauto,x-systemd.automount,mount-s3-options=read-only,allow-other,max-threads=24`

This parameter has to be put at the end of the the fstab options so everything following the `mount-s3-options=` can be considered mount-s3 arguments and parameters.

Regarding the "already mounted check" it could be disabled implicitly with the -o parameter but it would be nice to have a separate flag for this behaviour e.g. when creating systemd automounter services other than by fstab.

muddyfish:
Support for mounting using fstab is now possible as of [Mountpoint 1.18.0](https://github.com/awslabs/mountpoint-s3/releases/tag/mountpoint-s3-1.18.0)!

See the docs on how to configure it: https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#automatically-mounting-an-s3-bucket-at-boot

PR #1762, opened 2026-02-03
mountpoint-s3-fs: allow mounting on top of autofs

kiron1:
Allow mounting on a directory if it is already a mountpoint as long as it is of type `autofs`.
The fs_type autofs is used by autofs (and therefor SystemD automount units) to enable the automatic mount functionallity.

Only slightly, it enables now to use mount-s3 in combination with auto mount.

user @StarlightSyndrome mentions this problem already in #44, but no solution was provided so far.

Should be added: mount-s3 can now be used with autofs.

---

### DIFF
--- a/mountpoint-s3-fs/CHANGELOG.md
+++ b/mountpoint-s3-fs/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+* Allow mounting on top of `autofs` managed directories. ([#1762](https://github.com/awslabs/mountpoint-s3/pull/1762))
 * Update to latest S3 client. ([#1793](https://github.com/awslabs/mountpoint-s3/pull/1793))
 
 ## v0.9.2 (March 20, 2026)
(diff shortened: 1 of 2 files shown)


kiron1:
@renanmagagnin why did the CI fail previously: https://github.com/awslabs/mountpoint-s3/actions/runs/21640434812/job/62378152093

All jobs are skipped and therefore *Verify jobs succeeded* failed, seems unrelated to this PR.

renanmagagnin:
Hi @kiron1, sorry for the delay and thank you for working on this change! Before we merge, would you be able to add a `CHANGELOG.md` entry and tests? We would like to validate:
1. Should not reject a directory that appears in `mountinfo` as `autofs`, and the mount is functional
2. Should still reject a directory that is actively mounted with a non-autofs fs type

Note: the CI fails are unrelated to this PR

renanmagagnin:
Thank you for addressing the feedback. I have also verified the changes with a manual test and should be enough to merge. As a follow-up, we would like to extend the FSTAB tests with a validation for this change. 

Note: I resolved the small conflict in the `CHANGELOG.md` to unblock running the CI.

muddyfish:
Verified manually with `... ... mount-s3 _netdev,nosuid,nodev,allow-other,nofail,profile=mp,noauto,x-systemd.automount` in the fstab file and it worked

muddyfish (approved):

PR #1767, opened 2026-02-09
s3io_benchmark: Add test object generator

yerzhan7:
**What changed and why?** Currently, `s3io_benchmark` requires s3 object to exist in the bucket (for `read` jobs) which is inconvenient.

- Added ability to automatically generate new test objects during initialization for read jobs.
- Uses separate S3 CRT/Uploader client stack to not influence benchmark jobs. 
  - Didn't use Rust AWS SDK as it's slower than CRT
  - Didn't use Rust Transfer Manager as it's not GA yet
- Added new config flag `generate_object` with default value of `true`.
- Manually tested generating 100GB object.
- Automatically overwrite write part size if it exceeds 10K max parts S3 MPU limit
- Do not perform upload if object with correct size already exists

Only benchmark which is currently not used anywhere.

No - benchmark script only.

---

### DIFF
--- a/mountpoint-s3-fs/examples/s3io_benchmark/config.rs
+++ b/mountpoint-s3-fs/examples/s3io_benchmark/config.rs
@@ -4,6 +4,8 @@ use std::path::Path;
 use std::time::Duration;
 use thiserror::Error;
 
+use crate::test_object_generator::generate_test_objects;
+
 /// Top-level configuration structure
 #[derive(Debug, Clone, Deserialize, PartialEq)]
 pub struct Config {
@@ -198,7 +200,7 @@ pub async fn prepare_jobs(config: Config) -> Result<Vec<ResolvedJobConfig>, Conf
         }
     }
 
-    crate::test_object_generator::generate_test_objects(&resolved_jobs, &config.global)
+    generate_test_objects(&resolved_jobs, &config.global)
         .await
         .map_err(|e| ConfigError::Validation(format!("Object generation failed: {}", e)))?;
 
--- a/mountpoint-s3-fs/examples/s3io_benchmark/test_object_generator.rs
+++ b/mountpoint-s3-fs/examples/s3io_benchmark/test_object_generator.rs
@@ -1,8 +1,6 @@
-use std::error::Error;
-use std::fmt;
-
 use mountpoint_s3_client::ObjectClient;
 use mountpoint_s3_client::types::HeadObjectParams;
+use thiserror::Error;
 
 use crate::config::{GlobalConfig, ResolvedJobConfig, WorkloadType};
 use crate::executor::Executor;
@@ -11,22 +9,13 @@ use crate::executor::Executor;
 /// https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html
 const MAX_PARTS: u64 = 10_000;
 
-#[derive(Debug)]
+#[derive(Debug, Error)]
 pub enum ObjectGenerationError {
+    #[error("Setup failed: {0}")]
     Setup(String),
+    #[error("Upload failed for '{key}': {reason}")]
     Upload { key: String, reason: String },
 }
-
-impl fmt::Display for ObjectGenerationError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            Self::Setup(msg) => write!(f, "Setup failed: {}", msg),
-            Self::Upload { key, reason } => write!(f, "Upload failed for '{}': {}", key, reason),
-        }
-    }
-}
-
-impl Error for ObjectGenerationError {}
 
 // Note: This intentionally creates a separate Executor instance to ensure
 // test object generation doesn't influence benchmark jobs


mountpoint-s3-fs/examples/s3io_benchmark/config.rs
```diff
@@ -194,7 +198,9 @@ pub fn prepare_jobs(config: Config) -> Result<Vec<ResolvedJobConfig>, ConfigErro
         }
     }
 
-    // TODO: Generate/upload test objects for read workloads
+    crate::test_object_generator::generate_test_objects(&resolved_jobs, &config.global)
```
passaro: nit: prefer "use" rather than fully qualify

mountpoint-s3-fs/examples/s3io_benchmark/test_object_generator.rs
```diff
@@ -0,0 +1,168 @@
...
+
+use crate::config::{ChecksumAlgorithm, GlobalConfig, ResolvedJobConfig, SseType, WorkloadType};
+
+#[derive(Debug)]
+pub enum ObjectGenerationError {
+    Setup(String),
+    Upload { key: String, reason: String },
+}
+
+impl fmt::Display for ObjectGenerationError {
```
passaro: nit: you can use `thiserror` to derive the `Display` impl

mountpoint-s3-fs/examples/s3io_benchmark/test_object_generator.rs
```diff
@@ -0,0 +1,168 @@
...
+
+impl Error for ObjectGenerationError {}
+
+// Note: This intentionally creates a separate S3 client stack from the Executor to ensure
+// test object generation doesn't influence benchmark jobs
+pub async fn generate_test_objects(
+    jobs: &[ResolvedJobConfig],
+    global: &GlobalConfig,
+) -> Result<(), ObjectGenerationError> {
+    let jobs_requiring_generation: Vec<&ResolvedJobConfig> = jobs
```
passaro, line 26: Are we re-uploading objects even when they already exist? Does that make sense as a default?

Have we considered only doing it if HeadObject fails / returns the "wrong" object size?

mountpoint-s3-fs/examples/s3io_benchmark/test_object_generator.rs
```diff
@@ -0,0 +1,168 @@
...
+        eprintln!(
+            "Generated test object for job '{}': key={}, size={} bytes",
+            job.name, job.object_key, job.object_size
+        );
+    }
+
+    Ok(())
+}
+
+fn create_uploader(global: &GlobalConfig) -> Result<Uploader<S3CrtClient>, ObjectGenerationError> {
```
passaro: Does this duplicates most of `Executor::new`? Could we reuse that (e.g. using a separate `Executor` instance)?

mountpoint-s3-fs/examples/s3io_benchmark/test_object_generator.rs
```diff
@@ -0,0 +1,168 @@
...
+    let region = global.region.as_deref().unwrap_or("us-east-1");
+    let write_part_size = global.write_part_size.unwrap_or(8 * 1024 * 1024);
```
passaro: We may need to be careful here: depending on the object size, we may need to set a larger write part size.

yerzhan7: Yeah, but should be fine if it would be a one-off thing as I will add check to skip uploading if object already exists.

mountpoint-s3-fs/examples/s3io_benchmark/test_object_generator.rs
yerzhan7: Yup, will create separate Executor instance. Thanks for reviewing!

mountpoint-s3-fs/examples/s3io_benchmark/test_object_generator.rs
yerzhan7: Yeah, we can do that, thanks for the suggestion.

mountpoint-s3-fs/examples/s3io_benchmark/test_object_generator.rs
passaro: My concern is that with the default part size, the maximum upload is limited to 80,000 MiB. If users define a job where that limit is not respected, the upload will fail eventually. We should at the very least check that and fail early. But also consider whether we should automatically increase the write part size (just for this executor) if required.

yerzhan7: Ah, sorry, I initially thought that CRT has some logic to adhere to the limits automatically (based on this comment), but looks like it's not working.

https://github.com/awslabs/mountpoint-s3/blob/6976976e401b6147b72ebc3eec4c891fd4a3bca2/mountpoint-s3-crt/src/s3/client.rs#L190-L191

I will add some logic to override write part size just for the object generator.

mountpoint-s3-fs/examples/s3io_benchmark/test_object_generator.rs
yerzhan7: Fixed, thanks!

mountpoint-s3-fs/examples/s3io_benchmark/config.rs
yerzhan7: Fixed!

mountpoint-s3-fs/examples/s3io_benchmark/test_object_generator.rs
```diff
@@ -43,14 +32,54 @@ pub async fn generate_test_objects(
...
         );
-        upload_test_object(&uploader, &job.bucket, &job.object_key, job.object_size, job.write_size).await?;
+    }
+
+    let executor = Executor::new(&adjusted_global).map_err(|e| ObjectGenerationError::Setup(e.to_string()))?;
+    for job in jobs_requiring_generation {
+        // Skip generation if object already exists with correct size
+        match executor
+            .client
+            .head_object(&job.bucket, &job.object_key, &HeadObjectParams::new())
```
passaro, line 57: nice-to-have: now we are potentially duplicating HeadObject for each job. It would be nice if we did it just once in the prepare phase and recorded size and etag for use in the read benchmark.

mountpoint-s3-fs/examples/s3io_benchmark/test_object_generator.rs
```diff
@@ -60,81 +89,15 @@ pub async fn generate_test_objects(
...
 async fn upload_test_object(
-    uploader: &Uploader<S3CrtClient>,
+    executor: &Executor,
     bucket: &str,
     key: &str,
     size: u64,
     write_size: usize,
 ) -> Result<(), ObjectGenerationError> {
-    let mut request = uploader
+    let mut request = executor
```
passaro, line 99: nice-to-have: this is very similar to `Executor::execute_multipart_upload_iteration`. We could do without the duplication.

PR #1768, opened 2026-02-17
Make `MemoryPool` buffer acquisition async-ready

yerzhan7:
**What changed and why?**

This PR makes the `MemoryPool` trait async-capable by adding a new `get_buffer_async()` method, and wires up the CRT bridge so that buffer acquisition on the read path is spawned on a CRT event loop. (Note that on write-path we must keep doing `get_buffer()` in a sync way due to CRT limitation [here](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_meta_request.c#L2618-L2631))

Currently, when the S3 client needs a memory buffer, it calls `get_buffer` and blocks until the buffer is allocated. This works fine now because allocation is instant. But we want to support a future scenario where the pool is full and needs to wait for memory to free up. Adding `get_buffer_async` lays the groundwork — the actual behavior doesn't change much (allocations are still instant, but they now execute on a separate Task/Future on the read path), but the infrastructure is in place.

The async future is spawned on existing CRT's `EventLoopGroup`. Note that there is a timing issue: we create the `CrtBufferPoolFactory` (wrapping `MemoryPoolFactory`) when building `S3ClientConfig`, but the `EventLoopGroup` doesn't exist yet — it's created inside `S3CrtClientInner::new()`. So the factory must be constructed before the thread pool it needs to reference. The solution is to create wrapper `MemoryPoolFactoryWrapper`.

- Get buffer on read-path is now fulfilled on a separate spawned Task/Future instead of immediately in a sync way.
- `MemoryPool` trait has now `'static + Send` bounds

Yes

---

### DIFF
--- a/mountpoint-s3-crt/src/s3/client.rs
+++ b/mountpoint-s3-crt/src/s3/client.rs
@@ -218,8 +218,8 @@ impl ClientConfig {
     ///
     /// When not set, the client will use the default pool with the configured memory limit.
     pub fn buffer_pool_factory(&mut self, pool_factory: CrtBufferPoolFactory) -> &mut Self {
-        // SAFETY: `pool_factory` is stored in `self.pool_factory` below, keeping it alive
-        // for as long as the client config (and thus the CRT client) exists.
+        // SAFETY: `pool_factory` is stored in `self.pool_factory` below so that it
+        // remains alive until the client is initialized.
         let (factory_fn, user_data) = unsafe { pool_factory.as_inner() };
         self.inner.buffer_pool_factory_fn = factory_fn;
         self.inner.buffer_pool_user_data = user_data;
--- a/mountpoint-s3-crt/src/s3/pool.rs
+++ b/mountpoint-s3-crt/src/s3/pool.rs
@@ -164,7 +164,6 @@ impl CrtBufferPoolFactory {
     /// Builds a CRT buffer pool factory from a type-erased wrapper and an [EventLoopGroup].
     ///
     /// The wrapper provides the type-erased factory pointer and C callback.
-    /// The [EventLoopGroup] is stored directly and passed to each [`CrtBufferPool`].
     pub fn new(wrapper: MemoryPoolFactoryWrapper, event_loop_group: EventLoopGroup) -> Self {
         Self(Box::new(CrtBufferPoolFactoryInner {
             event_loop_group,


mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -15,19 +16,34 @@ use mountpoint_s3_crt_sys::{
...
 
     /// Get a buffer of at least the requested size.
     fn get_buffer(&self, size: usize, meta_request_type: MetaRequestType) -> Self::Buffer;
 
+    /// Get a buffer of at least the requested size asynchronously.
+    /// Returns a future that resolves to the buffer.
+    ///
+    /// The default implementation wraps [`get_buffer`](Self::get_buffer) in a
+    /// ready future. Override this when the pool needs to wait for memory to free up.
+    fn get_buffer_async(
```
passaro, line 43: Could we simplify and only require the async method? The caller could be responsible for blocking on it when required (`can_block` case).

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -440,6 +440,7 @@ impl S3CrtClientInner {
         client_config.memory_limit_in_bytes(config.memory_limit_in_bytes);
 
         if let Some(pool_factory) = &config.buffer_pool_factory {
+            pool_factory.set_event_loop_group(event_loop_group.clone());
```
passaro: This will misbehave if we reuse the configuration for multiple clients. We need a better approach.

mountpoint-s3-crt/src/s3/pool.rs
passaro: Since this type now also needs the ELG from the client, we could carve out a "MemoryPoolFactoryWrapper" that just erases the implementing type and can be stored in `S3ClientConfig`. We would then only create a `CrtBufferPoolFactory` when initializing a new client.

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -250,42 +293,28 @@ impl<Pool: MemoryPool> CrtBufferPool<Pool> {
...
+            // The write path in s3_meta_request.c expects the ticket to be fulfilled
+            // synchronously — it treats an unfulfilled future as error.
+            let buffer = self.pool.get_buffer(size, meta_request_type);
+            let ticket = CrtTicket::new(buffer, ticket_vtable);
+            future.set(ticket);
+        } else {
+            // Read path: spawn on background thread, return unfulfilled future.
+            let pool = self.pool.clone();
+            let future_clone = future.clone();
+            let _handle = self.event_loop_group.spawn_future(async move {
```
passaro, line 349: What happens on cancellation? Do `aws_future`s even provide a cancellation mechanism?

mountpoint-s3-crt/src/s3/pool.rs
yerzhan7: Are you proposing something like this inside `reserve()` method?

```
if can_block {
    let buffer = futures::executor::block_on(self.pool.get_buffer_async(size, meta_request_type));
    ...
} else {
    // spawn, .await inside the spawned task
}
```

I am a bit worried that this may introduce some deadlock scenarios.

passaro: Yes, that's the idea. But you're right, let's consider what could block.

Also consider:
* we should probably pass on the `can_block` flag (or equivalent) to the `MemoryPool` 
* we can also discuss alternatives to `can_block` with the CRT (longer term)

yerzhan7: > we should probably pass on the can_block flag (or equivalent) to the MemoryPool

The deadlock risk will still be there, right? We just move it into MemoryPool. 

I think it's fine for now to have 2 explicit methods - sync and async. We can refactor trait later as it's still marked as experimental anyway.

> we can also discuss alternatives to can_block with the CRT (longer term)

Agree.

mountpoint-s3-crt/src/s3/pool.rs
yerzhan7: I have looked at CRT code, and looks like currently CRT does not provide cancellation mechanism.

We may want it in order to avoid doing wasteful work allocating unneeded buffer (in case of meta_request cancellation). However, that over-allocated buffer will be immediately dropped. Not sure yet what's the best approach would be for tracking/doing cancellations (need discussion with CRT team). 

I think for now it should be fine. I will add comment about this.

Also, I've checked what would happen if future/ticket is cancelled by CRT - there won't be memory leak: buffer (and ticket) will be properly destroyed: 

https://github.com/awslabs/aws-c-io/blob/main/source/future.c#L323-L324


---

Also, I checked whether we (in Rust) should cancel/fail the future/ticket in the future MemoryPool implementation (in case of being over memory limit) and I think we should not because it will result in user I/O error (due to `AWS_ERROR_S3_BUFFER_ALLOCATION_FAILED`). CRT default memory pool also never fails the future. (it uses though different mechanism for memory limit enforcement with ticket queue)

mountpoint-s3-crt/src/s3/pool.rs
yerzhan7: Added `MemoryPoolFactoryWrapper` type in https://github.com/awslabs/mountpoint-s3/pull/1768/commits/10f50e174d08ac27c8c77c71c45f458aa3998b54

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -94,51 +98,101 @@ impl MemoryPoolFactoryOptions {
...
-    /// Set the [EventLoopGroup] to be passed to CrtBufferPool instances created by this factory.
+    /// Returns the factory pointer.
+    pub(crate) fn factory_ptr(&self) -> NonNull<libc::c_void> {
+        self.0.factory_ptr
+    }
+
+    /// Returns the C callback function pointer.
+    pub(crate) fn factory_fn(&self) -> aws_s3_buffer_pool_factory_fn {
+        self.0.factory_fn
+    }
```
passaro: These should not be required.

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -78,47 +98,111 @@ impl MemoryPoolFactoryOptions {
...
+}
+
+/// Factory used by [Client](`super::client::Client`) to create CRT wrappers for [MemoryPool] implementations.
+#[derive(Debug, Clone)]
+pub struct CrtBufferPoolFactory(Arc<CrtBufferPoolFactoryInner>);
+
+#[derive(Debug)]
+struct CrtBufferPoolFactoryInner {
+    /// Copied from the wrapper for direct access in the C callback
```
passaro: Why do we need to copy? We could just use the wrapper.

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -78,47 +98,111 @@ impl MemoryPoolFactoryOptions {
...
+    factory_fn: aws_s3_buffer_pool_factory_fn,
+    event_loop_group: EventLoopGroup,
+    /// Keeps the wrapper alive so its `Drop` impl frees the factory pointer exactly once.
+    _wrapper: MemoryPoolFactoryWrapper,
+}
+
+// SAFETY: `CrtBufferPoolFactoryInner` is safe to transfer across threads because it wraps a [MemoryPoolFactory] implementation that is [Send].
+unsafe impl Send for CrtBufferPoolFactoryInner {}
+// SAFETY: `CrtBufferPoolFactoryInner` is safe to share across threads because it wraps a [MemoryPoolFactory] implementation that is [Sync].
+unsafe impl Sync for CrtBufferPoolFactoryInner {}
```
passaro: No need for these once we remove the copied fields.

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -78,47 +98,111 @@ impl MemoryPoolFactoryOptions {
...
+        }))
+    }
+
     /// Returns the factory callback and user_data pointer to pass to the CRT.
+    ///
+    /// The user_data pointer points to the `CrtBufferPoolFactoryInner` so that the
+    /// C callback can access both the pool factory and the [EventLoopGroup].
+    /// The caller MUST keep this `CrtBufferPoolFactory` alive for as long as the CRT
+    /// may invoke the callback.
     pub(crate) fn as_inner(&self) -> (aws_s3_buffer_pool_factory_fn, *mut ::libc::c_void) {
```
passaro: Pre-existing, but this should be unsafe.

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -78,47 +98,111 @@ impl MemoryPoolFactoryOptions {
...
+
+    /// Returns the C callback function pointer.
+    pub(crate) fn factory_fn(&self) -> aws_s3_buffer_pool_factory_fn {
+        self.0.factory_fn
+    }
+}
+
+/// Factory used by [Client](`super::client::Client`) to create CRT wrappers for [MemoryPool] implementations.
+#[derive(Debug, Clone)]
+pub struct CrtBufferPoolFactory(Arc<CrtBufferPoolFactoryInner>);
```
passaro: Does this need to be `Clone`? Consider simplifying and removing the Arc<Inner>.

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -78,47 +98,111 @@ impl MemoryPoolFactoryOptions {
...
+
+#[derive(Debug)]
+struct CrtBufferPoolFactoryInner {
```
passaro, line 157: Do we need to make this not `Unpin`?

yerzhan7: Discussed offline: Using `Box` for now, no need for `!Unpin`.

mountpoint-s3-crt/src/s3/pool.rs
yerzhan7: Removed in https://github.com/awslabs/mountpoint-s3/pull/1768/changes/8c589668f020ecf7a41d31e60e0e4c5ef3d713a5

mountpoint-s3-crt/src/s3/pool.rs
yerzhan7: Removed in https://github.com/awslabs/mountpoint-s3/pull/1768/changes/8c589668f020ecf7a41d31e60e0e4c5ef3d713a5

mountpoint-s3-crt/src/s3/pool.rs
yerzhan7: Removed in https://github.com/awslabs/mountpoint-s3/pull/1768/changes/8c589668f020ecf7a41d31e60e0e4c5ef3d713a5

Thanks for reviewing!

mountpoint-s3-crt/src/s3/pool.rs
yerzhan7: Added unsafe in https://github.com/awslabs/mountpoint-s3/pull/1768/changes/8c589668f020ecf7a41d31e60e0e4c5ef3d713a5

mountpoint-s3-crt/src/s3/pool.rs
yerzhan7: Removed `Clone`and made it `Box` instead of `Arc`  in https://github.com/awslabs/mountpoint-s3/pull/1768/changes/8c589668f020ecf7a41d31e60e0e4c5ef3d713a5

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -218,7 +218,9 @@ impl ClientConfig {
     ///
     /// When not set, the client will use the default pool with the configured memory limit.
     pub fn buffer_pool_factory(&mut self, pool_factory: CrtBufferPoolFactory) -> &mut Self {
-        let (factory_fn, user_data) = pool_factory.as_inner();
+        // SAFETY: `pool_factory` is stored in `self.pool_factory` below, keeping it alive
+        // for as long as the client config (and thus the CRT client) exists.
```
passaro: I don't think this is true: it should be while the client is initialized and the factory created rather than for the whole client lifetime (but please double-check!).

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -147,62 +147,44 @@ impl MemoryPoolFactoryWrapper {
...
-// SAFETY: `CrtBufferPoolFactoryInner` is safe to transfer across threads because it wraps a [MemoryPoolFactory] implementation that is [Send].
-unsafe impl Send for CrtBufferPoolFactoryInner {}
-// SAFETY: `CrtBufferPoolFactoryInner` is safe to share across threads because it wraps a [MemoryPoolFactory] implementation that is [Sync].
-unsafe impl Sync for CrtBufferPoolFactoryInner {}
-
 impl CrtBufferPoolFactory {
     /// Builds a CRT buffer pool factory from a type-erased wrapper and an [EventLoopGroup].
     ///
     /// The wrapper provides the type-erased factory pointer and C callback.
     /// The [EventLoopGroup] is stored directly and passed to each [`CrtBufferPool`].
```
passaro: let's fix the doc warning. Also not sure what information the comment is adding.

mountpoint-s3-crt/src/s3/client.rs
yerzhan7: You are right - I've check the code and it looks like it only used during `aws_s3_client_new()`.

I will update the comment to: 

```
// SAFETY: `pool_factory` is stored in `self.pool_factory` below so that it
// remains alive until the client is initialized.
```

mountpoint-s3-crt/src/s3/pool.rs
yerzhan7: Will remove this line, thanks

PR #1769, opened 2026-02-18
Add fast fail override for perfomance benchmarks

tadiwa-aizen:
**What changed and why?**
Added configurable fail-fast mode for benchmark sweeps to enable early termination on failures.

Previously, benchmark sweeps would run all parameter combinations regardless of failures, making it difficult to catch configuration errors early in long-running sweeps. This change
adds a fail_fast parameter to the smart benchmark sweeper that, when enabled, runs benchmarks sequentially and stops immediately upon the first failure. This allows quick 
identification of issues during development without wasting resources on remaining benchmarks.

Implementation details:
- When fail_fast=true: Benchmarks run one at a time (batch_size=1), checking status after each job
- When fail_fast=false (default): All benchmarks run in one batch, maintaining existing behavior
- On failure with fail_fast=true: Re-raises the original exception with full context (stack trace, error message, failed configuration)

No breaking changes. Default behavior (`fail_fast`=false) maintains existing functionality where all benchmarks run regardless of failures.

### Testing

Unit tests in tests/test_execute_batches.py:

- **test_fail_fast_true_stops_on_first_failure**: Verifies that with fail_fast=true, the sweeper stops after the first failed job and doesn't execute remaining benchmarks (validates call count = 2 for 3 jobs when 2nd fails).

- **test_fail_fast_false_continues_through_failures**: Confirms that with fail_fast=false, all benchmarks run in a single batch and failures are captured without stopping execution.

Run tests with:
```bash
cd mountpoint-s3/benchmark
uv run pytest tests/test_execute_batches.py -v  
```

No - this only affects internal benchmark tooling, not the Mountpoint binary or published crates.

---


benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -86,8 +90,22 @@ def sweep(self, arguments: List[str]) -> Any:
...
+
+            for i in range(0, len(all_combinations), batch_size):
+                batch = all_combinations[i : i + batch_size]
+                results = self.launcher.launch(batch, initial_job_idx=initial_job_idx)
+
+                # Check results immediately if fail_fast enabled
+                # Accessing return_value raises an exception if the job failed (hydra/core/utils.py:251-258)
+                if self.fail_fast:
+                    for r in results:
+                        _ = r.return_value  # Raises on failure, stopping the sweep
```
Priyankakarumuru1: What happens if a job fails ? Does the exception message or logs get captured 
clearly so users know which benchmark failed and why?

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -21,15 +21,19 @@ class SmartBenchmarkSweeperConf:
     _target_: str = "hydra_plugins.smart_sweeper.smart_benchmark_sweeper.SmartBenchmarkSweeper"
     max_batch_size: Optional[int] = None
     params: Optional[Dict[str, str]] = None
+    fail_fast: bool = False
```
Priyankakarumuru1, line 35: Is this worth documenting somewhere? Users should know when and why they'd want to enable fail_fast=true. Consider this if it feels reasonable.

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -86,8 +90,22 @@ def sweep(self, arguments: List[str]) -> Any:
         initial_job_idx = 0
         if all_combinations:
             self.validate_batch_is_legal(all_combinations)
-            results = self.launcher.launch(all_combinations, initial_job_idx=initial_job_idx)
-            returns.append(results)
+
+            # Determine batch size: run all at once (fail_fast=False) or one at a time (fail_fast=True)
+            batch_size = 1 if self.fail_fast else len(all_combinations)
```
muddyfish: Why should `fail_fast` impact the batch size? Wouldn't it be faster to run in parallel but cancel on failure?

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
muddyfish: Does accessing `return_value` raise on failure?

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -86,8 +90,22 @@ def sweep(self, arguments: List[str]) -> Any:
...
+                batch = all_combinations[i : i + batch_size]
+                results = self.launcher.launch(batch, initial_job_idx=initial_job_idx)
+
+                # Check results immediately if fail_fast enabled
+                # Accessing return_value raises an exception if the job failed (hydra/core/utils.py:251-258)
+                if self.fail_fast:
+                    for r in results:
+                        _ = r.return_value  # Raises on failure, stopping the sweep
+
+                initial_job_idx += len(batch)
```
muddyfish: Is `initial_job_idx` the same as `i`?

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -86,8 +90,22 @@ def sweep(self, arguments: List[str]) -> Any:
         initial_job_idx = 0
         if all_combinations:
             self.validate_batch_is_legal(all_combinations)
-            results = self.launcher.launch(all_combinations, initial_job_idx=initial_job_idx)
-            returns.append(results)
+
+            # Determine batch size: run all at once (fail_fast=False) or one at a time (fail_fast=True)
+            batch_size = 1 if self.fail_fast else len(all_combinations)
+
+            for i in range(0, len(all_combinations), batch_size):
```
muddyfish: There should be a test for this.

benchmark/tests/test_fail_fast.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # Test default (fail_fast=False)
+        sweeper_default = SmartBenchmarkSweeper()
+        assert not sweeper_default.fail_fast, "Default should be fail_fast=False"
+
+        # Test fail_fast=True
+        sweeper_fast_fail = SmartBenchmarkSweeper(fail_fast=True)
+        assert sweeper_fast_fail.fail_fast
+
+        print("✓ fail_fast configuration test passed")
```
muddyfish: Why do we need a print here?

benchmark/tests/test_fail_fast.py
```diff
@@ -0,0 +1,126 @@
...
+    - fail_fast=True: Combinations run one at a time, stopping on first error
+    """
+
+    def setup_method(self):
+        self.parser = OverridesParser.create()
+
+    def test_fail_fast_configuration(self):
+        """Test that fail_fast defaults to False and can be set to True"""
+
+        # Test default (fail_fast=False)
```
muddyfish: Nit: useless comments

benchmark/tests/test_fail_fast.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def test_fail_fast_affects_batch_size(self):
+        """
+        Test that fail_fast controls batch size:
+        - fail_fast=False → batch_size = all combinations
+        - fail_fast=True → batch_size = 1
+        """
+
+        # Create mock launcher and setup
+        mock_launcher = Mock()
```
muddyfish: Extract to function `build_smart_sweeper`?

benchmark/tests/test_fail_fast.py
```diff
@@ -0,0 +1,126 @@
...
+        # Generate combinations (should be 3: workers=1,2,3)
+        combinations = sweeper_normal._generate_combinations_for_type("fio", parsed)
+        assert len(combinations) == 3, f"Expected 3 combinations, got {len(combinations)}"
+
+        # Mock the sweep to test batching
+        sweeper_normal._load_benchmark_params = Mock(return_value=[])
+        sweeper_normal._extract_benchmark_types = Mock(return_value=['fio'])
+
+        # Test the batching logic in the sweeper
+        all_combinations = combinations
```
muddyfish: Why renaming the variable?

benchmark/tests/test_fail_fast.py
```diff
@@ -0,0 +1,126 @@
...
+        batch_size = 1 if sweeper_normal.fail_fast else len(all_combinations)
```
muddyfish: This isn't testing anything useful.

benchmark/tests/test_fail_fast.py
```diff
@@ -0,0 +1,126 @@
...
+
+        assert batch_size == 3, f"With fail_fast=False, batch_size should be {len(all_combinations)}, got {batch_size}"
+
+        # Test with fail_fast=True (should batch one at a time)
+        sweeper_fast_fail = SmartBenchmarkSweeper(fail_fast=True)
+        batch_size_fast_fail = 1 if sweeper_fast_fail.fail_fast else len(all_combinations)
```
muddyfish: Similarly, this isn't testing anything useful

benchmark/tests/test_fail_fast.py
```diff
@@ -0,0 +1,126 @@
...
+        sweeper_fast_fail = SmartBenchmarkSweeper(fail_fast=True)
+
+        # Test 1: Successful job should not raise - return_value returns normally
+        mock_result_success = Mock()
+        mock_result_success.return_value = None  # Success
+
+        results = [mock_result_success]
+
+        # Accessing return_value on a successful job should NOT raise an exception
+        try:
```
muddyfish: This is not useful.

muddyfish (changes requested):

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
tadiwa-aizen: Our `SmartBenchmarkSweeper` uses Hydra's BasicLauncher by default, which executes jobs sequentially 
not in parallel. With `fail_fast=True` and `batch_size=1`, the sweeper passes one job at a 
time to the launcher, gets the result back immediately, and can check if it failed before launching the next job. If we used `batch_size=len(all_combinations)`, the launcher would run all jobs sequentially and only return after completing every single one, meaning we'd waste time running jobs after a failure instead of stopping early. This fits our requirements.

(For your parralel jobs question..: Hydra does support parallel launchers like 
JoblibLauncher, but they don't support canceling in-flight jobs on first failure—they run all submitted jobs to completion. I'm also not sure the effect that parallel might have on the benchmark results in general)

benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -86,7 +101,37 @@ def sweep(self, arguments: List[str]) -> Any:
         initial_job_idx = 0
         if all_combinations:
             self.validate_batch_is_legal(all_combinations)
-            results = self.launcher.launch(all_combinations, initial_job_idx=initial_job_idx)
+            returns = self._execute_batches(all_combinations, initial_job_idx)
+
+        return returns
```
muddyfish: Is this meant to return nothing if `all_combinations` is empty?

benchmark/tests/test_execute_batches.py
```diff
@@ -0,0 +1,73 @@
...
+        ["benchmark_type=fio", "mountpoint.stub_mode=off", "network.maximum_throughput_gbps=100"],
+        [
+            "benchmark_type=fio",
+            "mountpoint.stub_mode=s3_client",
+            "network.maximum_throughput_gbps=100",
+        ],  # Invalid config
+        ["benchmark_type=fio", "mountpoint.stub_mode=off", "network.maximum_throughput_gbps=100"],
+    ]
+
+    def test_fail_fast_true_stops_on_first_failure(self):
```
muddyfish, line 20: These tests look better, though have you verified that it actually behaves this way when called from the CLI?

benchmark/tests/test_execute_batches.py
```diff
@@ -0,0 +1,73 @@
...
+        ]
+
+        # Test 1: Should not raise any exception
+        results = sweeper._execute_batches(self.INVALID_COMBINATIONS, initial_job_idx=0)
+
+        # Test 2: Should call launcher ONCE with all 3 jobs
+        assert mock_launcher.launch.call_count == 1
+        assert len(results) == 1
+        assert len(results[0]) == 3
+        assert results[0][1].status == JobStatus.FAILED  # Verify failure is captured
```
muddyfish, line 72: Missing verification that we have success in the other 2 cases

benchmark/tests/test_execute_batches.py
```diff
@@ -0,0 +1,73 @@
...
+    ]
+
+    def test_fail_fast_true_stops_on_first_failure(self):
+        sweeper = SmartBenchmarkSweeper(fail_fast=True)
+
+        # We create Mock launcher that returns JobReturn objects to simulate a set of benchmarking jobs
+        mock_launcher = Mock()
+        sweeper.launcher = mock_launcher
+
+        # side_effect makes mock return different values on each call: 1st call gets 1st item, 2nd call gets 2nd item, etc.
```
muddyfish: We don't need excess comments describing what the code does directly.

benchmark/tests/test_execute_batches.py
tadiwa-aizen: Yes, verified with a real CLI run. Configured a benchmark sweep with 36 total combinations and intentionally used an invalid network interface name (ens129 instead of ens33) to trigger a failure.

Results with fail_fast=true:
- Job # 0: SUCCESS (fio, single NIC)
- Job # 1: SUCCESS (fio, single NIC, direct_io=True)
- Job # 2: FAILED (fio, dual NIC with invalid ens129 interface)
- **Stopped immediately** - remaining 33 jobs never executed

The framework correctly caught the CalledProcessError from the mount-s3 command failing with AWS_ERROR_INVALID_ARGUMENT, and stopped execution as expected.

PR #1790, opened 2026-03-16
Implement content type detection for uploaded objects

### LINKED ISSUE #632: How to keep or specify ContentType of object to upload?
### Mountpoint for Amazon S3 version

mount-s3 1.2.0

### AWS Region

ap-northeast-1

### Describe the running environment

Running in EC2 on Amazon Linux 2 using instance profile credentials against an S3 Bucket in the another account.

### What happened?

I use S3 as a CloudFront backend for website hosting purposes.
Therefore, the html file placed in S3 must be displayed as a web page when accessed by a browser, but if placed via mount-s3, the ContentType will be `binary/octet-stream`, so the html file will be downloaded.

Is there any way to modify the ContentType of the object being uploaded to `text/html`?

```
[ec2-user@ip-10-0-54-13 ~]$ mount-s3 --version
mount-s3 1.2.0
[ec2-user@ip-10-0-54-13 ~]$ sudo mount-s3 mybucket /s3 --uid 900 --gid 900 --allow-other --log-directory /var/log/mounts3 --debug
[ec2-user@ip-10-0-54-13 ~]$ file --mime-type index.html
index.html: text/html
[ec2-user@ip-10-0-54-13 ~]$ sudo cp index.html /s3/b.html
[ec2-user@ip-10-0-54-13 ~]$ aws s3api head-object --bucket mybucket --key b.html --query ContentType
"binary/octet-stream"
```

This is probably the same behavior as #471 .

### Relevant log output

_No response_

Zxilly:
Fixes #632

Add `--infer-content-type` flag to automatically set Content-Type on uploaded objects.

When enabled, Mountpoint infers the Content-Type from the uploaded object's file extension.

No breaking change, new behavior will only act if `--infer-content-type` was set.

Yes

---

### DIFF
--- a/mountpoint-s3-fs/Cargo.toml
+++ b/mountpoint-s3-fs/Cargo.toml
@@ -30,10 +30,10 @@ futures = "0.3.32"
 hdrhistogram = { version = "7.5.4", default-features = false }
 hex = "0.4.3"
 humansize = "2.1.3"
-mime_guess = "2.0.5"
 libc = "0.2.186"
 linked-hash-map = "0.5.6"
 metrics = "0.24.3"
+mime_guess = "2.0.5"
 nix = { version = "0.31.2", default-features = false, features = ["fs", "process", "signal", "user"] }
 rand = "0.10.1"
 regex = "1.12.3"
--- a/mountpoint-s3-fs/Cargo.toml
+++ b/mountpoint-s3-fs/Cargo.toml
@@ -30,6 +30,8 @@ futures = "0.3.32"
 hdrhistogram = { version = "7.5.4", default-features = false }
 hex = "0.4.3"
 humansize = "2.1.3"
+infer = "0.19"
+mime_guess = "2.0.5"
 libc = "0.2.186"
 linked-hash-map = "0.5.6"
 metrics = "0.24.3"
--- a/mountpoint-s3-fs/Cargo.toml
+++ b/mountpoint-s3-fs/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "mountpoint-s3-fs"
 # See `/doc/PUBLISHING_CRATES.md` to read how to publish new versions.
-version = "0.9.3"
+version = "0.9.4"
 edition = "2024"
 license = "Apache-2.0"
 repository = "https://github.com/awslabs/mountpoint-s3"
--- a/mountpoint-s3-fs/Cargo.toml
+++ b/mountpoint-s3-fs/Cargo.toml
@@ -30,7 +30,6 @@ futures = "0.3.32"
 hdrhistogram = { version = "7.5.4", default-features = false }
 hex = "0.4.3"
 humansize = "2.1.3"
-infer = "0.19"
 mime_guess = "2.0.5"
 libc = "0.2.186"
 linked-hash-map = "0.5.6"
--- a/mountpoint-s3-fs/src/content_type.rs
+++ b/mountpoint-s3-fs/src/content_type.rs
@@ -1,0 +1,113 @@
+/// Controls how content types are detected for uploaded objects.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub enum ContentTypeDetection {
+    /// No content type detection.
+    #[default]
+    Disabled,
+    /// Detect content type by file extension first, falling back to magic bytes via `infer`.
+    Auto,
+}
+
+/// Try to detect MIME type by file extension.
+fn detect_by_extension(key: &str) -> Option<String> {
+    mime_guess::from_path(key).first().map(|m| m.to_string())
+}
+
+/// Try to detect MIME type by magic bytes.
+fn detect_by_magic(bytes: &[u8]) -> Option<String> {
+    infer::get(bytes).map(|t| t.mime_type().to_owned())
+}
+
+/// Infer the content type of a file.
+///
+/// When `mode` is [`ContentTypeDetection::Auto`], detection uses file extension first
+/// (via `mime_guess`), falling back to magic bytes (via `infer`) when available.
+/// Extension-based detection is preferred because magic bytes misclassify text-based
+/// formats like SVG as `text/plain`.
+pub fn infer_content_type(key: &str, initial_bytes: Option<&[u8]>, mode: ContentTypeDetection) -> Option<String> {
+    match mode {
+        ContentTypeDetection::Disabled => return None,
+        ContentTypeDetection::Auto => {}
+    }
+
+    if let Some(mime) = detect_by_extension(key) {
+        return Some(mime);
+    }
+
+    if let Some(bytes) = initial_bytes {
+        return detect_by_magic(bytes);
+    }
+
+    None
+}
+
+/// Infer the content type of a file (async).
+///
+/// Same logic as [`infer_content_type`]. No blocking work is involved, so this is
+/// a thin async wrapper for API compatibility.
+pub async fn infer_content_type_async(
+    key: &str,
+    initial_bytes: Option<&[u8]>,
+    mode: ContentTypeDetection,
+) -> Option<String> {
+    infer_content_type(key, initial_bytes, mode)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn disabled_returns_none() {
+        let result = infer_content_type("photo.png", None, ContentTypeDetection::Disabled);
+        assert_eq!(result, None);
+    }
+
+    #[test]
+    fn auto_prefers_extension() {
+        // Even with JPEG bytes, .png extension wins
+        let jpeg_header = b"\xFF\xD8\xFF\xE0\x00\x10JFIF";
+        let result = infer_content_type("photo.png", Some(jpeg_header), ContentTypeDetection::Auto);
+        assert_eq!(result, Some("image/png".to_owned()));
+    }
+
+    #[test]
+    fn auto_falls_back_to_magic_bytes() {
+        let jpeg_header = b"\xFF\xD8\xFF\xE0\x00\x10JFIF";
+        let result = infer_content_type("no_extension", Some(jpeg_header), ContentTypeDetection::Auto);
+        assert_eq!(result, Some("image/jpeg".to_owned()));
+    }
+
+    #[test]
+    fn auto_extension_without_bytes() {
+        let result = infer_content_type("photo.png", None, ContentTypeDetection::Auto);
+        assert_eq!(result, Some("image/png".to_owned()));
+    }
+
+    #[test]
+    fn returns_none_when_unknown() {
+        let result = infer_content_type("no_extension", None, ContentTypeDetection::Auto);
+        assert_eq!(result, None);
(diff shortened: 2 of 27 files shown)


alecrubin:
@dannycjones @passaro any chance this could get a review? I could really use this feature.

mountpoint-s3-fs/src/content_type.rs
```diff
@@ -0,0 +1,113 @@
...
+    match mode {
+        ContentTypeDetection::Disabled => return None,
+        ContentTypeDetection::Auto => {}
+    }
+
+    if let Some(mime) = detect_by_extension(key) {
+        return Some(mime);
+    }
+
+    if let Some(bytes) = initial_bytes {
```
vladem: Thanks for working on this! We'd recommend dropping the `infer` crate and keeping detection extension-based only. AFAIK, AWS CLI follows the same approach. Inspecting object's data is something we'd like to avoid (doesn't work now for atomic uploads) and also we'd like to keep the number of dependencies minimal. Extension based detection should cover text-based formats (like HTML, CSS) described in the issue and well-known binary formats with extension specified in the file name (like [JPEG](https://github.com/abonander/mime_guess/blob/master/src/mime_types.rs#L572), PNG).

mountpoint-s3/src/cli.rs
```diff
@@ -400,6 +401,13 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
     )]
     pub upload_checksums: Option<UploadChecksums>,
 
+    #[clap(
+        long,
+        help = "Automatically detect and set Content-Type for uploaded objects based on file content or extension",
+        help_heading = BUCKET_OPTIONS_HEADER,
+    )]
+    pub content_type_detection: bool,
```
vladem, line 409: Just a note for now: the decision on UX is pending. Here we make the feature opt-in (you need to specify the flag to enable content type detection), AWS CLI detects content type by default (with `--no-guess-mime-type` to opt-out). Lets keep as-is (opt-in) now, but we may reconsider.

mountpoint-s3-fs/src/content_type.rs
```diff
@@ -0,0 +1,113 @@
...
+/// a thin async wrapper for API compatibility.
+pub async fn infer_content_type_async(
+    key: &str,
+    initial_bytes: Option<&[u8]>,
+    mode: ContentTypeDetection,
+) -> Option<String> {
+    infer_content_type(key, initial_bytes, mode)
+}
+
+#[cfg(test)]
```
vladem, line 19: This change requires an e2e test, consider adding a new one to [mountpoint-s3-fs/tests/fuse_tests/write_test.rs](https://github.com/vladem/mountpoint-s3/blob/otel_fuse_total_bytes/mountpoint-s3-fs/tests/fuse_tests/write_test.rs#L652):
- The existing write tests already use `TestSessionConfig` which contains `filesystem_config: S3FilesystemConfig`, and the PR adds content_type_detection to `S3FilesystemConfig`.
- The S3 test session (`S3TestSession`) already has sdk_client access and a head_object helper pattern (see `get_object_size`). You'd need to add a helper like `get_object_content_type` that does `head_object` and returns content type.
- The test would:
  1. Create a `TestSessionConfig` with `filesystem_config.content_type_detection = ContentTypeDetection::Auto`
  2. Mount via `S3TestSession::new()`
  3. Write a file like `image.png` with some content through the FUSE mount
  4. Call `head_object` on the S3 object and assert `content_type == "image/png"`
  5. Test for both `ATOMIC_UPLOAD` (S3 [standard](https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-class-intro.html#sc-freq-data-access) buckets) and `INCREMENTAL_UPLOAD` ([directory buckets](https://docs.aws.amazon.com/AmazonS3/latest/userguide/directory-buckets-objects-append.html) / `s3express_tests` feature flag)

mountpoint-s3-fs/src/content_type.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    #[test]
+    fn auto_detects_by_extension() {
+        let result = infer_content_type("photo.png", ContentTypeDetection::Auto);
+        assert_eq!(result, Some("image/png".to_owned()));
+    }
+
+    #[test]
+    fn auto_detects_svg() {
```
vladem: nit: may use `test_case` here

mountpoint-s3/src/cli.rs
```diff
@@ -400,6 +401,13 @@ Learn more in Mountpoint's configuration documentation (CONFIGURATION.md).\
     )]
     pub upload_checksums: Option<UploadChecksums>,
 
+    #[clap(
+        long,
+        help = "Automatically detect and set Content-Type for uploaded objects based on file extension",
+        help_heading = BUCKET_OPTIONS_HEADER,
+    )]
+    pub content_type_detection: bool,
```
vladem: Please rename `--content-type-detection` to `--infer-content-type`. Just the CLI flag. I think `content_type_detection` in configs is OK.

mountpoint-s3-fs/src/fs/config.rs
```diff
@@ -2,6 +2,7 @@ use std::time::Duration;
```
vladem, line 2: also please update the `mountpoint-s3/CHANGELOG.md`, my suggestion is:

```
## Unreleased (v1.22.1)

### New features

* Add support for automatic content type detection on file uploads. When the `--infer-content-type` flag is specified, Mountpoint will infer the `Content-Type` of new objects based on their file extension, instead of using the default `binary/octet-stream`. ([#1790](https://github.com/awslabs/mountpoint-s3/pull/1790))
```

vladem (commented):
thanks for working on this! a few final comments, otherwise LGTM.

mountpoint-s3-fs/src/fs/config.rs
```diff
@@ -43,6 +44,8 @@ pub struct S3FilesystemConfig {
     pub mem_limit: u64,
     /// Prefetcher configuration
     pub prefetcher_config: PrefetcherConfig,
+    /// Content type inference mode for uploaded objects
+    pub infer_content_type: ContentTypeDetection,
```
vladem: Please revert the field name back to `content_type_detection` for fields with type `ContentTypeDetection` (apologies for the confusion). The idea is:

1. The **CLI flag** is `--infer-content-type`, which reads naturally as an imperative command: `mount-s3 bucket /mnt --infer-content-type`.
2. **Struct fields** use the `snake_case` form of the type name (e.g., `content_type_detection: ContentTypeDetection`).
3. `CLIArgs` translates the `--infer-content-type` flag into the appropriate `content_type_detection` mode.

vladem: Also if you could please update the PR description to mention the correct flag name `--infer-content-type` that would be awesome.

vladem (dismissed):

mountpoint-s3/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mountpoint-s3"
-version = "1.22.3"
+version = "1.22.4"
```
yerzhan7: Shouldn't it be `1.23.0` instead as we are adding new feature?

mountpoint-s3-fs/Cargo.toml
```diff
@@ -30,6 +30,7 @@ futures = "0.3.32"
 hdrhistogram = { version = "7.5.4", default-features = false }
 hex = "0.4.3"
 humansize = "2.1.3"
+mime_guess = "2.0.5"
```
yerzhan7: This is inserted in non-alphabetical order.

mountpoint-s3/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+* Add support for automatic content type detection on file uploads. When the `--infer-content-type` flag is specified, Mountpoint will infer the `Content-Type` of new objects based on their file extension instead of using the default `binary/octet-stream`. ([#1790](https://github.com/awslabs/mountpoint-s3/pull/1790))
```
yerzhan7, line 4: We should also update `CONFIGURATOIN.md` doc?

yerzhan7:
np: Users enabling `--infer-content-type`  and renaming `foo.txt` → `foo.md` will see old Content-type and might think it's a bug.

Should we add a note somewhere like that "Renaming an object does not re-evaluate its Content-Type, even when `--infer-content-type` is enabled."?

mountpoint-s3/CHANGELOG.md
vladem: Good catch. I've also though about it, but I think we tend not to document each CLI option, for example `--upload-checksums` and `--storage-class`, which are similar in spirit, are not documented. 

I think the help message is enough, I suggest to mention rename limitation there:

```
target/debug/mount-s3 --help | grep -A 1 "infer-content-type"
      --infer-content-type
          Automatically infer the Content-Type of uploaded objects from their file extension. Content-Type is not updated on rename.
```

Please let me know what you think.

vladem: Alternatively can put this after [File modifications and deletions](https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#file-modifications-and-deletions) section:

```
### Content type detection

By default, the `Content-Type` of new objects is set to `binary/octet-stream`. If you want Mountpoint to automatically infer the Content-Type from the file extension, use the `--infer-content-type` flag at mount time. Content-Type is not updated on rename.
```

yerzhan7: Yes, let's go with that - documenting it in that doc is better for discoverability.

np: StorageClass is actually documented [here](https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#s3-storage-classes). Upload checksums though is a miss.

Zxilly:
I am currently implementing automatic updates to the content type during renaming.

yerzhan7:
> I am currently implementing automatic updates to the content type during renaming.

@Zxilly I don't think it's possible at the moment because `RenameObject` API does not support setting content-type: https://docs.aws.amazon.com/AmazonS3/latest/API/API_RenameObject.html

Zxilly:
My plan is `CopyObject` and `x-amz-metadata-directive: REPLACE`

vladem:
Hey @Zxilly, we discussed this internally with @yerzhan7, and for now we'd like to avoid introducing a `CopyObject` API call on the rename path. We chose to document the limitation in the documentation for `--infer-content-type`. Many thanks for working on this and implementing this feature. I am going to push small final changes to the documentation and merge the PR shortly.

yerzhan7 (dismissed):

Zxilly:
I implemented an update to the content type during renaming in https://github.com/awslabs/mountpoint-s3/pull/1790/commits/9d24fc9ad2563a528b93ff28cdf58200ef805360. It has been reverted, but you might want to check if this meets your needs.

vladem (approved):

yerzhan7 (approved):

PR #1799, opened 2026-03-26
Update Mountpoint crates to track dependencies between each other via workspace manifest, update dependencies

dannycjones:
This is a simple change to where the dependencies between Mountpoint crates resides. Instead of this being scattered and duplicated across the crate manifests, each manifest instead declares a dependency where the specifics of the dependency are defined on the workspace. In the workspace, we define the version, path, and features.

The motivation is to reduce the burden when making changes to the version numbers of our crates.

This change also updates some dependencies via `cargo update`.

No.

No, just Cargo manifest updates.

---


passaro (approved):
LGTM!

PR #1802, opened 2026-04-02
Remove last usages of forbidden CI bucket

dannycjones:
This change removes the last of the 'forbidden bucket' test dependencies. The tests were previously dependent on a bucket with a policy banning almost all S3 operations. Instead, we now use session policies to restrict access on a per-test basis.

Test change only. Removes the need to supply a forbidden bucket for tests.

No, test change only.

---


passaro (approved):
LGTM!

PR #1806, opened 2026-04-14
fix: Use cgroup-aware memory detection in mem limiter

yerzhan7:
**What changed and why?** 

Replace direct `sysinfo::System::total_memory()` call with a new `effective_total_memory()` helper that respects `cgroup` memory limits. This fixes incorrect memory limit detection when running inside containers with memory constraints.

Changes:
- Add `effective_total_memory()` to `mem_limiter` that checks `sysinfo`'s `cgroup_limits()` before falling back to total physical memory
  - Library automatically handles cgroup v1 vs v2, path resolution, no limit, Linux vs non-Linux, and other edge cases: https://github.com/GuillaumeGomez/sysinfo/blob/main/src/unix/linux/cgroup.rs
- Update `cli.rs` and `benchmark` examples to use the new helper
- Add CI job to run cgroup memory detection test in a memory-limited container

No - it prevents potential OOM in containers with memory constraints.

Done.

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -384,6 +384,55 @@ jobs:
           push: false
           file: docker/Dockerfile.source
 
+  cgroup-mem-limit:
+    name: Cgroup memory limit detection
+    runs-on: ubuntu-22.04
+
+    container:
+      image: ubuntu:22.04
+      options: >-
+        --memory=2g
+        --memory-swap=2g
+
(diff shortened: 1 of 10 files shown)


passaro (commented):
Looks good. Needs a rebase/merge from main and an entry in `mountpoint-s3` CHANGELOG.

passaro (approved):

PR #1808, opened 2026-04-16
bench: Add memory-limited FIO benchmarks with breach detection

yerzhan7:
## Summary

Extend the existing FIO throughput benchmarks CI so that each supported workload also runs with Mountpoint built using `--features mem_limiter` and mounted with `--max-memory-target=512` (MiB - fixed). Under this extra pressure, Mountpoint must stay within the configured budget; the new jobs surface per-test peak memory usage in a GitHub Actions step summary table so regressions/improvements in the memory limiter are easy to spot.

Non-latency only. Same FIO job definitions are used for both the regular and memory-limited variants.

- The existing `bench` (S3 Standard), `cache-bench`, and S3 Express `bench` jobs are extended with a `strategy.matrix` that fans each one out to two variants:
  - `default` — unchanged behaviour.
  - `mem-limited` — builds with `--features mem_limiter`, runs with `S3_MAX_MEMORY_TARGET_MIB=512`, and emits an extra GitHub Actions summary table.

  The matrix also drives per-variant job name suffixes, gh-pages chart sub-paths, and S3 results sub-prefixes, so the two variants don't collide.
- New `.github/actions/scripts/render-mem-summary.sh` that renders a Markdown table to `$GITHUB_STEP_SUMMARY` with per-test peak RSS, the limit, a breach flag, and peak reserved memory per area/kind. Gated on `matrix.variant == 'mem-limited'`.
- Shared benchmark scripts (`fs_bench.sh`, `fs_cache_bench.sh`) are now parameterised via the `S3_MAX_MEMORY_TARGET_MIB` env variable. When set, they:
  - Build with `--features mem_limiter`.
  - Mount with `--max-memory-target=<N>`.
  - Ask `mount-s3-log-analyzer` for an additional JSON file via `--mem-limit-mib=<N> --extra-metrics-out=<PATH>`.

  When unset, behaviour is unchanged.
- `mount-s3-log-analyzer` gains two optional flags, `--mem-limit-mib` and `--extra-metrics-out`, wired together with `clap`'s `requires` so either both are set or neither. When both are set, the analyzer also parses Mountpoint metric log lines for:
  - `mem.bytes_reserved[area=prefetch]`
  - `mem.bytes_reserved[area=upload]`
  - `pool.reserved_bytes[kind=get_object]`
  - `pool.reserved_bytes[kind=put_object]`

  and writes JSON with the test name, peak RSS in MiB, memory limit in MiB, a `breached = peak_rss_mib

## Example GitHub Actions summary

https://github.com/awslabs/mountpoint-s3/actions/runs/24720419069

```
| Test | Peak RSS (MiB) | Memory Limit (MiB) | Status | Peak Prefetch Reserved (MiB) | Peak Upload Reserved (MiB) | Peak Pool GetObject (MiB) | Peak Pool PutObject (MiB) |
|---|---|---|---|---|---|---|---|
| mix_1r4w | 1562.546875 | 512 | ❌ BREACHED | 32 | N/A | 32 | 1376 |
| rand_read_4t_direct | 22.0625 | 512 | ✅ OK | 68.5 | N/A | 64 | N/A |
```

Notes:
- Breach is **non-fatal**: the ❌ is informational; the CI job does not fail on a breach.
- A metric is rendered as `N/A` only when Mountpoint never emitted it in the logs (e.g. `pool.reserved_bytes[kind=get_object]` in a write-only workload). If the metric was emitted with value 0, the column shows `0.0`.
- The `_extra_metrics.json` file is consumed **only** by the memory summary step. It is not fed into the gh-pages benchmark charts.

## Where results are stored

Memory-limited results are stored under distinct `mem_limited` sub-paths so they don't collide with the existing charts:

| Workload                        | Throughput chart path                         | Peak-memory chart path                                      |
|---------------------------------|-----------------------------------------------|-------------------------------------------------------------|
| S3 Standard throughput          | `dev/bench/mem_limited`                       | `dev/bench/mem_limited/peak_mem_usage`                      |
| S3 Standard cache               | `dev/cache_bench/mem_limited`                 | `dev/cache_bench/mem_limited/peak_mem_usage`                |
| S3 Express One Zone throughput  | `dev/s3-express/bench/mem_limited`            | `dev/s3-express/bench/mem_limited/peak_mem_usage`           |

## Why a separate `_extra_metrics.json`?

The existing `<test>_peak_mem.json` follows the `{name, value, unit}` schema required by `benchmark-action/github-action-benchmark` and feeds the gh-pages charts. Adding more fields there would pollute the charts for non-memory-limited runs. Keeping the file separate lets each consumer (benchmark-action vs. GH Actions summary) receive only what it needs.

No - adding new benchmarks only.

No.

---


.github/workflows/bench.yml
```diff
@@ -275,3 +275,164 @@ jobs:
       run: |
         .github/actions/scripts/copy-static-files.sh dev/cache_bench dev/cache_bench/peak_mem_usage
         git push 'https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }}.git' gh-pages:gh-pages
+
+  bench-mem-limited:
```
passaro: Is there any opportunity to reduce code duplication here? Could add a matrix to the original jobs?

mountpoint-s3/scripts/fs_bench.sh
```diff
@@ -30,6 +30,20 @@ if [[ -n "${S3_DEBUG}" ]]; then
...
+mem_target_arg=""
+log_analyzer_mem_arg=""
+if [[ -n "${S3_MAX_MEMORY_TARGET_MIB}" ]]; then
+  # Memory-limited variant: build with the mem_limiter feature, cap Mountpoint at the
+  # requested MiB, and ask the log analyzer to additionally emit
+  # results/<job>_extra_metrics.json which is consumed only by the GitHub Actions
+  # memory summary table (render-mem-summary.sh). These files are not fed into the
+  # gh-pages benchmark charts.
+  cargo_feature_args="--features mem_limiter"
+  mem_target_arg="--max-memory-target=${S3_MAX_MEMORY_TARGET_MIB}"
```
passaro: Can't we just append to `optional_args` here?

.github/workflows/bench.yml
```diff
@@ -275,3 +275,164 @@ jobs:
...
+        tool: 'customBiggerIsBetter'
+        output-file-path: results/output.json
+        benchmark-data-dir-path: dev/bench/mem_limited
+        alert-threshold: "200%"
+        fail-on-alert: false
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        auto-push: false
+        comment-on-alert: false
+        max-items-in-chart: ${{ env.MAX_VALUES_IN_BENCH_CHART }}
+        summary-always: true
```
passaro: Will the summary contain 3 tables (Throughput, Peak Memory, and new memory summary)?

mountpoint-s3/src/bin/mount-s3-log-analyzer.rs
```diff
@@ -24,56 +32,120 @@ struct CliArgs {
 
     #[clap(help = "Test name to be reported in JSON file")]
     test_name: String,
+
+    #[clap(
+        long,
+        help = "If set, also write <out_file_dir>/<test_name>_extra_metrics.json with extra memory metrics and a breach flag against this limit",
```
passaro: Shouldn't we take the json file name as another parameter?

mountpoint-s3/src/bin/mount-s3-log-analyzer.rs
```diff
@@ -24,56 +32,120 @@ struct CliArgs {
...
-                            return Err(anyhow!("Unable to parse metric value: {}", parsed_result.unwrap_err()));
-                        };
-                        metric_values.push(value);
+            } else if args.mem_limit_mib.is_some() {
+                for (i, (_, re)) in mem_metrics.iter().enumerate() {
+                    if let Some(cap) = re.captures(&line) {
+                        let value: u64 = cap[1]
+                            .parse()
+                            .map_err(|e| anyhow!("Unable to parse metric value: {}", e))?;
+                        mem_metric_peaks[i] = Some(mem_metric_peaks[i].unwrap_or(0).max(value));
```
passaro: nit: Can't you just do:

```suggestion
                        mem_metric_peaks[i] = mem_metric_peaks[i].max(Some(value)));
```

yerzhan7: Fixed, thanks!

mountpoint-s3/src/bin/mount-s3-log-analyzer.rs
yerzhan7: Done

.github/workflows/bench.yml
yerzhan7: I believe so - github appends summaries from each step.

mountpoint-s3/scripts/fs_bench.sh
yerzhan7: Yes, that would be better, thanks

.github/workflows/bench.yml
yerzhan7: Refactored to a matrix in https://github.com/awslabs/mountpoint-s3/pull/1808/changes/064b811aba3d88190210b95d57c750cd3b1d55b2

passaro (approved):
LGTM!

PR #1812, opened 2026-04-20
Thread request ID from meta-request options into CRT buffer pool reservations

Priyankakarumuru1:
We need to be able to identify which meta-request a buffer allocation belongs to at the CRT pool layer.

This PR adds an optional `custom_id` to `MetaRequestOptionsInner` and exposes it through `MetaRequest`, which is now passed directly to `MemoryPool::get_buffer` / `get_buffer_async`. The identifier can be set via `GetObjectParams::custom_id` and `PutObjectParams::custom_id` in the client crate.

No.

The `MemoryPool` trait signature changed, but the trait is marked experimental. No user-visible behavior change otherwise.

---

### DIFF
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -240,6 +240,9 @@ pub struct GetObjectParams {
     pub range: Option<Range<u64>>,
     pub if_match: Option<ETag>,
     pub checksum_mode: Option<ChecksumMode>,
+    /// An optional caller-supplied identifier passed through to the memory pool on buffer
+    /// allocations for this request. Not related to the S3 request ID returned by the service.
+    pub custom_id: Option<u64>,
 }
 
 impl GetObjectParams {
@@ -263,6 +266,12 @@ impl GetObjectParams {
     /// Set option to retrieve checksum as part of the GetObject request
     pub fn checksum_mode(mut self, value: Option<ChecksumMode>) -> Self {
         self.checksum_mode = value;
+        self
+    }
+
+    /// Set the custom identifier for memory pool buffer allocations.
+    pub fn custom_id(mut self, value: Option<u64>) -> Self {
+        self.custom_id = value;
         self
     }
 }
@@ -553,6 +562,9 @@ pub struct PutObjectParams {
     pub custom_headers: Vec<(String, String)>,
     /// User-defined object metadata
     pub object_metadata: ObjectMetadata,
+    /// An optional caller-supplied identifier passed through to the memory pool on buffer
+    /// allocations for this request. Not related to the S3 request ID returned by the service.
+    pub custom_id: Option<u64>,
 }
 
 impl PutObjectParams {
@@ -594,6 +606,12 @@ impl PutObjectParams {
     /// Set user defined object metadata.
     pub fn object_metadata(mut self, value: ObjectMetadata) -> Self {
         self.object_metadata = value;
+        self
+    }
+
+    /// Set the custom identifier for memory pool buffer allocations.
+    pub fn custom_id(mut self, value: Option<u64>) -> Self {
+        self.custom_id = value;
         self
     }
 }
--- a/mountpoint-s3-client/src/object_client.rs
+++ b/mountpoint-s3-client/src/object_client.rs
@@ -269,7 +269,8 @@ impl GetObjectParams {
         self
     }
 
-    /// Set the custom identifier for memory pool buffer allocations.
+    /// Set an optional caller-supplied identifier passed through to the memory pool on buffer
+    /// allocations for this request. Not related to the S3 request ID returned by the service.
     pub fn custom_id(mut self, value: Option<u64>) -> Self {
         self.custom_id = value;
         self
@@ -609,7 +610,8 @@ impl PutObjectParams {
         self
     }
 
-    /// Set the custom identifier for memory pool buffer allocations.
+    /// Set an optional caller-supplied identifier passed through to the memory pool on buffer
+    /// allocations for this request. Not related to the S3 request ID returned by the service.
     pub fn custom_id(mut self, value: Option<u64>) -> Self {
         self.custom_id = value;
         self
--- a/mountpoint-s3-client/tests/common/memory_pool.rs
+++ b/mountpoint-s3-client/tests/common/memory_pool.rs
@@ -1,4 +1,4 @@
-use mountpoint_s3_client::config::{MemoryPool, MetaRequestType};
+use mountpoint_s3_client::config::{MemoryPool, MetaRequest};
 
 /// Creates a memory pool to use in tests.
 pub fn new_for_tests() -> impl MemoryPool {
@@ -12,7 +12,7 @@ struct NoReusePool();
 impl MemoryPool for NoReusePool {
     type Buffer = Box<[u8]>;
 
-    fn get_buffer(&self, size: usize, _type: MetaRequestType, _request_id: u64) -> Self::Buffer {
+    fn get_buffer(&self, size: usize, _meta_request: &MetaRequest) -> Self::Buffer {
         vec![0u8; size].into_boxed_slice()
     }
 
--- a/mountpoint-s3-client/tests/common/memory_pool.rs
+++ b/mountpoint-s3-client/tests/common/memory_pool.rs
@@ -12,7 +12,7 @@ struct NoReusePool();
 impl MemoryPool for NoReusePool {
     type Buffer = Box<[u8]>;
 
-    fn get_buffer(&self, size: usize, _type: MetaRequestType) -> Self::Buffer {
+    fn get_buffer(&self, size: usize, _type: MetaRequestType, _request_id: u64) -> Self::Buffer {
         vec![0u8; size].into_boxed_slice()
     }
 
--- a/mountpoint-s3-client/tests/common/memory_pool.rs
+++ b/mountpoint-s3-client/tests/common/memory_pool.rs
@@ -17,7 +17,7 @@ struct NoReusePool();
 impl MemoryPool for NoReusePool {
     type Buffer = Box<[u8]>;
 
-    fn get_buffer(&self, size: usize, _meta_request: MetaRequest) -> Self::Buffer {
+    fn get_buffer(&self, size: usize, _meta_request: &MetaRequest) -> Self::Buffer {
         vec![0u8; size].into_boxed_slice()
     }
 
@@ -33,12 +33,6 @@ pub struct RecordingMemoryPool {
 }
 
 impl RecordingMemoryPool {
-    pub fn new() -> Self {
-        Self {
-            observed_custom_ids: Arc::new(Mutex::new(Vec::new())),
-        }
-    }
-
     pub fn observed_custom_ids(&self) -> Vec<Option<u64>> {
         self.observed_custom_ids.lock().unwrap().clone()
     }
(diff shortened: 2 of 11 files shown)


mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -506,6 +510,30 @@ impl<'a> MetaRequestOptions<'a> {
         options.inner.copy_source_uri = unsafe { options.copy_source_uri.as_mut().unwrap().as_aws_byte_cursor() };
         self
     }
+
+    /// Set an opaque identifier for this meta request. This value can be retrieved
+    /// from the buffer pool reserve path via [`meta_request_id`].
+    pub fn request_id(&mut self, id: u64) -> &mut Self {
```
passaro: `request_id` could create confusion with S3 request ids. Let's consider other options, e.g. `custom_id` or `custom_tag`.

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -29,7 +30,8 @@ pub trait MemoryPool: Clone + Send + Sync + 'static {
     type Buffer: AsMut<[u8]> + Send;
 
     /// Get a buffer of at least the requested size.
-    fn get_buffer(&self, size: usize, meta_request_type: MetaRequestType) -> Self::Buffer;
+    /// `request_id` is a lightweight per-request identifier for tracking/correlation.
+    fn get_buffer(&self, size: usize, meta_request_type: MetaRequestType, request_id: u64) -> Self::Buffer;
```
passaro: Instead of adding more parameters here and in the async variant, we should either:
- pack them into a new `GetBufferParams` struct,
- or just pass a reference to the `MetaRequest` and expose methods on it to retrieve the type and the tag/id.

I'm leaning towards the latter, which more closely reflects the CRT API.

passaro (changes requested):
A few observations:

- The new identifier is not wired up in the `client` crate. How do users set it when calling `get_object` or `put_object`?
- This is missing any test: at the very least, check that a value set on a request is propagated in the calls to the memory pool.
- See comments below about the name of the identifier and the `MemoryPool` API.

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -286,6 +286,9 @@ struct MetaRequestOptionsInner<'a> {
     /// Finish callback, if provided (and not already called, since it's FnOnce).
     on_finish: Option<FinishCallback>,
 
+    /// Opaque identifier for this meta request, readable from the buffer pool reserve path.
+    request_id: u64,
```
passaro: Shouldn't this be optional?

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -1758,10 +1787,28 @@ impl UploadReviewPart {
...
 mod tests {
+    use std::pin::Pin;
+
     use test_case::test_case;
 
     use crate::aws_s3_request_type;
-    use crate::s3::client::RequestType;
+    use crate::s3::client::{MetaRequestOptions, RequestType};
+
+    #[test]
```
passaro: What are we testing in these 2 functions?

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -1758,10 +1787,28 @@ impl UploadReviewPart {
...
+    fn meta_request_options_custom_id_round_trip() {
+        let mut options = MetaRequestOptions::new();
+        options.custom_id(42);
+        // SAFETY: directly accessing inner struct for testing, not moving out of it.
+        let inner = unsafe { Pin::get_unchecked_mut(Pin::as_mut(&mut options.0)) };
+        assert_eq!(inner.custom_id, Some(42));
```
passaro: What is unsafe about this? Why not use `options.0.custom_id`?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -42,13 +40,8 @@ pub trait MemoryPool: Clone + Send + Sync + 'static {
...
     /// The default implementation wraps [`get_buffer`](Self::get_buffer) in a
     /// ready future. Override this when the pool needs to wait for memory to free up.
-    fn get_buffer_async(
-        &self,
-        size: usize,
-        meta_request_type: MetaRequestType,
-        request_id: u64,
-    ) -> impl Future<Output = Self::Buffer> + Send {
-        ready(self.get_buffer(size, meta_request_type, request_id))
+    fn get_buffer_async(&self, size: usize, meta_request: MetaRequest) -> impl Future<Output = Self::Buffer> + Send {
```
passaro, line 45: Why is this taking ownership of the meta request while the sync version only borrows it?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -373,12 +360,12 @@ unsafe extern "C" fn pool_reserve<Pool: MemoryPool>(
...
     let crt_pool = unsafe { CrtBufferPool::<Pool>::ref_from_raw(&pool) };
 
-    // SAFETY: `meta.meta_request` is a pointer to a valid `aws_s3_meta_request`.
-    let request_type = unsafe { (*meta.meta_request).type_ };
-    // SAFETY: `meta.meta_request` is a pointer to a valid `aws_s3_meta_request` whose
-    // user_data was initialized by `MetaRequestOptions`.
-    let request_id = unsafe { meta_request_id(meta.meta_request) };
-    let future = crt_pool.reserve(meta.size, request_type.into(), request_id, meta.can_block);
+    // SAFETY: `meta.meta_request` is a valid pointer to an `aws_s3_meta_request` whose
+    // `user_data` was initialised by `MetaRequestOptions`. `from_raw_acquire` increments
```
passaro: The first sentence fully justifies the safety of the invocation. The second just describes what `MetaRequest::from_raw_acquire` does (already in its doc) and should be removed.

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -707,6 +711,31 @@ pub struct MetaRequest {
 }
 
 impl MetaRequest {
+    /// Acquires a reference to an `aws_s3_meta_request` raw pointer, incrementing its ref-count.
```
passaro: Should mention that it creates a new `MetaRequest` instance. E.g.

> Creates a new instance from an `aws_s3_meta_request` raw pointer, incrementing its ref-count.

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -707,6 +711,31 @@ pub struct MetaRequest {
...
+    ///
+    /// # Safety
+    ///
+    /// `raw` must be a valid, non-null pointer to an `aws_s3_meta_request` whose `user_data`
+    /// was initialised by [`MetaRequestOptions`].
+    pub(crate) unsafe fn from_raw_acquire(raw: *mut aws_s3_meta_request) -> Self {
+        // SAFETY: caller guarantees `raw` is valid and non-null.
```
passaro: This justifies calling `aws_s3_meta_request_acquire`, but we should also mention why we can use `NonNull::new_unchecked` on the returned pointer. See the `Clone` implementation.

mountpoint-s3-client/tests/put_object.rs
```diff
@@ -12,14 +13,44 @@ use rand::RngExt;
...
+use mountpoint_s3_client::config::{MemoryPool, MetaRequest, S3ClientConfig};
 use mountpoint_s3_client::error::{GetObjectError, ObjectClientError};
 use mountpoint_s3_client::types::{
     ChecksumAlgorithm, GetObjectParams, HeadObjectParams, ObjectClientResult, PutObjectParams, PutObjectResult,
     PutObjectTrailingChecksums,
 };
 use mountpoint_s3_client::{ObjectClient, PutObjectRequest, S3RequestError};
 
+#[derive(Clone)]
+struct RecordingMemoryPool {
```
passaro: Avoid duplicated code.

mountpoint-s3-client/src/object_client.rs
```diff
@@ -265,6 +268,12 @@ impl GetObjectParams {
         self.checksum_mode = value;
         self
     }
+
+    /// Set the custom identifier for memory pool buffer allocations.
```
passaro: The doc should match the one for the field (besides starting with "Set ..").

passaro: Same for PUT

passaro (changes requested):

mountpoint-s3-crt/src/s3/client.rs
Priyankakarumuru1: These were basic sanity checks. Happy to remove them since the integration tests in get_object.rs and put_object.rs already cover the full end-to-end propagation

mountpoint-s3-crt/src/s3/pool.rs
Priyankakarumuru1: I'll make them consistent.

mountpoint-s3-client/tests/get_object.rs
```diff
@@ -112,7 +82,7 @@ async fn test_get_object_custom_id_propagates_to_memory_pool() {
     let client_config = S3ClientConfig::new()
         .endpoint_config(get_test_endpoint_config())
         .memory_pool(recording_pool.clone());
-    let client = get_test_client_with_config(client_config);
+    let client = S3CrtClient::new(client_config).expect("could not create test client");
```
passaro, line 87: Add a comment to document that we are not using `get_test_client_with_config` because we want to pass a specific pool.

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -711,14 +711,16 @@ pub struct MetaRequest {
...
     ///
     /// # Safety
     ///
     /// `raw` must be a valid, non-null pointer to an `aws_s3_meta_request` whose `user_data`
     /// was initialised by [`MetaRequestOptions`].
     pub(crate) unsafe fn from_raw_acquire(raw: *mut aws_s3_meta_request) -> Self {
-        // SAFETY: caller guarantees `raw` is valid and non-null.
+        // SAFETY: caller guarantees `raw` is a valid `aws_s3_meta_request`, and
+        // `aws_s3_meta_request_acquire` increments the reference count for it
+        // (and always returns a copy of the input, which is non-null).
```
passaro: "`aws_s3_meta_request_acquire` increments the reference count for it" is not relevant for the safety justification, only that it returns its input (which is not null).

mountpoint-s3-crt/src/s3/pool.rs
passaro: Do either of these methods require ownership of the meta request?

mountpoint-s3-crt/src/s3/pool.rs
```diff
@@ -29,7 +29,9 @@ pub trait MemoryPool: Clone + Send + Sync + 'static {
     type Buffer: AsMut<[u8]> + Send;
 
     /// Get a buffer of at least `size` bytes for the given meta request.
-    fn get_buffer(&self, size: usize, meta_request: &MetaRequest) -> Self::Buffer;
+    ///
+    /// The `meta_request` is reference-counted and may be cloned if needed.
```
passaro: Why do we need the additional comment about reference-counting? That is a property of the `MetaRequest` type, not of this method in particular.

mountpoint-s3-client/tests/common/memory_pool.rs
```diff
@@ -21,3 +26,33 @@ impl MemoryPool for NoReusePool {
         false
     }
 }
+
+#[derive(Clone, Default)]
```
passaro, line 30: You're deriving `Default` here, but not using it.

passaro: For example, you could use it to avoid implementing `new`.

passaro (changes requested):
A few more suggestions. Let's discuss before the next iteration.

passaro (approved):
Looks good!

PR #1813, opened 2026-04-21
bench: Add incremental-upload throughput benchmark to S3 Express CI

yerzhan7:
## Summary

Extend the existing S3 Express throughput benchmark CI with two new `--incremental-upload` variants, folded into the existing `bench` matrix introduced by #1808 rather than as a separate job:

- `incremental-upload` — default memory budget.
- `incremental-upload-mem-limited` — `--features mem_limiter` + `--max-memory-target=512`.

Both new variants only run the `write` and `mix` fio categories (read is skipped since incremental upload is an upload-path feature).

This PR also isolates `S3_BUCKET_TEST_PREFIX` per matrix leg on the throughput `bench` jobs (S3 Standard and S3 Express). The single workflow-level prefix previously caused all matrix legs to race for the same fio scratch keys in the benchmark bucket. This was latent (silent overlapping MPUs) for non-incremental legs but fatal for incremental upload: the append pipeline conditions each `PutObject` on the object's current ETag, and a sibling leg's `unlink=1` between iterations aborts the upload with `NoSuchKey`.

gh-pages paths follow the `data_path_suffix` convention from #1808:

| Variant | Throughput chart path |
|---|---|
| Incremental Upload | `dev/s3-express/bench/incremental_upload` |
| Incremental Upload, Memory-Limited | `dev/s3-express/bench/incremental_upload/mem_limited` |

No - only benchmark prefix changes generating/using new objects.

No — CI-only change.

---

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -268,7 +268,6 @@ jobs:
       env:
         S3_MOUNT_LOCAL_STORAGE: yes
         S3_MAX_MEMORY_TARGET_MIB: ${{ matrix.max_memory_target_mib }}
-        S3_BUCKET_TEST_PREFIX: ${{ vars.S3_BUCKET_BENCH_PREFIX || 'mountpoint-benchmark/' }}cache-bench${{ matrix.results_subdir_suffix }}/
       run: mountpoint-s3/scripts/fs_cache_bench.sh
     - name: Render memory summary
       if: matrix.variant == 'mem-limited'
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -24,7 +24,6 @@ env:
   RUST_BACKTRACE: 1
   CARGO_TERM_COLOR: always
   S3_BUCKET_NAME: ${{ vars.S3_BENCH_BUCKET_NAME }}
-  S3_BUCKET_TEST_PREFIX: ${{ vars.S3_BUCKET_BENCH_PREFIX || 'mountpoint-benchmark/' }}
   S3_BUCKET_BENCH_FILE: ${{ vars.BENCH_FILE_NAME || 'bench100GB.bin' }}
   S3_BUCKET_SMALL_BENCH_FILE: ${{ vars.SMALL_BENCH_FILE_NAME || 'bench5MB.bin' }}
   # Optional endpoint url, can be empty
@@ -182,6 +181,8 @@ jobs:
     - name: Build
       run: cargo build --release
     - name: Run Benchmark
+      env:
+        S3_BUCKET_TEST_PREFIX: ${{ vars.S3_BUCKET_BENCH_PREFIX || 'mountpoint-benchmark/' }}latency-bench/
       run: mountpoint-s3/scripts/fs_latency_bench.sh
     - name: Save benchmark results in S3
       if: inputs.s3_bench_results_prefix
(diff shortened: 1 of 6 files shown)


.github/workflows/bench.yml
```diff
@@ -94,6 +94,7 @@ jobs:
     - name: Run Benchmark
       env:
         S3_MAX_MEMORY_TARGET_MIB: ${{ matrix.max_memory_target_mib }}
+        S3_BUCKET_TEST_PREFIX: ${{ vars.S3_BUCKET_BENCH_PREFIX || 'mountpoint-benchmark/' }}bench${{ matrix.results_subdir_suffix }}/
```
passaro, line 96: Why are we doing this only here? And not for the cache one? Should this replace line 27?

mountpoint-s3/scripts/fs_bench.sh
```diff
@@ -41,6 +41,10 @@ if [[ -n "${S3_MAX_MEMORY_TARGET_MIB}" ]]; then
   optional_args+=" --max-memory-target=${S3_MAX_MEMORY_TARGET_MIB}"
 fi
 
+if [[ -n "${S3_INCREMENTAL_UPLOAD}" ]]; then
```
passaro: nit: what if we expose a more generic ~`S3_MOUNTPOINT_EXTRA_ARGS`? It would also avoid the potential issue with `S3_INCREMENTAL_UPLOAD="no"`.

mountpoint-s3/src/bin/mount-s3-log-analyzer.rs
```diff
@@ -54,7 +54,7 @@ struct CliArgs {
 /// JSON field name for the peak (in MiB) and the regex matching lines ending in that
 /// metric's value (bytes). Absent entries are omitted from `_extra_metrics.json` and
 /// rendered as "N/A" in the GH step summary table.
-fn mem_metric_patterns() -> [(&'static str, Regex); 4] {
+fn mem_metric_patterns() -> [(&'static str, Regex); 5] {
```
passaro: tiny nit: shall we just return a `Vec` here? and on line 92?

passaro (commented):
I can't seem to find all the new sections in the summary, e.g. [here](https://github.com/awslabs/mountpoint-s3/actions/runs/24995610064?pr=1813). Am I looking at the wrong place?

Besides that, just a few minor notes.

yerzhan7:
> I can't seem to find all the new sections in the summary, e.g. [here](https://github.com/awslabs/mountpoint-s3/actions/runs/24995610064?pr=1813). Am I looking at the wrong place?

Memory summary is there for the new `incremental-upload-mem-limited`. Others are missing because it's a new benchmark and there is no baseline from `main` to compare from.

Screenshot 2026-05-01 at 10 32 14

.github/workflows/bench.yml
yerzhan7: I did revert the cache one because it was failing due to missing the pre-seeded object at the new prefix.

I will add back the cache prefix and manually move object to the new prefix as one-off.

mountpoint-s3/src/bin/mount-s3-log-analyzer.rs
yerzhan7: Fixed, thanks

mountpoint-s3/scripts/fs_bench.sh
yerzhan7: Good point, fixed!

passaro (approved):

PR #1816, opened 2026-04-24
Reconcile memory limiter reservations with pool allocations

renanmagagnin:
The `MemoryLimiter` and `PagedPool` tracked memory independently — the limiter tracked intent (read window reservations) while the pool tracked actual buffer allocations. Download buffers were invisible to the limiter's budget check, the two systems could diverge, and reservations were only tracked as a global total with no way to attribute memory to individual cursors.

This PR makes the limiter aware of all pool allocations. The budget check now includes all buffer kinds (downloads, uploads, disk cache) via `pool.total_reserved_bytes()`. The pool notifies the limiter on each allocation via a callback that decrements both `mem_reserved` and `mem_reserved_per_cursor`, so `mem_reserved` represents only the unallocated portion of the read window. Per-cursor reservation tracking enables accurate cleanup on cursor drop.

The seek window reservation (at file open) and the backward seek re-injection reserve (in `PartQueue::push_front`) were removed — both double-counted memory already tracked by the pool.

The upload path is simplified to rely on pool stats directly instead of calling `reserve`/`release`.

No breaking changes to the public API.

No, these are internal accounting changes to the memory limiter with no user-visible behavior change or API change. The memory limiter is behind a feature flag (`mem_limiter`) and not yet stabilized.

---

### DIFF
--- a/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
+++ b/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
@@ -266,7 +266,7 @@ impl DiskBlock {
         }
 
         let size = header.block_len as usize;
-        let mut buffer = pool.get_buffer_mut(size, BufferKind::DiskCache);
+        let mut buffer = pool.get_buffer_mut(size, BufferKind::DiskCache, None);
         buffer.fill_from_reader(reader)?;
         let data = buffer.into_bytes();
 
--- a/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
+++ b/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
@@ -26,6 +26,7 @@ use crate::metrics::defs::{
     CACHE_PUT_ERRORS, CACHE_PUT_IO_SIZE, CACHE_PUT_LATENCY, CACHE_TOTAL_SIZE,
 };
 use crate::object::ObjectId;
+use crate::prefetch::HandleId;
 use crate::sync::Mutex;
 
 use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheResult};
@@ -262,7 +263,7 @@ impl DiskBlock {
         reader: &mut impl Read,
         block_size: u64,
         pool: &PagedPool,
-        custom_id: Option<u64>,
+        handle_id: Option<HandleId>,
     ) -> Result<Self, DiskBlockReadWriteError> {
         let header: DiskBlockHeader = bincode::decode_from_std_read(reader, BINCODE_CONFIG)?;
 
@@ -271,7 +272,7 @@ impl DiskBlock {
         }
 
         let size = header.block_len as usize;
-        let mut buffer = pool.get_buffer_mut(size, BufferKind::DiskCache, custom_id);
+        let mut buffer = pool.get_buffer_mut(size, BufferKind::DiskCache, handle_id);
         buffer.fill_from_reader(reader)?;
         let data = buffer.into_bytes();
 
@@ -342,7 +343,7 @@ impl DiskDataCache {
         cache_key: &ObjectId,
         block_idx: BlockIndex,
         block_offset: u64,
-        custom_id: Option<u64>,
+        handle_id: Option<HandleId>,
     ) -> DataCacheResult<Option<ChecksummedBytes>> {
         trace!(
             key = ?cache_key.key(),
@@ -366,7 +367,7 @@ impl DiskDataCache {
             return Err(DataCacheError::InvalidBlockContent);
         }
 
-        let block = DiskBlock::read(&mut file, self.block_size(), &self.pool, custom_id)
+        let block = DiskBlock::read(&mut file, self.block_size(), &self.pool, handle_id)
             .inspect_err(|e| warn!(path = ?path.as_ref(), "block could not be deserialized: {:?}", e))?;
         let bytes = block
             .data(cache_key, block_idx, block_offset)
@@ -470,7 +471,7 @@ impl DataCache for DiskDataCache {
         block_idx: BlockIndex,
         block_offset: u64,
         _object_size: usize,
-        custom_id: Option<u64>,
+        handle_id: Option<HandleId>,
     ) -> DataCacheResult<Option<ChecksummedBytes>> {
         if block_offset != block_idx * self.config.block_size {
             return Err(DataCacheError::InvalidBlockOffset);
@@ -478,7 +479,7 @@ impl DataCache for DiskDataCache {
         let start = Instant::now();
         let block_key = DiskBlockKey::new(cache_key, block_idx);
         let path = self.get_path_for_block_key(&block_key);
-        let result = match self.read_block(&path, cache_key, block_idx, block_offset, custom_id) {
+        let result = match self.read_block(&path, cache_key, block_idx, block_offset, handle_id) {
             Ok(None) => {
                 // Cache miss.
                 Ok(None)
--- a/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
+++ b/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
@@ -258,7 +258,12 @@ impl DiskBlock {
     }
 
     /// Deserialize an instance from `reader`.
-    fn read(reader: &mut impl Read, block_size: u64, pool: &PagedPool) -> Result<Self, DiskBlockReadWriteError> {
+    fn read(
+        reader: &mut impl Read,
+        block_size: u64,
+        pool: &PagedPool,
+        custom_id: Option<u64>,
+    ) -> Result<Self, DiskBlockReadWriteError> {
         let header: DiskBlockHeader = bincode::decode_from_std_read(reader, BINCODE_CONFIG)?;
 
         if header.block_len > block_size {
@@ -266,7 +271,7 @@ impl DiskBlock {
         }
 
         let size = header.block_len as usize;
-        let mut buffer = pool.get_buffer_mut(size, BufferKind::DiskCache, None);
+        let mut buffer = pool.get_buffer_mut(size, BufferKind::DiskCache, custom_id);
         buffer.fill_from_reader(reader)?;
         let data = buffer.into_bytes();
 
@@ -337,6 +342,7 @@ impl DiskDataCache {
         cache_key: &ObjectId,
         block_idx: BlockIndex,
         block_offset: u64,
+        custom_id: Option<u64>,
     ) -> DataCacheResult<Option<ChecksummedBytes>> {
         trace!(
             key = ?cache_key.key(),
@@ -360,7 +366,7 @@ impl DiskDataCache {
             return Err(DataCacheError::InvalidBlockContent);
         }
 
-        let block = DiskBlock::read(&mut file, self.block_size(), &self.pool)
+        let block = DiskBlock::read(&mut file, self.block_size(), &self.pool, custom_id)
             .inspect_err(|e| warn!(path = ?path.as_ref(), "block could not be deserialized: {:?}", e))?;
         let bytes = block
             .data(cache_key, block_idx, block_offset)
@@ -464,6 +470,7 @@ impl DataCache for DiskDataCache {
         block_idx: BlockIndex,
         block_offset: u64,
         _object_size: usize,
+        custom_id: Option<u64>,
     ) -> DataCacheResult<Option<ChecksummedBytes>> {
         if block_offset != block_idx * self.config.block_size {
             return Err(DataCacheError::InvalidBlockOffset);
@@ -471,7 +478,7 @@ impl DataCache for DiskDataCache {
         let start = Instant::now();
         let block_key = DiskBlockKey::new(cache_key, block_idx);
         let path = self.get_path_for_block_key(&block_key);
-        let result = match self.read_block(&path, cache_key, block_idx, block_offset) {
+        let result = match self.read_block(&path, cache_key, block_idx, block_offset, custom_id) {
             Ok(None) => {
                 // Cache miss.
                 Ok(None)
@@ -769,7 +776,7 @@ mod tests {
         );
 
         let block = cache
-            .get_block(&cache_key_1, 0, 0, object_1_size)
+            .get_block(&cache_key_1, 0, 0, object_1_size, None)
             .await
             .expect("cache should be accessible");
         assert!(
@@ -783,7 +790,7 @@ mod tests {
             .await
             .expect("cache should be accessible");
         let entry = cache
-            .get_block(&cache_key_1, 0, 0, object_1_size)
+            .get_block(&cache_key_1, 0, 0, object_1_size, None)
             .await
             .expect("cache should be accessible")
             .expect("cache entry should be returned");
@@ -798,7 +805,7 @@ mod tests {
             .await
             .expect("cache should be accessible");
         let entry = cache
-            .get_block(&cache_key_2, 0, 0, object_2_size)
+            .get_block(&cache_key_2, 0, 0, object_2_size, None)
             .await
             .expect("cache should be accessible")
             .expect("cache entry should be returned");
@@ -813,7 +820,7 @@ mod tests {
             .await
             .expect("cache should be accessible");
         let entry = cache
-            .get_block(&cache_key_1, 1, block_size, object_1_size)
+            .get_block(&cache_key_1, 1, block_size, object_1_size, None)
             .await
             .expect("cache should be accessible")
             .expect("cache entry should be returned");
@@ -824,7 +831,7 @@ mod tests {
 
         // Entry 1's first block still intact
         let entry = cache
-            .get_block(&cache_key_1, 0, 0, object_1_size)
+            .get_block(&cache_key_1, 0, 0, object_1_size, None)
             .await
             .expect("cache should be accessible")
             .expect("cache entry should be returned");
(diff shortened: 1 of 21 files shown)


mountpoint-s3-fs/src/prefetch/part_queue.rs
```diff
@@ -105,7 +110,8 @@ impl<Client: ObjectClient> PartQueue<Client> {
         metrics::gauge!("prefetch.bytes_in_queue").increment(part.len() as f64);
         // The backpressure controller is not aware of the parts from backwards seek,
         // so we have to reserve memory for them here.
-        self.mem_limiter.reserve(BufferArea::Prefetch, part.len() as u64);
+        self.mem_limiter
```
passaro: Why is this required? The justification above should not hold anymore.

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -26,46 +29,30 @@ impl BufferArea {
...
-///
-/// backwards_seek_start        part_queue_start         in_part_queue                 window_end_offset      preferred_window_end_offset
-///  │                              │                           │                               │                            │
-/// ─┼──────────────────────────────┼───────────────────────────┼───────────────────────────────┼────────────────────────────┼───────────-►
-///  │                              │                                                           │
-///  └──────────────────────────────┤                                                           │
-///  mem reserved by the part queue │                                                           │
-///                                 └───────────────────────────────────────────────────────────┤
-///                                     mem reserved by the backpressure controller
-///                                     (on `BackpressureFeedbackEvent`)
```
passaro, line 91: Missing: when is "reserved" memory decremented now?

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -129,11 +135,20 @@ impl MemoryLimiter {
     }
 
     /// Release the reserved memory.
-    pub fn release(&self, area: BufferArea, size: u64) {
-        self.mem_reserved.fetch_sub(size, Ordering::SeqCst);
+    pub fn release(&self, handle_id: HandleId, area: BufferArea, size: u64) {
```
passaro: When is this needed?

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -144,15 +159,136 @@ impl MemoryLimiter {
...
+        assert_eq!(limiter.mem_reserved.load(Ordering::SeqCst), 2048);
+
+        // Pool allocates 1024 — callback decrements global by 1024
+        let _buffer = pool.get_buffer_mut(1024, BufferKind::GetObject);
+        assert_eq!(limiter.mem_reserved.load(Ordering::SeqCst), 1024);
+
+        // release_handle releases the remaining 2048 from per-handle (not aware of pool decrement)
+        // Global goes to 1024 - 2048 which would underflow — this is the known limitation
+        // that per-handle tracking doesn't account for pool callback decrements.
+        // For now, release_handle releases the per-handle balance.
```
passaro: "known limitation"? Shouldn't we address it in this PR?

mountpoint-s3-fs/src/memory/stats.rs
```diff
@@ -1,23 +1,46 @@
...
 use crate::sync::Arc;
 use crate::sync::atomic::{AtomicUsize, Ordering};
 
 /// Usage stats for a pool.
-#[derive(Debug, Default)]
+#[derive(Default)]
 pub struct PoolStats {
     reserved_bytes: [AtomicUsize; BUFFER_KIND_COUNT],
+    /// Optional callback invoked whenever bytes are reserved in the pool.
+    on_reserve: OnceLock<Arc<dyn Fn(usize) + Send + Sync>>,
```
passaro: Let's consider options for this notification

passaro (commented):
A few initial comments. It would be useful to mention what you are planning to do before leaving draft.

mountpoint-s3-fs/src/mem_limiter.rs
renanmagagnin: It's used by `PrefetchGetObject::drop()` to release the seek window reservation, I'll consider merging the two

renanmagagnin: Removed seek window reservation and this was removed with it, thanks!

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -79,23 +72,46 @@ impl MemoryLimiter {
...
             formatter(reserved_mem)
         );
+        let mem_reserved = Arc::new(AtomicU64::new(0));
+        let mem_reserved_cb = mem_reserved.clone();
+        let mem_reserved_per_handle: Arc<DashMap<HandleId, AtomicU64>> = Arc::new(DashMap::new());
+        let per_handle_cb = mem_reserved_per_handle.clone();
+        pool.set_on_reserve(Arc::new(move |bytes, custom_id| {
+            // Saturating subtraction prevents underflow if this callback races with
+            // release_handle during meta-request cancellation. In that case the global
+            // clamps to 0 — a harmless under-count corrected when the pool buffer is dropped.
```
passaro: How is it corrected?

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -1,10 +1,13 @@
 use std::{sync::atomic::Ordering, time::Instant};
 
+use dashmap::DashMap;
 use humansize::make_format;
 use metrics::atomics::AtomicU64;
```
passaro: pre-existing issue: all `atomic` imports should come from `crate::sync`.

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -144,15 +167,124 @@ impl MemoryLimiter {
...
+    fn test_pool_allocation_decrements_mem_reserved() {
+        let (limiter, pool) = new_limiter(1024, MINIMUM_MEM_LIMIT);
+        let handle = HandleId::new(1);
+
+        // Reserve 1024 bytes of intent
+        limiter.reserve(handle, BufferArea::Prefetch, 1024);
+        assert_eq!(limiter.mem_reserved.load(Ordering::SeqCst), 1024);
+
+        // Pool allocation triggers on_reserve callback, decrementing mem_reserved
+        let _buffer = pool.get_buffer_mut(1024, BufferKind::GetObject, None);
```
passaro: We're not passing the handle here. What happens to the per-handle reservation?

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -110,16 +110,29 @@ impl PagedPool {
...
+    /// Register a callback to be invoked whenever bytes are reserved in the pool.
+    /// Must be called before any pool allocations occur to ensure accurate accounting.
+    pub fn set_on_reserve(&self, callback: Arc<dyn Fn(usize, Option<u64>) + Send + Sync>) {
+        self.inner.stats.set_on_reserve(callback);
+    }
+
     /// Get a new empty mutable buffer from the pool with the requested capacity.
-    pub fn get_buffer_mut(&self, capacity: usize, kind: BufferKind) -> PoolBufferMut {
-        let buffer = self.get_buffer(capacity, kind);
+    /// If `custom_id` is provided, it will be passed to the on_reserve callback
```
passaro: Why `custom_id: Option<u64>` instead of `handle_id: HandleId`?

mountpoint-s3-fs/src/mem_limiter.rs
renanmagagnin: Fixed this by avoiding both decrements after the handle specific counter entry has been removed, thanks!

mountpoint-s3-fs/src/memory/pool.rs
renanmagagnin: To match the CRT's `MetaRequest::custom_id()` type and avoid coupling the pool layer to `HandleId` from the prefetch module. The mapping from `u64` to `HandleId` is happening now in the `on_reserve` callback in `MemoryLimiter::new()`, seems more natural but let me know if you think we should reconsider. I'll also remove handle from the comment for consistency.

mountpoint-s3-fs/src/mem_limiter.rs
renanmagagnin: Fixed, test now passes the handle since in the real code path the CRT always gets `custom_id` from `GetObjectParams`. Without it, the callback would skip the decrement and mem_reserved stays inflated, making the budget check overly conservative.

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -79,23 +72,56 @@ impl MemoryLimiter {
...
+        let mem_reserved_per_handle: Arc<DashMap<HandleId, AtomicU64>> = Arc::new(DashMap::new());
+        let per_handle_cb = mem_reserved_per_handle.clone();
+        pool.set_on_reserve(Arc::new(move |bytes, custom_id| {
+            // Called by the pool on every buffer allocation. For download buffers with a
+            // known handle, this converts reservation from "intent" (mem_reserved) to
+            // "actual allocation" (pool_total) by decrementing both the global and
+            // per-handle counters.
+            //
+            // This is a no-op when:
+            // - custom_id is None (e.g. disk cache, uploads): these allocations have no
```
passaro: disk cache allocations should have a mem_reserved entry

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -79,23 +72,56 @@ impl MemoryLimiter {
...
+            //   corresponding mem_reserved entry and are tracked only via pool_total.
+            // - The handle has been removed (cancelled request): release_handle already
+            //   subtracted the full balance, so we skip to avoid double-decrementing.
+            //
+            // Saturating subtraction on the per-handle counter prevents wrapping if a late
+            // callback from a cancelled request hits a re-created entry for the same handle.
```
passaro: `saturating_sub` handles cases where we allocate more bytes than we originally reserved.
But isn't there still a race condition when an entry is re-created? The effect is that we may incorrectly reduce the reservation. Can we accept it? How else could we handle it?

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -79,23 +72,56 @@ impl MemoryLimiter {
...
+            if let Some(handle_id) = custom_id
+                && let Some(reservation) = per_handle_cb.get(&HandleId::new(handle_id))
+            {
+                let mut current = reservation.load(Ordering::SeqCst);
```
passaro: Could we use `AtomicU64::update` here?

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -266,7 +266,7 @@ impl DiskBlock {
         }
 
         let size = header.block_len as usize;
-        let mut buffer = pool.get_buffer_mut(size, BufferKind::DiskCache);
+        let mut buffer = pool.get_buffer_mut(size, BufferKind::DiskCache, None);
```
passaro: We should pass the handle_id here, right?

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -157,7 +170,7 @@ impl MemoryPool for PagedPool {
     type Buffer = PoolBuffer;
 
     fn get_buffer(&self, size: usize, meta_request: &MetaRequest) -> Self::Buffer {
-        self.get_buffer(size, meta_request.meta_request_type().into())
+        self.get_buffer(size, meta_request.meta_request_type().into(), meta_request.custom_id())
```
passaro: Here we should wrap in `HandleId`.

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -79,23 +72,56 @@ impl MemoryLimiter {
...
+            // - custom_id is None (e.g. disk cache, uploads): these allocations have no
+            //   corresponding mem_reserved entry and are tracked only via pool_total.
+            // - The handle has been removed (cancelled request): release_handle already
+            //   subtracted the full balance, so we skip to avoid double-decrementing.
+            //
+            // Saturating subtraction on the per-handle counter prevents wrapping if a late
+            // callback from a cancelled request hits a re-created entry for the same handle.
+            if let Some(handle_id) = custom_id
+                && let Some(reservation) = per_handle_cb.get(&HandleId::new(handle_id))
+            {
```
passaro: We'll be holding a lock for this scope. Are we okay with it?

passaro: In fact, let's review the choice of data structure (`DashMap<_, AtomicU64>`).

mountpoint-s3-fs/src/mem_limiter.rs
renanmagagnin: I think so but It seems to still be unstable (`atomic_try_update`) so I'm assuming not for now

passaro: I see. Not for long, though: #1823.

renanmagagnin: `AtomicU64::update` is now stable in Rust 1.95, but shuttle's `AtomicU64` doesn't implement it yet. As an alternative, `fetch_update` works with shuttle but requires `Some(...)` + `.unwrap()` since we never conditionally abort, which feels like unnecessary ceremony. Thinking we can keep the manual CAS loop until shuttle adds support, let me know if you disagree!

mountpoint-s3-fs/src/mem_limiter.rs
renanmagagnin: Aligned with offline discussion, replaced `HandleId` with `RequestId` for tracking of memory allocation. This fixes the race condition.

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -74,6 +76,36 @@ impl HandleId {
     }
 }
 
+/// Opaque identifier for a single prefetch request (i.e., one `BackpressureController` lifetime).
```
passaro, line 79: Let's reconsider the name, since "request" is overloaded. The description here does not help either: it explains why you made the change rather than what this new type is.

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -74,6 +76,36 @@ impl HandleId {
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl RequestId {
+    /// Generate a new unique request ID.
+    pub fn new() -> Self {
+        use crate::sync::atomic::{AtomicU64, Ordering};
+        static NEXT_ID: AtomicU64 = AtomicU64::new(1);
```
passaro: Rather than a static counter, could this belong to the memory limiter, which would hand out new ids?

mountpoint-s3-fs/src/prefetch/backpressure_controller.rs
```diff
@@ -119,11 +122,12 @@ pub struct BackpressureLimiter {
 pub fn new_backpressure_controller(
     config: BackpressureConfig,
     mem_limiter: Arc<MemoryLimiter>,
-) -> (BackpressureController, BackpressureLimiter) {
+) -> (BackpressureController, BackpressureLimiter, RequestId) {
```
passaro: Could we add the new id to `BackpressureLimiter` and simplify how it's passed around? Where else is it needed?

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -112,31 +104,70 @@ pub struct MemoryLimiter {
...
+            // Called by the pool on every buffer allocation. For download buffers with a
+            // known request, this converts reservation from "intent" (mem_reserved) to
+            // "actual allocation" (pool_total) by decrementing both the global and
+            // per-request counters.
+            //
+            // This is a no-op when:
+            // - request_id is None (e.g. uploads): these allocations have no
+            //   corresponding mem_reserved entry and are tracked only via pool_total.
+            // - The request has been removed (cancelled request): release_request already
+            //   subtracted the full balance, so we skip to avoid double-decrementing.
```
passaro: What if `remove` happens just after `get`? I believe there is still a race condition.

mountpoint-s3-fs/src/prefetch.rs
renanmagagnin: As discussed offline, renamed to CursorId

mountpoint-s3-fs/src/mem_limiter.rs
renanmagagnin: As discussed offline, this race was addressed with a swap in `release_cursor()` (previously `release_request()`)

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -208,7 +223,46 @@ impl MemoryLimiter {
             .unwrap_or(false)
     }
 
-    /// Get reserved memory from the memory pool for buffers not tracked by this limiter.
+    /// Increment both the global total and the per-cursor reservation.
+    fn add_reservation(&self, cursor_id: CursorId, size: u64) {
```
passaro, line 227: Can we inline this in `reserve`?

passaro: Happy to leave for a later PR.

passaro (approved):

PR #1820, opened 2026-04-28
Add stress tests

yerzhan7:
### Summary

- Adds a stress test harness under `mountpoint-s3-fs/tests/stress_tests/` with four scenarios (`sustained_reads`, `sustained_writes`, `mixed_rw`, `idle_and_churn`) that drive real S3 traffic under the 512 MiB memory limit to shake out deadlocks, per-worker stalls, tail-latency regressions, and memory issues.
- Asserts at teardown: reservation gauges return to zero, per-op p100 latency within a configurable ceiling (20sec default), and per-worker watchdog against stalls.
- Adds a feature flag `stress_tests`, an aggregated metrics recorder in `tests/common/test_recorder.rs`  and GitHub workflows when a PR is labelled `stress` (similar to benchmarks/performance workflows).
- Runs on the same hosts as benchmarks
- Example stress test run and CI logs from my fork: https://github.com/yerzhan7/mountpoint-s3/actions/runs/25390606910
- For now set it to run for 15min in CI

No

No

---


mountpoint-s3-fs/tests/stress_tests/README.md
```diff
@@ -0,0 +1,76 @@
...
+| `S3_REGION`             | yes      | Region of the bucket                                |
+| `S3_BUCKET_TEST_PREFIX` | no       | Defaults to `mountpoint-test/`; must end in `/`     |
+| AWS credentials         | yes      | Standard SDK resolution (`AWS_PROFILE`, static keys, instance role, etc.) |
+| `STRESS_DURATION_SECS`  | no       | Per-scenario duration in seconds; default 30        |
+
+## How to run
+
+The `stress_tests` feature transitively enables `s3_tests`, so no extra
+feature flag is needed. Tests are `#[ignore]` by default — always pass
+`--run-ignored only`.
```
passaro: Why the `#[ignore]`? Isn't the feature enough?

mountpoint-s3-fs/tests/common/test_recorder.rs
```diff
@@ -1,25 +1,44 @@
...
+            let actual: Vec<_> = key.labels().map(|l| (l.key(), l.value())).collect();
+            if actual.len() != labels.len() {
+                return false;
+            }
+            labels.iter().all(|(k, v)| actual.contains(&(*k, *v)))
+        })
+        .map(|entry| Arc::clone(entry.value()))
+}
+
+/// Simple `Mutex`-backed metrics storage; intended for low-contention functional tests where
```
passaro, line 27: Why do we need 2 implementations?

mountpoint-s3-fs/tests/mod.rs
```diff
@@ -3,3 +3,5 @@ pub mod common;
 #[cfg(feature = "fuse_tests")]
 mod fuse_tests;
 mod reftests;
+#[cfg(feature = "stress_tests")]
+mod stress_tests;
```
passaro: Could we move the new tests into their own test binary? (Separately, I think we should do the same for the other modules).

mountpoint-s3-fs/tests/stress/scenarios.rs
passaro, line 1: Prefer `scenarios.rs` in the parent directory for consistency. But note that the first level under the `tests` directory is an exception because every file is compiled into a separate binary.

mountpoint-s3-fs/tests/stress_tests/harness/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+fn stress_mount_s3_path() -> S3Path {
+    let bucket = crate::common::s3::get_test_bucket();
+    let prefix = std::env::var("S3_BUCKET_TEST_PREFIX").unwrap_or_else(|_| String::from("mountpoint-test/"));
+    assert!(prefix.ends_with('/'), "S3_BUCKET_TEST_PREFIX should end in '/'");
+    S3Path::new(Bucket::new(bucket).unwrap(), Prefix::new(&prefix).unwrap())
+}
+
+/// Run the given scenario. Reads `STRESS_DURATION_SECS` from env; default 30s.
+pub fn run<S: Scenario + 'static>(scenario: S) {
+    // Make sure the snapshotting recorder is installed even if the common-module ctor has not
```
passaro: Not clear why we need 2 different mechanisms. Also, do we rely on this function to be called only once per process? If so, we should make it clear and possibly fail if that's not the case.

mountpoint-s3-fs/tests/stress_tests/harness/invariants.rs
```diff
@@ -0,0 +1,265 @@
...
+            out.push((label_value, metric));
+        }
+    });
+    out.sort_by(|a, b| a.0.cmp(&b.0));
+    out
+}
+
+/// Assert each individual reserved-memory gauge stayed below the effective budget the
+/// limiter enforces against (`mem_limit - additional_mem_reserved`)
+pub(super) fn assert_peak_reserved_invariant(scenario_name: &str, mem_limit: f64) {
```
passaro: Just `pub` should be enough since this module is itself not public.

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -0,0 +1,265 @@
...
+            "stress: no recorder installed, skipping peak-reserved invariant"
+        );
+        return;
+    };
+    let mem_limit_u64 = mem_limit as u64;
+    let additional_mem_reserved = (mem_limit_u64 / 8).max(128 * 1024 * 1024);
+    let effective_budget = mem_limit_u64.saturating_sub(additional_mem_reserved);
+
+    let mut violations: Vec<String> = Vec::new();
+    for (area, metric) in collect_gauges_by_label(recorder, "mem.bytes_reserved", "area") {
```
passaro, line 58: not sure this is a valid invariant: do we want to guarantee never to "reserve" (vs allocate) more than the limit?

mountpoint-s3-fs/tests/stress_tests/scenarios/idle_and_churn.rs
```diff
@@ -0,0 +1,109 @@
...
+                read_to_eof_once("idle_and_churn churn", &path, &mut buf, progress, latencies, stop);
+            } else {
+                idle_cycle(&path, &mut buf, iter, progress, latencies, stop);
+            }
+        }
+    }
+}
+
+/// One idle-worker iteration: open, read the small object fully (so the prefetcher reserves
+/// memory against this handle), hold idle for 5-15s, then close.
```
passaro: Not sure what this is testing. What is the prefetcher keeping in memory after a full read?

mountpoint-s3-fs/tests/stress_tests/harness/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+pub use latency::{FileOp, FileOpLatencies};
+use memory_monitor::spawn_memory_monitor;
+use report::dump_summary;
+use watchdog::{NO_STALL, spawn_watchdog};
+
+/// Default scenario duration if `STRESS_DURATION_SECS` is unset.
+const DEFAULT_DURATION_SECS: u64 = 30;
+
+/// A stress-test scenario. Implementations describe a load shape; the harness drives it.
+pub trait Scenario: Send + Sync {
```
passaro: I wonder whether the `Scenario` trait is the best abstraction to build different tests. Looking at the current usage, it may be clearer to define each "worker" (e.g. "sustained_read", "churn", etc) and then build a test with a vector of workers.

mountpoint-s3-fs/tests/stress_tests/test_objects.rs
```diff
@@ -0,0 +1,180 @@
...
+        .memory_pool(pool.clone());
+    let client = S3CrtClient::new(client_config).expect("failed to build S3CrtClient for test object upload");
+    let runtime = Runtime::new(client.event_loop_group());
+    let uploader = Uploader::new(client, runtime, pool, mem_limiter, UploaderConfig::new(part_size));
+
+    (uploader, bucket, part_size)
+}
+
+/// HEAD-skip-if-size-matches, else stream-upload a shared test object.
+fn upload_test_object_with_uploader(
```
passaro: Should this whole function be `async`, rather than using multiple `block_on`?

mountpoint-s3-fs/tests/stress/test_objects.rs
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+    }
+}
+
+const DEFAULT_WRITE_PART_SIZE: usize = 8 * 1024 * 1024;
+const MAX_PARTS: u64 = 10_000;
+const TEST_OBJECT_FILL_BYTE: u8 = 0xA5;
+
+/// Memory budget for the test object uploader: 95% of total system memory, floored at
```
passaro, line 91: Won't this affect the peak memory measurement for the test workload? Should we move it out-of-process?

passaro (commented):
I have not looked at everything in detail yet, but here are a few initial comments.

mountpoint-s3-fs/tests/common/test_recorder.rs
yerzhan7: 1. `TestRecorder` stores every histogram sample in a `Mutex<Vec<f64>>` and every counter in a `Mutex<u64>`. That's fine for functional/integration tests with a handful of observations where we want exact values back, but in a long-running stress run, the `Vec` grows unboundedly.
1. `HdrRecorder` uses `hdrhistogram` for bounded-memory percentile storage, lock-free `AtomicU64` counters, and stores a full HDR `history` alongside the gauge's `current` value. That history is what lets the teardown invariants (`assert_p100_latency`, `assert_peak_reserved_invariant`) assert against true peaks/percentiles without sampling races, which `TestRecorder`'s single‑value gauge can't give us.

Let me know if you have better suggestions.

mountpoint-s3-fs/tests/mod.rs
yerzhan7: Will do, thanks. Will probably keep fuse_tests and reftests 'as-is' for now.

mountpoint-s3-fs/tests/stress/harness/invariants.rs
yerzhan7: Hmm, yeah. I thought that limiting "reserve" **strictly** is also part of our long-term plan, but now I realize it is more tricky.

I can opt to log WARN instead of error for now for informational purposes? (for the "mem.bytes_reserved" only, "pool.reserved_bytes" would still emit ERROR log) 

Or would you prefer removing this invariant completely? (metric will still be visible log like [here](https://github.com/yerzhan7/mountpoint-s3/actions/runs/25058706305/job/73406343314#step:7:122))

mountpoint-s3-fs/tests/stress_tests/scenarios/idle_and_churn.rs
yerzhan7: The idea was to validate (via file op latency emtrics) that "48 churn workers" (active readers) are prioritized over idle file handlers (non-active readers).

mountpoint-s3-fs/tests/stress/test_objects.rs
yerzhan7: Pattern was copied from `s3io_benchmark`. The Uploader is dropped before `spawn_memory_monitor` starts, so setup allocations aren't sampled directly but there might be  some small delay when kernel/allocator actually releases memory.

Haven't noticed issues so far, but happy to move it out-of-process (via `aws` CLI probably) if you prefer. Let me know.

mountpoint-s3-fs/tests/stress_tests/test_objects.rs
yerzhan7: Fixed

mountpoint-s3-fs/tests/stress_tests/harness/mod.rs
yerzhan7: Good suggestion, refactored.

mountpoint-s3-fs/tests/stress_tests/harness/invariants.rs
yerzhan7: Fixed

mountpoint-s3-fs/tests/stress_tests/README.md
yerzhan7: You are right. Removed `ignore`.

mountpoint-s3-fs/tests/stress_tests/harness/mod.rs
yerzhan7: Oops, leftover from my previous iteration. Left only 1 mechanism and added failure if called twice.

mountpoint-s3-fs/tests/stress_tests/scenarios/idle_and_churn.rs
yerzhan7: Sorry, you are right - I've just realized it was reading file fully. I now have adjusted the scenario in the following way: https://github.com/awslabs/mountpoint-s3/pull/1820/changes/4fd0294b06457c0791317d1a4da9028fbe4d8ae5
- Increased object size to 4MB
- Idle handles read 1MB to trigger prefetching and holding data in part queue for 1min before closing
- 48 idle handles and 8 churn handles

mountpoint-s3-fs/tests/stress_tests.rs
```diff
@@ -0,0 +1,14 @@
+//! Long-running stress tests.
+#![cfg(feature = "stress_tests")]
+
+#[path = "common/mod.rs"]
```
passaro: The path should not be necessary.

mountpoint-s3-fs/tests/stress_tests.rs
```diff
@@ -0,0 +1,14 @@
...
+mod common;
+
+#[path = "stress_tests"]
+mod _stress_tests {
```
passaro: Could we just name the module `stress_tests` and remove the custom path? We don't want the test to show up as `_stress_tests::scenarios::idle_and_churn::idle_and_churn`.

passaro: Alternative:
* Keep `stress_tests` as the test binary name (this file)
* Rename the module to e.g. `stress`
* Refer to the submodules everywhere as `crate::stress:harness` etc.

We can later do something similar for `fuse_tests` and `ref_tests`.

mountpoint-s3-fs/Cargo.toml
```diff
@@ -102,6 +102,7 @@ s3_tests = []
 s3express_tests = []
 shuttle = []
 second_account_tests = []
+stress_tests = ["s3_tests"]
```
passaro: Don't they also require `fuse_tests`?

passaro: I now see line 9 in `common/mod.rs`. Either approach works.

mountpoint-s3-fs/tests/stress_tests/workers/churn.rs
```diff
@@ -0,0 +1,51 @@
...
+                mount_path
+                    .join(SHARED_OBJECTS_PREFIX)
+                    .join(self.pool.key(pick_index(iter, instance, self.pool.count)));
+            read_to_eof_once("churn", &path, &mut buf, progress, latencies, stop);
+        }
+    }
+}
+
+/// Deterministic pseudo-random pick in `0..count`. Shared with [`super::idle::Idle`].
+pub(super) fn pick_index(iter: u64, instance: usize, count: usize) -> usize {
```
passaro: nit: method on SharedObjectPool?

mountpoint-s3-fs/tests/stress_tests/scenarios/idle_and_churn.rs
passaro: Thanks, it makes more sense now. I suspect we may want to try with larger file sizes and potentially with different "idle" patterns, e.g. read, idle, read, etc, in order to trigger even more prefetching. Happy to review it later though.

mountpoint-s3-fs/tests/stress/test_objects.rs
passaro: Let's note it down in a comment for now.

passaro (commented):
A few more notes:

mountpoint-s3-fs/tests/stress_tests.rs
yerzhan7: Thanks refactored to module `stress` in https://github.com/awslabs/mountpoint-s3/pull/1820/commits/6196a02475a49806f1dea51755da92b82a4cfd79

mountpoint-s3-fs/tests/stress_tests.rs
yerzhan7: Fixed

mountpoint-s3-fs/tests/stress/test_objects.rs
yerzhan7: Added comment, thanks for reviewing.

mountpoint-s3-fs/Cargo.toml
yerzhan7: Chose to add to `Cargo.toml` as it's cleaner, thanks!

mountpoint-s3-fs/tests/stress_tests/workers/churn.rs
yerzhan7: Good point, refactored.

passaro (approved):
LGTM!

PR #1822, opened 2026-04-29
Buffer Pruning: Add Per-Handle Active Read Tracking

Priyankakarumuru1:
The pruning system currently has no way to tell which buffers a user is actively waiting on vs. which are just prefetched ahead.

This PR adds per-handle active read tracking to the memory limiter. When a FUSE read comes in, we record the offset and size of what the user is waiting for. The allocation queue can later check if a buffer overlaps with an active read and prioritize accordingly. A concurrent map makes this safe to query from other threads, and an RAII guard ensures cleanup even on panics or early returns.

The active read range is set before any blocking. For forward seeks, the range is widened to include the skipped bytes the prefetcher must consume.

No

No, these are internal changes to the memory limiter with no user-visible behavior change.

---

### DIFF
--- a/mountpoint-s3-fs/src/prefetch.rs
+++ b/mountpoint-s3-fs/src/prefetch.rs
@@ -380,7 +380,7 @@ where
         // the skipped bytes the prefetcher must consume to reach the requested offset.
         let active_start = self.next_sequential_read_offset.min(offset);
         let active_size = (offset + length as u64 - active_start) as usize;
-        let mut _active_read_guard = self
+        let _active_read_guard = self
             .mem_limiter
             .set_active_read(self.handle_id, active_start, active_size);
 
@@ -400,13 +400,6 @@ where
                 // This is an approximation, tolerating some seeking caused by concurrent readahead.
                 self.record_contiguous_read_metric();
                 self.reset_prefetch_to_offset(offset);
-
-                // A failed forward seek resets the prefetcher to `offset`, so the widened
-                // active range is no longer valid. Replace with just the requested range.
-                if active_start != offset {
-                    drop(_active_read_guard);
-                    _active_read_guard = self.mem_limiter.set_active_read(self.handle_id, offset, length);
-                }
             }
         }
         assert_eq!(self.next_sequential_read_offset, offset);
--- a/mountpoint-s3-fs/src/prefetch.rs
+++ b/mountpoint-s3-fs/src/prefetch.rs
@@ -379,7 +379,7 @@ where
         // Set the active read range before any blocking. For forward seeks, widen to include
         // the skipped bytes the prefetcher must consume to reach the requested offset.
         let active_start = self.next_sequential_read_offset.min(offset);
-        let active_size = (offset + length as u64 - active_start) as usize;
+        let active_size = length + offset.saturating_sub(self.next_sequential_read_offset) as usize;
         let _active_read_guard = self
             .mem_limiter
             .set_active_read(self.handle_id, active_start, active_size);
--- a/mountpoint-s3-fs/src/prefetch.rs
+++ b/mountpoint-s3-fs/src/prefetch.rs
@@ -332,6 +332,12 @@ where
             "read"
         );
 
+        // Record the active read range so the pruning/allocation logic knows
+        // this handle has a user waiting for data at this specific range.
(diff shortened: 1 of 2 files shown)


mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -475,6 +484,33 @@ where
...
+    /// Act on the seek decision. This may block waiting for S3 data (forward seek)
+    /// or reset the prefetcher (random seek). The active read must already be set
+    /// before calling this method.
+    async fn perform_seek(
+        &mut self,
+        seek_kind: SeekKind,
+        offset: u64,
+    ) -> Result<(), PrefetchReadError<Client::ClientError>> {
+        if seek_kind != SeekKind::None {
+            if self.try_seek(offset).await? {
```
passaro: The active read range may be wrong when this fails on a forward seek.

passaro (changes requested):
This approach does not seem to work for all forward seek cases. An alternative may be to set the active read later in the forward seek path and return the guard.

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -387,23 +387,17 @@ where
         }
         let mut to_read = (length as u64).min(remaining);
 
-        // Determine the seek kind and active read range (cheap, no awaits).
-        let (seek_kind, active_start) = if self.next_sequential_read_offset == offset {
-            (SeekKind::None, offset)
+        // Determine the seek kind (cheap, no awaits).
+        let seek_kind = if self.next_sequential_read_offset == offset {
```
passaro: Why do we need this outside `perform_seek`?

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -484,31 +478,55 @@ where
...
+            actual = offset,
+            "out-of-order read, resetting prefetch"
+        );
+        counter!(PREFETCH_RESET_STATE).increment(1);
+
+        // This is an approximation, tolerating some seeking caused by concurrent readahead.
+        self.record_contiguous_read_metric();
+        self.reset_prefetch_to_offset(offset);
+
+        // Forward-seek failure changes the effective read to start at `offset` after reset.
```
passaro: To sum up the strategy: set a broader range as active if we may do a forward seek, then if it aborts and we reset the prefetcher, drop the guard and set just the read request range.

As written, it's not very clear: consider looking back at the original code (pre `perform_seek`) and how we would implement this strategy from there.

mountpoint-s3-fs/src/prefetch.rs
passaro: Why do we need this before `try_seek`?

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -391,8 +399,14 @@ where
 
                 // This is an approximation, tolerating some seeking caused by concurrent readahead.
                 self.record_contiguous_read_metric();
-
                 self.reset_prefetch_to_offset(offset);
+
+                // A failed forward seek resets the prefetcher to `offset`, so the widened
```
passaro: Is this required? What's the downside of leaving the wide range? After a reset, no allocation will be requested except for the smaller (offset, length) range anyway.

Priyankakarumuru1: Right

mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -376,6 +376,14 @@ where
         }
         let mut to_read = (length as u64).min(remaining);
 
+        // Set the active read range before any blocking. For forward seeks, widen to include
+        // the skipped bytes the prefetcher must consume to reach the requested offset.
+        let active_start = self.next_sequential_read_offset.min(offset);
+        let active_size = (offset + length as u64 - active_start) as usize;
```
passaro: The size calculation is off for backward seeks and could be clearer. What about this for example?

```suggestion
        let active_size = length + offset.saturating_sub(self.next_sequential_read_offset) as usize;
```

passaro (commented):
One remaining issue and needs to be merged into `feature/memory-limit`.

passaro (approved):

PR #1825, opened 2026-05-07
Extract Cursor and track active reads per cursor

passaro:
Extract a `Cursor` abstraction from `PrefetchGetObject` to encapsulate read position, backward seek window, and request task management. Each cursor is assigned a unique `CursorId` (generated by `MemoryLimiter`) which replaces `HandleId` for memory reservation tracking and active read detection, giving the buffer pruning logic accurate per-request visibility into which ranges are actively being waited on.

Rustdoc comments updated to reflect the new cursor-based model.

No.

No.

---

### DIFF
--- a/mountpoint-s3-fs/src/prefetch/cursor.rs
+++ b/mountpoint-s3-fs/src/prefetch/cursor.rs
@@ -29,6 +29,7 @@ where
     sequential_read_start_offset: u64,
     next_sequential_read_offset: u64,
     object_id: ObjectId,
+    cursor_id: CursorId,
 }
 
 impl<Client> Cursor<Client>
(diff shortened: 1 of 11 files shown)


mountpoint-s3-fs/src/prefetch/cursor.rs
```diff
@@ -0,0 +1,232 @@
...
+        Ok(true)
+    }
+}
+
+impl<Client> Drop for Cursor<Client>
+where
+    Client: ObjectClient + Clone + Send + Sync + 'static,
+{
+    fn drop(&mut self) {
+        histogram!("prefetch.contiguous_read_len").record((self.current_offset - self.start_offset) as f64);
```
yerzhan7, line 215: This will now record metric even in read error case? Was this intended?

passaro: Well spotted. Not intended, but I think it is the correct behavior. Previously, we would reset on errror and lose track of the contiguous read.

yerzhan7 (approved):

PR #1828, opened 2026-05-11
Fix skip ranges in metrics tests

passaro:
The script testing metrics emissions tries to exercise a mix of sequential and "random" reads to then verify the expected metrics are recorded. This change simplifies the skip pattern used to drive the reads and avoids moving past the size of the test file.

Unrelated minor change: add a random pattern to the temporary folders in line with similar scripts. 

No.

No. Test only.

---


sahityadg (approved):
Thanks for fix this. LGTM.

PR #1829, opened 2026-05-12
Add basic buffer pruner

yerzhan7:
Skeleton for the background buffer pruner that will reclaim memory from idle prefetch handles under memory pressure.

**What changed:**
- New `WakeSignal` primitive (`Mutex<bool>` + `Condvar`): coalescing single-consumer wake with optional timeout.
- New `maintenance` thread that runs both periodic pool trim *and* on-demand pruning rounds. Replaces the standalone `schedule_trim` thread — one thread now handles both responsibilities.
- `MemoryLimiter` gains `trigger_pruning()` (notify) and `pruning_signal()`; on `Drop` it notifies the signal so the maintenance thread observes the dead `Weak` and exits.
- `run_pruning_round` skeleton: pool trim → check queue → defer to natural release path → drop one idle prefetch handle. Queue inspection is stub for now. Will be wired up later.
- Dropping 1 idle LRU cursor. Tracks LRU via new `last_read_at` Atomic field that is updated on each `set_active_read`

**Why `Mutex<bool>` + `Condvar`**
- Mountpoint does not have dependency on tokio (except in unit tests).
- Coalesces many notifies into one wake (no thrashing under bursty pressure).
- Polls every 1ms while pressure is sustained (no CPU burn — the thread sleeps between rounds).
- Stops polling and parks the thread when pressure clears.
- No lost wakeups: notifies before `wait_timeout` set the pending flag and are observed on the next call.
- Periodic 60s trim runs on the same thread when idle — same code path, no extra thread.

**Sleep behavior:**
- Outer 60s idle wait is *interruptible* via `trigger_pruning()` — pressure starts → immediate wake.
- Inner 1ms tick between drain rounds is *not* interruptible — under sustained pressure the loop already polls every 1ms, so an extra wake has nothing useful to do.

No

No

---

### DIFF
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -194,14 +194,7 @@ impl MemoryLimiter {
         let Some(state) = self.cursors.get(&cursor_id).and_then(|r| r.upgrade()) else {
             return false;
         };
-        match *state.read_state.lock().unwrap() {
-            ReadState::Active { offset: o, size: s } => {
-                let self_end = o + s as u64;
-                let other_end = offset + size as u64;
-                o < other_end && offset < self_end
-            }
-            ReadState::Last { .. } => false,
-        }
+        state.read_state.lock().unwrap().is_active(offset, size)
     }
 
     /// Called by the pool on every buffer allocation. For download buffers with a known cursor,
@@ -350,6 +343,22 @@ enum ReadState {
     /// No read is in flight. `tick` is the monotonic counter stamped when
     /// the last read ended, or `0` if the cursor has never been read from.
     Last { tick: u64 },
+}
+
+impl ReadState {
+    /// `true` iff there is an in-flight read whose range overlaps `[offset, offset + size)`.
+    fn is_active(&self, offset: u64, size: usize) -> bool {
+        let Self::Active {
+            offset: active_offset,
+            size: active_size,
+        } = *self
+        else {
+            return false;
+        };
+        let active_end = active_offset + active_size as u64;
+        let other_end = offset + size as u64;
+        active_offset < other_end && offset < active_end
+    }
 }
 
 type ResetFn = Box<dyn Fn() -> bool + Send + Sync>;
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -8,7 +8,7 @@ use tracing::{debug, trace};
 use crate::prefetch::CursorId;
 use crate::sync::atomic::{AtomicU64, Ordering};
 use crate::sync::{Arc, Mutex, Weak};
-use crate::wake_signal::WakeSignal;
+use crate::util::wake_signal::WakeSignal;
 
 use super::PagedPool;
 use super::stats::PoolStats;
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -293,7 +293,11 @@ impl MemoryLimiter {
         };
 
         if self.request_reset(cursor_id) {
-            debug!(?cursor_id, last_read_tick = tick, "reset idle cursor under memory pressure");
+            debug!(
+                ?cursor_id,
+                last_read_tick = tick,
+                "reset idle cursor under memory pressure"
+            );
             return true;
         }
         false
@@ -780,9 +784,7 @@ mod tests {
     fn install_test_reset_fn(cursor: &CursorHandle) -> Arc<std::sync::atomic::AtomicBool> {
         let was_reset = Arc::new(std::sync::atomic::AtomicBool::new(false));
         let flag = was_reset.clone();
-        cursor.register_reset_fn(Box::new(move || {
-            !flag.swap(true, std::sync::atomic::Ordering::SeqCst)
-        }));
+        cursor.register_reset_fn(Box::new(move || !flag.swap(true, std::sync::atomic::Ordering::SeqCst)));
         was_reset
     }
 
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -197,6 +197,15 @@ impl MemoryLimiter {
         state.read_state.lock().unwrap().is_active(offset, size)
     }
 
+    /// Check if the given cursor currently has an active FUSE read in progress.
+    pub fn has_active_read(&self, cursor_id: CursorId) -> bool {
+        self.cursors
+            .get(&cursor_id)
+            .and_then(|r| r.upgrade())
+            .map(|s| matches!(*s.read_state.lock().unwrap(), ReadState::Active { .. }))
+            .unwrap_or(false)
+    }
+
     /// Called by the pool on every buffer allocation. For download buffers with a known cursor,
     /// this converts reservation from "intent" (`mem_reserved`) to "actual allocation" (pool stats)
     /// by decrementing both the global and per-cursor counters.
@@ -437,6 +446,8 @@ impl Drop for CursorState {
         // The limiter holds a weak reference to `CursorState`, so this `Drop` runs when all strong
         // references are gone. `release_cursor` removes the reference and decrements the global reservation counter.
         self.pool.limiter().release_cursor(self.cursor_id, &self.mem_reserved);
+        // Releasing the cursor's reservation may have freed memory — wake any pending allocations.
+        self.pool.try_wake_pending();
     }
 }
 
@@ -459,8 +470,13 @@ impl CursorHandle {
 
     /// Record an active FUSE read. Returns a guard that clears it on drop and
     /// stamps the cursor's LRU tick at that point.
+    ///
+    /// Also upgrades any pending speculative allocation queue entries for this
+    /// cursor to high priority, since a read is now actively waiting.
     pub fn set_active_read(&self, offset: u64, size: usize) -> ActiveReadGuard {
         *self.state.read_state.lock().unwrap() = ReadState::Active { offset, size };
+        // Promote any speculative queue entries to high priority.
+        self.state.pool.upgrade_pending(self.state.cursor_id);
         ActiveReadGuard { state: self.state() }
     }
 
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -61,6 +61,10 @@ pub struct MemoryLimiter {
     cursors: Arc<DashMap<CursorId, Weak<CursorState>>>,
     /// Counter for generating unique [CursorId]s.
     next_cursor_id: AtomicU64,
+    /// Monotonic counter stamped onto a cursor's `last_read_at` whenever it
+    /// receives a new active read, so the pruner can pick the LRU one.
+    /// Pure ordering — does not represent wall-clock time.
+    next_read_tick: AtomicU64,
     /// Additional reserved memory for other non-buffer usage like storing metadata
     additional_mem_reserved: u64,
     /// Wakes the background pruning loop's outer wait when memory pressure starts.
@@ -84,6 +88,7 @@ impl MemoryLimiter {
             mem_reserved: Default::default(),
             cursors: Default::default(),
             next_cursor_id: AtomicU64::new(1),
+            next_read_tick: AtomicU64::new(1),
             additional_mem_reserved,
             pruning_signal: Arc::new(WakeSignal::new()),
         }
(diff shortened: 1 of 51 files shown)


mountpoint-s3-fs/src/memory/maintenance.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// One idle prefetch handle was dropped this round.
+    Acted,
+}
+
+/// Period of the pruning loop's inner tick while under memory pressure.
+pub(crate) const PRUNING_TICK: Duration = Duration::from_millis(1);
+/// If the head of the allocation queue has been waiting longer than this, the pruner
+/// will drop an idle prefetch handle even if uploads/active reads are in flight.
+/// Acts as a starvation backstop.
+pub(crate) const PRUNING_STARVATION_THRESHOLD: Duration = Duration::from_millis(5);
```
Priyankakarumuru1: nit: is the 5ms value based on any benchmarking or just a starting guess? Maybe helpful to have a comment explaining why 5ms specifically.

passaro:
> Mountpoint does not have dependency on tokio.

It does actually, but why is it relevant?

mountpoint-s3-fs/src/wake_signal.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// `notify` must wake a waiter blocked in `wait_timeout` well before
+    /// the timeout elapses. This is the primitive's defining behavior.
+    #[test]
+    fn notify_wakes_waiter_before_timeout() {
+        let signal = Arc::new(WakeSignal::new());
+        let signal_for_waiter = signal.clone();
+        let started = Arc::new(std::sync::Mutex::new(false));
+        let started_for_waiter = started.clone();
+        let waiter_started = Arc::new(std::sync::Condvar::new());
+        let waiter_started_clone = waiter_started.clone();
```
passaro: These add a lot of noise to the test. Could we use a oneshot channel instead? Assuming it is what you need here...

mountpoint-s3-fs/src/wake_signal.rs
```diff
@@ -0,0 +1,148 @@
...
+        while !*g {
+            g = waiter_started.wait(g).unwrap();
+        }
+        drop(g);
+        std::thread::sleep(Duration::from_millis(10));
+
+        signal.notify();
+        let elapsed = waiter.join().expect("waiter panicked");
+        assert!(
+            elapsed < TEST_WAIT_TIMEOUT,
```
passaro: Can we clarify what all the durations are for? (10ms, 1s, 60s)

mountpoint-s3-fs/src/wake_signal.rs
```diff
@@ -0,0 +1,148 @@
+//! A coalescing wake signal for parking a single consumer thread until an
```
passaro: nit: shall we introduce a top level `util` module to move this into? we could later use it for other "generic" types, e.g. `async_util`.

mountpoint-s3-fs/src/wake_signal.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// flag is set (consuming it) or once the timeout elapses, whichever
+    /// comes first. Spurious wakeups are handled by looping on the flag.
+    pub fn wait_timeout(&self, timeout: Duration) {
+        let pending = self.pending.lock().unwrap();
+        let (mut pending, _) = self.cvar.wait_timeout_while(pending, timeout, |p| !*p).unwrap();
+        *pending = false;
+    }
+}
+
+#[cfg(all(test, not(feature = "shuttle")))]
```
passaro: This is suprising: `not(feature = "shuttle")`. Why is it needed?

mountpoint-s3-fs/src/memory/maintenance.rs
```diff
@@ -0,0 +1,198 @@
...
+/// When there is no pressure, the drain loop exits immediately after one
+/// round (which still does the cheap pool trim), so the periodic trim is
+/// implicit in the same code path.
+///
+/// `thread::sleep` inside the inner drain is deliberately uninterruptible by
+/// notifies — under pressure the loop polls every [`PRUNING_TICK`] anyway,
+/// so there is nothing useful for an extra wake to do.
+fn maintenance_loop(pool_inner: Weak<PagedPoolInner>, signal: Arc<WakeSignal>, idle_interval: Duration) {
+    loop {
+        signal.wait_timeout(idle_interval);
```
passaro, line 68: maybe future improvement, but we could surface whether this was a timeout or a trigger

yerzhan7:
> It does actually

Oh, I thought we only have/use it in tests? https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-fs/Cargo.toml#L86

> , but why is it relevant?

I thought instead of `wake_signal.rs` thingy we could have used Tokio's Notify.

mountpoint-s3-fs/src/wake_signal.rs
yerzhan7: Decided to refactor unit tests a little bit in https://github.com/awslabs/mountpoint-s3/pull/1829/commits/10a92f1b9a2c4580c36bb815644dbe2d49a4aec5 to rely on return value of `wait_timeout`. Also, documented the magic duration numbers.

mountpoint-s3-fs/src/wake_signal.rs
yerzhan7: Good point, done in https://github.com/awslabs/mountpoint-s3/pull/1829/commits/a143f5e3073cbbdb20169a1de1f9a19f7c706e93

mountpoint-s3-fs/src/memory/maintenance.rs
yerzhan7: Added return value for now. We can later add metrics or logs.

mountpoint-s3-fs/src/wake_signal.rs
yerzhan7: Refactored to oneshot in https://github.com/awslabs/mountpoint-s3/pull/1829/commits/ae380202ffa9de603ecaa6bc3bf4846f9ba6a86f

mountpoint-s3-fs/src/wake_signal.rs
yerzhan7: Good point. Should not be needed, removed.

mountpoint-s3-fs/src/memory/limiter.rs
```diff
@@ -308,6 +358,12 @@ pub struct CursorState {
     mem_reserved: AtomicU64,
     /// Active FUSE read range. None when no read is in progress.
     active_read: Mutex<Option<ActiveRead>>,
+    /// Monotonic tick of the last `set_active_read`. `0` until the first read.
+    /// Used by the pruner to pick the least-recently-used idle cursor.
+    last_read_at: AtomicU64,
```
passaro: This is accessed always together with `active_read`. Could we consolidate into one mutex? Maybe:
```rust
enum ReadState {
  Active { offset, length },
  Last { tick }
}
```

passaro (commented):
Looks good. One (optional) note.

mountpoint-s3-fs/src/memory/limiter.rs
yerzhan7: Agree, I think this would be better. Refactored in https://github.com/awslabs/mountpoint-s3/pull/1829/commits/1b833eaf1914d52f1edf5e478ec5c4aa5e777d61

Thanks for reviewing!

mountpoint-s3-fs/src/memory/limiter.rs
```diff
@@ -191,11 +191,17 @@ impl MemoryLimiter {
...
-            .get(&cursor_id)
-            .and_then(|r| r.upgrade())
-            .and_then(|s| s.active_read.lock().unwrap().map(|r| r.overlaps(offset, size)))
-            .unwrap_or(false)
+        let Some(state) = self.cursors.get(&cursor_id).and_then(|r| r.upgrade()) else {
+            return false;
+        };
+        match *state.read_state.lock().unwrap() {
+            ReadState::Active { offset: o, size: s } => {
+                let self_end = o + s as u64;
```
passaro: nit: rename to `active_*`

passaro: Or even better, implement `is_active(offset,size)` on `ReadState`

passaro (commented):
Now we use the tick when the read ended. Does that have impact on pruning?

mountpoint-s3-fs/src/memory/limiter.rs
yerzhan7: Added `is_active(offset,size)` method, thanks

yerzhan7:
> Now we use the tick when the read ended. Does that have impact on pruning?

Should be fine in practice due to pruner's frequent periodic invocation (and there is always another "is active check" during reset). In some way it is more semantically correct - "least-recently-used".

mountpoint-s3-fs/src/memory/maintenance.rs
```diff
@@ -0,0 +1,330 @@
...
+    /// "queue non-empty", so an empty queue means pressure has already cleared.
+    Idle,
+    /// In-flight uploads or active reads will release buffers naturally; wait.
+    WaitingForRelease,
+    /// One idle cursor was reset this round.
+    Acted,
+}
+
+/// Period of the pruning loop's inner tick while under memory pressure.
+pub(crate) const PRUNING_TICK: Duration = Duration::from_millis(1);
```
passaro: nit: `pub(crate)` and `pub(super)` seem superfluous since the `maintenance` module is not `pub` in its parent (`memory` module).

mountpoint-s3-fs/src/memory/maintenance.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    // 2. If no waiter is queued, there's no pressure — exit the inner tick.
+    if !pool_inner.is_memory_pressure() {
+        return PruningOutcome::Idle;
+    }
+
+    let starving = pool_inner
+        .head_waited()
+        .is_some_and(|d| d >= PRUNING_STARVATION_THRESHOLD);
```
passaro, line 115: I wonder if we should always invoke `try_wake_pending` here and have it return some state, e.g.:
- no memory pressure
- "head_waited" interval

It would simplify handling of cancelled entries, which `head_waited` could end up skipping multiple times.

passaro: (not a blocker)

mountpoint-s3-fs/src/memory/maintenance.rs
```diff
@@ -0,0 +1,330 @@
...
+/// Each write handle holds at most one "filling" buffer (FUSE write data
+/// being staged) at any time without an in-flight request, so excess
+/// reserved bytes above that baseline indicate at least one part actually
+/// uploading. Requires tracking `upload_handles_holding_buffers` (likely
+/// the active-write-handles counter).
+fn has_uploads_in_flight(_pool_inner: &PagedPoolInner) -> bool {
+    false
+}
+
+#[cfg(all(test, not(feature = "shuttle")))]
```
passaro: "shuttle" feature should not be relevant

mountpoint-s3/src/run.rs
```diff
@@ -205,9 +205,10 @@ fn mount(args: CliArgs, client_builder: impl ClientBuilder) -> anyhow::Result<Fu
         ])
         .with_memory_limit(args.mem_limit())
         .build();
-    // Schedule trimming of empty memory pages every minutes. We should consider
-    // event-based triggers and/or a configurable interval in the future.
-    pool.schedule_trim(Duration::from_secs(60));
+    // Spawn the background pool maintenance thread. Performs periodic trim
+    // (every 60s) when idle, and runs pruning rounds on demand under memory
+    // pressure (woken by `MemoryLimiter::trigger_pruning`).
```
passaro: nit: no need to mention the trigger here, it's an implementation detail.

mountpoint-s3-fs/src/memory/maintenance.rs
yerzhan7: Looks like without this shuttle compilation fails: https://github.com/awslabs/mountpoint-s3/actions/runs/27288197157/job/80601173169?pr=1829

```

error[E0599]: no method named `is_finished` found for struct `shuttle::thread::JoinHandle<T>` in the current scope
Error:    --> mountpoint-s3-fs/src/memory/maintenance.rs:212:23
    |
212 |         while !handle.is_finished() {
    |                       ^^^^^^^^^^^ method not found in `shuttle::thread::JoinHandle<()>`

```

Added it back.

mountpoint-s3/src/run.rs
yerzhan7: Agree. I removed comment altogether. The code is self-explanatory enough.

mountpoint-s3-fs/src/memory/maintenance.rs
yerzhan7: Fixed, thanks

passaro (approved):

Issue #785, opened 2024-02-26
Stale file handle during reading rewritten object with cached metadata


mhnap:
### Mountpoint for Amazon S3 version

mount-s3 1.4.0

### AWS Region

us-east-2

### Describe the running environment

Running in local Ubuntu 22.04.4 LTS.

### Mountpoint options

```shell
mount-s3 mhnap-bucket /media/mount-s3 --read-only --log-directory=/tmp/mount-s3/logs --debug --debug-crt --cache /tmp/mount-s3 --metadata-ttl 1000000
```

### What happened?

This issue is similar to https://github.com/awslabs/mountpoint-s3/issues/782 which I previously created, but here I rewrite the object.

I am not sure whether this is expected behavior, so I would be grateful for clarifying.

The test I ran:
```
$ dd if=/dev/urandom of=file bs=1 count=10
10+0 records in
10+0 records out
10 bytes copied, 0.000198077 s, 50.5 kB/s
$ aws s3 cp file s3://mhnap-bucket/dir/file
upload: ./file to s3://mhnap-bucket/dir/file
$ mount-s3 mhnap-bucket /media/mount-s3 --read-only --log-directory=/tmp/mount-s3/logs --debug --debug-crt --cache /tmp/mount-s3 --metadata-ttl 1000000
bucket mhnap-bucket is mounted at /media/mount-s3
$ ls -lh /media/mount-s3/dir/file
-rw-r--r-- 1 mhnap mhnap 10 Feb 26 11:31 /media/mount-s3/dir/file
$ dd if=/dev/urandom of=file bs=1 count=100
100+0 records in
100+0 records out
100 bytes copied, 0.000530112 s, 189 kB/s
$ aws s3 cp file s3://mhnap-bucket/dir/file
upload: ./file to s3://mhnap-bucket/dir/file
$ ./read /media/mount-s3/dir/file
Err(Os { code: 116, kind: StaleNetworkFileHandle, message: "Stale file handle" })
$ ./read /media/mount-s3/dir/file
[...]

PR #1830, opened 2026-05-13
Add versionId support to metadata cache for versioning-enabled S3 buckets

millin:
## What changed and why?

Implement caching of S3 object `versionId` alongside metadata to improve read consistency when metadata cache TTL is valid but data cache is evicted.

### Problem

When metadata cache TTL is still valid but data cache is evicted, reads use cached metadata(size, ETag) to determine object boundaries. If the object was replaced in S3 between the initial metadata lookup and the read, the prefetcher uses incorrect cached size, potentially causing reads to silently return empty data instead of the new version.

For **versioning-enabled S3 buckets**, this problem can now be solved by pinning reads to the specific object version using the S3 `versionId` parameter.

### Solution

1. **Extract versionId from S3 responses**: Parse `x-amz-version-id` header from HeadObject responses in the client layer
2. **Cache versionId with metadata**: Store versionId in `InodeStat` alongside other cached metadata
3. **Pin reads to version**: When opening files for read, use `ObjectId::with_version()` to create an ObjectId that includes the versionId
4. **Inject into GetObject requests**: The prefetcher automatically includes `?versionId=` query parameter when available, ensuring reads fetch the exact version

### Result

For versioning-enabled buckets: Reads are pinned to the specific object version at the time of open, ensuring consistent data even if metadata TTL is valid but data cache is evicted or the object is replaced in S3.

For non-versioned buckets: Behavior unchanged — versionId is None, requests identical to before this change.

**No breaking changes.**

- All new fields are `Option<T>` defaulting to `None`
- `ObjectId::new()` signature unchanged; `with_version()` is additive only
- When `version_id` is `None`, requests are identical to pre-change behavior
- Non-versioned buckets continue to work exactly as before

## Does this change need a changelog entry?

**Yes.** 

This improves correctness/consistency for versioning-enabled S3 buckets. Changelog should note:
- Metadata cache now includes S3 object versionId for versioning-enabled buckets
- Reads on versioning-enabled buckets are now pinned to the version at open time, improving consistency when metadata TTL is valid but data cache is evicted

## Related Issues

Addresses the metadata cache staleness issue in #785 **for versioning-enabled S3 buckets**. 
Note: #785 also affects non-versioned buckets, which would require a different solution (e.g., more aggressive cache invalidation or alternative versioning strategies).

---

### DIFF
--- a/mountpoint-s3-client/src/mock_client.rs
+++ b/mountpoint-s3-client/src/mock_client.rs
@@ -549,6 +549,8 @@ pub struct MockObject {
     ///
     /// Typically, at most one of the checksums should be set.
     checksum: Checksum,
+    /// Version ID of the object (for versioning-enabled buckets)
+    version_id: Option<String>,
 }
 
 impl MockObject {
@@ -569,6 +571,7 @@ impl MockObject {
             parts: None,
             object_metadata: HashMap::new(),
             checksum: Checksum::empty(),
+            version_id: None,
         }
     }
 
@@ -583,7 +586,14 @@ impl MockObject {
             parts: None,
             object_metadata: HashMap::new(),
             checksum: Checksum::empty(),
+            version_id: None,
         }
+    }
+
+    /// Builder method to set the version ID
+    pub fn with_version_id(mut self, version_id: impl Into<String>) -> Self {
+        self.version_id = Some(version_id.into());
+        self
     }
 
     pub fn ramp(seed: u8, size: usize, etag: ETag) -> Self {
@@ -607,6 +617,7 @@ impl MockObject {
             parts: None,
             object_metadata: HashMap::new(),
             checksum: Checksum::empty(),
+            version_id: None,
         }
     }
 
@@ -959,6 +970,16 @@ impl ObjectClient for MockClient {
                 )));
             }
 
+            // Validate versionId if specified. Return NoSuchKey if version doesn't exist,
+            // which matches AWS S3 behavior when a specific version is requested but not found.
+            if let Some(version_id) = params.version_id.as_ref() {
+                if object.version_id.as_ref() != Some(version_id) {
+                    return Err(ObjectClientError::ServiceError(GetObjectError::NoSuchKey(
+                        Default::default(),
+                    )));
+                }
+            }
+
             let (next_offset, length) = if let Some(range) = params.range.as_ref() {
                 if range.start >= object.len() as u64 || range.end > object.len() as u64 {
                     return mock_client_error(format!("invalid range, length={}", object.len()));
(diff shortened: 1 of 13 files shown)


passaro:
> If the object was replaced in S3 between the initial metadata lookup and the read, the prefetcher uses incorrect cached size, potentially causing reads to silently return empty data instead of the new version.

Could you provide a bit more clarity on the problem? Mountpoint returns consistent data from reads on the same file handle, or errors out if unable to do so (i.e the object in the bucket has changed). In which cases are you seeing reads silently returning empty data?

millin:
@passaro Frankly, "silently return empty data" is a poor description. If the object is modified, the read will fail.
I was actually getting the "[Errno 116] Stale file handle" error in my python application.

millin:
@passaro Have you had a chance to review the proposed changes?

passaro:
@millin, sorry I couldn't reply earlier.

> If the object is modified, the read will fail.

That is intended behavior: once Mountpoint discovers that the object was modified, it will not return stale data. When using a longer TTL and the disk cache, it is possible for Mountpoint to serve data entirely from the cache without ever needing to access the bucket. In those cases, it will not discover that the object was modified and just return the cached data.

With that in mind, the change attempts to "extend" this behavior by fetching data from the bucket at a fixed version when possible. While I understand that this is preferable to an error in your case, it is not achieving the stated goal to improve consistency and may in fact be a breaking change in general.

In addition, the implementation highlights how the versionId is only discovered in the case where the metadata was obtained through a HeadObject request and is left blank when the metadata originated from a ListObjectV2 request. This gap creates another inconsistency which could be very confusing for users: they could get errors or data from previous object versions depending on whether they had previously issued an `ls` command. 

Without a way to mitigate the latter issue, I don't think we can accept this contribution, even as an opt-in change.

millin:
@passaro Thanks for the feedback!
You're right about the gap in ListObjectsV2—it was indeed an issue. We solved it by always creating a HeadObject when opening a read, regardless of the TTL setting. This ensures that the versionId is always current when opening, so the behavior no longer depends on whether ls was previously called. As a result, reads are pinned to the version that existed at open time — replacing the object on S3 won't cause ESTALE mid-read.
The only tradeoff is one additional HeadObject per read when using --metadata-ttl, which we think is acceptable.

passaro:
>  The only tradeoff is one additional HeadObject per read when using --metadata-ttl, which we think is acceptable.

This may not be a warranted trade off, especially as the default behavior when using the cache:
* the additional cost, as implemented, will actually be a pair of ListObjectV2 + HeadObject requests,
* it results in a known stale version of the object being served, even when accessed for the first time rather than previously cached (breaking change).

passaro:
@millin, thanks for your effort, but we cannot accept this contribution.

I would recommend opening an issue with your original problem so we can discuss potential solutions and the trade off involved before focusing on the implementation.

PR #1831, opened 2026-05-14
Add concurrent write-handle limit returning ENOMEM on open()

yerzhan7:
Introduce `WriteHandleLimiter`, an admission-control type that caps the number
of files open for write at the same time. The cap is derived from the
configured memory target and write part size:

`max_concurrent_writes = (memory_target - additional_mem_reserved) / write_part_size`

When the cap is reached, `open()` for write returns ENOMEM before any inode
state is mutated, so retries are clean. The reserved slot is held by the
returned `WriteHandleSlot`; its `Drop` releases the slot when the file handle is
closed.

The limiter is enforced inside `Metablock::open_handle` so the slot acquire
happens before `start_writing` mutates the inode.

Note on ordering: the write-handle limiter check runs *before* the inode-conflict
check inside `open_handle`. This is intentional — the limiter check is a lock-free
atomic and produces no state to roll back, so it's the natural fail-fast point.
The trade-off is that when both errors apply (cap exhausted AND same file
already open for write), the user sees `ENOMEM` rather than the more specific
`EPERM`. We accept this; it's a rare double-failure case and the user retries
either way.

Yes.

Yes. Intentionally skipped changelog entry and version bumps for now to avoid merge conflicts and because we are working on a feature branch.

---

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -295,6 +295,18 @@ At mount time, Mountpoint automatically selects appropriate defaults to provide 
 * By default, Mountpoint can serve up to 16 concurrent file or directory operations, and automatically scales up to reach this limit. If your application makes more than this many concurrent reads and writes (including to the same or different files), you can improve performance by increasing this limit with the `--max-threads` command-line argument. Higher values of this flag might cause Mountpoint to use more of your instance's resources.
 * When reading or writing files to S3, Mountpoint divides them into parts and uses parallel requests to improve throughput. You can change the part size Mountpoint uses for these parallel requests using the `--read-part-size` and `--write-part-size` command-line arguments, providing a maximum number of bytes per part for reading or writing respectively. For Mountpoint v1.7.2 or earlier, use `--part-size` instead. The default value for these arguments is 8 MiB (8,306,688 bytes), which in our testing is the largest value that achieves maximum throughput. Larger values can reduce the number of billed requests Mountpoint makes, but also reduce the throughput of object reads and writes to S3.
 
+### Maximum number of files open for write
+
+Mountpoint enforces a cap on the number of files that may be open for write at the same time, to prevent out-of-memory crashes. The cap is computed at startup from the configured memory target and write part size:
+
+```
+max_concurrent_writes = (memory_target − additional_mem_reserved) / write_part_size
+```
+
+`memory_target` is set with `--memory-target` and `write_part_size` is set with `--write-part-size` (or with `--part-size`). `additional_mem_reserved` is `max(128 MiB, memory_target / 8)` and is held back from data buffers for Mountpoint's own overhead. The minimum supported `memory_target` is 512 MiB, which allows 48 concurrent writers at the default 8 MiB write part size.
+
+Once the cap is reached, `open()` calls for write return `ENOMEM` ("Cannot allocate memory") until an existing write handle is closed. To raise the cap, increase `--memory-target` or decrease `--write-part-size`.
+
 ### Maximum object size
 
 In its default configuration, there is no maximum on the size of objects Mountpoint can read. However, Mountpoint uses [multipart upload](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html) when writing new objects, and multipart upload allows a maximum of 10,000 parts for an object. This means Mountpoint can only upload objects up to 80,000 MiB (78.1 GiB) in size. If your application tries to write objects larger than this limit, writes will fail with an out of space error.
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -295,9 +295,9 @@ At mount time, Mountpoint automatically selects appropriate defaults to provide 
 * By default, Mountpoint can serve up to 16 concurrent file or directory operations, and automatically scales up to reach this limit. If your application makes more than this many concurrent reads and writes (including to the same or different files), you can improve performance by increasing this limit with the `--max-threads` command-line argument. Higher values of this flag might cause Mountpoint to use more of your instance's resources.
 * When reading or writing files to S3, Mountpoint divides them into parts and uses parallel requests to improve throughput. You can change the part size Mountpoint uses for these parallel requests using the `--read-part-size` and `--write-part-size` command-line arguments, providing a maximum number of bytes per part for reading or writing respectively. For Mountpoint v1.7.2 or earlier, use `--part-size` instead. The default value for these arguments is 8 MiB (8,306,688 bytes), which in our testing is the largest value that achieves maximum throughput. Larger values can reduce the number of billed requests Mountpoint makes, but also reduce the throughput of object reads and writes to S3.
 
-### Maximum number of files open for write
+### Maximum number of files open for writing
 
-Mountpoint enforces a cap on the number of files that may be open for write at the same time, to control memory usage. The cap is computed at startup from the configured memory target and write part size:
+Mountpoint enforces a cap on the number of files that may be open for writing at the same time, to control memory usage. The cap is computed at startup from the configured memory target and write part size:
 
 ```
 max_concurrent_writes = (memory_target − additional_mem_reserved) / write_part_size
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -297,13 +297,13 @@ At mount time, Mountpoint automatically selects appropriate defaults to provide 
 
 ### Maximum number of files open for write
 
-Mountpoint enforces a cap on the number of files that may be open for write at the same time, to prevent out-of-memory crashes. The cap is computed at startup from the configured memory target and write part size:
+Mountpoint enforces a cap on the number of files that may be open for write at the same time, to control memory usage. The cap is computed at startup from the configured memory target and write part size:
 
 ```
 max_concurrent_writes = (memory_target − additional_mem_reserved) / write_part_size
 ```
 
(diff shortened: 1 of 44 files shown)


mountpoint-s3-fs/src/superblock.rs
```diff
@@ -814,6 +815,7 @@ impl<OC: ObjectClient + Send + Sync + Clone> Metablock for Superblock<OC> {
         fh: u64,
         write_mode: &WriteMode,
         flags: OpenFlags,
+        write_handle_limiter: Option<&Arc<WriteHandleLimiter>>,
```
passaro: nit: could we avoid the reference to `Arc`? e.g. by the limiter holding an `Arc<AtomicU64>` internally?

mountpoint-s3-fs/src/write_handle_limiter.rs
```diff
@@ -0,0 +1,213 @@
...
+    fn test_max_concurrent_writes_at_512mb_8mb() {
+        let limiter = WriteHandleLimiter::new(MIN_LIMIT, MIN_BUDGET, PART_SIZE_8MIB);
+        assert_eq!(limiter.max_concurrent_writes(), 48);
+    }
+
+    #[test]
+    fn test_max_concurrent_writes_at_4gb_8mb() {
+        let limiter = WriteHandleLimiter::new(LARGE_LIMIT, LARGE_BUDGET, PART_SIZE_8MIB);
+        assert_eq!(limiter.max_concurrent_writes(), 448);
+    }
```
passaro: nit: use `test_case`

mountpoint-s3-fs/tests/fuse_tests/write_test.rs
```diff
@@ -1380,7 +1380,7 @@ fn write_with_sse_settings_test(policy: &str, sse: ServerSideEncryption, should_
 }
 
 #[cfg(feature = "s3_tests")]
-#[test_case(200)]
+#[test_case(48)]
```
passaro: Consider explicitly setting the memory limit in this test, so we can make clear that we want to test the maximum number of writers.

mountpoint-s3-fs/src/lib.rs
```diff
@@ -21,6 +21,7 @@ pub mod s3;
 mod superblock;
 mod sync;
 pub mod upload;
+pub mod write_handle_limiter;
```
passaro: Consider moving under `memory` or `fs`.

mountpoint-s3-fs/src/fs.rs
```diff
@@ -1090,4 +1104,134 @@ mod tests {
...
+    /// the formula gives `(256 - 128) / 32 = 4` concurrent writers.
+    #[tokio::test]
+    async fn test_open_for_write_returns_enomem_when_cap_exhausted() {
+        let test_name = "test_open_for_write_returns_enomem_when_cap_exhausted";
+        let bucket = Bucket::new("bucket").unwrap();
+        let client = MockClient::config()
+            .bucket(bucket.to_string())
+            .enable_backpressure(true)
+            .initial_read_window_size(1024 * 1024)
+            .part_size(32 * 1024 * 1024)
```
passaro: Could we define this and other values as constants at the top of this function? And move the calculation described in the rustdoc there.

mountpoint-s3-fs/src/fs/handles.rs
```diff
@@ -23,6 +24,11 @@ where
     pub state: AsyncMutex<FileHandleState<Client>>,
     /// Process that created the handle
     pub open_pid: u32,
+    /// Slot reserved on the [`MemoryLimiter`] for this handle. `Some` for write handles, `None`
+    /// for read handles. Released automatically when the `FileHandle` is dropped — held purely
+    /// for that `Drop` side effect, so the field is never read directly.
+    #[expect(dead_code, reason = "held for its Drop side effect")]
+    pub(super) write_slot: Option<WriteHandleSlot>,
```
passaro: Shouldn't this be in `FileHandleState::Write`?

passaro: Also, prefer `_write_slot` to the dead code expect.

mountpoint-s3-fs/src/fs.rs
```diff
@@ -349,13 +357,18 @@ where
 
         let fh = self.next_handle(); // TODO: can we delay obtaining the next handle until we know we are creating a new file handle?
         let write_mode = self.config.write_mode();
-        let new_handle = self.metablock.open_handle(ino, fh, &write_mode, flags).await?;
+        let mut new_handle = self
+            .metablock
+            .open_handle(ino, fh, &write_mode, flags, Some(&self.write_handle_limiter))
+            .await?;
+        let write_slot = new_handle.write_slot.take();
```
passaro: Why `take()` and `mut new_handle`? We don't actually want to mutate anything.

If this was just to "fix" lifetimes, consider instead unpacking or copying the data you will need later.

doc/CONFIGURATION.md
```diff
@@ -295,6 +295,18 @@ At mount time, Mountpoint automatically selects appropriate defaults to provide
...
 * When reading or writing files to S3, Mountpoint divides them into parts and uses parallel requests to improve throughput. You can change the part size Mountpoint uses for these parallel requests using the `--read-part-size` and `--write-part-size` command-line arguments, providing a maximum number of bytes per part for reading or writing respectively. For Mountpoint v1.7.2 or earlier, use `--part-size` instead. The default value for these arguments is 8 MiB (8,306,688 bytes), which in our testing is the largest value that achieves maximum throughput. Larger values can reduce the number of billed requests Mountpoint makes, but also reduce the throughput of object reads and writes to S3.
 
+### Maximum number of files open for write
+
+Mountpoint enforces a cap on the number of files that may be open for write at the same time, to prevent out-of-memory crashes. The cap is computed at startup from the configured memory target and write part size:
```
passaro: Can we say "to control memory usage" instead of oom crashes?

passaro: We should start mentioning what's the default for memory target (and thus max writes).

doc/METRICS.md
```diff
@@ -86,6 +86,7 @@ Mountpoint emits the following metrics:
 
 | Metric | Type | Dimensions | Description |
 |--------|------|------------|-------------|
+| `fs.write_handle_limit_exceeded` | Counter | | Number of `open()` calls for write rejected because the [concurrent-writers cap](CONFIGURATION.md#maximum-number-of-files-open-for-write) was reached |
```
passaro: Before documenting the new metric, we should add it to `metrics::defs`. Or we can leave it for a later review of all new metrics.

mountpoint-s3-fs/tests/fuse_tests/write_test.rs
yerzhan7: Done, thanks. Also, added back the 200+ test case to ensure no regression.

mountpoint-s3-fs/src/superblock.rs
yerzhan7: Yes, thanks!

mountpoint-s3-fs/src/lib.rs
yerzhan7: Done

mountpoint-s3-fs/src/fs/handles.rs
yerzhan7: Good suggestion - done.

mountpoint-s3-fs/src/fs.rs
yerzhan7: Make sense, done.

doc/CONFIGURATION.md
yerzhan7: Done. Added mentioning of the default value.

doc/METRICS.md
yerzhan7: Didn't know about the `defs` file.

Removed from docs for now - we can review all new metrics later as you've suggested.

mountpoint-s3-fs/src/fs.rs
yerzhan7: Done

doc/SEMANTICS.md
```diff
@@ -89,6 +89,8 @@ Your application should not write to the same object from multiple instances at
 
 By default, Mountpoint ensures that new file uploads to a single key are atomic. As soon as an upload completes, other clients are able to see the new key and the entire content of the object. If the `--incremental-upload` flag is set, however, Mountpoint may issue multiple separate uploads during file writes to append data to the object. After each upload, the appended object in your S3 bucket will be visible to other clients.
 
+Mountpoint enforces a cap on the number of files that may be open for write at the same time, derived from `--memory-target` and `--write-part-size`. When the cap is reached, `open()` for write returns `ENOMEM` until an existing write handle is closed. See [CONFIGURATION.md](CONFIGURATION.md#maximum-number-of-files-open-for-write) for more details.
```
Priyankakarumuru1: nit: "open for write" vs "open for writing" - the rest of the doc uses "open for writing" (e.g., "open a file for writing"). This sentence uses "open for write". Minor inconsistency.

mountpoint-s3-fs/src/memory/write_handle_limiter.rs
```diff
@@ -0,0 +1,214 @@
...
+            warn!(
+                mem_limit_mib,
+                write_part_size_mib,
+                "no concurrent open-for-write file handles allowed: memory target leaves no \
+                 budget for write buffers. All write opens will fail with ENOMEM. Increase \
+                 --memory-target or decrease --write-part-size. See CONFIGURATION.md \
+                 (\"Maximum number of files open for write\") for details."
+            );
+        } else {
+            info!(
```
Priyankakarumuru1, line 87: Nit: The warn! above uses structured fields, maybe the info! could also match that style for consistency.

Priyankakarumuru1 (approved):

doc/SEMANTICS.md
yerzhan7: Good catch, thanks, fixed in https://github.com/awslabs/mountpoint-s3/pull/1831/changes/dcb118d927a70a7f6d61cd57696afb48d2562c71

PR #1832, opened 2026-05-14
Introduce CursorState to consolidate per-cursor memory tracking

passaro:
Consolidates per-cursor state in the memory limiter into a single `CursorState` struct, eliminating hot-path lookups from the backpressure controller and cursor read path. 

Previously, per-cursor data was split across two maps in MemoryLimiter: one for reservation balances and one for active FUSE read ranges. Every `reserve`, `try_reserve`, `set_active_read`, and `clear_active_read` call performed an independent lookup on the hot path. This change introduces `CursorState`, a shared struct holding both the reservation counter and active read state, and a `CursorHandle` wrapper for setting active reads. The `BackpressureController` and `Cursor` directly hold a reference to `CursorState`, so reservation and active-read operations no longer require any lookup. The map is now only accessed on cursor creation, release, and in the `on_pool_reserve` callback (same as before, but unified into a single map).

No behavioral changes. Existing unit tests adapted to the new API.

No.

---

### DIFF
--- a/mountpoint-s3-fs/src/memory.rs
+++ b/mountpoint-s3-fs/src/memory.rs
@@ -5,6 +5,8 @@ mod pool;
 mod stats;
 
 pub use buffers::{PoolBuffer, PoolBufferMut};
-pub use limiter::{ActiveRead, ActiveReadGuard, BufferArea, MINIMUM_MEM_LIMIT, effective_total_memory};
+pub use limiter::{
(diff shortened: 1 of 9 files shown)


yerzhan7 (approved):
Nice refactor! LGTM

renanmagagnin (approved):
LGTM

Issue #1480, opened 2025-06-24
Support configuring the trusted certificate bundle for Mountpoint


dannycjones:
### Tell us more about this new feature.

Many AWS clients support configuring the trusted certificate bundle through the environment variable `AWS_CA_BUNDLE`, allowing for Mountpoint to use a CA bundle different from the one configured by the operating system. This is documented here: https://docs.aws.amazon.com/sdkref/latest/guide/feature-gen-config.html

This can be useful for testing, the use of proxies, etc..

There is an open feature request for this functionality on the CSI driver repo: https://github.com/awslabs/mountpoint-s3-csi-driver/issues/514. Note, that it is not possible at this time to configure the environment for the Mountpoint pod used by CSI Driver V2 - so simply adding `AWS_CA_BUNDLE` support to Mountpoint would not be enough.

yerzhan7:
`--ca-bundle` CLI flag and `AWS_CA_BUNDLE` has been added in `v1.23.0`.

CSI Driver support is tracked separately in https://github.com/awslabs/mountpoint-s3-csi-driver/issues/514

PR #1834, opened 2026-05-18
Add --ca-bundle and AWS_CA_BUNDLE support

yerzhan7:
### Description

Issue: https://github.com/awslabs/mountpoint-s3/issues/1480

Add `--ca-bundle` and `AWS_CA_BUNDLE` support (similarly to AWS CLI) as per https://docs.aws.amazon.com/sdkref/latest/guide/feature-gen-config.html

- Allows customers to specify the path to a custom certificate bundle (a file with a .pem extension) to use when establishing SSL/TLS connections. 
- This overrides OS default trust store.
- Used when constructing S3 Client and Credential Client
- Same precedence as in AWS CLI (flag > env variable)
- No support for `ca_bundle` from AWS config file (maybe added later as non-breaking change if there is usecase)

No - new opt-in feature/flag.

Yes - done.

---


mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -272,6 +276,109 @@ impl S3ClientConfig {
...
+    #[error("{role} '{}' is not a regular file", path.display())]
+    NotARegularFile {
+        /// A human-friendly role for the path ("CA bundle", ...).
+        role: &'static str,
+        /// The path that failed validation.
+        path: PathBuf,
+    },
+}
+
+fn validate_readable_file(path: &Path, role: &'static str) -> Result<(), TlsConfigValidationError> {
```
rajdchak, line 345: Do we want to add a validation for misconfigured path per OS? something which doc talks about https://docs.aws.amazon.com/sdkref/latest/guide/feature-gen-config.html ``` differences in how operating systems handle paths and escaping of path characters ```

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -1942,4 +2071,75 @@ mod tests {
         );
         assert_eq!(operation_name_to_static_metrics_string(None), "Unknown");
     }
+
+    // ---- TlsConfig tests ----
```
rajdchak: why are tests part of this file and not a separate test file?

mountpoint-s3-client/src/s3_crt_client.rs
yerzhan7: I think we're already covered here. The "escaping" caveat in that doc is guidance for the user. Mountpoint receives escaped path anyway already from clap/shell.

mountpoint-s3-client/src/s3_crt_client.rs
yerzhan7: Good point, I moved it into `tls.rs` file.

doc/CONFIGURATION.md
```diff
@@ -215,6 +215,18 @@ We also support the `AWS_ENDPOINT_URL` environment variable. The endpoint determ
 - Use `AWS_ENDPOINT_URL` if provided.
 - Fallback to automatically inferring the endpoint.
 
+### Custom CA trust bundle
+
+To verify HTTPS connections against a private certificate authority, pass a PEM file containing the CA certificate chain with `--ca-bundle`:
+
+```
+mount-s3 amzn-s3-demo-bucket /path/to/mount --ca-bundle /path/to/ca.pem
```
rajdchak, line 223: can we document how to configure ca-bundle when using mountpoint via csi driver?

doc/CONFIGURATION.md
```diff
@@ -215,6 +215,18 @@ We also support the `AWS_ENDPOINT_URL` environment variable. The endpoint determ
...
+```
+
+When set, this bundle is used in place of the operating-system default trust store for all HTTPS connections Mountpoint makes.
```
rajdchak, line 226: just for my knowledge, if the pem files are rotated after mount will the mountpoint process use the new rotated certs?

mountpoint-s3-client/src/s3_crt_client.rs
rajdchak: I was thinking do we need to validate whether the pem file contains invalid/expired certs?

doc/CONFIGURATION.md
yerzhan7: No - PEM file is only read once at mount time.

mountpoint-s3-client/src/s3_crt_client.rs
yerzhan7: Do not see a reason to duplicate the logic that CRT already does for us.

doc/CONFIGURATION.md
yerzhan7: CSI Driver is out of scope of this PR/repo.

CSI Driver might need separate design first to support this feature. For now, we will probably denylist this flag as it's not usable - customers currently cannot supply PEM file inside Mountpoint Pods.

rajdchak: will you do a followup for denylisting the flag? 
Also, should we update the doc to explicitly mention that this flag is not usable via CSI driver

rajdchak: AI: To create an issue in CSI driver to track CA support.

rajdchak (approved):

doc/CONFIGURATION.md
yerzhan7: CSI Driver issue already exist: https://github.com/awslabs/mountpoint-s3-csi-driver/issues/514

PR #1836, opened 2026-05-22
Add priority-ordered allocation queue

Priyankakarumuru1:
This adds a priority-ordered allocation queue (memory/allocation_queue.rs) that will be used to manage buffer allocation requests when memory is full. Instead of failing when there's no memory available, requests can now wait in the queue and get served in priority order - high priority and low priority.
  
The queue exposes `push_high`, `push_low`, `pop_front_if`, `upgrade`, `has_pending` and `fulfill`. It doesn't allocate buffers or check available memory itself, the pool will drive the wake loop by calling `pop_front_if` with a `can_allocate` predicate, then allocating and delivering the buffer. The wiring into the pool is separate work.

No. This is a new standalone module with no callers yet. Existing behavior is unchanged.

No, these are internal changes to the memory limiter with no user-visible behavior change.

---

### DIFF
--- a/mountpoint-s3-fs/src/memory.rs
+++ b/mountpoint-s3-fs/src/memory.rs
@@ -1,5 +1,6 @@
+// TODO(memory-limiter): remove once wired into PagedPool
 #[allow(unused)]
-pub(crate) mod allocation_queue;
+mod allocation_queue;
 mod buffers;
 mod limiter;
 mod pages;
--- a/mountpoint-s3-fs/src/memory.rs
+++ b/mountpoint-s3-fs/src/memory.rs
@@ -1,3 +1,5 @@
+#[allow(unused)]
+pub(crate) mod allocation_queue;
 mod buffers;
 mod limiter;
 mod pages;
--- a/mountpoint-s3-fs/src/memory/allocation_queue.rs
+++ b/mountpoint-s3-fs/src/memory/allocation_queue.rs
@@ -17,7 +17,7 @@ use super::stats::BufferKind;
 /// [`Low`](Self::Low) entries (FIFO). A [`Low`] request can be promoted to
 /// the back of [`High`] via [`AllocationQueue::upgrade`].
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum AllocationPriority {
+enum AllocationPriority {
     Low,
     High,
 }
@@ -86,19 +86,33 @@ impl AllocationQueue {
         }
     }
 
-    /// Enqueue a buffer allocation request. Returns a receiver that resolves to the
-    /// allocated [PoolBuffer] when fulfilled, or `Err(Canceled)` if the sender is dropped.
-    pub fn push(
+    /// Enqueue a high-priority buffer allocation request.
+    /// Returns a receiver that resolves to the allocated [PoolBuffer] when fulfilled,
+    /// or `Err(Canceled)` if the sender is dropped.
+    pub fn push_high(
+        &self,
+        cursor_id: Option<CursorId>,
+        size: usize,
+        kind: BufferKind,
+    ) -> oneshot::Receiver<PoolBuffer> {
+        self.push_inner(AllocationPriority::High, cursor_id, size, kind)
+    }
+
+    /// Enqueue a low-priority buffer allocation request.
+    /// Low-priority requests always have a cursor (speculative prefetch).
+    /// Returns a receiver that resolves to the allocated [PoolBuffer] when fulfilled,
+    /// or `Err(Canceled)` if the sender is dropped.
+    pub fn push_low(&self, cursor_id: CursorId, size: usize, kind: BufferKind) -> oneshot::Receiver<PoolBuffer> {
+        self.push_inner(AllocationPriority::Low, Some(cursor_id), size, kind)
+    }
+
+    fn push_inner(
         &self,
         priority: AllocationPriority,
         cursor_id: Option<CursorId>,
         size: usize,
         kind: BufferKind,
     ) -> oneshot::Receiver<PoolBuffer> {
-        assert!(
-            priority == AllocationPriority::High || cursor_id.is_some(),
-            "Low priority requests must have a cursor_id"
-        );
         let (sender, receiver) = oneshot::channel();
         let entry = PendingAllocation {
             cursor_id,
@@ -119,9 +133,10 @@ impl AllocationQueue {
 
     /// Atomically peeks at the front entry and removes it if `predicate` returns `true`.
     ///
-    /// Checks the high-priority list first, then low. The predicate receives the
-    /// entry's size, kind, and cursor_id so the caller can decide (e.g., check
-    /// `can_allocate(size)`) without a race between peek and pop.
+    /// Checks the high-priority list first, then low. Cancelled entries (where the
+    /// receiver was dropped) are pruned from the front of each queue before checking
+    /// the predicate — this prevents a large cancelled entry from blocking smaller
+    /// live entries behind it.
     ///
     /// Returns `None` if both queues are empty or the predicate returns `false`.
     /// Sets `has_pending` to `false` if both queues become empty after removal.
@@ -130,7 +145,21 @@ impl AllocationQueue {
         predicate: impl FnOnce(usize, BufferKind, Option<CursorId>) -> bool,
     ) -> Option<PendingAllocation> {
         let mut inner = self.inner.lock().unwrap();
-        let front = inner.high.front().or_else(|| inner.low.front())?;
+
+        // Prune cancelled entries from the front of each queue.
+        while inner.high.front().is_some_and(|e| e.sender.is_canceled()) {
+            inner.high.pop_front();
+        }
+        while inner.low.front().is_some_and(|e| e.sender.is_canceled()) {
+            inner.low.pop_front();
+        }
+
+        let front = inner.high.front().or_else(|| inner.low.front());
+        let Some(front) = front else {
+            self.has_pending.store(false, Ordering::Release);
+            return None;
+        };
+
         if !predicate(front.size, front.kind, front.cursor_id) {
             return None;
         }
@@ -151,11 +180,17 @@ impl AllocationQueue {
         let n = inner.low.len();
         for _ in 0..n {
             let entry = inner.low.pop_front().unwrap();
+            if entry.sender.is_canceled() {
+                continue;
+            }
             if entry.cursor_id == Some(cursor_id) {
                 inner.high.push_back(entry);
             } else {
                 inner.low.push_back(entry);
             }
+        }
+        if inner.high.is_empty() && inner.low.is_empty() {
+            self.has_pending.store(false, Ordering::Release);
         }
     }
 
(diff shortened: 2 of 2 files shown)


mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -0,0 +1,332 @@
+//! Priority-ordered allocation queue for buffer requests under memory pressure.
+
+use std::collections::VecDeque;
+use std::sync::Mutex;
+use std::sync::atomic::{AtomicBool, Ordering};
```
yerzhan7: I think we should import from `crate::sync` so that Shuttle tests work.

I.e.

```rs
  use crate::sync::Mutex;
  use crate::sync::atomic::{AtomicBool, Ordering};
```

mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -0,0 +1,332 @@
...
+/// This queue does not allocate buffers or check available memory — it only manages
+/// ordering, signaling, and lifecycle. The pool drives the wake loop by calling
+/// [`pop_front_if`](Self::pop_front_if) with a predicate (e.g., `can_allocate`),
+/// then performing the allocation and delivering the buffer via the entry's sender.
+pub struct AllocationQueue {
+    /// Queue state protected by a mutex.
+    inner: Mutex<AllocationQueueInner>,
+    /// `true` if either queue has pending entries. Allows a lock-free check
+    /// via [`is_memory_pressure`](Self::is_memory_pressure).
+    memory_pressure: AtomicBool,
```
yerzhan7: I think we should call it `is_empty` instead of `memory_pressure`? And then callers can decide if it means "memory pressure" or not. What do you think?

mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -0,0 +1,332 @@
...
+use super::buffers::PoolBuffer;
+use super::stats::BufferKind;
+
+/// Priority levels for allocation requests.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AllocationPriority {
+    /// Speculative prefetch — nobody is waiting on this data yet.
+    Low,
+    /// Active read or upload — a syscall is blocked waiting for this buffer.
+    High,
```
yerzhan7: The comment leaks the caller concepts. Let's make comment for generic for callers/consumers of the queue?

```
  /// Priority for an allocation request.
  ///
  /// The queue serves all [High] entries (FIFO) before any [Low] entries (FIFO).
  /// A [Low] request can be promoted to the back of [High] via
  /// [`AllocationQueue::upgrade`].
  #[derive(Debug, Clone, Copy, PartialEq, Eq)]
  struct AllocationPriority {
```

mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -0,0 +1,332 @@
...
+        let mut inner = self.inner.lock().unwrap();
+        let front = inner.high.front().or_else(|| inner.low.front())?;
+        if !predicate(front.size, front.kind, front.cursor_id) {
+            return None;
+        }
+        let entry = inner.high.pop_front().or_else(|| inner.low.pop_front());
+        if inner.high.is_empty() && inner.low.is_empty() {
+            self.memory_pressure.store(false, Ordering::Release);
+        }
+        entry
```
yerzhan7, line 170: Should we return `entry` here if Receiver was cancelled? (`entry.sender.is_canceled()`, e.g. when speculative prefetch is dropped)

Or are we planning to have separate `cancel(CursorID)` API on this queue?

mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -0,0 +1,332 @@
...
+
+    /// Moves all entries for `cursor_id` from the low-priority queue to the back
+    /// of the high-priority queue. No-op if no entries match.
+    ///
+    /// Called when a FUSE read arrives for a cursor that has pending speculative
+    /// allocations — those allocations are now urgent.
+    pub fn upgrade(&self, cursor_id: CursorId) {
+        let mut inner = self.inner.lock().unwrap();
+        let mut to_promote = Vec::new();
+        let mut remaining = VecDeque::with_capacity(inner.low.len());
```
yerzhan7: This will overallocate when not strictly needed. What if just do in-place rotation?

```
let mut inner = self.inner.lock().unwrap();
let n = inner.low.len();
for _ in 0..n {
    let entry = inner.low.pop_front().unwrap();
    if entry.cursor_id == Some(cursor_id) {
        inner.high.push_back(entry);
    } else {
        inner.low.push_back(entry);
    }
}
```

mountpoint-s3-fs/src/memory.rs
```diff
@@ -1,12 +1,12 @@
+#[allow(unused)]
+pub(crate) mod allocation_queue;
```
yerzhan7: Should it be just `mod allocation_queue;` to be consistent with modules below?

mountpoint-s3-fs/src/memory.rs
```diff
@@ -1,12 +1,12 @@
+#[allow(unused)]
```
yerzhan7, line 2: Could we add TODO comment here so that we don't forget? e.g. `// TODO(memory-limiter): remove once wired into PagedPool`

mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -0,0 +1,332 @@
...
+    /// Which cursor is requesting this buffer. `None` for upload requests
+    /// (uploads have no associated cursor).
+    pub cursor_id: Option<CursorId>,
+    /// Size of the buffer the waiter needs, in bytes.
+    pub size: usize,
+    /// The kind of buffer to allocate (e.g. [BufferKind::GetObject], [BufferKind::PutObject]).
+    pub kind: BufferKind,
+    /// Channel sender used to deliver the allocated [PoolBuffer] to the waiter.
+    /// When dropped, the receiver resolves with `Err(Canceled)`.
+    pub sender: oneshot::Sender<PoolBuffer>,
```
yerzhan7: np: `sender` could be private behind a `fulfill(self, buffer) -> Result<(), PoolBuffer>` method on `PendingAllocation` struct.

mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -0,0 +1,332 @@
...
+            kind,
+            sender,
+        };
+
+        let mut inner = self.inner.lock().unwrap();
+        match priority {
+            AllocationPriority::High => inner.high.push_back(entry),
+            AllocationPriority::Low => inner.low.push_back(entry),
+        }
+        self.memory_pressure.store(true, Ordering::Release);
```
yerzhan7: np: `self.memory_pressure.store(true, Ordering::Release);` can be moved after `let mut inner = self.inner.lock().unwrap();`

mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -0,0 +1,332 @@
...
+        assert_eq!(second.cursor_id, Some(CursorId::new_from_raw(2)));
+    }
+
+    #[test]
+    fn test_upgrade_moves_low_to_high() {
+        let queue = AllocationQueue::new();
+        let cursor_a = CursorId::new_from_raw(1);
+        let cursor_b = CursorId::new_from_raw(2);
+
+        let _rx_a = queue.push(AllocationPriority::Low, Some(cursor_a), 1024, BufferKind::GetObject);
```
yerzhan7: Can we also test/assert that if we push ` cursor_id: None` (uploads) and then call `upgrade(some_cursor)` the upload request is still in the correct place in the queue?

mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -0,0 +1,332 @@
...
+            cursor_id,
+            size,
+            kind,
+            sender,
+        };
+
+        let mut inner = self.inner.lock().unwrap();
+        match priority {
+            AllocationPriority::High => inner.high.push_back(entry),
+            AllocationPriority::Low => inner.low.push_back(entry),
```
yerzhan7, line 128: Should we fail if we trying to push Low priority request without cursor ID? I.e. such request must always be high priority?

mountpoint-s3-fs/src/memory/allocation_queue.rs
Priyankakarumuru1: Good idea. Added a test for that. Thanks!

mountpoint-s3-fs/src/memory/allocation_queue.rs
Priyankakarumuru1: Good one.

mountpoint-s3-fs/src/memory/allocation_queue.rs
Priyankakarumuru1: For now we aren't checking for cancelled receivers here. What happens is that if a cursor is dropped while its entry is in the queue, the receiver goes away. When pop_front_if pops that entry and the pool calls fulfill(buffer), the send fails and the buffer is dropped immediately. We can add a cancel API later as an optimization if needed.

mountpoint-s3-fs/src/memory/allocation_queue.rs
Priyankakarumuru1: Good point. A low priority request without a cursor_id shouldn't happen. Low = speculative prefetch which always has a cursor. Added an assert! that low priority requests must have cursor_id.

mountpoint-s3-fs/src/memory/allocation_queue.rs
Priyankakarumuru1: Makes sense, I've renamed it to `has_pending` to keep it generic. I didn't go with is_empty because most callers want to know 'is there someone waiting?', with `is_empty` they'd have to write !is_empty() every time which feels a bit backwards. `has_pending` gives you a direct yes/no.

mountpoint-s3-fs/src/memory/allocation_queue.rs
Priyankakarumuru1: Good point, updated the comment to be generic.

mountpoint-s3-fs/src/memory/allocation_queue.rs
yerzhan7: I think we should remove/prune all canceled receivers (from the front of the queue) here, while we hold queue mutex. It's not only about wasted buffer allocations. E.g. if the front entry in the queue was cancelled (and e.g. it's of size 64MB), then the predicate (when only 8MB is free) will never actually return anything from the queue. And live low-priority 8MB request (next item after 64MB request) stays blocked in the queue.

We should remove/prune cancelled receivers in the `upgrade()` method as well while we iterate through low priority queue.

mountpoint-s3-fs/src/memory/allocation_queue.rs
yerzhan7: Instead of `assert` what if we split `push` into `push_low()` and `push_high()` methods to enforce this at compile time instead of runtime? 

```rs
  pub fn push_high(&self, cursor_id: Option<CursorId>, size: usize, kind: BufferKind)
      -> oneshot::Receiver<PoolBuffer>
  {
      self.push_inner(AllocationPriority::High, cursor_id, size, kind)
  }

  pub fn push_low(&self, cursor_id: CursorId, size: usize, kind: BufferKind)
      -> oneshot::Receiver<PoolBuffer>
  {
      self.push_inner(AllocationPriority::Low, Some(cursor_id), size, kind)
  }
```

Priyankakarumuru1: Done

mountpoint-s3-fs/src/memory/allocation_queue.rs
Priyankakarumuru1: Good one.

yerzhan7 (approved):

PR #1838, opened 2026-05-26
Add CRC64NVME option for upload checksums

alecrubin:
Adds CRC64NVME as a third value (alongside `crc32c` and `off`) for `--upload-checksums`, wired end-to-end through the atomic and incremental upload paths.

CRC32C keeps using S3's composite (checksum-of-checksums) format on multipart uploads. CRC64NVME uses S3's `FULL_OBJECT` mode (S3 rejects composite for it), so the object-level checksum is computed by Mountpoint's local hasher and sent on `CompleteMultipartUpload`. That value travels through the existing upload-review callback: `review_and_complete` returns an `UploadReviewOutcome` (`Proceed(Option<UploadChecksum>)` / `Abort`), and the CRT sends the returned checksum on completion. (An earlier revision used a separate `FullObjectChecksumHandle`; that's been folded into the review API per review feedback, so there's no handle to populate-before-complete and forget.)

**Mountpoint end users:** no breaking changes. `crc32c` and `off` keep identical semantics, `crc64nvme` is additive and opt-in, and defaults are unchanged. Objects uploaded with `crc64nvme` report `ChecksumType: FULL_OBJECT` from `GetObjectAttributes` and a bare-base64 `x-amz-checksum-crc64nvme` value (no `-<part-count>` suffix), versus `COMPOSITE` + `<base64>-<N>` for CRC32C. Downstream tools that strip a composite suffix or branch on `ChecksumType == COMPOSITE` should handle the FULL_OBJECT case, but only ever see it if a user opts in.

**Library consumers of `mountpoint-s3-client` / `mountpoint-s3-fs`:**
- `PutObjectTrailingChecksums` carries the algorithm in its variants: `Disabled` / `Composite(algo)` / `FullObject(algo)` / `ReviewOnly(algo)`.
- `PutObjectRequest::review_and_complete`'s callback returns `UploadReviewOutcome` instead of `bool`; `complete()` is unchanged.
- `S3FilesystemConfig::upload_checksum_algorithm` changes from `Option<ChecksumAlgorithm>` to the narrower `Option<UploadChecksumAlgorithm>` (`Crc32c` | `Crc64nvme`).

These are breaking changes for direct API consumers.

### Notable internals

- A narrower `UploadChecksumAlgorithm` on the FS-config/CLI boundary and an `AtomicUploadHasher` newtype make the FS-side checksum paths total, removing the runtime `expect`/guard chains (the `unimplemented!` arm in `atomic.rs` stays as defense-in-depth, now unreachable from any sanctioned path).
- In full-object mode the CRT stashes the checksum returned by the upload-review callback and sends it on `CompleteMultipartUpload`. This relies on the prepare-step ordering where the upload-review callback runs immediately before the full-object checksum callback on the same thread.
- `write_checksums_test` asserts the object-level checksum shape on both upload modes: bare base64 for CRC64NVME and for the CRC32C append/incremental path, and `<base64>-<N>` for CRC32C composite (multipart). A regression that flipped CRC64NVME to composite, or quietly broke CRC32C composite, fails loudly. Runs on the mock and real-S3 matrices.

The `mountpoint-s3/CHANGELOG.md` entry (new flag value + read-side semantics) is included. Crate version bumps aren't included yet. Happy to add once scope is settled:
- `mountpoint-s3-crt`, `mountpoint-s3`: additive (new public API / CLI value), patch / minor.
- `mountpoint-s3-client`, `mountpoint-s3-fs`: breaking API changes, minor.

### Local verification

`fmt-check`, `cargo check --workspace --tests`, and clippy pass locally, using a stub `fuse.pc` to satisfy the `mountpoint-s3-fuser` build script (macFUSE isn't installed on the dev machine, and `cargo check` doesn't link). Unit tests are green on the non-FUSE crates and the FS upload path:
- `cargo test -p mountpoint-s3-crt --lib`
- `cargo test -p mountpoint-s3-client --lib --features mock` (incl. `crc64nvme_full_object_checksum_lands_on_object`)
- `cargo test -p mountpoint-s3-fs --lib upload::atomic`

The FUSE integration tests (`write_checksums_test` and the rest of the `fuse_tests` suite) need real libfuse to link, so they run in CI rather than locally. The object-level shape assertion lives in `write_checksums_test`, which runs in the mock, `s3_tests`, and `s3express_tests` matrices.

---


Mattral (commented):
Hello there, thank you for the contribution and I am trying to learn something from your PR, and I must say that is great work wiring CRC64NVME end‑to‑end and tightening the type system. One thing I’d like to clarify: since CRC64NVME uses S3’s FULL_OBJECT mode rather than the composite checksum format, have you considered how this will interact with existing multipart upload monitoring tools or client libraries that currently expect composite checksums? Documenting or testing that compatibility could help downstream users avoid surprises.

alecrubin:
Good callout @Mattral. Worth stressing up front that this is opt-in. Nothing changes for existing users. CRC32C still uses composite, and the default (no `--upload-checksums` flag set) is unchanged. The CRC64NVME path only activates if someone explicitly passes `--upload-checksums crc64nvme`, so downstream tools that haven't been updated won't see FULL_OBJECT objects show up unless the user opts in.

When someone does opt in, you're right that the wire shape is different. On multipart objects `GetObjectAttributes` reports `ChecksumType: FULL_OBJECT` for CRC64NVME, vs `COMPOSITE` for CRC32C, and the `x-amz-checksum-crc64nvme` value is the bare full-object hash with no `-<part-count>` suffix. Tools that strip a composite `-N` suffix or branch on `ChecksumType == COMPOSITE` need to handle the FULL_OBJECT case for CRC64NVME-uploaded objects.

I pushed two follow-ups:

* CHANGELOG: the existing bullet now spells out the `ChecksumType` and `-N` suffix difference for downstream consumers.
* Test: `write_checksums_test` now asserts the object-level checksum shape on multipart uploads. Bare base64 for CRC64NVME, `<base64>-<part-count>` for CRC32C. A regression that flipped CRC64NVME to composite, or quietly broke CRC32C composite, would fail loudly. Runs on both the mock and real-S3 matrices.

One thing to flag: S3 itself enforces the split. It rejects `CompleteMultipartUpload` with CRC64NVME + COMPOSITE, so the algorithm-to-mode mapping isn't really a choice on our end. But the read-side surprise for downstream tools is real and explicit docs were a good call.

passaro:
HI @alecrubin, thanks for submitting this PR! Overall this is a great addition and it fits very nicely as a new option in Mountpoint.

Before going into a detailed review, I'd like to discuss the API shape on the S3 client for the full-object checksum. The PR introduces a `FullObjectChecksumHandle`, which the caller needs to use to set the checksum before completing the upload. Could we instead integrate it with the existing `review_and_complete` API? We could modify `review_callback` so it can return the full object checksum and that is internally used before completing the upload.

alecrubin:
> HI @alecrubin, thanks for submitting this PR! Overall this is a great addition and it fits very nicely as a new option in Mountpoint.
> 
> Before going into a detailed review, I'd like to discuss the API shape on the S3 client for the full-object checksum. The PR introduces a `FullObjectChecksumHandle`, which the caller needs to use to set the checksum before completing the upload. Could we instead integrate it with the existing `review_and_complete` API? We could modify `review_callback` so it can return the full object checksum and that is internally used before completing the upload.

Thanks @passaro! Glad it fits well, and good call on the API.

Agreed, folding this into `review_and_complete` is cleaner than the separate handle. The handle always bugged me a bit since you have to remember to set it before completing and nothing stops you from forgetting, so having `review_callback` return the checksum like you suggested gets rid of that.

I checked the CRT to make sure returning it from the review callback works timing wise, and it does. In `s_s3_prepare_complete_multipart_upload` the `upload_review_callback` runs right before the `full_object_checksum_callback`, same prepare step, same thread, so whatever the review callback returns can just be stashed and used for the full object checksum on complete. Basically what you described.

I'm thinking `review_callback` returns the proceed flag plus an optional full object checksum, something like a small struct `{ proceed, full_object_checksum: Option<String> }`. Then stash that for the existing full object shim and drop `FullObjectChecksumHandle` / `trailing_with_full_object` / `with_full_object_handle`, so `FullObject` would just carry the algorithm.

Any preference on the shape, struct vs an enum like `Proceed(Option<String>)` / `Abort`? Happy either way, I'll push it up once you confirm.

passaro:
> Any preference on the shape, struct vs an enum like `Proceed(Option<String>)` / `Abort`? Happy either way, I'll push it up once you confirm.

I'd definitely go with an enum like you suggest. Also consider whether we can use a type like `UploadChecksum` instead of `String` for the optional full-object checksum.

alecrubin:
> > Any preference on the shape, struct vs an enum like `Proceed(Option<String>)` / `Abort`? Happy either way, I'll push it up once you confirm.
> 
> I'd definitely go with an enum like you suggest. Also consider whether we can use a type like `UploadChecksum` instead of `String` for the optional full-object checksum.

@passaro Done, pushed it up.

Went with the enum like you suggested. `review_and_complete`'s callback now returns an `UploadReviewOutcome` (`Proceed(Option<UploadChecksum>`) / `Abort`), and `UploadChecksum` rather than a `String`, so the FS layer hands back the same typed checksum it already computed and the base64 rendering stays at the CRT boundary.

That let me drop `FullObjectChecksumHandle` entirely, plus `trailing_with_full_object` / `with_full_object_handle` on the CRT side. `PutObjectTrailingChecksums::FullObject` now just carries the algorithm. In full-object mode the CRT stashes whatever the review callback returns and hands it to the existing full-object checksum callback on complete, which works cleanly because the upload-review callback runs right before it in the same prepare step, same thread.

No more set-the-handle-before-completing footgun, which is what was bugging me about the old shape. Ready for the detailed review whenever you are.

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -1301,17 +1305,62 @@ impl MockPutObjectRequest {
...
+            sha256_to_base64(&sha256::checksum(part).expect("SHA256 hashing should not fail in mock client"))
+        }
+        other => unimplemented!("mock client does not yet support checksum algorithm {other}"),
+    }
+}
+
+/// Build the whole-object `Checksum` for an upload with `algorithm`. For algorithms with a
+/// well-defined S3 composite ("checksum of checksums"), produce it; otherwise leave the
+/// whole-object slot empty and rely on per-part checksums.
+fn compute_whole_object_checksum(algorithm: &ChecksumAlgorithm, parts: &[MockObjectPartAttributes]) -> Checksum {
```
passaro: This function seems misleading: it really only "works" for Crc32c. Consider inlining in `complete_inner`.

mountpoint-s3-client/src/mock_client.rs
```diff
@@ -1301,17 +1305,62 @@ impl MockPutObjectRequest {
...
+    if let ChecksumAlgorithm::Crc32c = algorithm {
+        checksum.checksum_crc32c = Some(compute_crc32c_of_crc32c_checksums(part_checksums));
+    }
+    checksum
+}
+
+/// Compute a checksum of checksums, mirroring how S3 computes the composite object checksum for an
+/// MPU. GetObjectAttributes returns this as a bare `<base64>` (the `-<part-count>` suffix only
+/// appears in the HeadObject representation), so we don't append it here.
 fn compute_crc32c_of_crc32c_checksums(individual_checksums: impl IntoIterator<Item = String>) -> String {
```
passaro: Not convinced we should change this. How do we handle HeadObject? Also, GetObjectAttributes does not guarantee to return the "bare" string.

mountpoint-s3-client/src/s3_crt_client/put_object.rs
```diff
@@ -251,19 +257,27 @@ impl std::fmt::Debug for ReviewCallbackBox {
...
         let Some(callback) = callback.take() else {
             error!("review callback was either never set or invoked twice");
-            return false;
+            return CrtUploadReviewOutcome::Abort;
         };
 
-        (callback)(review)
+        match (callback)(review) {
+            UploadReviewOutcome::Proceed(checksum) => {
+                CrtUploadReviewOutcome::Proceed(checksum.map(|checksum| checksum.to_base64().into_bytes()))
```
passaro, line 277: Ideally, we would not need this mapping here and just use a common `UploadReviewOutcome` type. However it may require some refactoring between the `client` and `crt` crates, so happy to leave it out of scope for this PR.

mountpoint-s3-client/src/object_client.rs
```diff
@@ -550,7 +550,9 @@ pub type ObjectMetadata = HashMap<String, String>;
 #[derive(Debug, Default, Clone)]
 #[non_exhaustive]
 pub struct PutObjectParams {
-    /// Enable Crc32c trailing checksums.
+    /// Trailing-checksum mode for the upload. Variants carry the algorithm (and, for full-object
+    /// mode, the handle the caller will populate) so invalid states like "Disabled + algorithm"
```
passaro: out of date with the latest changes

mountpoint-s3-client/src/object_client.rs
```diff
@@ -618,16 +620,44 @@ impl PutObjectParams {
...
     #[default]
     Disabled,
+    /// S3 composite mode. Per-part trailers are sent; the object-level checksum on the response
+    /// is `composite(part_checksums)`. Supported for algorithms S3 composes (CRC32C, CRC32, SHA1,
+    /// SHA256). For CRC64NVME, use [`Self::FullObject`] instead — S3 rejects composite for it.
+    Composite(ChecksumAlgorithm),
+    /// S3 FULL_OBJECT mode. Per-part trailers are still sent (so upload review works), but the
+    /// object-level checksum sent on `CompleteMultipartUpload` is the one the
+    /// [`review_and_complete`](PutObjectRequest::review_and_complete) callback returns via
+    /// [`UploadReviewOutcome::Proceed`] (typically computed by the FS layer's local hasher).
```
passaro: nit: remove reference to the FS layer

mountpoint-s3-client/src/object_client.rs
```diff
@@ -618,16 +620,44 @@ impl PutObjectParams {
...
+        match self {
+            Self::Disabled => None,
+            Self::Composite(algorithm) => Some(algorithm),
+            Self::FullObject(algorithm) => Some(algorithm),
+            Self::ReviewOnly(algorithm) => Some(algorithm),
+        }
+    }
+
+    /// `true` for any variant that causes checksums to be computed on the client (i.e. not `Disabled`).
+    pub fn is_computing(&self) -> bool {
```
passaro: This could just be `.algorithm().is_none()`. But it does not seem used anyway. Remove?

mountpoint-s3-fs/src/upload/atomic.rs
```diff
@@ -25,7 +28,11 @@ pub struct UploadRequest<Client: ObjectClient> {
     bucket: String,
     key: String,
     next_request_offset: u64,
-    hasher: crc32c::Hasher,
+    hasher: AtomicUploadHasher,
+    /// `true` when the upload uses S3 full-object checksum mode (e.g. CRC64NVME). In that case
+    /// `complete` returns the final object-level checksum from the `review_and_complete` callback,
+    /// and the CRT sends it on `CompleteMultipartUpload`.
+    full_object_checksum: bool,
```
passaro: This works, but we could also just implicitly "encode" it in the review callback where we only return the full object checksum if the algorithm is  CRC64NVME.

mountpoint-s3-fs/src/upload/atomic.rs
```diff
@@ -129,12 +158,21 @@ where
     pub async fn complete(self) -> Result<PutObjectResult, UploadError<Client::ClientError>> {
         let size = self.size();
         let checksum = self.hasher.finalize();
+        // In full-object mode (CRC64NVME on multipart), hand the final object-level checksum back
+        // to the CRT via the review callback so it can send it on CompleteMultipartUpload.
+        let full_object_checksum = self.full_object_checksum.then(|| checksum.clone());
```
passaro: The clone is awkward here (it's the same checksum!). See suggestion above on `full_object_checksum` flag.

mountpoint-s3-fs/tests/fuse_tests/write_test.rs
```diff
@@ -1469,19 +1464,61 @@ fn write_checksums_test(
...
+                    .all(|checksum| checksum.as_ref().is_some_and(|c| extract(c).is_some()));
+            // CRC64NVME must be a full-object checksum (S3 doesn't support composite for CRC64NVME),
+            // so its value must land on the object itself, not just the parts, and GetObjectAttributes
+            // always reports it as a bare `<base64>` with no part-count suffix.
+            //
+            // CRC32C is composite (carried by the per-part checksums) and can land on the object or
+            // only the parts. The object-level representation returned by GetObjectAttributes differs
+            // by backend: S3 Standard returns a bare `<base64>`, while S3 Express One Zone returns the
+            // composite form `<base64>-<part-count>`. Accept either, and when the suffix is present
+            // assert the part count matches the number of parts we observed.
```
passaro: It may be worth simplifying here. I'd suggest adding a top level case `Some(UploadChecksumAlgorithm::Crc64nvme)` checking only the full-object checksum and renaming the previous `UploadChecksumsMode::Enabled` to `Some(UploadChecksumAlgorithm::Crc32c)`. The remaining `Some(other)` can be explicitly not supported.

mountpoint-s3/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Unreleased
 
+* Add support for CRC64NVME full-object checksums on uploads. The `--upload-checksums` argument now accepts `crc64nvme` in addition to the existing `crc32c` and `off` values. CRC32C continues to use S3's composite (checksum-of-checksums) format on multipart uploads; CRC64NVME uses S3's `FULL_OBJECT` mode (sent on `CompleteMultipartUpload` via the CRT's `full_object_checksum_callback`). Both algorithms work end-to-end on atomic and incremental upload paths. Note for downstream consumers: on multipart objects, `GetObjectAttributes` reports `ChecksumType: FULL_OBJECT` for CRC64NVME (vs. `COMPOSITE` for CRC32C), and the `x-amz-checksum-crc64nvme` header value is the full-object hash with no `-<part-count>` suffix. Tooling that strips the composite `-N` suffix or assumes `COMPOSITE` should branch on the algorithm.
```
passaro: Remove the CRT reference: 


```suggestion
* Add support for CRC64NVME full-object checksums on uploads. The `--upload-checksums` argument now accepts `crc64nvme` in addition to the existing `crc32c` and `off` values. CRC32C continues to use S3's composite (checksum-of-checksums) format on multipart uploads; CRC64NVME uses S3's `FULL_OBJECT` mode. Both algorithms work end-to-end on atomic and incremental upload paths. Note for downstream consumers: on multipart objects, `GetObjectAttributes` reports `ChecksumType: FULL_OBJECT` for CRC64NVME (vs. `COMPOSITE` for CRC32C), and the `x-amz-checksum-crc64nvme` header value is the full-object hash with no `-<part-count>` suffix. Tooling that strips the composite `-N` suffix or assumes `COMPOSITE` should branch on the algorithm.
```

passaro (changes requested):
A few suggestions, but looks good overall:

mountpoint-s3-client/src/mock_client.rs
alecrubin: Yeah good call, inlined it into `complete_inner`. It was only ever doing anything for CRC32C anyway, so the standalone helper read as more general than it was.

mountpoint-s3-client/src/mock_client.rs
alecrubin: Fair, reverted the doc back to the original one liner. I was overstating what GetObjectAttributes guarantees and pulling HeadObject into a comment that didn't need it. Mock behavior is unchanged, just the comment.

mountpoint-s3-client/src/s3_crt_client/put_object.rs
alecrubin: Agreed, I'd rather keep it out of this PR. The crt crate can't see `UploadChecksum`, so a shared type means moving it (or something like it) across the crate boundary, which feels like its own change. Happy to do it as a follow-up.

mountpoint-s3-fs/src/upload/atomic.rs
alecrubin: Done, dropped the flag. The review callback now returns the checksum only when it's CRC64NVME and `None` otherwise, so the mode is implicit in the algorithm like you suggested. Also kills the clone you flagged below.

mountpoint-s3-fs/src/upload/atomic.rs
alecrubin: Gone with the flag. `matches!` on the checksum borrows it and then it moves straight into `Proceed`, so there's no clone anymore.

mountpoint-s3-fs/tests/fuse_tests/write_test.rs
alecrubin: Done, split it into explicit `Crc64nvme` / `Crc32c` / `None` arms. Reads a lot cleaner, thanks.

mountpoint-s3-client/src/object_client.rs
alecrubin: Went with removing it, nothing was calling it. If we ever need the check, `.algorithm().is_none()` is the better spelling like you said.

mountpoint-s3/CHANGELOG.md
alecrubin: Applied

passaro (approved):
Looks good. Thanks again for the contribution!

PR #1839, opened 2026-05-29
Gate incremental upload buffers through the allocation queue

yerzhan7:
**What changed and why?** Gate incremental upload buffers through the allocation queue

Yes

N/A

---

### DIFF
--- a/mountpoint-s3-fs/src/upload/incremental.rs
+++ b/mountpoint-s3-fs/src/upload/incremental.rs
@@ -162,8 +162,6 @@ struct AppendUploadQueue<Client: ObjectClient> {
     checksum_algorithm: Option<Option<ChecksumAlgorithm>>,
     /// Stores the last successful result to return in [join].
     last_known_result: Option<PutObjectResult>,
-    /// Tracks the requests pushed to the queue but still pending a response.
-    requests_in_queue: usize,
 }
 
 impl<Client> AppendUploadQueue<Client>
@@ -198,7 +196,6 @@ where
             buffer_sender,
             event_receiver,
             last_known_result: None,
-            requests_in_queue: 0,
             checksum_algorithm: None,
             pool,
             _task_handle: task_handle,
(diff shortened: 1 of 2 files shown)


Mattral (commented):
Thank you, and I'm trying to learn something here, thats interesting approach gating incremental upload buffers through the allocation queue. One thing I’d like to clarify: since can_allocate now checks only pool stats and ignores per‑cursor reservations, how do you see this affecting fairness between concurrent writers under heavy memory pressure? For example, could one handle monopolize buffers while others starve, or does WriteHandleLimiter admission control fully prevent that? Documenting or testing this interaction might help maintainers validate the new async allocation path.

mountpoint-s3-fs/src/upload/incremental.rs
```diff
@@ -256,20 +256,34 @@ where
...
-                        return Err(UploadError::UploadAlreadyTerminated);
-                    }
-                }
+        // Three-way decision per iteration:
+        //   1. No outstanding requests for this handle (nothing pushed-but-not-acked) →
+        //      force-allocate via `get_buffer_mut`. Forward-progress guarantee for the
+        //      write() FUSE thread: with nothing pending for this handle to wait on, the
+        //      next buffer is allocated unconditionally rather than blocking on memory held
+        //      by unrelated handles. Each admitted writer is guaranteed a buffer this way;
+        //      the resulting overshoot is bounded by `WriteHandleLimiter` admission control.
```
passaro: Is that true? What if the prefetcher has already saturated the memory?

I would have expected that we would just call `get_buffer_mut_async` unconditionally here. That already ensure that upload requests are prioritized (although on par with active reads), while `WriteHandleLimiter` ensures that writes alone don't lead to starvation.

yerzhan7: Discussed offline: Changed to always call `get_buffer_mut_async` in https://github.com/awslabs/mountpoint-s3/pull/1839/commits/1c869392a1c1a079db39deced176a7c639249a06

passaro (approved):

PR #1840, opened 2026-06-01
Gate pool allocations through limiter

Priyankakarumuru1:
This is the change that actually enforces the memory limit for buffer allocations. Previously, get_buffer_async always allocated immediately regardless of how much memory was already in use. Now, when memory is full, requests wait in the allocation queue and get served in priority order when memory frees up, active reads and uploads
first and speculative prefetch last.

The memory limit check uses a compare-and-swap loop on a pending_allocations counter - no mutex needed. Multiple threads can claim bytes concurrently, only one wins per available slot. The actual allocation proceeds without any external lock since both are already concurrent-safe.
  
The queue is woken whenever memory is freed: when a buffer is dropped (primary or secondary), when a cursor is released, or when the pool is trimmed.

Unit tests added.
  
get_buffer (the sync write path used by uploads) bypasses the queue intentionally - the CRT write path requires synchronous buffer allocation and cannot wait. Upload memory is bounded by the concurrent write-handle limit instead.

Yes, async buffer allocations now block under memory pressure instead of allocating unboundedly.

No, the memory limit isn't user-visible yet. No behavior change for existing users.

---

### DIFF
--- a/mountpoint-s3-fs/src/memory/allocation_queue.rs
+++ b/mountpoint-s3-fs/src/memory/allocation_queue.rs
@@ -219,11 +219,10 @@ mod tests {
     use futures::executor::block_on;
 
     use crate::memory::pages::Page;
-    use crate::sync::Weak;
 
     fn make_buffer(size: usize) -> PoolBuffer {
         let page = Page::new_for_tests(size);
-        let ptr = page.try_reserve(BufferKind::Other, Weak::new()).unwrap();
+        let ptr = page.try_reserve(BufferKind::Other).unwrap();
         PoolBuffer::new_primary(ptr, size)
     }
 
--- a/mountpoint-s3-fs/src/memory/allocation_queue.rs
+++ b/mountpoint-s3-fs/src/memory/allocation_queue.rs
@@ -68,6 +68,14 @@ struct AllocationQueueInner {
     low: VecDeque<PendingAllocation>,
 }
 
+impl std::fmt::Debug for AllocationQueue {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("AllocationQueue")
+            .field("has_pending", &self.has_pending.load(Ordering::Relaxed))
+            .finish_non_exhaustive()
+    }
+}
+
 impl Default for AllocationQueue {
     fn default() -> Self {
         Self::new()
@@ -175,6 +183,7 @@ impl AllocationQueue {
     ///
     /// Called when a FUSE read arrives for a cursor that has pending speculative
     /// allocations — those allocations are now urgent.
+    #[allow(dead_code)] // TODO(memory-limiter): wire into set_active_read
     pub fn upgrade(&self, cursor_id: CursorId) {
         let mut inner = self.inner.lock().unwrap();
         let n = inner.low.len();
@@ -210,10 +219,11 @@ mod tests {
     use futures::executor::block_on;
 
     use crate::memory::pages::Page;
+    use crate::sync::Weak;
 
     fn make_buffer(size: usize) -> PoolBuffer {
         let page = Page::new_for_tests(size);
-        let ptr = page.try_reserve(BufferKind::Other).unwrap();
+        let ptr = page.try_reserve(BufferKind::Other, Weak::new()).unwrap();
         PoolBuffer::new_primary(ptr, size)
     }
 
--- a/mountpoint-s3-fs/src/memory/allocation_queue.rs
+++ b/mountpoint-s3-fs/src/memory/allocation_queue.rs
@@ -183,7 +183,6 @@ impl AllocationQueue {
     ///
     /// Called when a FUSE read arrives for a cursor that has pending speculative
     /// allocations — those allocations are now urgent.
-    #[allow(dead_code)] // TODO(memory-limiter): wire into set_active_read
     pub fn upgrade(&self, cursor_id: CursorId) {
         let mut inner = self.inner.lock().unwrap();
         let n = inner.low.len();
--- a/mountpoint-s3-fs/src/memory/pages.rs
+++ b/mountpoint-s3-fs/src/memory/pages.rs
@@ -1,7 +1,8 @@
 use std::alloc::{self, Layout};
 
-use crate::sync::{Arc, Mutex};
+use crate::sync::{Arc, Mutex, Weak};
 
+use super::pool::{PagedPool, PagedPoolInner};
 use super::stats::{BufferKind, SizePoolStats};
 
 /// Page in a size pool.
@@ -78,7 +79,7 @@ impl Page {
     /// Try to reserve a buffer from this page.
     ///
     /// Returns [None] if the page is already full.
-    pub fn try_reserve(&self, kind: BufferKind) -> Option<PagedBufferPtr> {
+    pub fn try_reserve(&self, kind: BufferKind, pool: Weak<PagedPoolInner>) -> Option<PagedBufferPtr> {
         let (index, page_status) = consume(&mut self.inner.reserved_bitmask.lock().unwrap())?;
         let offset = index * self.inner.stats.buffer_size;
         if let PageStatus::Empty = page_status {
@@ -93,6 +94,7 @@ impl Page {
             ptr,
             pool_page: self.clone(),
             kind,
+            pool,
         })
     }
 
@@ -183,6 +185,8 @@ pub(super) struct PagedBufferPtr {
     ptr: *mut u8,
     pool_page: Page,
     kind: BufferKind,
+    /// Weak reference back to the pool so this buffer's drop can wake pending allocations.
+    pool: Weak<PagedPoolInner>,
 }
 
 // SAFETY: access to the buffer and release back to the pool page on drop can be performed from another thread.
@@ -191,6 +195,9 @@ unsafe impl Send for PagedBufferPtr {}
 impl Drop for PagedBufferPtr {
     fn drop(&mut self) {
         self.pool_page.release(self.ptr, self.kind);
+        if let Some(pool) = self.pool.upgrade() {
+            PagedPool::from_inner(pool).try_wake_pending();
+        }
     }
 }
 
@@ -214,24 +221,24 @@ mod tests {
         let mut buffers = Vec::new();
         for _ in 0..Page::BUFFERS_PER_PAGE {
             let buffer_ptr = page
-                .try_reserve(BufferKind::Other)
+                .try_reserve(BufferKind::Other, Weak::new())
                 .expect("reserving up to 16 buffers from a new page should succeed");
             buffers.push(buffer_ptr);
         }
 
         assert!(
-            page.try_reserve(BufferKind::Other).is_none(),
+            page.try_reserve(BufferKind::Other, Weak::new()).is_none(),
             "reserving the 17th buffer from a page should return None"
         );
 
         _ = buffers.pop().expect("drop one of the reserved buffers");
 
         buffers.push(
-            page.try_reserve(BufferKind::Other)
+            page.try_reserve(BufferKind::Other, Weak::new())
                 .expect("reserving after dropping 1 buffer should succeed"),
         );
         assert!(
-            page.try_reserve(BufferKind::Other).is_none(),
+            page.try_reserve(BufferKind::Other, Weak::new()).is_none(),
             "reserving when all 16 buffers are in use should return None"
         );
 
@@ -247,7 +254,7 @@ mod tests {
             "invalidation of a new, empty page should succeed"
         );
         assert!(
-            page.try_reserve(BufferKind::Other).is_none(),
+            page.try_reserve(BufferKind::Other, Weak::new()).is_none(),
             "reserving from an invalidated page should return None"
         );
     }
@@ -256,7 +263,7 @@ mod tests {
     fn test_invalidate_in_use() {
         let page = Page::new_for_tests(1024);
         let buffer_ptr = page
-            .try_reserve(BufferKind::Other)
+            .try_reserve(BufferKind::Other, Weak::new())
             .expect("reserving from a new page should succeed");
         assert!(!page.invalidate_if_empty(), "invalidation of a page in use should fail");
         drop(buffer_ptr);
@@ -265,7 +272,7 @@ mod tests {
             "invalidation of an empty page should succeed"
         );
         assert!(
-            page.try_reserve(BufferKind::Other).is_none(),
+            page.try_reserve(BufferKind::Other, Weak::new()).is_none(),
             "reserving from an invalidated page should return None"
         );
     }
--- a/mountpoint-s3-fs/src/memory/pages.rs
+++ b/mountpoint-s3-fs/src/memory/pages.rs
@@ -97,7 +97,6 @@ impl Page {
             ptr,
             pool_page: self.clone(),
             kind,
-            pool: self.inner.pool.clone(),
         })
     }
 
@@ -118,6 +117,14 @@ impl Page {
         self.inner.stats.remove_reserved_buffer(kind);
         if *bitmask == 0 {
             self.inner.stats.add_empty_page();
+        }
+        // Release the bitmask lock before waking pending allocations.
+        // try_wake_pending may try to allocate from this same page which would deadlock
+        // if we held the bitmask lock.
+        drop(bitmask);
+        // A primary buffer was freed — wake any pending allocation requests.
+        if let Some(pool) = self.inner.pool.upgrade() {
+            pool.try_wake_pending();
         }
     }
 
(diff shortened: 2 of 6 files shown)


mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -181,6 +193,59 @@ impl PagedPool {
...
+        let rx = match cursor_id {
+            Some(id) if self.inner.limiter.has_active_read(id) => {
+                self.inner.allocation_queue.push_high(cursor_id, size, kind)
+            }
+            Some(id) => self.inner.allocation_queue.push_low(id, size, kind),
+            None => self.inner.allocation_queue.push_high(None, size, kind), // uploads are urgent
+        };
+        // The only error is Canceled (cursor dropped while waiting). The CRT must always
+        // receive a buffer, so fall back to a direct allocation.
+        rx.await.unwrap_or_else(|_| self.get_buffer(size, kind, cursor_id))
```
yerzhan7, line 293: I think we should call `try_wake_pending` here? Edge case:
1. Thread A - Enters `acquire_buffer_async`, sees that it cannot allocate, decides to push to queue
2. Thread B - dropped buffer, called `try_wake_pending` - queue is still empty, so no-op
3. Thread A - pushed to the queue, memory is available, but no one calls `try_wake_pending` and it is stuck?

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -181,6 +193,59 @@ impl PagedPool {
...
+        // everything else is Low (speculative prefetch or upload without a cursor).
+        let rx = match cursor_id {
+            Some(id) if self.inner.limiter.has_active_read(id) => {
+                self.inner.allocation_queue.push_high(cursor_id, size, kind)
+            }
+            Some(id) => self.inner.allocation_queue.push_low(id, size, kind),
+            None => self.inner.allocation_queue.push_high(None, size, kind), // uploads are urgent
+        };
+        // The only error is Canceled (cursor dropped while waiting). The CRT must always
+        // receive a buffer, so fall back to a direct allocation.
```
yerzhan7: `CRT must always receive a buffer`. I don't this is actually true. 

I think in CRT ticket future we can do [`aws_future_s3_buffer_ticket_set_error`](https://github.com/awslabs/aws-c-s3/blob/f1a52b5e960c06bd9392cb5e982c6fe04f1ce253/source/s3_meta_request.c#L1970) and do not allocate unneeded buffer? (Might require a separate change/PR though)

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -181,6 +193,59 @@ impl PagedPool {
...
+    /// Attempt to fulfill pending allocation requests from the queue.
+    ///
+    /// Called whenever memory is freed — on buffer drop, cursor release, or pool trim.
+    /// Loops until no more entries can be fulfilled or the queue is empty.
+    pub(super) fn try_wake_pending(&self) {
+        loop {
+            let entry = self
+                .inner
+                .allocation_queue
+                .pop_front_if(|size, _, _| self.can_allocate(size));
```
yerzhan7: `pop_front_if` - acquires queue mutex always.

I think we should have fast path here by checking `self.inner.allocation_queue.has_pending()` (similar to fast path in `acquire_buffer_async` method below)

yerzhan7 (commented):
(posting partial review comments)

mountpoint-s3-fs/src/memory/pool.rs
Priyankakarumuru1: Good one. Added try_wake_pending(). Thanks!

mountpoint-s3-fs/src/memory/pool.rs
Priyankakarumuru1: Added a has_pending() check. Thanks!

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -181,6 +195,63 @@ impl PagedPool {
...
+    ///
+    /// **Slow path:** pushes the request into the allocation queue and awaits a buffer.
+    /// Priority is determined from the cursor's active-read state:
+    /// - cursor with an active FUSE read → high priority
+    /// - cursor without an active read (speculative prefetch) → low priority
+    /// - no cursor (upload) → high priority (write syscall is blocked)
+    async fn acquire_buffer_async(&self, size: usize, kind: BufferKind, cursor_id: Option<CursorId>) -> PoolBuffer {
+        // Fast path: if the queue is empty and there is room, allocate immediately.
+        if !self.inner.allocation_queue.has_pending() && self.can_allocate(size) {
+            return self.get_buffer(size, kind, cursor_id);
```
passaro: This looks like a race condition: concurrent calls to `acquire_buffer_async` may both short-circuit and go over the limit. We should also add a test to try and reproduce it.

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -181,6 +195,63 @@ impl PagedPool {
...
+        }
+
+        // Determine priority: a cursor with an active FUSE read is High (urgent),
+        // everything else is Low (speculative prefetch or upload without a cursor).
+        let rx = match cursor_id {
+            Some(id) if self.inner.limiter.has_active_read(id) => {
```
passaro, line 281: Are we just assuming this allocation is to serve the active read, rather than looking at the range? We can review later, but let's add a comment at least.

mountpoint-s3-fs/src/memory/pool.rs
passaro: Seems like an ad-hoc fix. Let's review the whole synchronization approach.

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -208,16 +279,24 @@ impl MemoryPool for PagedPool {
         )
     }
 
+    fn get_buffer_async(&self, size: usize, meta_request: &MetaRequest) -> impl Future<Output = Self::Buffer> + Send {
```
passaro: doesn't `async fn` work here?

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -516,4 +596,103 @@ mod tests {
...
+            let mut blockers = Vec::new();
+            while pool.can_allocate(BUF) {
+                blockers.push(pool.get_buffer(BUF, BufferKind::Other, None));
+            }
+
+            // get_buffer_async should now queue and not resolve.
+            let mut fut = Box::pin(pool.acquire_buffer_async(BUF, BufferKind::GetObject, None));
+            let waker = futures::task::noop_waker();
+            let mut cx = std::task::Context::from_waker(&waker);
+            assert!(std::future::Future::poll(fut.as_mut(), &mut cx).is_pending());
```
passaro: Resorting to a manual poll can make the test fragile. 
* Isn't there a different way to write the same test? 
* If not, we should document more clearly what the expectations are.

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -516,4 +596,103 @@ mod tests {
...
+            }
+
+            let pool_clone = pool.clone();
+            let handle =
+                std::thread::spawn(move || block_on(pool_clone.acquire_buffer_async(BUF, BufferKind::GetObject, None)));
+
+            std::thread::sleep(std::time::Duration::from_millis(1));
+            blockers.pop(); // free one buffer — should wake the pending request
+
+            let buffer = handle.join().unwrap();
```
passaro: At least one failure mode for this test is that it will never return. Could we add a timeout? I'd think about using a channel rather than the thread handle directly.

passaro: Same for other tests.

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -181,6 +195,63 @@ impl PagedPool {
...
+    /// Returns `true` if the pool can accommodate an additional allocation of `size` bytes.
+    fn can_allocate(&self, size: usize) -> bool {
+        self.available_mem() >= size as u64
+    }
+
+    /// Attempt to fulfill pending allocation requests from the queue.
+    ///
+    /// Called whenever memory is freed — on buffer drop, cursor release, or pool trim.
+    /// Loops until no more entries can be fulfilled or the queue is empty.
+    pub(super) fn try_wake_pending(&self) {
```
passaro, line 253: Consider moving to `PagedPoolInner`. It may help avoiding the `from_inner` dance in a few (all?) places.

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -183,6 +185,8 @@ pub(super) struct PagedBufferPtr {
     ptr: *mut u8,
     pool_page: Page,
     kind: BufferKind,
+    /// Weak reference back to the pool so this buffer's drop can wake pending allocations.
+    pool: Weak<PagedPoolInner>,
```
passaro: The original pool got away with a simple ownership approach where pages and buffers did not need to point back to the whole pool. Now it seems that has changed. Rather than passing weak pointers around, shall we consider adding them to intermediate types (e.g. `Page` here)?

passaro (changes requested):

mountpoint-s3-fs/src/memory/pages.rs
Priyankakarumuru1: Good suggestion. Moved the Weak to Page.

mountpoint-s3-fs/src/memory/pool.rs
Priyankakarumuru1: Done.

mountpoint-s3-fs/src/memory/pool.rs
Priyankakarumuru1: Fixed with an `alloc_gate` mutex where the check and allocate happen atomically so two threads can't both take the same memory. Also added a test to verify.

mountpoint-s3-fs/src/memory/pool.rs
Priyankakarumuru1: Done!

mountpoint-s3-fs/src/memory/pool.rs
Priyankakarumuru1: Yes, done!

mountpoint-s3-fs/src/memory/pool.rs
Priyankakarumuru1: Yes, removed the manual poll. It now spawns a thread and checks `has_pending()` to verify th erequest is in the queue.

mountpoint-s3-fs/src/memory/pool.rs
Priyankakarumuru1: Done! Added 5 second timeouts using channels.

mountpoint-s3-fs/src/memory/pages.rs
passaro: But why it's still here?

mountpoint-s3-fs/src/memory/pages.rs
```diff
@@ -191,6 +198,9 @@ unsafe impl Send for PagedBufferPtr {}
 impl Drop for PagedBufferPtr {
     fn drop(&mut self) {
         self.pool_page.release(self.ptr, self.kind);
+        if let Some(pool) = self.pool.upgrade() {
```
passaro: I'd expect that this could be moved to `Page::release` (or even to the stats?)

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -200,23 +200,24 @@ impl PagedPool {
...
 
+    /// Check available memory and, if there is room, allocate and return a buffer. Returns `None`
+    /// if the pool cannot accommodate `size` bytes.
+    ///
+    /// Both the check and the allocation are performed while holding `alloc_gate`, so this
+    /// check-and-allocate is atomic with respect to other gated allocations: a plain `can_allocate`
+    /// check followed by `get_buffer` would let concurrent callers both observe room for the same
+    /// bytes and overshoot the limit.
+    fn try_allocate(&self, size: usize, kind: BufferKind, cursor_id: Option<CursorId>) -> Option<PoolBuffer> {
+        let _gate = self.inner.alloc_gate.lock().unwrap();
```
passaro: In this way we will block on all allocations. Have you looked at `SizePool::reserve`? Let's discuss offline.

passaro (changes requested):
Partial review. Let's think about the synchronization approach.

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -176,19 +191,107 @@ impl PagedPool {
...
+        // `try_allocate` uses CAS on `pending_allocations`, so concurrent fast-path callers cannot both
+        // observe room for the same bytes and overshoot the limit.
+        if !self.inner.allocation_queue.has_pending()
+            && let Some(buffer) = self.try_allocate(size, kind, cursor_id)
+        {
+            return buffer;
+        }
+
+        // Determine priority: a cursor with an active FUSE read is High (urgent),
+        // everything else is Low (speculative prefetch or upload without a cursor).
```
yerzhan7: `everything else is Low ...` `...or upload without a cursor`.  But Upload is High priority.

mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -175,6 +183,7 @@ impl AllocationQueue {
     ///
     /// Called when a FUSE read arrives for a cursor that has pending speculative
     /// allocations — those allocations are now urgent.
+    #[allow(dead_code)] // TODO(memory-limiter): wire into set_active_read
```
passaro: Can we wire this up and resolve the TODO?

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -269,6 +383,54 @@ impl PagedPoolInner {
...
+                .get_pool_for_size(entry.size)
+                .map(|p| p.buffer_size())
+                .unwrap_or(entry.size) as u64;
+
+            // Allocate without any lock — SizePool handles concurrency with its own RwLock.
+            let buffer = pool.get_buffer(entry.size, entry.kind, entry.cursor_id);
+            self.pending_allocations.fetch_sub(alloc_size, Ordering::SeqCst);
+
+            if let Err(buffer) = entry.fulfill(buffer) {
+                undelivered.push(buffer);
```
passaro, line 436: What's the point of keeping the "undelivered" buffers alive? These are buffers that we just got from the pool, but could not send to their requester because it was cancelled. 
When this happens, we should really drop the buffer immediately, so the next pending allocation can use the memory. However, the drop will recursively call `try_wake_pending` again.

I suspect the fix will be to decouple `try_wake_pending` by running it concurrently.
For now, let's add a TODO here.

passaro (commented):
As discussed offline, the main issue is that we are not looking at the memory actually allocated, but rather at the buffers in use. Relatedly, synchronizing the availability check should really happen on `Page::new` rather than when calling `get_buffer`.

However, that may be a larger change that we implement in a separate PR. In the meantime, let's address the other remaining issues and merge this one.

mountpoint-s3-fs/src/memory/pool.rs
Priyankakarumuru1: Right. Fixed the comment.

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -771,40 +787,80 @@ mod tests {
...
+                    break;
+                }
+                std::thread::sleep(std::time::Duration::from_millis(1));
+            }
+
+            // Simulate a FUSE read arriving — should upgrade the entry to high priority.
+            let _guard = cursor.set_active_read(0, BUF);
+
+            // Verify the entry is now in the high-priority queue by freeing one buffer
+            // and checking it gets served (would time out if stuck in low queue behind nothing).
```
passaro, line 856: Isn't this the only pending allocation? Why would it ever get stuck?

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -771,40 +787,80 @@ mod tests {
                 blockers.push(pool.get_buffer(BUF, BufferKind::Other, None));
             }
 
-            // Enqueue a low-priority request (speculative, cursor with no active read).
-            let (low_tx, low_rx) = std::sync::mpsc::channel();
-            let pool1 = pool.clone();
+            // Push both requests directly to the queue so ordering is deterministic.
```
passaro, line 790: I'm not convinced this is any better. Can we guarantee that the buffers are enqueued in order if we don't await these calls?

passaro (approved):
The tests need reworking, in particular the additional changes in the last push. However, for now, let's merge the change into the branch as is, so we can unblock other PRs.

PR #1841, opened 2026-06-02
Implement cursor reset mechanism

passaro:
Allow the memory limiter to externally reset a cursor, immediately dropping its in-flight S3 request and buffered data so that reserved memory can be reclaimed under pressure.

No.

No.

---

### DIFF
--- a/mountpoint-s3-fs/src/memory/pool.rs
+++ b/mountpoint-s3-fs/src/memory/pool.rs
@@ -177,8 +177,9 @@ impl PagedPool {
     }
 
     /// Reset a cursor: immediately drop its in-flight request and buffered data,
-    /// reclaiming all reserved memory. Returns `true` if the cursor was actually
-    /// reset, `false` if already dropped or already reset.
+    /// reclaiming all reserved memory.
+    ///
+    /// Returns whether the cursor was actually reset.
     pub fn reset_cursor(&self, cursor_id: CursorId) -> bool {
         self.inner.limiter.request_reset(cursor_id)
     }
--- a/mountpoint-s3-fs/src/memory/pool.rs
+++ b/mountpoint-s3-fs/src/memory/pool.rs
@@ -176,6 +176,13 @@ impl PagedPool {
         self.inner.limiter.create_cursor(self)
     }
 
+    /// Reset a cursor: immediately drop its in-flight request and buffered data,
+    /// reclaiming all reserved memory. Returns `true` if the cursor was actually
+    /// reset, `false` if already dropped or already reset.
+    pub fn reset_cursor(&self, cursor_id: CursorId) -> bool {
+        self.inner.limiter.request_reset(cursor_id)
+    }
+
     /// Query available memory.
     pub fn available_mem(&self) -> u64 {
         self.inner.limiter.available_mem(&self.inner.stats)
--- a/mountpoint-s3-fs/src/prefetch.rs
+++ b/mountpoint-s3-fs/src/prefetch.rs
@@ -311,22 +311,27 @@ where
         let max_preferred_part_size = 1024 * 1024;
         self.preferred_part_size = self.preferred_part_size.max(length).min(max_preferred_part_size);
 
-        loop {
-            // Try to use the current cursor, if present. Allows for limited non sequential reads.
-            if let Some(ref mut cursor) = self.cursor
-                && let Some(result) = cursor.try_read(offset, length).await?
-            {
-                return Ok(result);
-            } else {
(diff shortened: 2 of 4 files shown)


mountpoint-s3-fs/src/prefetch.rs
```diff
@@ -316,15 +311,16 @@ where
...
-        // Try to use the current cursor, if present. Allows for limited non sequential reads.
-        if let Some(ref mut cursor) = self.cursor
-            && let Some(result) = cursor.try_read(offset, length).await?
-        {
-            Ok(result)
-        } else {
-            // Otherwise, create a new cursor at `offset` and read from it.
-            let cursor = self.cursor.insert(self.create_cursor(offset)?);
-            cursor.read(length).await
+        loop {
```
yerzhan7: Sorry, not sure I understand why we need unbounded loop now here? (code comment might be useful here)
Are we sure it's safe to loop indefinitely? Maybe 2 attempts is enough before returning error?

mountpoint-s3-fs/src/prefetch/cursor.rs
```diff
@@ -182,9 +226,12 @@ where
         Ok(true)
     }
 
-    async fn try_seek_backward(&mut self, offset: u64) -> Result<bool, PrefetchReadError<Client::ClientError>> {
+    async fn try_seek_backward(
+        &mut self,
+        _cursor: &Cursor<Client>,
```
yerzhan7: unused parameter?

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -176,6 +176,13 @@ impl PagedPool {
         self.inner.limiter.create_cursor(self)
     }
 
+    /// Reset a cursor: immediately drop its in-flight request and buffered data,
+    /// reclaiming all reserved memory. Returns `true` if the cursor was actually
+    /// reset, `false` if already dropped or already reset.
```
yerzhan7: I think this is slightly incorrect? It also returns `false` if cursor is used by active reader FUSE thread due to `let Some(mut guard) = lock.try_lock()` returning `None` in case of `try_read` holding a lock?

mountpoint-s3-fs/src/prefetch/cursor.rs
```diff
@@ -140,30 +176,37 @@ where
...
         } else {
-            self.try_seek_backward(offset).await
+            self.try_seek_backward(cursor, offset).await
         }
     }
 
-    async fn try_seek_forward(&mut self, offset: u64) -> Result<bool, PrefetchReadError<Client::ClientError>> {
+    async fn try_seek_forward(
+        &mut self,
+        cursor: &Cursor<Client>,
```
yerzhan7: If we only use `cursor.max_forward_seek_wait_distance` field, then we can pass only that instead of the whole cursor?

mountpoint-s3-fs/src/prefetch/cursor.rs
```diff
@@ -50,64 +58,92 @@ where
...
+        let inner = Arc::new(AsyncMutex::new(None));
+        let inner_reset = Arc::downgrade(&inner);
 
-    /// Read at the current offset.
-    pub async fn read(
-        &mut self,
-        length: usize,
-    ) -> Result<(ChecksummedBytes, bool), PrefetchReadError<Client::ClientError>> {
-        let _active_read_guard = self.cursor_handle.set_active_read(self.current_offset, length);
+        // Populated the mutex and set up the reset callback.
```
yerzhan7: Could we register reset callback on `cursor_handle` before we construct `cursor_inner`? This way we can avoid doing `lock_blocking` which carries some warning in https://docs.rs/async-lock/latest/async_lock/struct.Mutex.html#method.lock_blocking about async context

mountpoint-s3-fs/src/memory/pool.rs
passaro: Simplified wording. We may extend this in the future and potentially return an enum.

mountpoint-s3-fs/src/prefetch.rs
passaro: Good point. Refactored so that the initial read cannot be reset, which was the loop was (badly) attempting to handle.

mountpoint-s3-fs/src/prefetch/cursor.rs
passaro: Since we now start reading immediately, we can use the async `lock()` method.

yerzhan7 (approved):

PR #1844, opened 2026-06-11
Limit memory allocations instead of buffer acquisitions

passaro:
Move memory limiting to the time of actual heap allocation, rather than on requesting a buffer. When returning an unused buffers from an already allocated page, the pool can avoid checking against the limit.

Also included in this change:
* Vocabulary cleanup in code and metrics: buffers handed out to consumers are now described as "acquired" / "in use" rather than "reserved" , reserve now means only the prefetch *intent* reservation (e.g. `mem_reserved`).
* Processing of pending allocations was moved to a background thread. Releasing a buffer or deallocating memory now only notifies the thread to trigger processing, rather than invoking it directly, thus avoiding recursive calls.
* Fallback mechanism to allocate smaller pages in the memory pool when under memory pressure. Memory pool pages are still created by default with a buffer count of 16, however if the allocation is blocked by the memory limiter, will iteratively scaled down by halving the buffer count, down to a single buffer.
* Merged the functionality of `PoolStats` into `MemoryLimiter`, which now controls all allocations and deallocations and maintains stats like `allocated_bytes` by handing out the new low-level type `ManagedBuffer`.

N/A

N/A

---

### DIFF
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -14,7 +14,7 @@ use super::PagedPool;
 use super::buffers::{BufferPtr, ManagedBuffer};
 use super::stats::{BUFFER_KIND_COUNT, BufferKind};
 
-pub const MINIMUM_MEM_LIMIT: u64 = 512 * 1024 * 1024;
+pub const MINIMUM_MEM_LIMIT: usize = 512 * 1024 * 1024;
 
 /// Buffer areas that can be managed by the memory limiter. This is used for updating metrics.
 #[derive(Debug)]
@@ -56,15 +56,15 @@ impl BufferArea {
 #[derive(Debug)]
 pub struct MemoryLimiter {
     /// Target memory limit.
-    mem_limit: u64,
+    mem_limit: usize,
     /// Additional reserved memory for other non-buffer usage like storing metadata
-    additional_mem_reserved: u64,
+    additional_mem_reserved: usize,
     /// Memory allocated by the pool.
     allocated_bytes: AtomicUsize,
     /// Memory in use by [`BufferKind`].
     acquired_bytes: [AtomicUsize; BUFFER_KIND_COUNT],
     /// Total reserved memory. Used in budget checks.
-    mem_reserved: Arc<AtomicU64>,
+    mem_reserved: Arc<AtomicUsize>,
     /// Unified per-cursor state. Cursors own a strong reference to their state.
     cursors: Arc<DashMap<CursorId, Weak<CursorState>>>,
     /// Counter for generating unique [CursorId]s.
@@ -82,7 +82,7 @@ pub struct MemoryLimiter {
 }
 
 impl MemoryLimiter {
-    pub fn new(mem_limit: u64, pending_signal: Arc<WakeSignal>) -> Self {
+    pub fn new(mem_limit: usize, pending_signal: Arc<WakeSignal>) -> Self {
         let min_reserved = 128 * 1024 * 1024;
         let additional_mem_reserved = (mem_limit / 8).max(min_reserved);
         let formatter = make_format(humansize::BINARY);
@@ -107,33 +107,33 @@ impl MemoryLimiter {
 
     /// The configured memory limit in bytes. Note this is the total memory target including
     /// non-buffer overhead, not the budget available for data buffers — see [`Self::data_buffer_budget`].
-    pub fn mem_limit(&self) -> u64 {
+    pub fn mem_limit(&self) -> usize {
         self.mem_limit
     }
 
     /// The static memory budget available for data buffers, i.e. `mem_limit - additional_mem_reserved`.
     /// This is the upper bound on buffer-backed allocations and is used by
     /// [`crate::memory::WriteHandleLimiter`] to derive its cap.
-    pub fn data_buffer_budget(&self) -> u64 {
+    pub fn data_buffer_budget(&self) -> usize {
         self.mem_limit.saturating_sub(self.additional_mem_reserved)
     }
 
     /// Reserve the memory for future uses. Always succeeds, even if it means going beyond
     /// the configured memory limit.
-    fn reserve(&self, area: BufferArea, size: u64) {
+    fn reserve(&self, area: BufferArea, size: usize) {
         self.mem_reserved.fetch_add(size, Ordering::SeqCst);
         metrics::gauge!("mem.bytes_reserved", "area" => area.as_str()).increment(size as f64);
     }
 
     /// Reserve the memory for future uses. If there is not enough memory returns `false`.
-    fn try_reserve(&self, area: BufferArea, size: u64) -> bool {
+    fn try_reserve(&self, area: BufferArea, size: usize) -> bool {
         let start = Instant::now();
         let mut mem_reserved = self.mem_reserved.load(Ordering::SeqCst);
         loop {
             let new_mem_reserved = mem_reserved.saturating_add(size);
-            let pool_mem_reserved = self.allocated_bytes() as u64;
+            let mem_allocated = self.allocated_bytes();
             let new_total_mem_usage = new_mem_reserved
-                .saturating_add(pool_mem_reserved)
+                .saturating_add(mem_allocated)
                 .saturating_add(self.additional_mem_reserved);
             if new_total_mem_usage > self.mem_limit {
                 trace!(new_total_mem_usage, "not enough memory to reserve");
@@ -160,7 +160,7 @@ impl MemoryLimiter {
     }
 
     /// Release all remaining reservation for a cursor and remove it from tracking.
-    fn release_cursor(&self, cursor_id: CursorId, cursor_reserved: &AtomicU64) {
+    fn release_cursor(&self, cursor_id: CursorId, cursor_reserved: &AtomicUsize) {
         if self.cursors.remove(&cursor_id).is_some() {
             let remaining = cursor_reserved.swap(0, Ordering::SeqCst);
             self.mem_reserved.fetch_sub(remaining, Ordering::SeqCst);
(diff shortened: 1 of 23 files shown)


mountpoint-s3-fs/src/lib.rs
```diff
@@ -41,3 +41,49 @@ fn init_crt() {
     mountpoint_s3_client::config::io_library_init(&mountpoint_s3_client::config::Allocator::default());
     mountpoint_s3_client::config::s3_library_init(&mountpoint_s3_client::config::Allocator::default());
 }
+
+// TODO: fix with merge
```
passaro: Will remove after rebasing.

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -61,47 +61,48 @@ pub struct PagedPool {
...
-            stats,
             limiter,
             allocation_queue: AllocationQueue::new(),
-            pending_allocations: AtomicU64::new(0),
-        };
-        Self { inner: Arc::new(inner) }
+        });
+
+        let weak = Arc::downgrade(&inner);
+        thread::spawn(move || {
```
yerzhan7: np: Shall we name new thread like we did `"mem-pool-maintenance"` thread?

https://github.com/awslabs/mountpoint-s3/blob/feature/memory-limit/mountpoint-s3-fs/src/memory/maintenance.rs#L45

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -61,47 +61,48 @@ pub struct PagedPool {
...
+            loop {
+                wake_signal.wait_timeout(Duration::from_secs(1));
+                if let Some(pool) = weak.upgrade() {
+                    pool.process_pending();
+                }
```
yerzhan7: We should exit the loop if `weak.upgrade()` returns `None`?

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -141,28 +142,12 @@ impl PagedPool {
...
-                PoolBuffer::new_primary(buffer_ptr, size)
-            }
-            None => {
-                metrics::histogram!("pool.reserved_bytes", "type" => "secondary", "kind" => kind.as_str())
-                    .record(size as f64);
-                self.inner.limiter.on_pool_reserve(size, cursor_id);
-                PoolBuffer::new_secondary(size, kind, self.inner.stats.clone(), Arc::downgrade(&self.inner))
-            }
-        }
+        self.inner.try_get_buffer(size, kind, cursor_id, true).unwrap()
```
yerzhan7: np: We could add `.expect("forced allocations cannot fail")` to enforce the invariant?

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -791,6 +692,8 @@ mod tests {
                 std::thread::sleep(std::time::Duration::from_millis(1));
             }
             panic!("request did not enter queue within 100ms");
+
+            // TODO(memory-limiter): review test
```
yerzhan7: Stale

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -410,6 +281,72 @@ impl PagedPoolInner {
...
+        // check above and the push (avoids the race where a buffer drop called trigger_process_pending
+        // before the entry was in the queue).
+        self.limiter.trigger_process_pending();
+        // Nudge the maintenance thread: it may be parked on its long idle interval, in which
+        // case it would otherwise sleep up to that interval before noticing the new waiter.
+        self.limiter.trigger_pruning();
+        // Await the buffer from the queue. Err(Canceled) can only happen during pool shutdown.
+        // TODO(memory-limiter): signal error to CRT via aws_future_s3_buffer_ticket_set_error.
+        rx.await
+            .unwrap_or_else(|_| self.try_get_buffer(size, kind, cursor_id, true).unwrap())
```
yerzhan7: np: Same here: `.expect("forced allocations cannot fail")`?

mountpoint-s3-fs/src/memory/limiter.rs
```diff
@@ -54,8 +55,15 @@ impl BufferArea {
...
 pub struct MemoryLimiter {
+    /// Target memory limit.
     mem_limit: u64,
-    /// Global total of reserved memory (lock-free). Used in budget checks.
+    /// Additional reserved memory for other non-buffer usage like storing metadata
+    additional_mem_reserved: u64,
+    /// Memory allocated by the pool.
+    allocated_bytes: AtomicUsize,
+    /// Memory in use by [`BufferKind`].
+    acquired_bytes: [AtomicUsize; BUFFER_KIND_COUNT],
```
yerzhan7, line 65: Shall we do AtomicU64 for `allocated_bytes` and `acquired_bytes`  here to be consistent with other fields in MemLimiter? (Easier to do comparisons/subs without casting)

passaro: Yes, we have not been consistent, but I think I'll converge everything to `usize`/`AtomicUsize` instead, since that's what we already use the most for memory / size. I'll push another commit soon.

mountpoint-s3-fs/src/memory/limiter.rs
```diff
@@ -149,29 +160,30 @@ impl MemoryLimiter {
...
     /// Query available memory tracked by the memory limiter.
-    pub fn available_mem(&self, stats: &PoolStats) -> u64 {
+    pub fn available_mem(&self) -> usize {
+        let mem_allocated = self.allocated_bytes();
         let mem_reserved = self.mem_reserved.load(Ordering::SeqCst);
-        let pool_mem_reserved = self.pool_mem_reserved(stats);
         self.mem_limit
             .saturating_sub(mem_reserved)
-            .saturating_sub(pool_mem_reserved)
+            .saturating_sub(mem_allocated)
```
yerzhan7: This means that cursor's reservation's no longer 100% guarantee buffer when actually requested by CRT? Is this intentional change?

yerzhan7:
I think i see some changes in benchmarks when comparing with #1840 results: https://github.com/awslabs/mountpoint-s3/actions/runs/27157395632
- `S3 Standard, Throughput, Memory-Limited`: `seq_read_4t` - from 384.8 to 1707.9 MiB Peak Memory

mountpoint-s3-fs/src/memory/pool.rs
yerzhan7: np: should be `on_pool_acquire` instead `on_pool_reserve`

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -435,58 +391,21 @@ impl PagedPoolInner {
...
-                undelivered.push(buffer);
-            }
+    /// Attempt to fulfill pending allocation requests from the queue.
+    ///
+    /// Called whenever memory is freed — on buffer drop, cursor release, or pool trim.
+    /// Loops until no more entries can be fulfilled or the queue is empty.
+    fn process_pending(&self) {
+        while self
+            .allocation_queue
+            .try_fulfill_front(|pending| self.try_get_buffer(pending.size, pending.kind, pending.cursor_id, false))
```
yerzhan7, line 392: This exits while loop if receiver was cancelled. I think previously `try_wake_pending` continued in that case. I.e. a cancelled entry stalls draining until the next `trigger_process_pending`.

mountpoint-s3-fs/src/memory/limiter.rs
yerzhan7: np: `on_pool_acquire`

mountpoint-s3-fs/src/memory/limiter.rs
passaro: That was never the case. However, I think here we should use the acquired memory rather than allocated. This is checking how much we can reserve within the limit, so memory in use matters. I'll rename the method as well to make that explicit.

mountpoint-s3-fs/src/memory/limiter.rs
```diff
@@ -231,12 +243,83 @@ impl MemoryLimiter {
...
+        forced: bool,
+    ) -> Option<ManagedBuffer> {
+        if forced {
+            self.allocated_bytes.fetch_add(size, Ordering::SeqCst);
+        } else {
+            let start = Instant::now();
+            let mut mem_allocated = self.allocated_bytes.load(Ordering::SeqCst);
+            loop {
+                let new_mem_allocated = mem_allocated.saturating_add(size);
+                let new_total_mem_usage = new_mem_allocated.saturating_add(self.additional_mem_reserved);
```
yerzhan7, line 262: np: So, now `try_allocate` gates on `allocated_bytes`, but `try_reserve` is still on `acquired_bytes` (which is what `BackpressureController` relies on). Might be worth documenting this discrepancy as a comment somewhere.

yerzhan7 (approved):

mountpoint-s3/src/bin/mount-s3-log-analyzer.rs
```diff
@@ -66,15 +66,15 @@ fn mem_metric_patterns() -> Vec<(&'static str, Regex)> {
         ),
         (
             "peak_pool_get_object_mib",
-            Regex::new(r"pool\.reserved_bytes\[kind=get_object\]:\s(\d+)$").unwrap(),
+            Regex::new(r"pool\.bytes_in_use\[kind=get_object\]:\s(\d+)$").unwrap(),
```
yerzhan7, line 69: Also, need to update in Stress Tests: https://github.com/passaro/mountpoint-s3/blob/mem/track-allocations/mountpoint-s3-fs/tests/stress/harness/invariants.rs#L86

yerzhan7 (approved):

yerzhan7 (approved):

PR #1848, opened 2026-06-19
stress: Add held_writes_vs_reads scenario

yerzhan7:
**What changed and why?** Need to test scenario when many writers hold unevictable buffers and do not stall buffers needed for active reads.

No

No

---

### DIFF
--- a/.github/workflows/stress.yml
+++ b/.github/workflows/stress.yml
@@ -9,7 +9,7 @@ on:
         required: true
         type: string
   push:
-    branches: [main]
+    branches: [main, feature/memory-limit]
   workflow_dispatch:
 
 permissions:
@@ -51,6 +51,7 @@ jobs:
           - sustained_writes
           - mixed_rw
           - idle_and_churn
+          - held_writes_vs_reads
     steps:
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v6
--- a/.github/workflows/stress.yml
+++ b/.github/workflows/stress.yml
@@ -9,7 +9,7 @@ on:
         required: true
         type: string
   push:
-    branches: [main, feature/memory-limit]
+    branches: [main, feature/memory-limit, "wf-changes/**"]
   workflow_dispatch:
 
 permissions:
--- a/mountpoint-s3-fs/tests/stress/scenarios/held_writes_vs_reads.rs
+++ b/mountpoint-s3-fs/tests/stress/scenarios/held_writes_vs_reads.rs
@@ -10,11 +10,12 @@ use crate::stress::workers::{HoldingWriter, LARGE_READ_OBJECT, SequentialReader}
 
 const NUM_WRITERS: usize = 48; // Matches WriteHandleLimit for MINIMUM_MEM_LIMIT memory target
 const NUM_READERS: usize = 16;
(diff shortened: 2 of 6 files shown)


mountpoint-s3-fs/tests/stress/scenarios/held_writes_vs_reads.rs
```diff
@@ -0,0 +1,37 @@
...
+use std::sync::Arc;
+use std::time::Duration;
+
+use mountpoint_s3_fs::memory::MINIMUM_MEM_LIMIT;
+
+use crate::common::fuse::TestSessionConfig;
+use crate::stress::harness::{self, Scenario, Worker, default_max_latency};
+use crate::stress::workers::{HoldingWriter, LARGE_READ_OBJECT, SequentialReader};
+
+const NUM_WRITERS: usize = 48; // Matches WriteHandleLimit for MINIMUM_MEM_LIMIT memory target
```
passaro, line 11: nit: how much work would it be to expose and use the calculation in `WriteHandleLimit`?

mountpoint-s3-fs/tests/stress/scenarios/held_writes_vs_reads.rs
```diff
@@ -0,0 +1,37 @@
...
+const NUM_READERS: usize = 16;
+const WRITE_BEFORE_HOLD: usize = 8 * 1024 * 1024 + 4 * 1024; // 8 MiB + 4 KiB
```
passaro: nit: here and below, worth explicitly setting part size?

.github/workflows/stress.yml
```diff
@@ -9,7 +9,7 @@ on:
         required: true
         type: string
   push:
-    branches: [main]
+    branches: [main, feature/memory-limit]
```
passaro: This is useful for now, but should we also add `wf-changes/**` as for benchmarks, so we can more easily test changes like this?

passaro (commented):
Looks good, but we should run the new test before merging.

yerzhan7:
> Looks good, but we should run the new test before merging.

I have run it locally manually. Cannot run it in PR because it requires merge to main first I think.

.github/workflows/stress.yml
yerzhan7: Added `wf-changes/**`, thanks

mountpoint-s3-fs/tests/stress/scenarios/held_writes_vs_reads.rs
yerzhan7: Done

mountpoint-s3-fs/tests/stress/scenarios/held_writes_vs_reads.rs
yerzhan7: Not super easy. Will leave as out-of-scope for now.

passaro:
> I have run it locally manually. Cannot run it in PR because it requires merge to main first I think.

The trick is to push to a `wf-changes/*` branch: https://github.com/awslabs/mountpoint-s3/actions/runs/28228537145/job/83626334025?pr=1848

yerzhan7:
> The trick is to push to a `wf-changes/*` branch

Got it, thanks! Here's the run: https://github.com/awslabs/mountpoint-s3/actions/runs/28230288017/job/83632133995

(Will delete branch afterwards)

PR #1850, opened 2026-06-24
Update aws-lc to 5.1.0 and invoke cleanup on exit

passaro:
Update the aws-lc CRT submodule to the latest release: `aws-lc v5.1.0`.

Also added cleanup for the CRT libraries to `mountpoint-s3-crt` in order to address an issue resulting in the address sanitizer test to fail:
- Register a single atexit handler (once, from every `aws_*_library_init`) that runs the full `aws_*_library_clean_up` sequence top-down. Each cleanup is self-guarded and idempotent, so it is safe regardless of which libraries were initialized, and joins all managed threads before aws-lc tears down.
- The init for CRT libraries were not originally paired with a cleanup. However, cleanups join the CRT's managed worker threads, and without them a worker thread could still be running a TLS handshake when the process runs its C runtime destructors, hitting a race condition. With the aws-lc 5.1.0 upgrade, that race hits an abort() call (rand.c:575) in aws-lc's FIPS RNG teardown, taking down `make test-asan` at exit.

<details>
  <summary>Full CRT changelog:</summary>

```
Submodule mountpoint-s3-crt-sys/crt/aws-lc d50ded59..6283365b:
  > Prepare v5.1.0 release (#3321)
  > Concurrency is not generally supported for EVP_AEAD_CTX_foo functions (#3318)
  > Add 'version -fips' to surface FIPS module version in openssl tool (#3315)
  > Bump AWSLC_FIPS_VERSION to 5 for next FIPS branch (#3316)
  > ci: opt in to allow-unsafe-pr-checkout for gated pull_request_target jobs (#3313)
  > ML-DSA: import and enable aarch64 assembly backend from mldsa-native (#3219)
  > Define OPENSSL_INIT_NO_ATEXIT as a no-op (#3311)
  > Bump the github-actions group across 1 directory with 4 updates (#3307)
  > Move TLS 1.3 KDF into the FIPS module and wire up ACVP (#3247)
  > Add WASIp2 build and test support (#3172)
  > Make OPENSSL_memcmp constant time (#3183)
  > Fix SSL_CTX_add_extra_chain_cert slot routing for chains added before the leaf (#3296)
  > Stabilize libgit2 integration test (#3300)
  > CI: Switch integration tests to nightly and raise alarm on failures (#3287)
  > Update MySQL CI integration to mysql-cluster-9.7.1 (#3302)
  > Support SSL_OP_IGNORE_UNEXPECTED_EOF option (#3294)
  > Fix PostgreSQL integration: match upstream revoked-cert alert regex (#3299)
  > ci: fix Windows CI breakage from VS2026 (MSVC 14.51) image bump (#3298)
  > ML-DSA: import and enable x86_64 assembly backend from mldsa-native (#3195)
  > Bump aws-cdk-lib from 2.251.0 to 2.255.0 in /tests/ci in the pip-ci group (#3283)
  > Bump the cargo-ci-lambda group in /tests/ci/lambda with 4 updates (#3284)
  > Bump cross-platform-actions/action from 0.32.0 to 1.2.0 in the github-actions group (#3285)
  > ci: move BSD actions to scheduled (#3276)
  > Add SSL_CTX_set_security_callback and related APIs for OpenSSL compat… (#3275)
  > Quote LIBCRYPTO_PATH for dynamic load test (#3259)
  > Make FIPS compiler wrapper unconditional (#3269)
  > ML-DSA support as a TLS 1.3 signature scheme (#3251)
  > Add inline documentation for API contracts (#3267)
  > use /map: linker flag to avoid running a binary to capture the hash (#3133)
  > Skip MariaDB socket conflict test unable to run as root (#3274)
  > Make rustfmt optional for Rust bindings generation (#3270)
  > Fix python 3.13 patch (#3271)
  > BoringSSL: Harden nc_email name constraint checking (#3266)
  > Gate Linux specific code to fix compilation on AIX (#3265)
  > Fix manylinux1 build: O_CLOEXEC fallback in getauxval shim (#3268)
  > Prefer CRLs with specific IDP match (#3264)
  > Add `getauxval` availability detection with `/proc/self/auxv` fallback for uclibc targets (#3250)
  > Enable Windows 7 compat path on MinGW builds (#3239)
  > Drop obsolete test_pkey_rsa.rb hunk from Ruby 3.4 patch (#3260)
  > Reject undersized buffer in pkey_dsa_sign (#3112)
  > tool-openssl/s_client: default SNI to -connect host to match OpenSSL (#3209)
  > Bump time from 0.3.36 to 0.3.47 in /tests/ci/lambda (#3248)
  > Bump the github-actions group with 2 updates (#3258)
  > Decouple the FIPS version number from the AWS-LC version number (#3211)
  > Document new versioning scheme and bump mainline to v5.0.0 (#3212)
  > Fix correctness findings from penpal testing (#3235)
  > Add SHRT_MAX caps to bound iteration and input lengths (#3240)
  > Tighten OCSP_parse_url URL parsing (#3238)
  > Log versioning and library names druing cmake build step (#3254)
  > Add new review workflow (#3230)
  > ci: declare contents: read on zig compiler workflow (#3249)
  > Release cipher_data on error path too for EVP_CTRL_INIT and EVP_CTRL_COPY (#3243)
  > Free existing responderId union arm in OCSP_RESPID setters (#3234)
  > Check parameters before comparing pqdsa public keys (#3229)
  > Reject negative pass_len in PEM_ASN1_write_bio (#3228)
  > Include trailing NUL in BIO_ADDR_rawaddress AF_UNIX length (#3236)
  > Ensure no trailing data for PKCS8 EVP_parse_private_key (#3242)
  > Free existing union arm by current type in PKCS7_set_type (#3231)
  > Reject len < -1 in ASN1_mbstring_ncopy (#3232)
  > Harden PKCS7 and OCSP error handling (#3237)
  > Fix wherelen handling in BIO_ADDR_rawmake AF_UNIX path (#3233)
  > Fix: Apply COHABITANT_HEADERS logic to location of tool binaries (#3116)
  > Add OPENSSL_cleanse to zero stack secrets before return (#3227)
  > BoringSSL: Test key import in EVPTest a bit more extensively (#3058)
  > Prepare 1.73.0 (#3226)
  > Fix shared library install on Windows: place DLLs in bin directory (#3225)
  > Fix thread-local DRBG cleanup deadlock at process exit (#3220)
  > Reject URIs containing '@' in name constraint checking (#3202)
  > ci: pin zig x86_64-windows job to windows-2022 (#3222)
  > Support non-empty context strings in ML-DSA EVP sign/verify (#3135)
  > Bump the pip-ci group across 1 directory with 4 updates (#3216)
  > Bump the cargo-ci-lambda group across 1 directory with 10 updates (#3217)
  > Bump the github-actions group with 17 updates (#3218)
  > Bump golang.org/x/crypto from 0.31.0 to 0.50.0 in the gomod-root group (#3214)
  > Fixes several issues across X509 and EVP parsing/comparison code (#3213)
  > ci: add Dependabot configuration (#3191)
  > ci: harden zig wrappers against libc++ ABI drift and opaque failures (#3190)
  > Ensure correct bio memory buffer type is assigned (#3204)
  > Rework order for initialisation of digest object. If memory allocation fails, object is now not in a corrupted state. (#3205)
  > Implement EVP_PKEY_get_private_seed to return seed representation of private key (#3200)
  > Align X.509 Limbo local patch with upstream changes (#3206)
  > Dilently drop handshake fragments at the far edge of the seq window (#3203)
  > Add `EVP_PKEY_kem_get_type` public accessor for KEM key NID (#3179)
  > Scope _STL_EXTRA_DISABLED_WARNINGS to C++ to fix Ninja + clang-cl builds (#3199)
  > Fix FIPS build under MSAN by broadening integrity-test guards (#3167)
  > tool/s_client: fix -verify depth and missing CA store (#3189)
  > Handle id-pkix-ocsp-nocheck in OCSP responder verification (#3169)
  > Add WaitForFileAccessible to fix intermittent Windows test failures (#3178)
  > Handle allocation failures in add_string's section strdup and stack push (#3187)
  > Fix grpc CI (#3196)
  > Silence two stringop-overflow false-positives (#3201)
  > Prepare v1.72.1 (#3192)
  > Update NID_rsaesOaep test certificate (#3194)
  > ci: pin cryptography to source builds in pyopenssl integration (#3193)
  > Bump MySQL version to 9.7.0 (#3185)
  > Map rsaesOaep SPKI to RSA in parse_key_type (#3181)
  > docs: update platform support tables (#3176)
  > ci: add gh-pages workflow for API documentation (#3177)
  > BoringSSL: Don't support parameterless DSA keys in SPKIs AND Set an EVP_PKEY's algorithm and data together (#3057)
  > Mitigate intermittent SSL runner timeouts on FreeBSD CI (#3171)
  > Fix intermittent `ImplDispatchTest.AEAD_AES_GCM` failure in gcc-4.8 CI (#3170)
  > Add CI for Zig compiler support (#3142)
  > Improve test portability for OPENSSL_NO_SOCK, OPENSSL_THREADS, and OPENSSL_NO_TTY (#3146)
  > Generalize SSL test runner idle-timeout retry to all tests (#3163)
  > Bump Go version in gcc-4.8 Docker image from 1.18.10 to 1.22.12 (#3168)
  > ssl: invalidate X509 leaf/chain caches in cert_set_chain_and_key and … (#3117)
  > Fix intermittent CA test failure on Windows CI when TEMP is unset (#3161)
  > Bump minimum Go version to 1.20 and update Go dependencies (#3159)
```
</details>

No.

No.

---

### DIFF
--- a/mountpoint-s3-crt/src/auth.rs
+++ b/mountpoint-s3-crt/src/auth.rs
@@ -21,5 +21,6 @@ fn auth_library_init(allocator: &Allocator) {
         unsafe {
             aws_auth_library_init(allocator.inner.as_ptr());
         }
+        crate::register_crt_cleanup_at_exit();
     });
 }
--- a/mountpoint-s3-crt/src/lib.rs
+++ b/mountpoint-s3-crt/src/lib.rs
@@ -18,9 +18,47 @@ pub mod io;
 pub mod s3;
 
 use std::ptr::NonNull;
+use std::sync::Once;
 use std::{ffi::OsStr, os::unix::prelude::OsStrExt};
 
 use crate::common::error::Error;
+
+static CRT_CLEANUP_REGISTER: Once = Once::new();
+
+/// Register a process-exit handler that cleans up every CRT library.
+///
+/// CRT library init is a demand-driven, boolean-guarded singleton with no automatic cleanup. Cleanup
+/// is what joins the CRT's managed worker threads; without it, a worker thread can still be running
+/// native code when the process runs its C runtime destructors, which aborts the process under the
+/// aws-lc FIPS RNG. Called from every `*_library_init` so we register regardless of entry point.
+fn register_crt_cleanup_at_exit() {
+    CRT_CLEANUP_REGISTER.call_once(|| {
+        // SAFETY: `crt_cleanup_at_exit` is a `'static` `extern "C"` function. Called at most once,
+        // and only from an init path, so the CRT libraries are initialized before it could run.
+        unsafe {
+            libc::atexit(crt_cleanup_at_exit);
+        }
+    });
+}
+
+/// Runs at process exit to tear down every CRT library that was initialized.
+extern "C" fn crt_cleanup_at_exit() {
+    // Each `aws_*_library_clean_up` is self-guarded and idempotent, so the full sequence is safe
+    // regardless of which libraries were initialized; top-down order lets higher layers release
+    // their references before lower layers tear down. The first step of each is
+    // `aws_thread_join_all_managed`, which blocks until CRT worker threads exit — so this must run
+    // on a non-CRT thread, which `atexit` guarantees (it runs on the thread that calls `exit`).
+    //
+    // SAFETY: each cleanup is safe whether or not its library was initialized (see above); they take
+    // no arguments and touch only process-global CRT state.
+    unsafe {
+        aws_s3_library_clean_up();
+        aws_auth_library_clean_up();
+        aws_http_library_clean_up();
+        aws_io_library_clean_up();
+        aws_common_library_clean_up();
+    }
+}
 
 pub(crate) mod private {
     /// Seals a trait to prevent clients from implementing it for their own types, since this trait
--- a/mountpoint-s3-client/CHANGELOG.md
+++ b/mountpoint-s3-client/CHANGELOG.md
@@ -3,6 +3,7 @@
 * **Breaking:** `ObjectClient::put_object_single` now requires `contents: impl AsRef<[u8]> + Send + 'static` (was `+ 'a`), so the body can be held until the request is fully torn down.
 * Update to latest CRT dependencies.
 * Add `TlsConfig` type and `S3ClientConfig::tls_config()` builder for configuring a custom CA trust store. ([#1834](https://github.com/awslabs/mountpoint-s3/pull/1834))
+* Run cleanup for CRT libraries on process exit. ([#1850](https://github.com/awslabs/mountpoint-s3/pull/1850))
 
 ## v0.20.0 (April 28, 2026)
 
--- a/mountpoint-s3-crt/CHANGELOG.md
+++ b/mountpoint-s3-crt/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Add `MetaRequestOptions::request_body`.
 * Remove the `io::stream` module (`InputStream`) and `Message::set_body_stream`, which are superseded by `MetaRequestOptions::request_body`.
 * `Message` and `MetaRequestOptions` are no longer generic over a lifetime.
+* Run cleanup for CRT libraries on process exit. ([#1850](https://github.com/awslabs/mountpoint-s3/pull/1850))
 
 ## v0.14.0 (April 28, 2026)
 
--- a/mountpoint-s3-crt/src/common.rs
+++ b/mountpoint-s3-crt/src/common.rs
@@ -25,5 +25,6 @@ fn common_library_init(allocator: &Allocator) {
         unsafe {
             aws_common_library_init(allocator.inner.as_ptr());
         }
+        crate::register_crt_cleanup_at_exit();
     });
 }
--- a/mountpoint-s3-crt/src/http.rs
+++ b/mountpoint-s3-crt/src/http.rs
@@ -17,5 +17,6 @@ fn http_library_init(allocator: &Allocator) {
         unsafe {
             aws_http_library_init(allocator.inner.as_ptr());
         }
+        crate::register_crt_cleanup_at_exit();
     });
 }
--- a/mountpoint-s3-crt/src/io.rs
+++ b/mountpoint-s3-crt/src/io.rs
@@ -22,5 +22,6 @@ pub fn io_library_init(allocator: &Allocator) {
         unsafe {
             aws_io_library_init(allocator.inner.as_ptr());
         }
+        crate::register_crt_cleanup_at_exit();
     });
 }
--- a/mountpoint-s3-crt/src/s3.rs
+++ b/mountpoint-s3-crt/src/s3.rs
@@ -22,5 +22,6 @@ pub fn s3_library_init(allocator: &Allocator) {
         unsafe {
             aws_s3_library_init(allocator.inner.as_ptr());
         }
+        crate::register_crt_cleanup_at_exit();
     });
 }
--- a/mountpoint-s3-fs/tests/common/fuse.rs
+++ b/mountpoint-s3-fs/tests/common/fuse.rs
@@ -126,48 +126,22 @@ impl TestSessionConfig {
     }
 }
 
-// Holds resources for the testing session and cleans them on drop.
-pub struct TestSession {
+/// A mounted FUSE session that unmounts and joins its worker threads when dropped.
+pub struct MountedSession {
     mount_dir: TempDir,
-    test_client: Box<dyn TestClient>,
-    prefix: String,
     // Option so we can explicitly unmount
     session: Option<FuseSession>,
     // Only set if `pass_fuse_fd` is true, will unmount the filesystem on drop.
     mount: Option<Mount>,
 }
 
-impl TestSession {
-    pub fn new(
-        mount_dir: TempDir,
-        session: FuseSession,
-        test_client: impl TestClient + 'static,
-        prefix: String,
-        mount: Option<Mount>,
-    ) -> Self {
-        Self {
-            mount_dir,
-            test_client: Box::new(test_client),
-            session: Some(session),
-            mount,
-            prefix,
-        }
-    }
-
-    pub fn mount_path(&self) -> &Path {
+impl MountedSession {
+    pub fn path(&self) -> &Path {
         self.mount_dir.path()
-    }
-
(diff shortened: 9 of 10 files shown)


mountpoint-s3-crt/src/lib.rs
```diff
@@ -18,10 +18,48 @@ pub mod io;
...
+/// CRT library init is a demand-driven, boolean-guarded singleton with no automatic cleanup. Cleanup
+/// is what joins the CRT's managed worker threads; without it, a worker thread can still be running
+/// native code when the process runs its C runtime destructors, which aborts the process under the
+/// aws-lc FIPS RNG. Called from every `*_library_init` so we register regardless of entry point.
+fn register_crt_cleanup_at_exit() {
+    CRT_CLEANUP_REGISTER.call_once(|| {
+        // SAFETY: `crt_cleanup_at_exit` is a `'static` `extern "C"` function. Called at most once,
+        // and only from an init path, so the CRT libraries are initialized before it could run.
+        unsafe {
+            libc::atexit(crt_cleanup_at_exit);
```
vladem, line 39: have we confirmed with CRT that calling `aws_*_library_clean_up` atexit is the recommended way? (I don't see any guidance in the doc, but `aws_thread_join_all_managed` looks prone to a deadlock, maybe we could use `aws_thread_set_managed_join_timeout_ns`? 

have we considered an alternative of calling cleanup in Drop impls for each of [the Rust structs](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-crt/src/s3/client.rs#L980)?

mountpoint-s3-crt/src/lib.rs
```diff
@@ -18,10 +18,48 @@ pub mod io;
...
+    // Each `aws_*_library_clean_up` is self-guarded and idempotent, so the full sequence is safe
+    // regardless of which libraries were initialized; top-down order lets higher layers release
+    // their references before lower layers tear down. The first step of each is
+    // `aws_thread_join_all_managed`, which blocks until CRT worker threads exit — so this must run
+    // on a non-CRT thread, which `atexit` guarantees (it runs on the thread that calls `exit`).
+    //
+    // SAFETY: each cleanup is safe whether or not its library was initialized (see above); they take
+    // no arguments and touch only process-global CRT state.
+    unsafe {
+        aws_s3_library_clean_up();
```
vladem, line 55: how did we figure this order? what could happen if we called these in wrong order?

mountpoint-s3-crt/src/auth.rs
```diff
@@ -21,5 +21,6 @@ fn auth_library_init(allocator: &Allocator) {
         unsafe {
             aws_auth_library_init(allocator.inner.as_ptr());
         }
+        crate::register_crt_cleanup_at_exit();
```
yerzhan7, line 24: Should we add a mountpoint-s3-crt CHANGELOG note?

Also, worth renaming PR title to mention that it also does the cleanup change (in addition to aws-lc update)

mountpoint-s3-crt/src/lib.rs
```diff
@@ -18,10 +18,48 @@ pub mod io;
...
+        }
+    });
+}
+
+/// Runs at process exit to tear down every CRT library that was initialized.
+extern "C" fn crt_cleanup_at_exit() {
+    // Each `aws_*_library_clean_up` is self-guarded and idempotent, so the full sequence is safe
+    // regardless of which libraries were initialized; top-down order lets higher layers release
+    // their references before lower layers tear down. The first step of each is
+    // `aws_thread_join_all_managed`, which blocks until CRT worker threads exit — so this must run
```
yerzhan7, line 49: np: Is there a timeout on thread join? What if CRT thread does not join and hangs the mountpoint process?

mountpoint-s3-crt/src/lib.rs
```diff
@@ -18,10 +18,48 @@ pub mod io;
...
+        // SAFETY: `crt_cleanup_at_exit` is a `'static` `extern "C"` function. Called at most once,
+        // and only from an init path, so the CRT libraries are initialized before it could run.
+        unsafe {
+            libc::atexit(crt_cleanup_at_exit);
+        }
+    });
+}
+
+/// Runs at process exit to tear down every CRT library that was initialized.
+extern "C" fn crt_cleanup_at_exit() {
```
yerzhan7, line 45: np: Would be nice to add some debug logging here (if it's possible). 

Especially, in case `libc::atexit` returns non-zero value.

mountpoint-s3-crt/src/lib.rs
```diff
@@ -18,10 +18,48 @@ pub mod io;
...
+    // on a non-CRT thread, which `atexit` guarantees (it runs on the thread that calls `exit`).
+    //
+    // SAFETY: each cleanup is safe whether or not its library was initialized (see above); they take
+    // no arguments and touch only process-global CRT state.
+    unsafe {
+        aws_s3_library_clean_up();
+        aws_auth_library_clean_up();
+        aws_http_library_clean_up();
+        aws_io_library_clean_up();
+        aws_common_library_clean_up();
```
yerzhan7, line 59: `checksums` is missing from the list?

yerzhan7 (dismissed):

mountpoint-s3-crt/src/lib.rs
passaro: we don't initialize it directly (only through s3)

yerzhan7 (approved):

mountpoint-s3-crt/src/lib.rs
passaro: The cleanup functions 1/ invoke each other in (inverted) dependency order and 2/ are noop after the first invocation or if the corresponding init had not been invoked.
So if we initialized s3, the first call cleans up everything and the others are noop. If we only initialized one of the other modules, only their cleanup will be acted upon.

mountpoint-s3-crt/src/lib.rs
passaro: I explored the Drop alternative: 1/ it does not fit the cleanup functions which are global, 2/ when limiting to a wait for the event loop group to shutdown, it was subject to deadlocks. This is because we expose the `EventLoopGroup` outside the client and even allow its thread to "own" the client itself, so unless we rethink that design, it's not a viable option.

PR #1852, opened 2026-06-26
Add `pool.allocated_bytes` and improve `pool.allocated_pages` metrics

renanmagagnin:
Adds observability for the memory the pool has actually committed, which no existing metric exposed. The limiter checks allocated_bytes + additional_mem_reserved <= mem_limit
  on every allocation, but allocated_bytes itself was never emitted — so you couldn't see how close the pool was to its ceiling, or the gap between committed and in-use memory
  (idle pages).

  Changes:

  1. Add the `pool.allocated_bytes` gauge, emitted wherever `allocated_bytes` changes (both `try_allocate` paths and `deallocate`). Stress invariants assert it stays within budget at peak and returns to zero at teardown.
  2. Label `pool.allocated_pages` and `pool.empty_pages` by `buffer_size` and `buffer_count`. With the small-page fallback, a size tier can hold pages of differing buffer counts, so a bare page count no longer maps to committed memory. The `buffer_count` dimension restores it as `sum(buffer_size * buffer_count * page_count)` and surfaces fallback activity. The size label is renamed to `buffer_size` across the affected pool metrics for clarity.
  3. Poll for the teardown gauges to settle before asserting. Freeing pool pages (which decrements `pool.allocated_bytes`) lags `drop(session)` by however long asynchronous CRT teardown takes to release its pool reference, so the teardown assertion now polls until the gauges settle rather than reading once.
  4. Promote the peak memory invariants to hard assertions. `pool.allocated_bytes` and `pool.bytes_in_use` peak breaches now fail the stress test rather than only logging, so the memory limit is enforced ahead of the feature release. Reservations may transiently overshoot, so `mem.bytes_reserved` remains informational.

No behavioral change — allocation and trim logic are untouched; only metric emission
is added or relabeled. Log-schema change for consumers: `size` becomes `buffer_size`
on the pool metrics, and `pool.allocated_pages` gains `buffer_count`.

No, internal.

---

### DIFF
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -254,7 +254,7 @@ impl MemoryLimiter {
     ) -> Option<ManagedBuffer> {
         if forced {
             self.allocated_bytes.fetch_add(size, Ordering::SeqCst);
-            metrics::gauge!("mem.allocated_bytes").increment(size as f64);
+            metrics::gauge!("pool.allocated_bytes").increment(size as f64);
         } else {
             let start = Instant::now();
             let mut mem_allocated = self.allocated_bytes.load(Ordering::SeqCst);
@@ -263,7 +263,7 @@ impl MemoryLimiter {
                 let new_total_mem_usage = new_mem_allocated.saturating_add(self.additional_mem_reserved);
                 if new_total_mem_usage > self.mem_limit {
                     trace!(new_total_mem_usage, "not enough memory to allocate");
-                    metrics::histogram!("mem.allocate_latency_us").record(start.elapsed().as_micros() as f64);
+                    metrics::histogram!("pool.allocate_latency_us").record(start.elapsed().as_micros() as f64);
                     return None;
                 }
                 // Check that the value we have read is still the same before updating it
@@ -274,8 +274,8 @@ impl MemoryLimiter {
                     Ordering::SeqCst,
                 ) {
                     Ok(_) => {
-                        metrics::gauge!("mem.allocated_bytes").increment(size as f64);
-                        metrics::histogram!("mem.allocate_latency_us").record(start.elapsed().as_micros() as f64);
+                        metrics::gauge!("pool.allocated_bytes").increment(size as f64);
+                        metrics::histogram!("pool.allocate_latency_us").record(start.elapsed().as_micros() as f64);
                         break;
                     }
                     Err(current) => mem_allocated = current, // another thread updated the atomic before us, trying again
@@ -294,7 +294,7 @@ impl MemoryLimiter {
         let size = ptr.size();
         drop(ptr);
         self.allocated_bytes.fetch_sub(size, Ordering::SeqCst);
-        metrics::gauge!("mem.allocated_bytes").decrement(size as f64);
+        metrics::gauge!("pool.allocated_bytes").decrement(size as f64);
         if let Some(kind) = kind {
             self.release_bytes(size, kind);
         } else {
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -254,6 +254,7 @@ impl MemoryLimiter {
     ) -> Option<ManagedBuffer> {
         if forced {
             self.allocated_bytes.fetch_add(size, Ordering::SeqCst);
+            metrics::gauge!("mem.allocated_bytes").increment(size as f64);
         } else {
             let start = Instant::now();
             let mut mem_allocated = self.allocated_bytes.load(Ordering::SeqCst);
@@ -273,6 +274,7 @@ impl MemoryLimiter {
                     Ordering::SeqCst,
                 ) {
                     Ok(_) => {
+                        metrics::gauge!("mem.allocated_bytes").increment(size as f64);
                         metrics::histogram!("mem.allocate_latency_us").record(start.elapsed().as_micros() as f64);
                         break;
                     }
@@ -292,6 +294,7 @@ impl MemoryLimiter {
         let size = ptr.size();
         drop(ptr);
         self.allocated_bytes.fetch_sub(size, Ordering::SeqCst);
+        metrics::gauge!("mem.allocated_bytes").decrement(size as f64);
         if let Some(kind) = kind {
             self.release_bytes(size, kind);
         } else {
--- a/mountpoint-s3-fs/src/memory/pool.rs
+++ b/mountpoint-s3-fs/src/memory/pool.rs
@@ -372,14 +372,14 @@ impl PagedPoolInner {
                     "free empty memory pool pages"
                 );
                 removed = true;
-                for (buffer_count, &n) in freed_by_count.iter().enumerate() {
-                    if n > 0 {
+                for (buffer_count, &page_count) in freed_by_count.iter().enumerate() {
+                    if page_count > 0 {
                         metrics::gauge!(
                             "pool.allocated_pages",
                             "buffer_size" => format!("{}", pool.stats.buffer_size),
                             "buffer_count" => format!("{}", buffer_count),
                         )
-                        .decrement(n as f64);
+                        .decrement(page_count as f64);
                     }
                 }
             }
--- a/mountpoint-s3-fs/src/memory/pool.rs
+++ b/mountpoint-s3-fs/src/memory/pool.rs
@@ -331,7 +331,7 @@ impl PagedPoolInner {
         {
             metrics::histogram!("pool.acquired_bytes", "type" => "primary", "kind" => kind.as_str())
                 .record(size as f64);
-            metrics::histogram!("pool.slack_bytes", "size" => format!("{}", pool.buffer_size()), "kind" => kind.as_str())
+            metrics::histogram!("pool.slack_bytes", "buffer_size" => format!("{}", pool.buffer_size()), "kind" => kind.as_str())
                     .record((pool.buffer_size() - size) as f64);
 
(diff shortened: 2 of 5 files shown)


mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -360,20 +360,38 @@ impl PagedPoolInner {
...
                 tracing::trace!(
-                    size = pool.stats.buffer_size,
+                    buffer_size = pool.stats.buffer_size,
                     pages_freed,
                     "free empty memory pool pages"
                 );
                 removed = true;
-                metrics::gauge!("pool.allocated_pages", "size" => format!("{}", pool.stats.buffer_size))
-                    .decrement(pages_freed as f64);
+                for (buffer_count, &n) in freed_by_count.iter().enumerate() {
```
passaro: nit: `n` -> `page_count` or similar

mountpoint-s3-fs/src/memory/stats.rs
```diff
@@ -34,19 +34,24 @@ impl SizePoolStats {
     }
 
     pub(super) fn add_empty_page(&self) {
-        metrics::gauge!("pool.empty_pages", "size" => format!("{}", self.buffer_size)).increment(1.0);
+        metrics::gauge!("pool.empty_pages", "buffer_size" => format!("{}", self.buffer_size)).increment(1.0);
```
passaro: no `buffer_count` here?

mountpoint-s3-fs/src/memory/limiter.rs
```diff
@@ -254,6 +254,7 @@ impl MemoryLimiter {
     ) -> Option<ManagedBuffer> {
         if forced {
             self.allocated_bytes.fetch_add(size, Ordering::SeqCst);
+            metrics::gauge!("mem.allocated_bytes").increment(size as f64);
```
passaro: Should this be `mem.` or `pool.`? Same applies to the latency to be fair.

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -85,6 +86,25 @@ pub fn assert_peak_reserved_invariant(scenario_name: &str, mem_limit: f64) {
         );
     }
 
+    if let Some(metric) = recorder.get(ALLOCATED_MEMORY_METRIC, &[]) {
```
passaro, line 128: I'm confused by the different pattern: 
- why are we adding to `mem_overshoots` here?
- why not use `check_peak_violation`?
- where are we reporting the FAIL/PASS?

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -195,7 +215,12 @@ pub fn assert_peak_rss_invariant(scenario_name: &str, ceiling_bytes: f64) {
     }
 }
 
+/// How long to wait for reservation gauges to settle to zero after the session is dropped.
+/// Freeing pool pages lags `drop(session)` by however long CRT teardown takes.
```
passaro: Why is that? Could it also affect other assertions? Isn't there any other way to wait for teardown to complete?

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -204,17 +229,48 @@ pub fn assert_teardown_invariants(scenario_name: &str) {
         );
         return;
     };
-    let mut leaks: Vec<String> = Vec::new();
+
+    // Collect any reservation gauge not back to zero.
+    let collect_leaks = || {
```
passaro: Let me recap:
- we collect 3 different types of metrics
- we check each for "leaks"
- we assemble a vector of strings
- try again multiple times within the timeout if there is any leak
- then we collect the 3 metrics **again** (just for logging?)
- finally we can `assert!`

Could we simplify? E.g. collect the metrics and keep them around while we wait, then only build the strings at the end.

passaro (changes requested):

mountpoint-s3-fs/tests/stress/harness/invariants.rs
renanmagagnin: Thanks! `pool.allocated_bytes` now uses `check_peak_violation` with its own vec, like the others. Also renamed the vecs per-metric, and promoted the two hard-bound checks to assert! for the release.

mountpoint-s3-fs/tests/stress/harness/invariants.rs
renanmagagnin: Done, thanks! Gauge handles are collected once into NamedGauge, polled in place while waiting (no string building), then leaks are built in a single final pass. Dropped the closure and the extra walk.

mountpoint-s3-fs/tests/stress/harness/invariants.rs
renanmagagnin: The CRT releases its pool references asynchronously, so gauges reach zero shortly after drop(session) returns, with no signal to await — hence polling. Other assertions are not affected: peak checks read gauge_history().max() (teardown only decrements, can't inflate a max) and the latency check doesn't touch these gauges. Let me know if I missed anything :)

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -68,24 +71,67 @@ pub fn assert_peak_reserved_invariant(scenario_name: &str, mem_limit: f64) {
...
         );
     }
 
-    let mut pool_violations: Vec<String> = Vec::new();
+    // The limiter enforces `allocated_bytes + additional_mem_reserved <= mem_limit`, so the
+    // allocated gauge is a hard bound (unlike reservations, which may transiently overshoot).
+    let mut allocated_violations: Vec<String> = Vec::new();
+    if let Some(metric) = recorder.get(ALLOCATED_MEMORY_METRIC, &[]) {
+        let peak = metric.gauge_history().max();
+        tracing::info!(
```
passaro: Could we push these `info!` into `check_peak_violation`? It's the same info (plus `scenario_name`) everywhere, right?

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -204,21 +287,57 @@ pub fn assert_teardown_invariants(scenario_name: &str) {
         );
         return;
     };
-    let mut leaks: Vec<String> = Vec::new();
+
+    // Each handle holds a live value that CRT teardown decrements, so we poll them in place.
+    let mut gauges: Vec<NamedGauge> = Vec::new();
     for (area, metric) in collect_gauges_by_label(recorder, RESERVED_MEMORY_METRIC, "area") {
```
passaro: Could `collect_gauges_by_label` return a `NamedGauge` directly?

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -204,21 +287,57 @@ pub fn assert_teardown_invariants(scenario_name: &str) {
...
-        tracing::info!(scenario = scenario_name, kind = %kind, value = v, "stress: teardown {}", IN_USE_MEMORY_METRIC);
-        check_teardown_leak(IN_USE_MEMORY_METRIC, "kind", &kind, v, &mut leaks);
+        gauges.push(NamedGauge {
+            metric_name: IN_USE_MEMORY_METRIC,
+            label: Some(("kind", kind)),
+            metric,
+        });
+    }
+
+    let any_leak = |gauges: &[NamedGauge]| gauges.iter().any(|g| g.metric.gauge() != 0.0);
```
passaro: Why use a closure here?

passaro: Consider adding a method to `NamedGauge` and then just iterate inline on line 318.

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -98,46 +144,59 @@ pub fn assert_peak_reserved_invariant(scenario_name: &str, mem_limit: f64) {
...
-/// Push a `{metric}{{{label_key}={label_value}}} peak ... exceeds effective budget ...`
-/// message onto `violations` iff `peak > effective_budget`.
+/// Push a `{metric_id} peak ... exceeds effective budget ...` message onto `violations`
+/// iff `peak > effective_budget`.
 fn check_peak_violation(
     metric_name: &str,
-    label_key: &str,
-    label_value: &str,
+    label: Option<(&str, &str)>,
     peak: u64,
```
passaro: Shall we pass a `NamedGauge` here as well?

passaro (changes requested):

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -175,28 +165,21 @@ pub fn assert_peak_reserved_invariant(scenario_name: &str, mem_limit: f64) {
...
-fn check_peak_violation(
-    metric_name: &str,
-    label: Option<(&str, &str)>,
-    peak: u64,
-    effective_budget: u64,
-    violations: &mut Vec<String>,
-) {
+/// Log a gauge's peak against the budget and, iff `peak > effective_budget`, push a
+/// `{gauge_id} peak ... exceeds effective budget ...` message onto `violations`.
+fn check_peak_violation(scenario_name: &str, gauge: &NamedGauge, effective_budget: u64, violations: &mut Vec<String>) {
```
passaro: nit: could be a method on `NamedGauge`

passaro (approved):
One minor suggestion (non-blocking)

passaro (approved):
LGTM!

PR #1853, opened 2026-06-26
Scale stress HDR bounds per metric unit

renanmagagnin:
Gauge histories in the stress test recorder reused the latency HDR histogram, whose upper bound (`HDR_HIGH`, 600M µs = 10 minutes) is sized for microseconds. Gauges record bytes (process RSS, committed pool memory), so any value above ~572 MiB was silently clamped to that ceiling — causing the peak-memory stress invariants (`gauge_history().max()`) to under-report memory peaks (e.g. a true 721.8 MiB RSS peak reported as 572.5 MiB).

  Instead of a single bound, the recorder now picks an HDR upper bound from each metric's declared `Unit` (`hdr_high_for`): microseconds keep the tight 10-minute latency bound, bytes get a 256 TiB bound (covering memory and disk-cache sizes), and counts get their own bound. The bound is chosen high enough to avoid clamping but no higher, so latency histograms stay tightly sized and byte metrics are never under-reported. Adds a regression test.

No.

No.

---

### DIFF
--- a/mountpoint-s3-fs/tests/common/test_recorder.rs
+++ b/mountpoint-s3-fs/tests/common/test_recorder.rs
@@ -184,12 +184,28 @@ pub mod stress {
     use hdrhistogram::Histogram as HdrHistogram;
     use std::sync::atomic::{AtomicU64, Ordering};
 
-    /// Upper bound of the HDR histogram — 10 minutes expressed as µs. Records beyond this
-    /// are clamped down so we never panic during recording.
-    pub const HDR_HIGH: u64 = 600_000_000;
+    /// Latency upper bound: 10 minutes in microseconds.
+    pub const LATENCY_HDR_HIGH: u64 = 600_000_000;
 
-    fn new_hdr() -> HdrHistogram<u64> {
-        HdrHistogram::<u64>::new_with_bounds(1, HDR_HIGH, 3).expect("HDR bounds valid")
+    /// Byte upper bound: 256 TiB, above any memory or disk-cache size a stress test reaches.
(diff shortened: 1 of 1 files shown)


mountpoint-s3-fs/tests/common/test_recorder.rs
```diff
@@ -184,14 +184,20 @@ pub mod stress {
     use hdrhistogram::Histogram as HdrHistogram;
     use std::sync::atomic::{AtomicU64, Ordering};
 
-    /// Upper bound of the HDR histogram — 10 minutes expressed as µs. Records beyond this
-    /// are clamped down so we never panic during recording.
+    /// Upper bound for latency HDR histograms, in microseconds (10 minutes).
     pub const HDR_HIGH: u64 = 600_000_000;
 
+    /// Upper bound for gauge-history HDR histograms, in bytes (1 TiB).
```
passaro: Why do we need 2 different values? Why is this "gauge"-specific? Why 1 TiB - I guess that's likely fine for memory, but what if tomorrow we try and measure e.g. disk space?

passaro: Could we set a single – maybe slightly higher limit?

passaro (changes requested):

mountpoint-s3-fs/tests/common/test_recorder.rs
renanmagagnin: I went per-unit rather than one shared limit so latency histograms don't over-allocate buckets, but happy to collapse to a single bound if you'd prefer the simplicity.

passaro (approved):

PR #1860, opened 2026-07-02
Use zero-copy request_body for single PutObject uploads

yerzhan7:
Bump aws-c-s3 to v0.12.8, which adds the `request_body` meta request option to send a body from caller-owned memory with no extra CRT-side allocation or copy. Wire it through `MetaRequestOptions::request_body` and use it in `put_object_single` (used by incremental/append uploads) instead of an input-stream body, so the CRT uploads directly from the pooled buffer.

No.

Yes.

---


mountpoint-s3-client/src/object_client.rs
```diff
@@ -109,12 +109,12 @@ pub trait ObjectClient {
...
 
     /// Put an object into the object store.
-    async fn put_object_single<'a>(
+    async fn put_object_single(
         &self,
         bucket: &str,
         key: &str,
         params: &PutObjectSingleParams,
-        contents: impl AsRef<[u8]> + Send + 'a,
+        contents: impl AsRef<[u8]> + Send + 'static,
```
passaro, line 117: I can't see why we need to make this static.

mountpoint-s3-client/src/s3_crt_client.rs
```diff
@@ -1036,12 +1035,6 @@ impl<'a> S3Message<'a> {
         self.checksum_config = checksum_config;
     }
 
-    /// Sets the body input stream for this message, and returns any previously set input stream.
-    /// If input_stream is None, unsets the body.
-    fn set_body_stream(&mut self, input_stream: Option<InputStream<'a>>) -> Option<InputStream<'a>> {
```
passaro, line 1041: I'd have expected this to be replaced by: `fn set_body_contents(&mut self, contents: Option<&'a [u8]>)` and called from the `put_object_single` method.

mountpoint-s3-crt/src/s3/client.rs
```diff
@@ -271,6 +271,9 @@ struct MetaRequestOptionsInner<'a> {
     /// Owned source uri for copy request, if provided.
     copy_source_uri: Option<String>,
 
+    /// Owned request body, if provided.
+    request_body: Option<Box<dyn AsRef<[u8]> + Send>>,
```
passaro, line 275: We should not need this: the body should by in `message` and just forwarded in `fn message`.

passaro (commented):
We should try and avoid the API change

mountpoint-s3-client/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Unreleased (v0.20.1)
```
passaro: Change to 0.21 here and in Cargo.toml.

mountpoint-s3-crt/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## Unreleased (v0.14.1)
```
passaro: As above, update version.

passaro (commented):
Looks good. Only needs updating the version numbers.

mountpoint-s3-client/CHANGELOG.md
yerzhan7: Fixed, but I was kinda intentionally not updating it as it would most likely result in merge conflicts (when we merge it to `main`). I wanted to update/review versions/changelogs for all updated crates in the final merge-to-main PR/commit.

passaro: I see. Any reason not to merge this change to `main` first?

yerzhan7: Yeah, might be better to merge to main first. I will create new PR to main, thanks for reviewing!

yerzhan7:
Moving to https://github.com/awslabs/mountpoint-s3/pull/1882

PR #1862, opened 2026-07-06
Add SHA pins for GitHub Actions dependencies

dannycjones:
Add SHA refs for all GitHub Action dependencies, pinning them to that specific commit. This mitigates the risk of the dependency being updated without us knowing, acting a bit like a lock file. This change implements the best practice for GHA dependencies.

Dependabot supports updating SHA pins, and thus will open PRs when required on the configured cadence.

CI only. It does not change version, only pins to the current version.

No, no customer facing changes.

---

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -65,13 +65,13 @@ jobs:
 
     steps:
     - name: Configure AWS credentials
-      uses: aws-actions/configure-aws-credentials@v6
+      uses: aws-actions/configure-aws-credentials@254c19bd240aabef8777f48595e9d2d7b972184b # v6.2.1
       with:
         role-to-assume: ${{ vars.ACTIONS_BENCH_IAM_ROLE }}
         aws-region: ${{ vars.S3_BENCH_REGION }}
         role-duration-seconds: 21600
     - name: Checkout code
-      uses: actions/checkout@v6
+      uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       with:
         ref: ${{ inputs.ref }}
         submodules: true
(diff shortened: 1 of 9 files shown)


jet-tong (commented):
LGTM, I understand the human-readable commented version should also be [updated by dependabot](https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/). 

Question we should think about too:
- Are there perhaps any reliable tools / processes that help us write these hashes, when we e.g. add a new workflow file and have to add dependency hashes without dependabot?
- Also, does pinning these versions mean we'll be getting [weekly](https://github.com/awslabs/mountpoint-s3/blob/f455c4df12fb70fc16c35cebe6727673bc42841b/.github/dependabot.yml#L6) updates to all hashes this way so we can consume e.g. security patches of github actions updates? Perhaps there are some rules we should add to dependabot.yml to make managing github actions dependency updates easier?

(Should also check with team before merging this)

dannycjones:
>Are there perhaps any reliable tools / processes that help us write these hashes, when we e.g. add a new workflow file and have to add dependency hashes without dependabot?

I was looking at this, there do exist tools to do the SHA automatically but I don't see anything universally recommended. Ultimately, it would be best to block PRs that add without the hash. However, that block could be pushed to local pre-PR checks to help catch things earlier.

I think an LLM assistant can easily remediate this too, if not using one of those tools.

>does pinning these versions mean we'll be getting [weekly](https://github.com/awslabs/mountpoint-s3/blob/f455c4df12fb70fc16c35cebe6727673bc42841b/.github/dependabot.yml#L6) updates to all hashes this way so we can consume e.g. security patches of github actions updates? Perhaps there are some rules we should add to dependabot.yml to make managing github actions dependency updates easier?

Yes, Dependabot will update according to policy. It updates major, minor, and patch versions at the moment.

I expect there will be an increase in Dependabot PRs as we'll now be updating minor/patch versions and not only major versions. You may want to consider aggregating the patch versions and maybe minor versions into batch PRs, and treating major version updates as we do today with their own PR each.

passaro (approved):

PR #1863, opened 2026-07-06
Add 7-day cooldown to GitHub Actions dependency updates

dannycjones:
Adds a cooldown to Dependabot, so PRs won't be opened until the version is at least 7 days old. This provides some mitigation to avoid picking up dependencies that may not be suitable due to bug or malicious behavior, as there is time for vetting or bug fixes.

This still respects Dependabot's cadence - for example, it will run weekly still but on that weekly run, the new versions must be at least 7 days old to be considered eligible.

Note, security updates do not respect this config and will open a PR as soon as an update is available. https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-

Zizmor would report no specification of cooldown (albeit as a pedantic finding): https://docs.zizmor.sh/audits/#dependabot-cooldown

This impacts dependencies updates only - dependencies will only be prompted to update if they are at least 7 days old.

No.

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,5 @@ updates:
   directory: "/"
   schedule:
     interval: "weekly"
+  cooldown:
+    default-days: 7


jet-tong (approved):

yerzhan7:
What happens if the dependency update is High CVE fix? Would dependabot delay posting PR by 7 days now?

jet-tong:
[Github blog](https://github.blog/changelog/2025-07-01-dependabot-supports-configuration-of-a-minimum-package-age/) says it should "Stay responsive to critical security patches.".

dannycjones:
> What happens if the dependency update is High CVE fix? Would dependabot delay posting PR by 7 days now?

No delay. This config doc expands more there. Security updates are mostly separate from dependency updates with Dependabot - it specifically calls out cooldown as not being respected by security updates. https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-

PR #1868, opened 2026-07-06
Fix false-positive stress test memory invariant failures

renanmagagnin:
These are fixes to the invariants in the stress tests. following up the metrics changes in #1852.

1. **Track the exact gauge peak.** Peak invariants read gauge_history().max(), but HDR bucketing could round a peak sitting on the ceiling up into a spurious breach. `GaugeState` now records an exact peak and the invariants read that; the histogram still backs the percentile report.

2. **Exclude worker I/O buffers from the peak-RSS invariant.** Stress workers allocate I/O buffers outside the pool, inflating RSS with memory the limiter doesn't govern. The invariant now subtracts `worker_io_buffer_bytes` before checking the ceiling.

No

No

---

### DIFF
--- a/mountpoint-s3-fs/tests/common/test_recorder.rs
+++ b/mountpoint-s3-fs/tests/common/test_recorder.rs
@@ -236,13 +236,14 @@ pub mod stress {
         }
     }
 
-    /// State behind a [`HdrMetric::Gauge`]. Holds both the latest point-in-time value and
-    /// a history of every value the gauge has taken on since installation. The history lets
-    /// tests assert true peak/percentiles of the gauge; the current
-    /// value is still needed for teardown invariants assertions.
+    /// State behind a [`HdrMetric::Gauge`]. Holds the latest point-in-time value, the exact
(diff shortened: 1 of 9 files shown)


mountpoint-s3-fs/tests/common/test_recorder.rs
```diff
@@ -270,8 +272,27 @@ pub mod stress {
             }
         }
 
+        /// Current value if this metric is a gauge, otherwise `None`. Unlike [`Self::gauge`],
+        /// this never panics on counters/histograms, so callers iterating every registered
+        /// metric (e.g. the time-series snapshotter) can filter to gauges safely.
+        pub fn try_gauge(&self) -> Option<f64> {
```
passaro: Where is this used?

renanmagagnin: This was being used to log metrics in stress tests but I haven't included those changes in this PR, thanks!

passaro (approved):

PR #1869, opened 2026-07-07
  Clamp read_part_size when exceeds memory budget

Priyankakarumuru1:
When read_part_size > data_buffer_budget, the memory pool cannot allocate even a single buffer, causing all read operations to fail.  This PR implements a hybrid validation approach. 

1. When the user explicitly sets both --max-memory-target and --read-part-size and they conflict, the mount fails immediately with an error message that directs the user to either increase the memory target or decrease the part size.
2. When the memory limit is using the default value (not explicitly set by the user), the implementation clamps read_part_size to data_buffer_budget, logs a warning about the clamping, and allows the mount to proceed with the clamped value.

Yes, introduces new mount failure when user explicitly sets conflicting values.

Yes

---

### DIFF
--- a/mountpoint-s3-fs/examples/mount_from_config.rs
+++ b/mountpoint-s3-fs/examples/mount_from_config.rs
@@ -129,13 +129,13 @@ impl ConfigOptions {
         Ok(fs_config)
     }
 
-    fn build_client_config(&self) -> Result<ClientConfig> {
+    fn build_client_config(&self, mem_limit: usize) -> Result<ClientConfig> {
         let user_agent_string = match &self.user_agent_prefix {
             Some(prefix) => format!("{prefix}/mp-exmpl"),
             None => "mountpoint-s3-example/mp-exmpl".to_string(),
         };
         let throughput_target = self.determine_throughput()?;
-        Ok(ClientConfig {
+        let mut config = ClientConfig {
             region: Region::new_user_specified(self.region.clone()),
             endpoint_url: self.endpoint_url.clone(),
             addressing_style: AddressingStyle::Automatic,
@@ -148,8 +148,13 @@ impl ConfigOptions {
             bind: None,
             part_config: PartConfig::with_part_size(self.part_size()),
             user_agent: UserAgent::new(Some(user_agent_string)),
-            memory_limit_user_specified: false,
-        })
+        };
+
+        // Validate and clamp read_part_size before returning
+        use mountpoint_s3_fs::s3::config::MemoryLimitSetting;
+        config.validate_and_clamp_read_part_size(MemoryLimitSetting::Default(mem_limit))?;
+
+        Ok(config)
     }
 
     fn build_logging_config(&self) -> LoggingConfig {
@@ -307,13 +312,11 @@ fn mount_filesystem(
         .error_logger(error_logger);
 
     // Create the client and runtime
-    let mut client_config = config.build_client_config()?;
     let mem_limit = config
         .memory_limit_bytes
         .unwrap_or(mountpoint_s3_fs::memory::MINIMUM_MEM_LIMIT);
 
-    // Validate and clamp read_part_size BEFORE creating the memory pool
-    client_config.validate_and_clamp_read_part_size(mem_limit)?;
+    let client_config = config.build_client_config(mem_limit)?;
 
     let pool = PagedPool::config()
         .with_candidate_sizes([client_config.part_config.read_size_bytes])
--- a/mountpoint-s3-fs/examples/mount_from_config.rs
+++ b/mountpoint-s3-fs/examples/mount_from_config.rs
@@ -307,10 +307,14 @@ fn mount_filesystem(
         .error_logger(error_logger);
 
     // Create the client and runtime
-    let client_config = config.build_client_config()?;
+    let mut client_config = config.build_client_config()?;
     let mem_limit = config
         .memory_limit_bytes
         .unwrap_or(mountpoint_s3_fs::memory::MINIMUM_MEM_LIMIT);
+
+    // Validate and clamp read_part_size BEFORE creating the memory pool
+    client_config.validate_and_clamp_read_part_size(mem_limit)?;
+
     let pool = PagedPool::config()
         .with_candidate_sizes([client_config.part_config.read_size_bytes])
         .with_memory_limit(mem_limit)
--- a/mountpoint-s3-fs/examples/mount_from_config.rs
+++ b/mountpoint-s3-fs/examples/mount_from_config.rs
@@ -148,6 +148,7 @@ impl ConfigOptions {
             bind: None,
             part_config: PartConfig::with_part_size(self.part_size()),
             user_agent: UserAgent::new(Some(user_agent_string)),
+            memory_limit_user_specified: false,
         })
     }
 
--- a/mountpoint-s3-fs/examples/mount_from_config.rs
+++ b/mountpoint-s3-fs/examples/mount_from_config.rs
@@ -135,7 +135,7 @@ impl ConfigOptions {
             None => "mountpoint-s3-example/mp-exmpl".to_string(),
         };
         let throughput_target = self.determine_throughput()?;
-        let config = ClientConfig {
+        Ok(ClientConfig {
             region: Region::new_user_specified(self.region.clone()),
             endpoint_url: self.endpoint_url.clone(),
             addressing_style: AddressingStyle::Automatic,
@@ -149,9 +149,7 @@ impl ConfigOptions {
             part_config: PartConfig::with_part_size(self.part_size())
                 .validate(MemoryLimitSetting::Default(mem_limit))?,
             user_agent: UserAgent::new(Some(user_agent_string)),
-        };
-
-        Ok(config)
+        })
     }
 
     fn build_logging_config(&self) -> LoggingConfig {
--- a/mountpoint-s3-fs/examples/mount_from_config.rs
+++ b/mountpoint-s3-fs/examples/mount_from_config.rs
@@ -16,7 +16,7 @@ use mountpoint_s3_fs::{
     manifest::{ChannelConfig, Manifest, ManifestMetablock, ingest_manifest},
     memory::PagedPool,
     metrics::{self, MetricsSinkHandle},
-    s3::config::{ClientConfig, PartConfig, Region, TargetThroughputSetting},
+    s3::config::{ClientConfig, MemoryLimitSetting, PartConfig, Region, TargetThroughputSetting},
 };
 use nix::sys::signal::{self, Signal};
 use nix::unistd::Pid;
@@ -135,7 +135,7 @@ impl ConfigOptions {
             None => "mountpoint-s3-example/mp-exmpl".to_string(),
         };
         let throughput_target = self.determine_throughput()?;
-        let mut config = ClientConfig {
+        let config = ClientConfig {
             region: Region::new_user_specified(self.region.clone()),
             endpoint_url: self.endpoint_url.clone(),
             addressing_style: AddressingStyle::Automatic,
@@ -146,13 +146,10 @@ impl ConfigOptions {
             expected_bucket_owner: self.expected_bucket_owner.clone(),
             throughput_target,
             bind: None,
-            part_config: PartConfig::with_part_size(self.part_size()),
+            part_config: PartConfig::with_part_size(self.part_size())
+                .validate(MemoryLimitSetting::Default(mem_limit))?,
             user_agent: UserAgent::new(Some(user_agent_string)),
         };
-
-        // Validate and clamp read_part_size before returning
-        use mountpoint_s3_fs::s3::config::MemoryLimitSetting;
-        config.validate_and_clamp_read_part_size(MemoryLimitSetting::Default(mem_limit))?;
 
         Ok(config)
     }
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -16,14 +16,16 @@ use super::stats::{BUFFER_KIND_COUNT, BufferKind};
 
 pub const MINIMUM_MEM_LIMIT: usize = 512 * 1024 * 1024;
 
+/// Calculate additional memory reserved for overhead (metadata and internal structures).
+/// Reserves 12.5% of total memory (or 128 MiB minimum) for non-buffer usage.
+fn additional_mem_reserved_for(mem_limit: usize) -> usize {
+    (mem_limit / 8).max(128 * 1024 * 1024)
+}
+
 /// Calculate the data buffer budget for a given memory limit.
-/// This is the memory available for actual data buffers after reserving overhead for metadata and internal structures.
-///
-/// The formula reserves 12.5% of total memory (or 128 MiB minimum) for non-buffer usage.
+/// This is the memory available for actual data buffers after reserving overhead.
 pub fn data_buffer_budget_for(mem_limit: usize) -> usize {
-    let min_reserved = 128 * 1024 * 1024;
-    let additional_mem_reserved = (mem_limit / 8).max(min_reserved);
-    mem_limit.saturating_sub(additional_mem_reserved)
+    mem_limit.saturating_sub(additional_mem_reserved_for(mem_limit))
 }
 
 /// Buffer areas that can be managed by the memory limiter. This is used for updating metrics.
@@ -93,7 +95,7 @@ pub struct MemoryLimiter {
 
 impl MemoryLimiter {
     pub fn new(mem_limit: usize) -> Self {
-        let additional_mem_reserved = mem_limit.saturating_sub(data_buffer_budget_for(mem_limit));
+        let additional_mem_reserved = additional_mem_reserved_for(mem_limit);
         let formatter = make_format(humansize::BINARY);
         debug!(
             "target memory usage is {} with {} reserved memory",
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -16,6 +16,16 @@ use super::stats::{BUFFER_KIND_COUNT, BufferKind};
 
 pub const MINIMUM_MEM_LIMIT: usize = 512 * 1024 * 1024;
 
+/// Calculate the data buffer budget for a given memory limit.
+/// This is the memory available for actual data buffers after reserving overhead for metadata and internal structures.
+///
+/// The formula reserves 12.5% of total memory (or 128 MiB minimum) for non-buffer usage.
+pub fn data_buffer_budget_for(mem_limit: usize) -> usize {
+    let min_reserved = 128 * 1024 * 1024;
+    let additional_mem_reserved = (mem_limit / 8).max(min_reserved);
+    mem_limit.saturating_sub(additional_mem_reserved)
+}
+
 /// Buffer areas that can be managed by the memory limiter. This is used for updating metrics.
 #[derive(Debug)]
 pub enum BufferArea {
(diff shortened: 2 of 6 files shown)


mountpoint-s3-fs/src/s3/config.rs
```diff
@@ -139,13 +142,46 @@ impl ClientConfig {
         memory_pool: PagedPool,
         validate_on_s3_path: Option<&S3Path>,
     ) -> anyhow::Result<S3CrtClient> {
+        // Check if read_part_size exceeds memory budget and apply hybrid approach:
+        // - Fail mount when both memory and part size are explicitly set and conflict
+        // - Clamp when memory is default (auto-detected)
```
yerzhan7: I think this clamping happens too late because Memory Pool will initialize paged pool of `read_size_bytes` instead of `effective_read_part_size`? We need to do it earlier so that memory pool creates correctly sized paged pool for read buffers.

mountpoint-s3-fs/src/s3/config.rs
```diff
@@ -139,13 +142,46 @@ impl ClientConfig {
...
+        if effective_read_part_size > data_buffer_budget {
+            let read_part_size_mib = effective_read_part_size / MIB;
+            let budget_mib = data_buffer_budget / MIB;
+
+            if self.memory_limit_user_specified {
+                // User explicitly set memory target - fail fast with clear error
+                anyhow::bail!(
+                    "read part size ({} MiB) exceeds memory budget ({} MiB). \
+                     Increase --max-memory-target or decrease --read-part-size.",
+                    read_part_size_mib,
```
yerzhan7: If I exceed limit just by 1 byte this would read `read part size (384 MiB) exceeds memory budget (384 MiB)`? Not very informative error message.

Consider using `humansize` helper library.

mountpoint-s3-fs/src/s3/config.rs
Priyankakarumuru1: Yes, moved the clamping to after config is built but before the pool is created.

mountpoint-s3-fs/src/s3/config.rs
```diff
@@ -133,12 +137,47 @@ impl TargetThroughputSetting {
...
+        Ok(())
+    }
+
     /// Create an [S3CrtClient]
     pub fn create_client(
         self,
         memory_pool: PagedPool,
         validate_on_s3_path: Option<&S3Path>,
     ) -> anyhow::Result<S3CrtClient> {
+        // Validation and clamping now happens before pool creation via validate_and_clamp_read_part_size()
```
yerzhan7: Let's remove this comment? Seems redundant

mountpoint-s3-fs/src/s3/config.rs
```diff
@@ -133,12 +137,47 @@ impl TargetThroughputSetting {
...
+    /// Validate and clamp read_part_size against memory budget.
+    /// Must be called BEFORE creating the PagedPool.
+    ///
+    /// - If user explicitly set both memory and part size AND they conflict → FAIL
+    /// - If memory is default (auto-detected) and conflict → CLAMP and warn
+    pub fn validate_and_clamp_read_part_size(&mut self, mem_limit: usize) -> anyhow::Result<()> {
+        // Calculate data_buffer_budget (same formula as MemoryLimiter)
+        let min_reserved = 128 * 1024 * 1024;
+        let additional_mem_reserved = (mem_limit / 8).max(min_reserved);
+        let data_buffer_budget = mem_limit.saturating_sub(additional_mem_reserved);
```
yerzhan7: Can we avoid this duplication? E.g. create common helper method `data_buffer_budget_for(mem_limit: usize)` in the `memory` module. (we can also use that new helper in unit tests)

mountpoint-s3-fs/src/s3/config.rs
```diff
@@ -51,6 +52,9 @@ pub struct ClientConfig {
 
     /// Value for the user-agent header
     pub user_agent: UserAgent,
+
+    /// Whether the memory limit was explicitly set by the user via --max-memory-target
+    pub memory_limit_user_specified: bool,
```
yerzhan7: Is it possible to create enum similar to our existing `TargetThroughputSetting` pattern? (to avoid this new bool)

```rs
 pub enum MemoryTarget {
      Default(usize), // 95% of system memory
      User(usize), // CLI flag
  }
```

mountpoint-s3-fs/src/s3/config.rs
```diff
@@ -133,12 +137,47 @@ impl TargetThroughputSetting {
...
+        let min_reserved = 128 * 1024 * 1024;
+        let additional_mem_reserved = (mem_limit / 8).max(min_reserved);
+        let data_buffer_budget = mem_limit.saturating_sub(additional_mem_reserved);
+
+        if self.part_config.read_size_bytes > data_buffer_budget {
+            if self.memory_limit_user_specified {
+                // User explicitly set both → FAIL IMMEDIATELY
+                anyhow::bail!(
+                    "read part size ({}) exceeds memory budget ({}). \
+                     Increase --max-memory-target or decrease --read-part-size.",
```
yerzhan7: Currently this will read: 

```
mount-s3 XXX --memory-target 512 --read-part-size 524288000
Error: read part size (500 MiB) exceeds memory budget (384 MiB)
```

Which will make it confusing to users about the 384 number.  (Similar case for the warning log below)

How about something like this?
```
      "read part size ({}) exceeds the memory available for data buffers ({budget}). \
       The memory target is {target}, of which {reserved} is reserved for Mountpoint overhead. \
       Increase --memory-target or decrease --read-part-size.",
```

For reference see also similar log message we do in `WriteHandleLimiter`: https://github.com/awslabs/mountpoint-s3/blob/feature/memory-limit/mountpoint-s3-fs/src/memory/write_handle_limiter.rs#L78-L90

mountpoint-s3-fs/src/s3/config.rs
```diff
@@ -243,3 +282,131 @@ fn validate_client_for_bucket(
...
+            endpoint_url: None,
+            addressing_style: AddressingStyle::Automatic,
+            dual_stack: false,
+            transfer_acceleration: false,
+            auth_config: S3ClientAuthConfig::NoSigning,
+            requester_pays: false,
+            expected_bucket_owner: None,
+            throughput_target: TargetThroughputSetting::Default,
+            bind: None,
+            part_config: PartConfig::with_part_size(excessive_size),
```
yerzhan7: Can we also have unit test with `with_read_write_sizes` to model a read-only `--read-part-size` override?

mountpoint-s3-fs/src/s3/config.rs
```diff
@@ -243,3 +282,131 @@ fn validate_client_for_bucket(
...
+        assert!(
+            result.is_ok(),
+            "Should succeed (with clamping) when memory is default and read_part_size > budget"
+        );
+
+        // Verify the read_size_bytes was clamped to budget
+        assert_eq!(
+            config.part_config.read_size_bytes, budget,
+            "read_size_bytes should be clamped to budget"
+        );
```
yerzhan7, line 378: Can we add assertion that write part was not clamped?

mountpoint-s3-fs/src/s3/config.rs
```diff
@@ -243,3 +282,131 @@ fn validate_client_for_bucket(
...
+        assert!(result.is_ok(), "Should succeed when read_part_size < budget");
+        assert_eq!(
+            config.part_config.read_size_bytes,
+            8 * 1024 * 1024,
+            "read_size_bytes should remain unchanged"
+        );
+    }
+
+    #[test]
+    fn test_read_part_size_exceeds_budget_user_specified_fails() {
```
yerzhan7, line 330: Can we create some unit tests in `cli.rs`? To test end-to-end with the CLI flags.

mountpoint-s3/src/run.rs
```diff
@@ -194,9 +194,12 @@ fn mount(args: CliArgs, client_builder: impl ClientBuilder) -> anyhow::Result<Fu
     let fuse_session_config = args.fuse_session_config()?;
     let sse = args.server_side_encryption()?;
 
-    let client_config = args.client_config(build_info::FULL_VERSION);
+    let mut client_config = args.client_config(build_info::FULL_VERSION);
 
-    // Set up a paged memory pool
+    // Validate and clamp read_part_size BEFORE creating the memory pool
+    client_config.validate_and_clamp_read_part_size(args.mem_limit())?;
```
yerzhan7: What if we compute the clamped the `part_config` inside `client_config()` constructor and return `Result<ClientConfig>`?

This way we won't forget to always call the validate method after constructing client config.

mountpoint-s3-fs/src/s3/config.rs
Priyankakarumuru1: Removed.

mountpoint-s3-fs/src/s3/config.rs
Priyankakarumuru1: Done.

mountpoint-s3-fs/src/s3/config.rs
Priyankakarumuru1: Makes sense, replaced the bool with MemoryLimitSetting enum (Default(usize) / User(usize)), same pattern as TargetThroughputSetting.

mountpoint-s3-fs/src/s3/config.rs
Priyankakarumuru1: Updated the message.

mountpoint-s3-fs/src/s3/config.rs
Priyankakarumuru1: Yes, added.

mountpoint-s3-fs/src/s3/config.rs
Priyankakarumuru1: Done.

mountpoint-s3-fs/src/s3/config.rs
Priyankakarumuru1: Yes, added unit tests.

mountpoint-s3/src/run.rs
Priyankakarumuru1: Makes sense. `client_config()` now returns Result`<ClientConfig>` and validates internally.

mountpoint-s3/src/cli.rs
```diff
@@ -822,7 +822,24 @@ impl CliArgs {
...
+        };
+
+        // Validate and clamp read_part_size before returning
+        config.validate_and_clamp_read_part_size(self.memory_limit_setting())?;
+
+        Ok(config)
+    }
+
+    pub fn memory_limit_setting(&self) -> mountpoint_s3_fs::s3::config::MemoryLimitSetting {
+        use mountpoint_s3_fs::s3::config::MemoryLimitSetting;
```
yerzhan7: `MemoryLimitSetting` is already in the same module as `ClientConfig`, which this file imports at the top (line 16). Can we add it to that group and drop the fully-qualified return type + inline use? (inconsistent style)

Same for the inline use in `mount_from_config.rs:154`

mountpoint-s3/src/cli.rs
```diff
@@ -822,7 +822,24 @@ impl CliArgs {
...
             user_agent,
+        };
+
+        // Validate and clamp read_part_size before returning
+        config.validate_and_clamp_read_part_size(self.memory_limit_setting())?;
+
+        Ok(config)
+    }
+
+    pub fn memory_limit_setting(&self) -> mountpoint_s3_fs::s3::config::MemoryLimitSetting {
```
yerzhan7: np: Can we move `memory_limit_setting` method closer to `mem_limit` method above?

Also, it would be cleaner to inverse the relationship and have the logic in 1 single place:

```rs
  pub fn mem_limit(&self) -> usize {
      self.memory_limit_setting().bytes()
  }

  pub fn memory_limit_setting(&self) -> MemoryLimitSetting {
      #[cfg(feature = "mem_limiter")]
      if let Some(max_mem_target) = self.max_memory_target {
          return MemoryLimitSetting::User(max_mem_target as usize * 1024 * 1024);
      }
  
      let default = (effective_total_memory() as f64 * 0.95) as usize;
      MemoryLimitSetting::Default(default.max(MINIMUM_MEM_LIMIT))
  }
```

mountpoint-s3-fs/src/memory/limiter.rs
```diff
@@ -83,8 +93,7 @@ pub struct MemoryLimiter {
 
 impl MemoryLimiter {
     pub fn new(mem_limit: usize) -> Self {
-        let min_reserved = 128 * 1024 * 1024;
-        let additional_mem_reserved = (mem_limit / 8).max(min_reserved);
+        let additional_mem_reserved = mem_limit.saturating_sub(data_buffer_budget_for(mem_limit));
```
yerzhan7: This seems to calculate `additional_mem_reserved` again. (it was calculated previously in `data_buffer_budget_for()`).

What if we create separate helper method for it?

```rs
 fn additional_mem_reserved_for(mem_limit: usize) -> usize {
      (mem_limit / 8).max(128 * 1024 * 1024)
  }
  pub fn data_buffer_budget_for(mem_limit: usize) -> usize {
      mem_limit.saturating_sub(additional_mem_reserved_for(mem_limit))
  }
```

mountpoint-s3-fs/src/s3/config.rs
```diff
@@ -132,7 +133,69 @@ impl TargetThroughputSetting {
...
+    }
+}
+
 impl ClientConfig {
+    /// Validate and clamp read_part_size against memory budget.
+    /// Must be called BEFORE creating the PagedPool.
+    ///
+    /// - If user explicitly set both memory and part size AND they conflict → FAIL
+    /// - If memory is default (auto-detected) and conflict → CLAMP and warn
+    pub fn validate_and_clamp_read_part_size(&mut self, memory_limit: MemoryLimitSetting) -> anyhow::Result<()> {
```
yerzhan7: Is it possible to move this to `PartConfig` method instead? And then the callers would do:

- `part_config: PartConfig::with_part_size(self.part_size()).validate(MemoryLimitSetting::Default(mem_limit))?`
- `part_config: self.part_config().validate(self.memory_limit_setting())?,`

mountpoint-s3/src/cli.rs
```diff
@@ -982,4 +1000,135 @@ mod tests {
...
+            err_msg.contains("reserved for Mountpoint overhead"),
+            "Error should explain reserved memory: {}",
+            err_msg
+        );
+    }
+
+    #[test]
+    fn validate_read_part_size_exceeds_default_memory_clamps() {
+        // Test that when default memory is used and part size exceeds, it clamps (doesn't fail)
+        // We'll use a small default by not setting --max-memory-target
```
yerzhan7: Let's delete this test as looks to be non-determenistic and depends on memory of CI machine. We already have this case covered in the `config.rs`  unit test.

mountpoint-s3/src/cli.rs
```diff
@@ -982,4 +1000,135 @@ mod tests {
...
+    #[cfg(feature = "mem_limiter")]
+    fn validate_read_part_size_within_budget() {
+        // Test that validation passes when read_part_size is within budget
+        let cli_args = CliArgs::try_parse_from([
+            "mount-s3",
+            "bucket",
+            "test/location",
+            "--max-memory-target",
+            "1024",
+            "--read-part-size",
```
yerzhan7: Can we add unit test case for `--part-size` variant?

mountpoint-s3/src/cli.rs
```diff
@@ -982,4 +1000,135 @@ mod tests {
...
+        assert_eq!(
+            memory_limit.bytes(),
+            1024 * 1024 * 1024,
+            "Should match the specified 1024 MiB"
+        );
+    }
+
+    #[test]
+    #[cfg(feature = "mem_limiter")]
+    fn validate_read_part_size_within_budget() {
```
yerzhan7: np: I think we can parameterize the test cases like this:

```rs
  #[cfg(feature = "mem_limiter")]
  #[test_case("--read-part-size", "8388608",   true  ; "read-part within budget ok")]
  #[test_case("--read-part-size", "524288000", false ; "read-part over budget fails")]
  #[test_case("--part-size",      "524288000", false ; "part-size over budget fails")]  // NEW coverage
  fn validate_read_part_size(flag: &str, value: &str, ok: bool) {
      let args = CliArgs::try_parse_from(["mount-s3", "bucket", "test/location",
          "--max-memory-target", "512", flag, value]).expect("parse");
      let result = args.client_config(build_info::FULL_VERSION);
      assert_eq!(result.is_ok(), ok);
      if !ok {
          let msg = result.unwrap_err().to_string();
          assert!(msg.contains("exceeds the memory available for data buffers"), "{msg}");
          assert!(msg.contains("reserved for Mountpoint overhead"), "{msg}");
      }
  }
```

mountpoint-s3-fs/examples/mount_from_config.rs
```diff
@@ -129,13 +129,13 @@ impl ConfigOptions {
...
 
-    fn build_client_config(&self) -> Result<ClientConfig> {
+    fn build_client_config(&self, mem_limit: usize) -> Result<ClientConfig> {
         let user_agent_string = match &self.user_agent_prefix {
             Some(prefix) => format!("{prefix}/mp-exmpl"),
             None => "mountpoint-s3-example/mp-exmpl".to_string(),
         };
         let throughput_target = self.determine_throughput()?;
-        Ok(ClientConfig {
+        let config = ClientConfig {
```
yerzhan7: np: no longer need to create `config` variable since we removed the mutation?

mountpoint-s3/src/cli.rs
```diff
@@ -803,13 +803,13 @@ impl CliArgs {
...
 
-    pub fn client_config(&self, version: &str) -> ClientConfig {
+    pub fn client_config(&self, version: &str) -> anyhow::Result<ClientConfig> {
         let instance_info = InstanceInfo::new();
         let user_agent = self.user_agent(&instance_info, version);
         let throughput_target = self.throughput_target_gbps(&instance_info);
         let region = autoconfigure::get_region(&instance_info, self.region.clone());
 
-        ClientConfig {
+        let config = ClientConfig {
```
yerzhan7: np: Same here, no longer need to create `config` variable.

mountpoint-s3/src/run.rs
yerzhan7: We now compute `mem_limit()` twice (here and inside `args.client_config()`) which is potentially expensive (due to cgroup scans, etc.). 

Let's compute the mem limit enum once, and then supply it both places:

```rs
  let memory_limit = args.memory_limit_setting();
  let client_config = args.client_config(build_info::FULL_VERSION, memory_limit)?;

  let pool = PagedPool::config()
      .with_candidate_sizes([...])
      .with_memory_limit(memory_limit.bytes())
```

Priyankakarumuru1: Good catch fixed.

mountpoint-s3-fs/examples/mount_from_config.rs
Priyankakarumuru1: you're right.

mountpoint-s3/src/cli.rs
```diff
@@ -493,17 +493,17 @@ impl CliArgs {
     }
 
     pub fn mem_limit(&self) -> usize {
```
yerzhan7: np: This method seems to be unused now and can be deleted.

yerzhan7 (approved):

mountpoint-s3/src/cli.rs
Priyankakarumuru1: Right, forgot to remove that! Fixed now.

yerzhan7 (approved):

PR #1879, opened 2026-07-13
Fix stress test hanging on cleanup when workers stuck

Priyankakarumuru1:
When a stress test worker is stuck in a FUSE syscall (e.g. blocked waiting for a memory allocation that can never be fulfilled), the test harness cleanup would hang indefinitely. The `drop(session)` call triggers `fuse_session_unmount` which waits for all in-flight FUSE requests to drain  but the stuck worker is an in-flight request that will never complete. This creates a circular dependency that causes the test to hang in CI instead of failing.
  
Replaced `drop()` with `std::mem::forget()` in the bounded-join timeout path so the test panics immediately. Process exit handles the unmount cleanup.
  
Confirmed on EC2: test now fails instead of hanging indefinitely.

No breaking change.

No, this is a test-only change.

---

### DIFF
--- a/mountpoint-s3-fs/tests/stress/harness.rs
+++ b/mountpoint-s3-fs/tests/stress/harness.rs
@@ -148,9 +148,9 @@ pub fn run(scenario: Scenario) {
 
     // Bounded join: workers observe `stop` between ops and should finish quickly. If any
     // are wedged in a kernel FUSE syscall they cannot observe `stop` and `JoinHandle` has
-    // no timeout API, so we poll `is_finished()` up to a deadline. On timeout we drop
-    // (unmount) the session to force EIO/EINTR on stuck syscalls, then panic with the
-    // stuck worker labels.
+    // no timeout API, so we poll `is_finished()` up to a deadline. On timeout we skip
+    // clean unmount (which would hang waiting for the stuck workers in a circular
+    // dependency) and let process exit force-unmount the filesystem.
     let join_deadline = Instant::now() + max_join_wait;
     while !handles.iter().all(|h| h.is_finished()) {
         if Instant::now() >= join_deadline {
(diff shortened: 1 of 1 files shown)


mountpoint-s3-fs/tests/stress/harness.rs
```diff
@@ -148,9 +148,9 @@ pub fn run(scenario: Scenario) {
 
     // Bounded join: workers observe `stop` between ops and should finish quickly. If any
     // are wedged in a kernel FUSE syscall they cannot observe `stop` and `JoinHandle` has
-    // no timeout API, so we poll `is_finished()` up to a deadline. On timeout we drop
-    // (unmount) the session to force EIO/EINTR on stuck syscalls, then panic with the
-    // stuck worker labels.
+    // no timeout API, so we poll `is_finished()` up to a deadline. On timeout we skip
+    // clean unmount (which would hang waiting for the stuck workers in a circular
+    // dependency) and let process exit force-unmount the filesystem.
```
yerzhan7, line 153: Are you sure that process exit will trigger force unmount? I thought it would just close the `/dev/fuse` connection, but the mount itself might still exist in `/proc/mounts`.

mountpoint-s3-fs/tests/stress/harness.rs
```diff
@@ -159,8 +159,12 @@ pub fn run(scenario: Scenario) {
                 .enumerate()
                 .filter_map(|(id, h)| (!h.is_finished()).then_some(labels[id].as_str()))
                 .collect();
-            drop(setup_guard);
-            drop(session);
+            // Workers are wedged (likely in uninterruptible FUSE syscalls). Clean unmount
+            // would hang in fuse_session_unmount waiting for the FUSE session thread, which
+            // is stuck waiting for the workers. Skip cleanup and panic immediately — process
+            // exit will close /dev/fuse fd and trigger kernel-side unmount.
```
yerzhan7: I am okay with the forget solution for now as stress tests are executed on ephemeral CI hosts. However, it leaks some resources like "skips the setup guard cleanup", mount directory is not deleted, and potentially some dead mount entry in `/proc/mount`.

Can we add a TODO comment (and maybe logging) about this? 

E.g.

```rs
             // TODO: Avoid resource leak
              tracing::warn!(
                  scenario = scenario_name,
                  "stress: workers wedged after stop; skipping clean unmount to avoid a deadlocked \
                   teardown. Leaking the mount dir, FUSE mount entry, and setup-guard objects"
              );
```

yerzhan7 (approved):

mountpoint-s3-fs/tests/stress/harness.rs
Priyankakarumuru1: You're correct. I've fixed the comment to not make that claim.

mountpoint-s3-fs/tests/stress/harness.rs
Priyankakarumuru1: Makes sense.

mountpoint-s3-fs/tests/stress/harness.rs
yerzhan7: np: I think it still mentions force unmount? https://github.com/awslabs/mountpoint-s3/pull/1879/changes#diff-72728c153851378a576c8a0f3b6f9fe496e48ce60a6432ac164fdef7770a0cadR153

yerzhan7 (approved):

PR #1880, opened 2026-07-13
Reserve buffer budget so writers don't starve reads

renanmagagnin:
Under memory pressure, concurrent writers could consume the entire data-buffer budget and starve active reads (the `held_writes_vs_reads` stress scenario stall). This reserves a slice of the pool budget for "prunable" allocations (reads and disk-cache read-backs, which are reclaimable/re-fetchable) that non-prunable allocations (writes) cannot take:

- `MemoryLimiter` reserves budget sized from the prunable candidate buffer sizes, and non-prunable allocations stop that many bytes short of the limit.
- The `WriteHandleLimiter` open-time cap is derived from the non-prunable budget, so an admitted writer is always backed by budget and never parks on memory reserved for reads.
- Buffer sizes are now configured as `CandidateSize` values carrying a `prunable` flag; reads and the disk cache block are marked prunable.

No

No

### DIFF
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -32,6 +32,17 @@ pub fn data_buffer_budget_for(mem_limit: usize) -> usize {
 /// prunable kind today) holds one part while it prefetches the next.
 const PRUNABLE_RESERVED_PARTS: usize = 1;
 
+/// The budget reserved for prunable allocations (0 if none is configured).
+fn prunable_reserved_for(prunable_buffer_size: usize) -> usize {
+    prunable_buffer_size * PRUNABLE_RESERVED_PARTS
+}
+
+/// The buffer budget available to writes: `data_buffer_budget_for(mem_limit)` minus the prunable
+/// reserve. Source of truth for the write-handle cap (see [`MemoryLimiter::write_buffer_budget`]).
+pub fn write_buffer_budget_for(mem_limit: usize, prunable_buffer_size: usize) -> usize {
+    data_buffer_budget_for(mem_limit).saturating_sub(prunable_reserved_for(prunable_buffer_size))
+}
+
 /// Buffer areas that can be managed by the memory limiter. This is used for updating metrics.
 #[derive(Debug)]
 pub enum BufferArea {
@@ -104,7 +115,7 @@ pub struct MemoryLimiter {
 impl MemoryLimiter {
     pub fn new(mem_limit: usize, prunable_buffer_size: usize) -> Self {
         let additional_mem_reserved = additional_mem_reserved_for(mem_limit);
-        let prunable_reserved = prunable_buffer_size * PRUNABLE_RESERVED_PARTS;
+        let prunable_reserved = prunable_reserved_for(prunable_buffer_size);
         let formatter = make_format(humansize::BINARY);
         debug!(
             "target memory usage is {} with {} reserved memory and {} reserved for prunable buffers",
@@ -139,11 +150,11 @@ impl MemoryLimiter {
         self.mem_limit.saturating_sub(self.additional_mem_reserved)
     }
 
-    /// The buffer budget available to non-evictable allocations, i.e. `data_buffer_budget - prunable_reserved`.
+    /// The buffer budget available to writes, i.e. `data_buffer_budget - prunable_reserved`.
     /// This is the ceiling writes actually see (reads may use the full `data_buffer_budget`), so
     /// [`crate::memory::WriteHandleLimiter`] derives its cap from it: an admitted writer is then
     /// always backed by budget and never waits on memory reserved for reads.
-    pub fn non_prunable_data_buffer_budget(&self) -> usize {
+    pub fn write_buffer_budget(&self) -> usize {
         self.data_buffer_budget().saturating_sub(self.prunable_reserved)
     }
 
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -633,12 +633,12 @@ mod tests {
     // TODO: Consider which tests are specific to the MemoryLimiter and which are testing the whole PagedPool.
 
     use super::*;
-    use crate::memory::{BufferKind, PagedPool};
+    use crate::memory::{BufferKind, CandidateSize, PagedPool};
     use crate::sync::atomic::Ordering;
 
     fn new_pool() -> PagedPool {
         PagedPool::config()
-            .with_candidate_sizes([1024])
+            .with_candidate_sizes([CandidateSize::new(1024)])
             .with_minimum_memory_limit()
             .build()
     }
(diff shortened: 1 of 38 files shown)


mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -482,6 +495,54 @@ impl SizePool {
...
+
+    pub fn is_prunable(&self) -> bool {
+        self.prunable
+    }
+}
+
+// Let callers pass bare sizes (owned or borrowed, e.g. from a `&[usize]` slice); these default to
+// non-prunable. Prunable sizes must be constructed explicitly via `CandidateSize::prunable`.
+//
+// TODO: remove these conversions and make every call site pass an explicit `CandidateSize`, so
```
yerzhan7: np: Worth fixing in this PR.

mountpoint-s3-fs/src/memory/pool.rs
yerzhan7: I suspect that this one can now be deleted?

mountpoint-s3-fs/src/memory/limiter.rs
yerzhan7, line 27: I think we need to update calculation logic in stress test as well? https://github.com/awslabs/mountpoint-s3/blob/feature/memory-limit/mountpoint-s3-fs/tests/stress/harness/setup/budget_hold.rs#L18

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -506,16 +567,23 @@ impl Default for PagedPoolConfig {
...
-        self.ordered_sizes.dedup();
+    /// Ignores invalid (0 or greater than [MAX_BUFFER_SIZE]) or duplicate sizes; falls back to
+    /// [DEFAULT_BUFFER_SIZE] if none are valid. On a size collision the prunable candidate wins.
+    pub fn with_candidate_sizes<I>(&mut self, buffer_sizes: I) -> &mut Self
+    where
+        I: IntoIterator,
+        I::Item: Into<CandidateSize>,
+    {
+        let mut sizes: Vec<CandidateSize> = buffer_sizes.into_iter().map(Into::into).collect();
+        sizes.retain(|s| s.bytes > 0 && s.bytes <= MAX_BUFFER_SIZE);
```
yerzhan7, line 558: What if read part size is e.g. 128MiB? Currently, it looks like the prunable reserve will be 1MiB instead of 128MiB. (you calculate `prunable_size` from  `ordered_sizes`)

Could we add a unit test for this case?

mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -968,5 +1040,48 @@ mod tests {
...
+        /// With a read reserve configured, non-read allocations must stop one read-part
+        /// short of the budget, leaving that slice allocatable only by reads. This is the
+        /// pool-level guarantee behind the `held_writes_vs_reads` starvation fix.
+        #[test]
+        fn read_reserve_keeps_last_slice_for_reads() {
+            const BUF: usize = 1024;
+            let additional_reserved = (BUF * 16).max(128 * 1024 * 1024);
+            let mem_limit = BUF * 16 + additional_reserved;
+            let pool = PagedPool::config()
+                .with_candidate_sizes([CandidateSize::prunable(BUF)])
```
yerzhan7, line 1064: Can we add unit tests with multiple prunable candidate sizes? (e.g. `[prunable(256 KiB read), prunable(1 MiB cache), new(8 MiB write)]`)

I think the filter/max logic is not tested at all at the moment.

mountpoint-s3-fs/src/memory/write_handle_limiter.rs
```diff
@@ -3,15 +3,15 @@
...
 //! forward progress — S3 multipart uploads require non-final parts to be at least 5 MiB, so this
 //! is a hard floor we cannot reduce by flushing more often. This module enforces a hard cap on
-//! the number of concurrently open write handles, derived from the available data-buffer budget
-//! and write part size:
+//! the number of concurrently open write handles, derived from the buffer budget available to
+//! writes and the write part size:
 //!
 //! ```text
-//! max_concurrent_writes = data_buffer_budget / write_part_size
+//! max_concurrent_writes = non_prunable_data_buffer_budget / write_part_size
```
yerzhan7: Could we update the customer docs as well? https://github.com/awslabs/mountpoint-s3/blob/feature/memory-limit/doc/CONFIGURATION.md#maximum-number-of-files-open-for-writing

mountpoint-s3-fs/src/memory/limiter.rs
```diff
@@ -123,12 +134,19 @@ impl MemoryLimiter {
...
+    /// This is the upper bound on buffer-backed allocations across all kinds.
     pub fn data_buffer_budget(&self) -> usize {
         self.mem_limit.saturating_sub(self.additional_mem_reserved)
     }
 
+    /// The buffer budget available to non-evictable allocations, i.e. `data_buffer_budget - prunable_reserved`.
+    /// This is the ceiling writes actually see (reads may use the full `data_buffer_budget`), so
+    /// [`crate::memory::WriteHandleLimiter`] derives its cap from it: an admitted writer is then
+    /// always backed by budget and never waits on memory reserved for reads.
+    pub fn non_prunable_data_buffer_budget(&self) -> usize {
```
yerzhan7: How about `write_buffer_budget()`? "non-prunable" is kinda long and double negative.

mountpoint-s3-fs/src/memory/write_handle_limiter.rs
```diff
@@ -216,4 +218,36 @@ mod tests {
...
     }
+
+    /// A prunable (read) reserve lowers the write cap, so every admitted writer gets a buffer and a
+    /// read can still allocate — the fix for the `held_writes_vs_reads` stall.
+    #[test]
+    fn read_reserve_lowers_the_write_cap() {
+        use crate::memory::{BufferKind, CandidateSize, PagedPool};
+
+        let pool = PagedPool::config()
+            .with_candidate_sizes([CandidateSize::prunable(PART_SIZE_8MIB)])
```
yerzhan7: np: it would be more realistic if we also add a non-prunable CandidateSize here.

mountpoint-s3-fs/src/memory/write_handle_limiter.rs
yerzhan7: Test drift: this parameter needs to be renamed to e.g. `write_buffer_budget`

(MIN_BUDGET/LARGE_BUDGET potentially also need to be adjusted/renamed)

mountpoint-s3-fs/src/memory/limiter.rs
```diff
@@ -257,22 +286,38 @@ impl MemoryLimiter {
...
             self.allocated_bytes.fetch_add(size, Ordering::SeqCst);
             metrics::gauge!("pool.allocated_bytes").increment(size as f64);
         } else {
             let start = Instant::now();
+            // Prunable kinds use the full limit; others must leave `prunable_reserved` bytes free.
+            let ceiling = if kind.is_prunable() {
+                self.mem_limit
+            } else {
+                self.mem_limit.saturating_sub(self.prunable_reserved)
+            };
```
yerzhan7, line 315: np: Would be cleaner if we re-use the helper methods:

```rs
  let budget = if kind.is_prunable() {
      self.data_buffer_budget()
  } else {
      self.write_buffer_budget()
  };
```

And then inside loop remove the `saturating_add(self.additional_mem_reserved)`:

```rs
  if new_mem_allocated > budget {
      return None;
  }
```

yerzhan7 (approved):

PR #1881, opened 2026-07-13
Improve stability of metrics tests 

passaro:
Improve the stability of the metrics tests by making cleanup in otel_export.sh more robust. The script now waits for the OTel collector to fully exit (time-box for 5s, then SIGKILL) so its port is released before the next iteration, and then clears the PID so the EXIT trap doesn't re-kill a stale PID. It also logs when cleanup runs after a failure.

No.

No.

---

### DIFF
--- a/mountpoint-s3/tests/metrics/otel_export.sh
+++ b/mountpoint-s3/tests/metrics/otel_export.sh
@@ -140,6 +140,8 @@ trigger_metrics() {
   done
 
   cat "${MOUNT_DIR}/nonexistent_file.txt" 2>/dev/null || true
+
+  echo "Completed file operations."
 }
 
 setup_mount() {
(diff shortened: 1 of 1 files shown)


mountpoint-s3/tests/metrics/otel_export.sh
```diff
@@ -48,12 +48,29 @@ if ! command -v jq &> /dev/null; then
...
 
 cleanup() {
+  local exit_code=$?
+
+  if [[ ${exit_code} -ne 0 ]]; then
+    echo "Cleaning up after failure (exit code ${exit_code})"
+  fi
+
   ! mountpoint -q "${MOUNT_DIR}" || sudo umount "${MOUNT_DIR}"
   rm -rf "${TMP_DIR}"
```
renanmagagnin: because of `set -e`, could a non-zero `umount` abort before the rest of cleanup? might be worth to add a `|| true` here

renanmagagnin (dismissed):

renanmagagnin (approved):

PR #1885, opened 2026-07-14
Clear active cursor's backward seek window under memory starvation

yerzhan7:
Move the backward seek window behind its own mutex so the pruner can clear it without the `inner` lock, and add a last-resort pruning step: when a waiter has starved past the threshold and no idle cursor is available to reset, clear one active cursor's seek window to return its pinned buffers to the pool.

Stress Test Run: https://github.com/awslabs/mountpoint-s3/actions/runs/29342006509

N/A

N/A

---


mountpoint-s3-fs/src/memory/limiter.rs
```diff
@@ -444,6 +444,36 @@ impl MemoryLimiter {
...
+    /// Clear one *active* cursor's backward seek window to release the part buffers it pins.
+    ///
+    /// Returns `true` if a window was cleared and freed at least one byte.
+    pub(super) fn clear_one_seek_window(&self) -> bool {
+        let cleared = self
+            .cursors
+            .iter()
+            .filter_map(|entry| entry.value().upgrade())
+            .filter(|state| matches!(*state.read_state.lock().unwrap(), ReadState::Active { .. }))
+            .find_map(|state| {
```
passaro: nit: I'd consider a loop here instead of the functional style. It would make it clearer that we attempt an action (clear_seek_window) and exit if it was effective.

mountpoint-s3-fs/src/memory/maintenance.rs
```diff
@@ -95,6 +97,8 @@ fn maintenance_loop(pool_inner: Weak<PagedPoolInner>, signal: Arc<WakeSignal>, i
 ///      natural release path do the work — unless the head of the queue
 ///      has been waiting beyond [`PRUNING_STARVATION_THRESHOLD`].
 ///   4. Otherwise reset one idle cursor.
+///   5. If no idle cursor was available and a waiter is starving, clear one active cursor's
```
passaro, line 101: Maybe a future optimization: should we clear the backward seek windows (starting with non-active cursors?) before resetting?

passaro (commented):
What is happening with the failed stress tests?

yerzhan7:
> What is happening with the failed stress tests?

Fixed the test by adding threshold parameter.

mountpoint-s3-fs/src/memory/limiter.rs
yerzhan7: Fixed, thanks!

mountpoint-s3-fs/src/memory/maintenance.rs
yerzhan7: Will leave as it is for now, we can do that in the future.

I was thinking that since in step 4 we already know that there no idle cursors available, it make sense to do step 5 with active cursor only.

passaro (approved):

PR #1887, opened 2026-07-15
fix: Use FUSE abort to prevent stress tests from hanging indefinitely

Priyankakarumuru1:
Stress tests hang indefinitely on CI when `fuse_session_unmount` waits for in-flight FUSE requests that never complete. Fixed by using the kernel's FUSE abort mechanism (/sys/fs/fuse/connections/*/abort) to fail stuck requests with EIO, followed by `libc::_exit(1)` to terminate immediately, with a nextest per-test timeout (45 min) as a safety net.

[Tested on the CI](https://github.com/awslabs/mountpoint-s3/actions/runs/30545339202/job/90908339229#logs).

No breaking change.

No, this is a test-only change.

---

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,3 +1,10 @@
 [profile.ci]
 # Do not cancel the test run on the first failure.
 fail-fast = false
+
+[[profile.default.overrides]]
+# Terminate stress tests that exceed 20 minutes. Stress tests run for 900s (15 min);
+# anything beyond 20 min is hung in cleanup (e.g. FUSE unmount waiting for in-flight requests).
+# grace-period = 0 sends SIGKILL immediately (SIGTERM can't kill processes in uninterruptible state).
+filter = 'test(/stress::scenarios/)'
+slow-timeout = { period = "60s", terminate-after = 20, grace-period = "0s" }
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -3,8 +3,9 @@
 fail-fast = false
 
 [[profile.default.overrides]]
-# Terminate stress tests that exceed 20 minutes. Stress tests run for 900s (15 min);
-# anything beyond 20 min is hung in cleanup (e.g. FUSE unmount waiting for in-flight requests).
+# Terminate stress tests that exceed 45 minutes. Stress tests run for 900s (15 min, see
+# STRESS_DURATION_SECS in .github/workflows/stress.yml); 45 min timeout allows buffer for
+# setup/cleanup phases that might take longer.
 # grace-period = 0 sends SIGKILL immediately (SIGTERM can't kill processes in uninterruptible state).
 filter = 'test(/stress::scenarios/)'
-slow-timeout = { period = "60s", terminate-after = 20, grace-period = "0s" }
+slow-timeout = { period = "60s", terminate-after = 45, grace-period = "0s" }
--- a/.github/workflows/stress.yml
+++ b/.github/workflows/stress.yml
@@ -28,6 +28,7 @@ env:
   S3_BUCKET_TEST_PREFIX: ${{ vars.S3_BUCKET_BENCH_PREFIX || 'mountpoint-benchmark/' }}
   S3_ENDPOINT_URL: ${{ vars.S3_BENCH_ENDPOINT_URL }}
   S3_REGION: ${{ vars.S3_BENCH_REGION }}
+  # Stress test duration. If changed, also update slow-timeout in .config/nextest.toml
   STRESS_DURATION_SECS: "900"
 
 jobs:
--- a/mountpoint-s3-fs/tests/stress/harness.rs
+++ b/mountpoint-s3-fs/tests/stress/harness.rs
@@ -1,6 +1,7 @@
 //! Core harness for stress scenarios.
 
 use std::collections::HashMap;
+use std::path::Path;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering};
 use std::thread;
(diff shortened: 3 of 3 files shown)


.github/workflows/stress.yml
```diff
@@ -42,6 +42,7 @@ jobs:
   stress:
     name: Stress (${{ matrix.scenario }})
     runs-on: [self-hosted, linux, x64, nvme-high-performance]
+    timeout-minutes: 25
```
renanmagagnin: Can we justify the choice of 25 minutes? Would be nice for confidence and also it seems like some tests (e.g. [sustained_writes](https://github.com/awslabs/mountpoint-s3/actions/runs/29505558811/job/87645301369#logs)) are taking around 22m which makes me think we might want to leave a larger margin.

yerzhan7:
I think the kernel/FUSE has special `/sys/fs/fuse/connections/<minor>/abort` file. Writing to it fails all in-flight requests, so wedged workers get `EIO` and we can fail fast, instead of a bare timed-out job. Could we explore if this is feasible and would fix the issue?

https://www.kernel.org/doc/html/next/filesystems/fuse.html#aborting-a-filesystem-connection

.github/workflows/stress.yml
yerzhan7: np: If we go timeout route, I think it's better to do timeout in test framework on the test cases. (to exclude build/setup time)

mountpoint-s3-fs/tests/stress/harness.rs
```diff
@@ -261,3 +283,21 @@ fn read_duration_env() -> Duration {
...
     Duration::from_secs(secs)
 }
+
+/// Abort all FUSE connections by writing to /sys/fs/fuse/connections/*/abort.
+/// This fails all in-flight FUSE operations with EIO, unblocking stuck threads.
+fn abort_fuse_connections(mount_path: &Path) {
+    use std::fs;
+    let connections_dir = Path::new("/sys/fs/fuse/connections");
+    if let Ok(entries) = fs::read_dir(connections_dir) {
+        for entry in entries.filter_map(Result::ok) {
```
yerzhan7: This abort all FUSE connections/mounts (migth be very bad for local dev machine). Is it possible to avoid that by finding correct FUSE mount's minor number?

mountpoint-s3-fs/tests/stress/harness.rs
```diff
@@ -261,3 +283,21 @@ fn read_duration_env() -> Duration {
         .unwrap_or(DEFAULT_DURATION_SECS);
     Duration::from_secs(secs)
 }
+
+/// Abort all FUSE connections by writing to /sys/fs/fuse/connections/*/abort.
+/// This fails all in-flight FUSE operations with EIO, unblocking stuck threads.
+fn abort_fuse_connections(mount_path: &Path) {
+    use std::fs;
```
yerzhan7: np: use top-of-file `use` is prefered

.config/nextest.toml
```diff
@@ -1,3 +1,10 @@
 [profile.ci]
 # Do not cancel the test run on the first failure.
-fail-fast = false
\ No newline at end of file
+fail-fast = false
+
+[[profile.default.overrides]]
+# Terminate stress tests that exceed 20 minutes. Stress tests run for 900s (15 min);
```
yerzhan7: Could we add a one-line cross-reference at the `STRESS_DURATION_SECS: "900"` definition in CI? (So that, we won't forget if we decide to increase duration in the future)

.config/nextest.toml
```diff
@@ -1,3 +1,10 @@
...
+# anything beyond 20 min is hung in cleanup (e.g. FUSE unmount waiting for in-flight requests).
```
yerzhan7: Let's do 45 min to be safe (some setup or cleaning phase might take longer)

mountpoint-s3-fs/tests/stress/harness.rs
```diff
@@ -184,7 +182,31 @@ pub fn run(scenario: Scenario) {
...
+            break;
+        }
+        if Instant::now() >= unmount_deadline {
+            eprintln!("stress: unmount hung after 30s, aborting FUSE connection");
+            // Abort FUSE connection - fails all in-flight requests with EIO
+            abort_fuse_connections(&mount_path);
+            thread::sleep(Duration::from_secs(2));
+            if session_thread.is_finished() {
+                let _ = session_thread.join();
+                break;
```
yerzhan7: I think we should fail here as well since normal unmount did not finish in 30s and required manual abort. What do you think?

Priyankakarumuru1: Good point! If unmount hangs for 30s and needs FUSE abort, that indicates something went wrong. We should fail the test in that case.

.config/nextest.toml
Priyankakarumuru1: Makes sense, will make it 45 minutes.

.config/nextest.toml
Priyankakarumuru1: Yes, will add a comment.

mountpoint-s3-fs/tests/stress/harness.rs
Priyankakarumuru1: Good point.

mountpoint-s3-fs/tests/stress/harness.rs
```diff
@@ -192,18 +193,13 @@ pub fn run(scenario: Scenario) {
...
-                let _ = session_thread.join();
-                break;
-            } else {
-                // Thread still stuck even after FUSE abort — terminate immediately.
-                eprintln!("stress: unmount thread still stuck after FUSE abort, terminating");
-                unsafe { libc::_exit(1) };
-            }
+            let _ = session_thread.join();
+            // Unmount hanging for 30s is a test failure even if FUSE abort unblocked it.
+            std::process::exit(1);
```
yerzhan7: Let's do `unsafe { libc::_exit(1) };` as it was before? Not sure why it was decided to be changed

mountpoint-s3-fs/tests/stress/harness.rs
```diff
@@ -192,18 +193,13 @@ pub fn run(scenario: Scenario) {
             break;
         }
         if Instant::now() >= unmount_deadline {
-            eprintln!("stress: unmount hung after 30s, aborting FUSE connection");
+            eprintln!("stress: unmount hung after 30s, aborting FUSE connection and failing test");
             // Abort FUSE connection - fails all in-flight requests with EIO
             abort_fuse_connections(&mount_path);
             thread::sleep(Duration::from_secs(2));
```
yerzhan7: This sleep is now redundant?

mountpoint-s3-fs/tests/stress/harness.rs
```diff
@@ -192,18 +193,13 @@ pub fn run(scenario: Scenario) {
...
-            if session_thread.is_finished() {
-                let _ = session_thread.join();
-                break;
-            } else {
-                // Thread still stuck even after FUSE abort — terminate immediately.
-                eprintln!("stress: unmount thread still stuck after FUSE abort, terminating");
-                unsafe { libc::_exit(1) };
-            }
+            let _ = session_thread.join();
```
yerzhan7: Let's remove this and just exit for simplicity - it can potentially still block indefinitely as you have removed the `is_finished()` check/guard.

mountpoint-s3-fs/tests/stress/harness.rs
Priyankakarumuru1: I initially thought after FUSE abort + join, normal exit would be safe. But you're right, if unmount hung for 30s, `exit(1) `atexit handlers might block on other stuck state. Better to use `_exit(1)`. Will update it.

mountpoint-s3-fs/tests/stress/harness.rs
Priyankakarumuru1: Yes, you're right. It's redundant now since `_exit` terminates immediately.

yerzhan7 (approved):
LGTM! Thanks. Let's validate last time via CI run before merging (after resolving the CI runners issue)

PR #1889, opened 2026-07-16
Add integration tests for STS web identity credential source

passaro:
Add new integration tests for STS web identity credential source, both configured directly and through a source profile. Both tests are gated under a new `web_identity_tests` crate feature.

No.

Adding changelog entries.

---


.github/workflows/integration.yml
```diff
@@ -170,6 +172,16 @@ jobs:
       with:
         # not required for client tests. TODO: make it optional.
         fuseVersion: 2
+    - name: Fetch a GitHub Actions OIDC token for web identity tests
```
yerzhan7, line 177: Should we do a post-test token cleanup step as per best security practices?

.github/workflows/integration.yml
```diff
@@ -170,6 +172,16 @@ jobs:
...
+      # Requests a fresh, real OIDC token distinct from the one used above to assume
+      # ACTIONS_IAM_ROLE, so tests can exercise the web_identity_token_file credential source
+      # by assuming S3_WEB_IDENTITY_TEST_ROLE_ARN with it.
+      run: |
```
yerzhan7, line 181: np: What is the token expiration? If it's very short, might be worth moving this step after the "Build tests" step as building can take some time.

mountpoint-s3-client/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Update to latest CRT dependencies.
 * Add `TlsConfig` type and `S3ClientConfig::tls_config()` builder for configuring a custom CA trust store. ([#1834](https://github.com/awslabs/mountpoint-s3/pull/1834))
 * Run cleanup for CRT libraries on process exit. ([#1850](https://github.com/awslabs/mountpoint-s3/pull/1850))
+* Fix credential resolution when using a source profile with STS Web Identity. ([#1834](https://github.com/awslabs/mountpoint-s3/pull/1834))
```
yerzhan7: PR link is incorrect

mountpoint-s3-client/tests/auth.rs
```diff
@@ -228,6 +230,107 @@ async fn test_profile_provider_assume_role_async() {
...
+    writeln!(config_file, "role_arn = {subsession_role}").unwrap();
+    writeln!(config_file, "source_profile = {source_profile}").unwrap();
+
+    // Set up the environment variables to use this new config file.
+    // SAFETY: This test is run in a forked process, so won't affect any other concurrently running tests.
+    unsafe {
+        std::env::set_var("AWS_CONFIG_FILE", config_file.path().as_os_str());
+        std::env::remove_var("AWS_ACCESS_KEY_ID");
+        std::env::remove_var("AWS_SECRET_ACCESS_KEY");
+        std::env::remove_var("AWS_SESSION_TOKEN");
```
yerzhan7, line 319: np: Should we also disable IMDS to prevent accidental auto-fallback by CRT to use Instance Profile credentials in case of misconfig of the test? `AWS_EC2_METADATA_DISABLED=true`(Maybe we also need to change `HOME` directory also to prevent `~/.aws` CRT lookup/fallback)

We probably safe though because we use Profile credentials I think.

.github/workflows/integration.yml
```diff
@@ -170,6 +172,16 @@ jobs:
...
+        token=$(curl -sf -H "Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
+          "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=sts.amazonaws.com" | jq -r '.value')
```
yerzhan7, line 184: Shall we do `set -euo pipefail` to ensure we catch errors during piping?

mountpoint-s3-client/tests/auth.rs
passaro: Yes, we are already setting the profile. This is the same approach for similar tests in the module.

mountpoint-s3-client/CHANGELOG.md
passaro: Oops, fixed.

.github/workflows/integration.yml
passaro: 5s, but good idea!

yerzhan7 (approved):

PR #1893, opened 2026-07-21
fix(stress): Add tolerance for pool memory metrics transient overshoot

Priyankakarumuru1:
The `pool.bytes_in_use` and `pool.allocated_bytes` invariant checks in stress tests were failing due to transient overshoots:
  
  - `pool.bytes_in_use`: peak 392.0 MiB exceeds budget 384.0 MiB. Caused by a race in buffer release/acquire - during `Page::release()`, the bitmask bit is cleared (making the slot available) before `bytes_in_use` is decremented. Another thread can acquire that slot and increment the metric in that window, causing a transient double-count of one buffer.
  - `pool.allocated_bytes`: peak 768.0 MiB exceeds budget 384.0 MiB. Caused by forced allocations, which bypass budget checks.
  
This change adds a 1-buffer tolerance to both assertions using the scenario's actual part size, so they no longer flake on these expected overshoots while still catching real memory regressions. 
 CI run: https://github.com/awslabs/mountpoint-s3/actions/runs/31604704634/job/94140652849
  
No breaking change.

No this is a test-only change.

---

### DIFF
--- a/mountpoint-s3-fs/tests/stress/harness/invariants.rs
+++ b/mountpoint-s3-fs/tests/stress/harness/invariants.rs
@@ -15,6 +15,33 @@ const RESERVED_MEMORY_METRIC: &str = "mem.bytes_reserved";
 const ALLOCATED_MEMORY_METRIC: &str = "pool.allocated_bytes";
 const IN_USE_MEMORY_METRIC: &str = "pool.bytes_in_use";
 
+/// Default S3 part size (matches client default).
+/// All stress scenarios currently use this part size.
+const DEFAULT_PART_SIZE: usize = 8 * 1024 * 1024; // 8 MiB
+
+/// Maximum number of parts that `pool.bytes_in_use` may overshoot the budget by due to
+/// concurrent release/acquire races.
+///
+/// # Background
+///
+/// The pool tracks memory in two phases:
+/// 1. `pool.allocated_bytes` — atomically enforced, never exceeds budget
+/// 2. `pool.bytes_in_use{kind}` — incremented on acquire, decremented on release
+///
+/// During buffer release, there's a small window where:
+/// - The releasing thread clears the bitmask bit (slot becomes available)
+/// - Another thread acquires that slot and increments `bytes_in_use`
+/// - The releasing thread finally decrements `bytes_in_use`
+///
+/// This creates a transient double-count. Under high concurrency (e.g., 48 writers),
+/// multiple buffers can be in this window simultaneously, causing `bytes_in_use` to
+/// temporarily spike above budget even though `allocated_bytes` remains compliant.
+///
+/// This constant allows test assertions to tolerate this known race without masking
+/// real memory regressions. Set conservatively: large enough to avoid flakes under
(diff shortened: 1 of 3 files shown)


mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -15,6 +15,33 @@ const RESERVED_MEMORY_METRIC: &str = "mem.bytes_reserved";
...
+const DEFAULT_PART_SIZE: usize = 8 * 1024 * 1024; // 8 MiB
+
+/// Maximum number of parts that `pool.bytes_in_use` may overshoot the budget by due to
+/// concurrent release/acquire races.
+///
+/// # Background
+///
+/// The pool tracks memory in two phases:
+/// 1. `pool.allocated_bytes` — atomically enforced, never exceeds budget
+/// 2. `pool.bytes_in_use{kind}` — incremented on acquire, decremented on release
```
yerzhan7: Could we add code comment *in source code (where this metric is actually emitted)* to make it more clearer for us that this metric is not accurate and it's having this issue. (Also, the current comment here is overly verbose - 2-3 sentences should be enough.)

And then here reference to that source code comment as for the reason of the tolerating overshoot logic in stress test.

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -15,6 +15,33 @@ const RESERVED_MEMORY_METRIC: &str = "mem.bytes_reserved";
...
+/// All stress scenarios currently use this part size.
+const DEFAULT_PART_SIZE: usize = 8 * 1024 * 1024; // 8 MiB
+
+/// Maximum number of parts that `pool.bytes_in_use` may overshoot the budget by due to
+/// concurrent release/acquire races.
+///
+/// # Background
+///
+/// The pool tracks memory in two phases:
+/// 1. `pool.allocated_bytes` — atomically enforced, never exceeds budget
```
yerzhan7: Actually, I think we need to do this also for the `pool.allocated_bytes` - https://github.com/awslabs/mountpoint-s3/actions/runs/30529922802/job/90829594952#step:7:1095

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -15,6 +15,33 @@ const RESERVED_MEMORY_METRIC: &str = "mem.bytes_reserved";
 const ALLOCATED_MEMORY_METRIC: &str = "pool.allocated_bytes";
 const IN_USE_MEMORY_METRIC: &str = "pool.bytes_in_use";
 
+/// Default S3 part size (matches client default).
+/// All stress scenarios currently use this part size.
```
yerzhan7: That's not true. For example this stress test uses 384MiB as part size - https://github.com/awslabs/mountpoint-s3/blob/feature/memory-limit/mountpoint-s3-fs/tests/stress/scenarios/many_readers_budget_part.rs#L17

yerzhan7 (approved):

PR #1895, opened 2026-07-24
fix: update instance throughput table

jet-tong:
### What changed and why?

Re-generates the stale (2025-04-17) EC2 instance throughput table which maps instance types to network throughput, for auto-configuring Mountpoint target throughput through [`autoconfigure::get_maximum_network_throughput`](https://github.com/awslabs/mountpoint-s3/blob/4f38ff8b6ef0bdee9cc6f9852a6cc8de0041439b/mountpoint-s3-fs/examples/mount_from_config.rs#L220).

Changes:
- Regenerated the table via `network_performance.sh`
- For instances no longer offered: removed entries after checking no regions offer them, and removed dl1.24xlarge test case
- Corrected `trn2.48xlarge` override throughput (`32000` → `3200`); also updated test case for it.
- Fixed multi-NIC overrides silently failing on bash <4.3 (e.g. `4x 100` → `4100`) and added bash version guard.
  - Multi-NIC instances now emit `None` + a warning instead of a wrong number, if not in THROUGHPUT_OVERRIDE table

No breaking changes.

No. (Just data refresh and script bug fixes).

### Possible follow-ups

- A GitHub Actions workflow to regenerate this table more often (at least once a year)?
- Use multiplication instead of override table for multi-NIC entries?
- Query more regions than the 3 in the script?

---

#### Testing

Randomly selected instance types and used aws ec2 describe-isntance-types to check for accuracy, and also used https://instances.vantage.sh/ to quick check accuracy for a small number of instance types.

#### Some instance types are no longer offered

12 instances are not offered anymore - verified gone in all regions (not just the 3 we query): `dl1.24xlarge`, `f1.2xlarge`, `f1.4xlarge`, `f1.16xlarge`, `i3.metal`, `p3.2xlarge`, `p3.8xlarge`, `p3.16xlarge`, `u-9tb1.112xlarge`, `u-12tb1.112xlarge`, `u-18tb1.112xlarge`, `u-24tb1.112xlarge`.

<details>
<summary>Verify all instances gone in all regions (click to expand)</summary>

```bash
for r in $(aws ec2 describe-regions --region us-east-1 --query "Regions[].RegionName" --output text); do
  aws ec2 describe-instance-type-offerings --region "$r" \
    --filters "Name=instance-type,Values=dl1.24xlarge,f1.2xlarge,f1.4xlarge,f1.16xlarge,i3.metal,p3.2xlarge,p3.8xlarge,p3.16xlarge,u-9tb1.112xlarge,u-12tb1.112xlarge,u-18tb1.112xlarge,u-24tb1.112xlarge" \
    --query "InstanceTypeOfferings[].[InstanceType]" --output text | sed "s/^/$r /"
done
# No output
```
</details>

#### [RESOLVED] g7.8xlarge / g7.4xlarge network performance number is unstable

> RESOLVED - they updated both g7.4xlarge and g7.8xlarge to "Up to 100 Gigabit" and "100 Gigabit" respectively, so I've updated the table again. They seemed to have changed during my commits / tests. 

<details>
<summary>Check those numbers are unstable (click to expand)</summary>

```bash
for i in $(seq 1 10); do
  aws ec2 describe-instance-types --region us-east-1 --instance-types g7.8xlarge \
    --query "InstanceTypes[0].NetworkInfo.NetworkPerformance" --output text
done | sort | uniq -c

# output: (80 appears more usually)
1 100 Gigabit
9 80 Gigabit

for i in $(seq 1 10); do
  aws ec2 describe-instance-types --region us-east-1 --instance-types g7.4xlarge \
    --query "InstanceTypes[0].NetworkInfo.NetworkPerformance" --output text
done | sort | uniq -c

# output:
9 60 Gigabit
1 Up to 100 Gigabit
```

</details>

Additional ref:
- [`ec2-instance-selector` comparators.go](https://github.com/aws/amazon-ec2-instance-selector/blob/71c31e5a8949f35ea0683ca1c27db9de00ae4fc3/pkg/selector/comparators.go#L302) throughput number parser

---


mountpoint-s3-fs/scripts/network_performance.sh
```diff
@@ -1,8 +1,19 @@
...
 # Please ensure 'jq' and AWS CLI v2 are installed before executing this script.
+# 
+# To run: 
+# cd mountpoint-s3-fs/scripts/
+# ./network_performance.sh
 
 set -euo pipefail
 
+# Requires bash 4.0+ for the `declare -A THROUGHPUT_OVERRIDE` associative array.
+# macOS ships bash 3.2, so run `brew install bash` and use that bash there.
```
passaro: The script still does not run correctly on macOS: on line 18, `date` does not support `--utc` (should be `-u`, I think).

I'd say either we don't support macOS or we need to fix that line. If the latter, let's also consider switching the script to `zsh`, which should be modern on most OSes.

passaro (changes requested):

mountpoint-s3-fs/scripts/network_performance.sh
jet-tong: I don't want to make big changes in a fix PR to support zsh, I'll just say not use macOS for the script for this.

On a tangent; noted override is only for legacy instances (newer multi NICs use already multiplied throughput); added a comment there.

passaro (dismissed):

passaro (approved):

Issue #976, opened 2024-08-08
Startup option to ignore mode change errors


pawalt:
### Tell us more about this new feature.

We currently use mountpoint-s3 to allow our users to mount buckets in their containers as regular volumes. This has been working well, but there's one footgun that some of our users have run into - the mode changing errors. We have some users with pre-written scripts that automatically do a `chmod` on their files even if the mode is already properly set. For these users, mountpoint breaks their scripts as it throws errors on the `chmod`.
 
Would folks be open to a change to allow mountpoint to ignore change mode errors, effectively making them a no-op? This would be opt-in startup behavior, and it would make our lives much easier.

dannycjones:
Hey @pawalt, thanks for opening this feature request. I'll bring this to the team, I can see why this could be useful.

helenweng-stripe:
hey @dannycjones any updates on prioritizing this? It would be useful for us as well if we want to replace the Hadoop fs with s3 mountpoint.

unexge:
Hey @helenweng-stripe, thanks for sharing your use-case. There is no update we can share as of now. I'd suggest adding thumps up reaction (👍) to the main comment to help us to track interest on this feature.

daveshm:
Hello, this also impacts django applications because the default storage backend health check runs chmod on a temp file https://github.com/django/django/blob/5a1cae3a5675c5733daf5949759476d65aa0e636/django/core/files/storage/filesystem.py#L141

taktakpeops:
Hello, my use-case for such feature is for GitHub self hosted runners. I use a the mountpoint S3 CSI driver to cache the tools for the runners. Because of the issue with chmod, I can't directly mount the `_tool` folder.

Issue #600, opened 2023-11-08
setattr should return error on unsupported attributes


monthonk:
### Tell us more about this new feature.

For some applications that require `setattr` on access time to work like `touch`, we allow users to call `setattr` operation to set access time and modified time when the files/directories are not persisted on S3 yet. With the fuse interface, we do accept other attributes too, e.g. uid and gid, but silently ignore them and return a success response:

```
    fn setattr(
        &self,
        _req: &Request<'_>,
        ino: u64,
        _mode: Option<u32>,
        _uid: Option<u32>,
        _gid: Option<u32>,
        _size: Option<u64>,
        atime: Option<TimeOrNow>,
        mtime: Option<TimeOrNow>,
```

The correct behavior might be explicitly return an error when we don't support them. That is what we defined in [Mountpoint tenets](https://github.com/awslabs/mountpoint-s3/blob/main/doc/SEMANTICS.md#behavior-tenets).

PR #1897, opened 2026-07-25
Return EPERM from setattr for unsupported attributes

akhileshvattumilli:
Addresses #600.

Mountpoint's `setattr` accepted `mode`, `uid`, and `gid` from FUSE and threw them away — they were bound as `_mode`, `_uid`, and `_gid` in the FUSE handler and never passed on. So a `chmod` or `chown` on a file that hadn't been uploaded yet returned success while Mountpoint quietly kept the mode and owner configured at mount time. The same call on a file already in S3 failed with `EPERM`, because `setattr` on a remote inode is rejected before it reaches that point. The result was that whether `chmod` appeared to work depended on whether the file happened to have been uploaded yet.

This makes the unsupported cases fail consistently. `S3Filesystem::setattr` now takes the requested `mode`, `uid`, and `gid`, and returns `EPERM` when any of them is set. Times are untouched, so `utimensat` on a not-yet-uploaded file still works and `touch` keeps behaving as it does today.

I put the check in `S3Filesystem::setattr` rather than in the FUSE handler for two reasons: it keeps the behavior consistent for anything using `mountpoint-s3-fs` as a library rather than through a mount, and it makes the case testable without mounting a filesystem, which is why the new tests live in `tests/fs.rs`.

`EPERM` is what `setattr` already returns for a remote inode via `InodeError::SetAttrNotPermittedOnRemoteInode`, so after this change `chmod` reports the same error regardless of upload state. This is also the behavior the [tenets](https://github.com/awslabs/mountpoint-s3/blob/main/doc/SEMANTICS.md#behavior-tenets) describe — failing explicitly rather than accepting an operation that will never be persisted — and `SEMANTICS.md` already documents `chmod`, `chown`, and `chgrp` as unsupported, so this brings the code in line with the documentation rather than changing what's promised.

Yes, deliberately — that is the point of the change. `chmod`, `chown`, and `chgrp` on a file that hasn't been uploaded yet now fail with `EPERM` instead of silently reporting success. Applications that create a file and then set its mode (`install`, `cp -p`, `tar -x`, or an `mkstemp` followed by `fchmod`) will now see the error at that call. For files already in S3 there is no change, since those already returned `EPERM`.

One thing worth raising: #976 asks for the opposite, an opt-in flag that turns mode-change errors into no-ops, and this change makes the current behavior slightly more visible by extending the error to the pre-upload case. I kept this PR to what #600 asks for rather than trying to settle that. If you'd rather have the escape hatch land first, or want `setattr` to accept a request whose values already match what Mountpoint reports, I'm happy to follow up.

Nothing else changes: setting times through `setattr`, truncation, and the `--allow-overwrite` hint all behave as before.

Yes to both, and both are included.

Changelog entries are added to `mountpoint-s3/CHANGELOG.md` for the user-visible `chmod`/`chown` change and to `mountpoint-s3-fs/CHANGELOG.md` for the crate.

For versions: `mountpoint-s3-fs` goes from `0.10.0` to `0.11.0`, since adding arguments to `S3Filesystem::setattr` is a breaking change for the crate and the middle number carries breaking changes for a `0.x` crate. `mountpoint-s3` goes from `1.23.0` to `1.24.0`, treating the behavior change as the "very minor breaking change" case rather than a plain bug fix. The workspace dependency on `mountpoint-s3-fs` and `Cargo.lock` are updated to match. If you read the Mountpoint change as a bug fix warranting `1.23.1` instead, that's an easy change.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2754,7 +2754,7 @@ dependencies = [
 
 [[package]]
 name = "mountpoint-s3"
-version = "1.23.0"
+version = "1.24.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
@@ -2876,7 +2876,7 @@ dependencies = [
 
 [[package]]
 name = "mountpoint-s3-fs"
-version = "0.10.0"
+version = "0.11.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
-mountpoint-s3-fs = { version = "0.10.0", path = "./mountpoint-s3-fs" }
+mountpoint-s3-fs = { version = "0.11.0", path = "./mountpoint-s3-fs" }
 mountpoint-s3-client = { version = "0.21.0", path = "./mountpoint-s3-client" }
 mountpoint-s3-crt = { version = "0.15.0", path = "./mountpoint-s3-crt" }
 mountpoint-s3-crt-sys = { version = "0.17.0", path = "./mountpoint-s3-crt-sys" }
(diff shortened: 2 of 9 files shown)


yerzhan7:
HI @akhileshvattumilli, thanks for the PR.

It looks like this is going to be a breaking change, and I am hesitant to merge it even though I do recognize that this is inconsistent behaviour in Mountpoint.

So, I'd like to first understand the concrete motivation before we take a breaking change here. Is there a specific use case or user issue where the current inconsistency is actually causing a problem? Right now it reads more like a tenet/consistency cleanup than a reported pain point.

Additionally, #600 was left open for 3 years and got zero customer interactions (issue was created by team member), whereas the opposite issue #976 has more customer demand. (To be clear, I'm not saying we should just implement #976 as-is either; that needs its own design discussion.)

PR #1904, opened 2026-07-28
Copy first part to heap in `do_read` to avoid pool deadlock

renanmagagnin:
On a FUSE read spanning two prefetch parts, `Cursor::do_read` accumulated the first part
into `response` via `ChecksummedBytes::extend`, whose empty-self fast path aliases the
part's pool buffer zero-copy, then blocked awaiting the next part's allocation while still
pinning that buffer. Under memory pressure the read reserve holds only one part of headroom,
so the second allocation could never be satisfied and the read wedged (the pruning backstop
clears the seek window, but `response` held the buffer). This copies the first part off its
pool buffer before blocking, leaving the reclaimable seek-window clone as the only pool
reference. The common single-part read is unchanged (early-returns zero-copy).

The copy is untracked heap, bounded to **`max_threads` × 1 MiB (16 MiB at defaults)**:
concurrent `do_read` calls are capped by FUSE worker threads (`max_threads`, default 16),
and each copy is at most one FUSE read (Linux caps a single read at 1 MiB). That fits inside
the `additional_mem_reserved` overhead reserve of `max(mem_limit / 8, 128 MiB)`. Raising
`--max-threads` raises the ceiling proportionally.

Benchmarks: https://github.com/awslabs/mountpoint-s3/actions/runs/30529923080
Stress tests: https://github.com/awslabs/mountpoint-s3/actions/runs/30529922802/job/90829594952

No. The single-part read path (common case) is untouched and still zero-copy; only multi-part
straddle reads change.

No

### DIFF
--- a/mountpoint-s3-fs/src/checksums.rs
+++ b/mountpoint-s3-fs/src/checksums.rs
@@ -233,6 +233,13 @@ impl ChecksummedBytes {
         Ok((self.buffer, self.checksum))
     }
 
+    /// Copy the bytes into a freshly allocated buffer sharing no storage with the original, e.g. to
+    /// release a pooled buffer while this data lives on. Reuses the checksum, so no CRC is recomputed.
+    pub fn into_detached(self) -> Result<Self, IntegrityError> {
+        let (bytes, checksum) = self.into_inner()?;
+        Ok(Self::new_from_inner_data(Bytes::copy_from_slice(&bytes), checksum))
+    }
+
     /// Return the slice of `buffer` corresponding to `range`.
     ///
     /// Note that no data is copied: the returned `Bytes` still points to a subslice of `buffer`.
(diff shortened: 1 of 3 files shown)


mountpoint-s3-fs/src/prefetch/cursor.rs
```diff
@@ -181,7 +182,15 @@ where
             }
 
             let part_len = part_bytes.len() as u64;
-            response.extend(part_bytes)?;
+            if response.is_empty() {
```
yerzhan7, line 184: Could we please add some unit test for this multi-part case (`read_size > part_size`)? Maybe we already have one in Prefetch or RequestTask?

mountpoint-s3-fs/src/prefetch/cursor.rs
```diff
@@ -181,7 +182,15 @@ where
...
+                // Copy off the pool buffer instead of letting `extend` alias it: this read spans
+                // parts, and holding a pool buffer across the next allocation below deadlocks under
+                // memory pressure. Reuse the checksum to avoid recomputing it.
+                let (bytes, checksum) = part_bytes.into_inner()?;
+                response = ChecksummedBytes::new_from_inner_data(Bytes::copy_from_slice(&bytes), checksum);
```
yerzhan7: Worth refactoring this into some `ChecksummedBytes` helper method (e.g. `deep_clone()` or `into_detached()`) to avoid leaking checksum detail.

yerzhan7 (approved):
LGTM. Let's validate benchmarks/stress CI run before merging

mountpoint-s3-fs/src/checksums.rs
```diff
@@ -233,6 +233,13 @@ impl ChecksummedBytes {
         Ok((self.buffer, self.checksum))
     }
 
+    /// Copy the bytes into a freshly allocated buffer sharing no storage with the original, e.g. to
+    /// release a pooled buffer while this data lives on. Reuses the checksum, so no CRC is recomputed.
```
passaro, line 237: > Reuses the checksum, so no CRC is recomputed.

Not true in all cases. See how `shrink_to_fit` is implemented.

mountpoint-s3-fs/src/prefetch/cursor.rs
```diff
@@ -181,7 +181,14 @@ where
             }
 
             let part_len = part_bytes.len() as u64;
-            response.extend(part_bytes)?;
+            if response.is_empty() {
+                // Copy off the pool buffer instead of letting `extend` alias it: this read spans
+                // parts, and holding a pool buffer across the next allocation below deadlocks under
+                // memory pressure.
+                response = part_bytes.into_detached()?;
```
passaro, line 188: Now we are always doing an extra copy for the first chunk, and then `extend` will do (at least) another allocation and copy both the old and the new data. 
Shall we instead consider a wider change where we directly allocate the whole buffer (if a single chunk is not enough) and only copy the data from each chunk once?

passaro: If we used a single allocation here, we could also consider going back to the pool with a "force" `get_buffer` for better tracking.

PR #1905, opened 2026-07-28
Fix Python lint errors in benchmark and packaging scripts

Priyankakarumuru1:
CI runs uvx ruff check . which picks up the latest ruff version. Ruff 0.16.0 expanded default rules from 59 to 413, which flagged existing code in `benchmark/ `and `package/`.
  
Commit 1: Ran` ruff check --fix `and `ruff format` - these are all mechanical, tool-applied changes (import sorting, type annotation modernization, etc.) with no behavioral impact.
  
Commit 2: Added `# noqa` suppressions for remaining rules where fixing them would change code behavior, which is not worth the risk for a lint-only PR. Also removed an unused import and fixed file permissions.
  
Tested uvx ruff check locally and passes clean in both directories.

No - commit 1 is tool-applied safe fixes, commit 2 only adds lint suppressions.

No - CI fix only, no user-facing changes.

---

### DIFF
--- a/benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
+++ b/benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
@@ -49,7 +49,7 @@ class SmartBenchmarkSweeper(Sweeper):
                 params = benchmark_config.get("params", {})
                 return [f"{key}={value}" for key, value in params.items()]
             return []
-        except Exception as e:
+        except Exception as e:  # noqa: BLE001
             log.error(f"Failed to load config for {benchmark_type}: {e}")
             return []
 
--- a/benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
+++ b/benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
@@ -1,16 +1,16 @@
-from dataclasses import dataclass
 import itertools
 import logging
+from dataclasses import dataclass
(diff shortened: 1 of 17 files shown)


benchmark/hydra_plugins/smart_sweeper/smart_benchmark_sweeper.py
```diff
@@ -41,19 +41,19 @@ def setup(self, *, hydra_context: HydraContext, task_function: TaskFunction, con
...
+    def _load_benchmark_params(self, benchmark_type: str) -> list[str]:
         try:
             config_path = Path("conf") / "hydra" / "sweeper" / f"{benchmark_type}.yaml"
             if config_path.exists():
                 benchmark_config = OmegaConf.load(config_path)
                 params = benchmark_config.get("params", {})
                 return [f"{key}={value}" for key, value in params.items()]
             return []
-        except Exception as e:
+        except Exception as e:  # noqa: BLE001
```
renanmagagnin: Can we avoid these suppressions? If not, can we justify in the PR or even maybe in the code?

Priyankakarumuru1: These are intentional catch-all handlers that log and continue, changing them would mean a bigger refactor with no real benefit for simple scripts. The `#noqa `acknowledges the rule but keeps the existing behavior. Will justify in the PR as well.

renanmagagnin (approved):

PR #1907, opened 2026-07-28
stress: Add cache_hit_vs_miss_held_budget scenario

yerzhan7:
Adding cache hit vs cache miss stress test scenario under single 8MB read part size available.

Sample run: https://github.com/awslabs/mountpoint-s3/actions/runs/30822959093/job/91717290367

No

No

---

### DIFF
--- a/.github/workflows/stress.yml
+++ b/.github/workflows/stress.yml
@@ -59,6 +59,7 @@ jobs:
           - single_reader_held_budget_direct_io
           - many_readers_held_budget_direct_io
           - single_reader_held_budget_misaligned_part
+          - cache_hit_vs_miss_held_budget
     steps:
       - name: Configure AWS credentials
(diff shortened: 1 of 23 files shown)


renanmagagnin (approved):

.github/workflows/stress.yml
```diff
@@ -89,6 +90,17 @@ jobs:
           rustflags: ""
       - name: Install cargo-nextest
         uses: taiki-e/install-action@nextest
+      - name: Set up local NVMe cache storage
+        if: contains(matrix.scenario, 'cache')
+        run: |
+          local_storage=/localstorage
+          sudo mkfs -t ext4 /dev/nvme1n1
```
renanmagagnin, line 97: should we check if it's already mounted?

yerzhan7: it cannot be already mounted - hosts are spin up On-Demand 1 per job AFAIK.

Thanks for reviewing!

PR #1908, opened 2026-07-28
Add incremental upload stress test

Priyankakarumuru1:
Added a new stress test for incremental uploads. The test has 47 workers that write 100 MiB objects concurrently using incremental upload mode while sharing a 512 MiB memory limit. This mirrors the existing `sustained_writes `scenario but with the incremental upload path enabled. 

Sample run: https://github.com/awslabs/mountpoint-s3/actions/runs/30802403323/job/91649829872

No. This only adds a new test

No. Test-only change

---

### DIFF
--- a/.github/workflows/stress.yml
+++ b/.github/workflows/stress.yml
@@ -59,6 +59,11 @@ jobs:
           - single_reader_held_budget_direct_io
           - many_readers_held_budget_direct_io
           - single_reader_held_budget_misaligned_part
+          - sustained_writes_incremental_upload
+        include:
+          - scenario: sustained_writes_incremental_upload
+            s3_bucket_name: ${{ vars.S3_EXPRESS_ONE_ZONE_BENCH_BUCKET_NAME }}
+            s3_endpoint_url: ${{ vars.S3_EXPRESS_ONE_ZONE_BENCH_ENDPOINT_URL }}
     steps:
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@254c19bd240aabef8777f48595e9d2d7b972184b # v6.2.1
@@ -84,6 +89,9 @@ jobs:
       - name: Install cargo-nextest
         uses: taiki-e/install-action@nextest
       - name: Run stress scenario
+        env:
+          S3_BUCKET_NAME: ${{ matrix.s3_bucket_name || vars.S3_BENCH_BUCKET_NAME }}
+          S3_ENDPOINT_URL: ${{ matrix.s3_endpoint_url || vars.S3_BENCH_ENDPOINT_URL }}
         run: |
           cargo nextest run \
             --release \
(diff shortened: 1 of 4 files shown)


mountpoint-s3-fs/tests/stress/workers/incremental_writer.rs
```diff
@@ -0,0 +1,80 @@
...
+/// incremental upload mode. `scope` is the scenario label used as the ephemeral-key scenario
+/// segment (e.g. `"incremental_upload"`) — it ends up in the S3 key so concurrent scenarios
+/// do not collide.
+pub struct IncrementalWriter {
+    pub scope: &'static str,
+    pub object_size: usize,
+    pub chunk: usize,
+}
+
+impl Worker for IncrementalWriter {
```
yerzhan7: This is exactly the same as our existing Writer? we should re-use to avoid duplication.

.github/workflows/stress.yml
yerzhan7, line 27: This needs to be changed to S3 express bucket? Check how benchmark CI is configured for S3 Express runs as a reference

yerzhan7, line 29: Presumably this as well?

mountpoint-s3-fs/tests/stress/scenarios/incremental_upload.rs
```diff
@@ -0,0 +1,40 @@
+//! `incremental_upload`: 32 workers concurrently writing 64 MiB objects using incremental upload
```
yerzhan7: Let's just duplicate existing `sustained_writes` scenario (same parameters/settings - just do incremental upload flag/config). And name it like `sustained_writes_incremental_upload` (for consistency with benchmark naming)

mountpoint-s3-fs/tests/stress/workers/incremental_writer.rs
Priyankakarumuru1: Good point! I created it separately for clearer observability (incremental_writer vs writer in logs/metrics), but you're right - will reuse Writer to avoid the duplication.

mountpoint-s3-fs/tests/stress/scenarios/incremental_upload.rs
Priyankakarumuru1: Done! Matched `sustained_writes` exactly - same parameters, just added` incremental_upload: true` config.

mountpoint-s3-fs/tests/stress/scenarios/sustained_writes_incremental_upload.rs
```diff
@@ -0,0 +1,40 @@
...
+use std::sync::Arc;
+
+use mountpoint_s3_fs::S3FilesystemConfig;
+use mountpoint_s3_fs::memory::MINIMUM_MEM_LIMIT;
+
+use crate::common::fuse::TestSessionConfig;
+use crate::stress::harness::{self, Scenario, Worker, default_max_latency};
+use crate::stress::workers::Writer;
+
+const NUM_WORKERS: usize = 48;
```
yerzhan7: Sorry, can we fix this to be 47? to match https://github.com/awslabs/mountpoint-s3/blob/feature/memory-limit/mountpoint-s3-fs/tests/stress/scenarios/held_writes_vs_reads.rs#L11

I think we forgot to update `sustained_writes` previously (when we updated formula).

yerzhan7 (commented):
Could you please push your commit to `wf-changes/XXX` branch (in `awslabs`, not fork) and paste a link here of a successful stress test run of the new scenario (as manual validation)?

(And we need to resolve the merge conflict)

yerzhan7 (approved):

PR #1912, opened 2026-08-06
Add allocation queue depth and wait-time metrics

yerzhan7:
Add allocation queue depth and block-time metrics: (will be added to defs.rs and METRICS.md in a separate PR)
- `pool.allocation_queue_depth`
- `pool.allocation_queue_wait`

No

N/A

---

### DIFF
--- a/mountpoint-s3-fs/src/memory/allocation_queue.rs
+++ b/mountpoint-s3-fs/src/memory/allocation_queue.rs
@@ -252,12 +252,10 @@ impl AllocationQueue {
             self.has_pending.store(false, Ordering::SeqCst);
         }
 
-        let waited = entry.as_ref().map(|e| e.queued_at.elapsed());
         drop(inner);
-        if let Some(waited) = waited {
-            metrics::histogram!(ALLOCATION_QUEUE_WAIT).record(waited.as_micros() as f64);
-        }
-        entry
+        entry.inspect(|e| {
+            metrics::histogram!(ALLOCATION_QUEUE_WAIT).record(e.queued_at.elapsed().as_micros() as f64);
+        })
     }
 
     /// Moves all entries for `cursor_id` from the low-priority queue to the back
--- a/mountpoint-s3-fs/src/memory/allocation_queue.rs
+++ b/mountpoint-s3-fs/src/memory/allocation_queue.rs
@@ -1,17 +1,25 @@
 //! Priority-ordered allocation queue for buffer requests under memory pressure.
 
 use std::collections::VecDeque;
+use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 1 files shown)


mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -73,6 +81,33 @@ struct AllocationQueueInner {
     low: VecDeque<PendingAllocation>,
 }
 
+/// Guard around the locked [`AllocationQueueInner`] that refreshes the
+/// [`ALLOCATION_QUEUE_DEPTH`] gauge from the queue lengths when it is dropped.
+struct DepthTrackingGuard<'a> {
```
passaro, line 86: Neat!

mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -135,7 +178,7 @@ impl AllocationQueue {
             sender,
         };
 
-        let mut inner = self.inner.lock().unwrap();
+        let mut inner = self.lock_tracked();
```
passaro, line 181: not blocking: if we really want to close the loop and prevent any accidental usage of `self.inner.lock()` on its own, we could also introduce a `DepthTrackingMutex` to wrap `self.inner` itself.

mountpoint-s3-fs/src/memory/allocation_queue.rs
```diff
@@ -208,6 +251,12 @@ impl AllocationQueue {
         if inner.high.is_empty() && inner.low.is_empty() {
             self.has_pending.store(false, Ordering::SeqCst);
         }
+
+        let waited = entry.as_ref().map(|e| e.queued_at.elapsed());
+        drop(inner);
+        if let Some(waited) = waited {
+            metrics::histogram!(ALLOCATION_QUEUE_WAIT).record(waited.as_micros() as f64);
+        }
```
passaro: np1: we can drop inner before calculating waited.

np2: map + if let are both picking inside an Option. Maybe we could just use `Option::inspect` here to update the metric.

passaro (approved):

mountpoint-s3-fs/src/memory/allocation_queue.rs
yerzhan7: I wanted to leave `self.inner.lock()` for the read methods like `head_queued_at()` that do not mutate queue length.

mountpoint-s3-fs/src/memory/allocation_queue.rs
yerzhan7: Fixed, thanks for reviewing!

PR #1913, opened 2026-08-06
Avoid pool-buffer aliasing deadlock and stitch multi-part reads in one buffer

renanmagagnin:
`ChecksummedBytes::extend` aliased the appended buffer when the accumulator was empty (`*self = extend`). At `CachingPartComposer` and `Cursor::do_read` that buffer is a pooled GetObject buffer, so the accumulator pinned it across the next part's allocation and deadlocked under memory pressure.

Both now stitch their output with a new `ChecksummedBytesBuilder`, which reserves the expected total length on first append and copies each piece exactly once, rather than reallocating and re-copying the accumulated prefix per piece as `extend` did. Nothing aliases a pooled buffer any more, so `ChecksummedBytes::extend` and the already-unused `Part::extend` are removed.

Stress tests: https://github.com/awslabs/mountpoint-s3/actions/runs/31591359789
Benchmarks: https://github.com/awslabs/mountpoint-s3/actions/runs/31591360134

No.

No.

---

### DIFF
--- a/mountpoint-s3-fs/src/checksums.rs
+++ b/mountpoint-s3-fs/src/checksums.rs
@@ -196,12 +196,16 @@ impl ChecksummedBytes {
 /// Accumulates several [ChecksummedBytes] into one buffer, copying each piece exactly once and
 /// combining their checksums rather than recomputing one over the result.
 ///
-/// [Self::new] reserves the expected total length up front, so appending that much data allocates
-/// only once. Appending more still succeeds, growing the buffer as [Vec] normally would.
+/// The buffer is only allocated on the first append, reserving the `capacity` given to [Self::new],
+/// so appending up to that much data allocates exactly once and a builder that is never appended to
+/// does not allocate at all. Appending more still succeeds, growing the buffer as [Vec] normally
+/// would.
 #[must_use]
 pub struct ChecksummedBytesBuilder {
-    /// Destination buffer, filled from the front.
+    /// Destination buffer, filled from the front. Empty until the first append.
     buffer: Vec<u8>,
+    /// Bytes to reserve for [Self::buffer] when it is first allocated.
+    capacity: usize,
     /// Combined checksum of [Self::buffer].
     checksum: Crc32c,
 }
@@ -210,7 +214,8 @@ impl ChecksummedBytesBuilder {
     /// Create a builder that can hold `capacity` bytes without reallocating.
     pub fn new(capacity: usize) -> Self {
         Self {
-            buffer: Vec::with_capacity(capacity),
+            buffer: Vec::new(),
+            capacity,
             checksum: Crc32c::new(0),
         }
     }
@@ -238,14 +243,19 @@ impl ChecksummedBytesBuilder {
         // Shrink to fit so `checksum` covers exactly `bytes`, as `combine_checksums` requires.
         let (bytes, checksum) = bytes.into_inner()?;
 
+        if self.buffer.is_empty() {
+            self.buffer.reserve_exact(self.capacity);
+        }
         self.buffer.extend_from_slice(&bytes);
         self.checksum = combine_checksums(self.checksum, checksum, bytes.len());
         Ok(())
     }
 
-    /// Consume the builder and return the accumulated data.
-    pub fn finish(self) -> ChecksummedBytes {
-        ChecksummedBytes::new_from_inner_data(Bytes::from_owner(self.buffer), self.checksum)
+    /// Take the accumulated data, leaving the builder empty and ready to accumulate again.
+    pub fn finish(&mut self) -> ChecksummedBytes {
+        let buffer = std::mem::take(&mut self.buffer);
+        let checksum = std::mem::replace(&mut self.checksum, Crc32c::new(0));
+        ChecksummedBytes::new_from_inner_data(Bytes::from(buffer), checksum)
     }
 }
 
--- a/mountpoint-s3-fs/src/checksums.rs
+++ b/mountpoint-s3-fs/src/checksums.rs
@@ -193,53 +193,41 @@ impl ChecksummedBytes {
     }
 }
 
-/// Accumulates several [ChecksummedBytes] into one caller-supplied buffer, copying each piece
-/// exactly once and combining their checksums rather than recomputing one over the result.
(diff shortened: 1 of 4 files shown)


mountpoint-s3-fs/src/checksums.rs
```diff
@@ -248,6 +249,79 @@ impl ChecksummedBytes {
...
+            // No op, but check that `bytes` was not corrupted.
+            bytes.validate()?;
+            return Ok(());
+        }
+
+        // Shrink to fit so `checksum` covers exactly `bytes`, as `combine_checksums` requires.
+        let (bytes, checksum) = bytes.into_inner()?;
+
+        let new_len = self.len + bytes.len();
+        assert!(
```
passaro: We should avoid the panic here and on finish. Options:
1. Return a new error here, and let the caller "assert". The (relative) improvement is in locality: the caller can prove that they set up enough capacity.
2. Remove the generic parameter and use a `Vec` directly. `new` takes the capacity so when used "correctly" there is still a single allocation, but it degrades to reallocations otherwise. 
 
I'm leaning towards (2), but happy to discuss.

mountpoint-s3-fs/src/checksums.rs
```diff
@@ -172,9 +172,10 @@ impl ChecksummedBytes {
         }
 
         if self.is_empty() {
-            // Replace with `extend`, but check that `self` was not corrupted
+            // Copy `extend` rather than alias it, so the result never keeps its backing buffer
+            // alive: a pooled buffer pinned here would deadlock allocation under memory pressure.
```
passaro: Not great to change a generic method behavior to account for an issue with a specific type of backing buffer.

passaro (changes requested):
Why is `CachingPartComposer` still relying on the `extend` method? Shouldn't it also use the new `ChecksummedBytesBuilder` (and thus a single allocation)?

As a plus, both `extend` and `into_detached` may be removed.

yerzhan7:
It looks like now in `caching_stream.rs` we would always do copy (even when it's not needed). Is it possible to improve that? (Similarly to how we do that in `Cursor::do_read()`)

mountpoint-s3-fs/src/checksums.rs
renanmagagnin: Went with option 2 for simplicity, assuming the goal of the builder is to minimise allocations rather than guarantee exactly one. We'd have to revisit if we allocate via the pool later since PoolBuffer is fixed-size, but that seems fine.

mountpoint-s3-fs/src/prefetch/caching_stream.rs
```diff
@@ -291,16 +291,19 @@ where
     ) -> Result<(), PrefetchReadError<E>> {
         let key = self.cache_key.key();
         let block_size = self.cache.block_size();
-        let mut buffer = ChecksummedBytes::default();
+        let object_size = self.original_range.object_size();
+
+        let mut block_builder: Option<ChecksummedBytesBuilder> = None;
```
passaro: tiny nit: let's just call this `buffer`

mountpoint-s3-fs/src/prefetch/caching_stream.rs
```diff
@@ -312,8 +315,11 @@ where
             // Split the body into blocks.
             let mut offset = offset;
             while !body.is_empty() {
-                let remaining = (block_size as usize).saturating_sub(buffer.len()).min(body.len());
-                let chunk: ChecksummedBytes = body.split_to(remaining).into();
+                // A full block, except for the object's last one which may be shorter.
+                let block_len = (object_size.saturating_sub(self.block_offset as usize)).min(block_size as usize);
```
passaro: We only care about the shorter block size (if ever) if we are allocating the buffer. Could we move this calculation there and reduce the noise here?

mountpoint-s3-fs/src/checksums.rs
```diff
@@ -248,6 +193,62 @@ impl ChecksummedBytes {
...
+    buffer: Vec<u8>,
+    /// Combined checksum of [Self::buffer].
+    checksum: Crc32c,
+}
+
+impl ChecksummedBytesBuilder {
+    /// Create a builder that can hold `capacity` bytes without reallocating.
+    pub fn new(capacity: usize) -> Self {
+        Self {
+            buffer: Vec::with_capacity(capacity),
```
passaro: Proposal: what if we stored the capacity separately here, started with an empty vector, and only replaced it with `Vec::with_capacity(self.capacity)` lazily on `append`?

That would dramatically simplify usage by removing the need for `Option`, but still avoiding any extra allocations.

passaro (commented):
Looks good overall. A couple of possible refinements:

mountpoint-s3-fs/src/checksums.rs
```diff
@@ -248,6 +193,72 @@ impl ChecksummedBytes {
...
+        if self.buffer.is_empty() {
+            self.buffer.reserve_exact(self.capacity);
+        }
+        self.buffer.extend_from_slice(&bytes);
+        self.checksum = combine_checksums(self.checksum, checksum, bytes.len());
+        Ok(())
+    }
+
+    /// Take the accumulated data, leaving the builder empty and ready to accumulate again.
+    pub fn finish(&mut self) -> ChecksummedBytes {
```
passaro, line 255: nit: `build`?

mountpoint-s3-fs/src/prefetch/caching_stream.rs
```diff
@@ -332,31 +336,45 @@ where
...
-                    .extend(chunk)
-                    .inspect_err(|e| warn!(key, error=?e, "integrity check for body part failed"))?;
-                if buffer.len() < block_size as usize {
-                    break;
-                }
 
-                // We have a full block: write it to the cache, send it to the queue, and flush the buffer.
-                self.update_cache(buffer, self.block_index, self.block_offset, &self.cache_key, range);
+                // A full block, except for the object's last one which may be shorter.
+                let block_len = (object_size.saturating_sub(self.block_offset as usize)).min(block_size as usize);
```
passaro, line 341: nit: start with the block size, `(block_size as usize).min(...)`

passaro (approved):
LGTM

PR #1914, opened 2026-08-07
Upgrade actions/checkout from 6.0.3 to 7.0.1

passaro:
Replaces #1901: by default, `actions/checkout` v7 disallows checking out fork PR code from `pull_request_target` workflows, which is a feature we rely on in our integration, bench, and stress workflows. This change opts in by setting `allow-unsafe-pr-checkout`, which is safe because every affected job sits behind `needs: approval`, a protected environment requiring maintainer review.

No.

No.

---

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -71,7 +71,7 @@ jobs:
         aws-region: ${{ vars.S3_BENCH_REGION }}
         role-duration-seconds: 21600
     - name: Checkout code
-      uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
         ref: ${{ inputs.ref }}
         submodules: true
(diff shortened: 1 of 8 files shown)


.github/workflows/bench.yml
```diff
@@ -71,11 +71,13 @@ jobs:
         aws-region: ${{ vars.S3_BENCH_REGION }}
         role-duration-seconds: 21600
     - name: Checkout code
-      uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
         ref: ${{ inputs.ref }}
         submodules: true
         persist-credentials: false
+        # Fork PR checkout gated by `needs: approval`, which requires maintainer review
```
jet-tong: Do we want to also justify by saying this code only has access to ephemeral test cluster?

Referencing https://github.com/awslabs/mountpoint-s3-csi-driver/pull/839 in CSI repo:
```
# Allow fork PR checkouts since we gate deployments and the code only has access to ephemeral test cluster
allow-unsafe-pr-checkout: true
```

passaro: Aligned.

jet-tong (approved):

PR #1915, opened 2026-08-10
Withhold the prunable reserve from paged allocations

yerzhan7:
The memory limiter keeps `prunable_reserved` bytes of headroom below `mem_limit` so a read can still allocate after writes have consumed the rest of the budget. `Prunable` kinds (`GetObject`, `DiskCache`) were allowed to spend that reserve. However, the ceiling was chosen purely from `BufferKind`, without regard to whether the allocation was a page in a `SizePool` or an exact-size one-off buffer.
  
The issue: a prunable page allocation could dip into the reserve. A page holds up to `MAX_BUFFERS_PER_PAGE (16)` buffers, is freed only when wholly empty, and any `BufferKind` may claim a free buffer inside it. So a page carved out of the reserve destroys it. I.e. the bytes stay committed to that one size pool, where a non-prunable holder (e.g. a write) can pin them indefinitely. The reserve stops functioning as read headroom.

N/A

N/A

---

### DIFF
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -286,20 +286,36 @@ impl MemoryLimiter {
         metrics::gauge!("mem.bytes_reserved", "area" => BufferArea::Prefetch.as_str()).decrement(decremented as f64);
     }
 
+    /// The memory ceiling [`Self::try_allocate`] enforces.
+    ///
+    /// Every allocation stops `prunable_reserved` bytes short of `mem_limit` — that slice is the
+    /// headroom that lets a read still allocate when writes have taken the rest of the budget — with
+    /// one exception: a prunable ([`BufferKind::is_prunable`]) one-off allocation (`is_paged` false)
+    /// may spend it. Such an allocation is the read the reserve exists for, and it takes the reserve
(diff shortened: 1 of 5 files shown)


mountpoint-s3-fs/src/memory/limiter.rs
```diff
@@ -286,33 +286,44 @@ impl MemoryLimiter {
...
-    /// individual buffers are tracked separately as they are acquired.
-    pub fn try_allocate(
+    /// `is_paged` distinguishes a page in a [`SizePool`](super::pool::SizePool) (`true`) from an
+    /// exact-size one-off buffer (`false`).
+    pub(super) fn try_allocate(
         self: &Arc<Self>,
         size: usize,
         kind: BufferKind,
-        track: bool,
+        is_paged: bool,
```
passaro: Now that we never use `kind` when `is_paged == true`, would it make sense to restore the original `Option<BufferKind>`?

yerzhan7: Good point! Fixed in https://github.com/awslabs/mountpoint-s3/pull/1915/commits/7ac604f6607acd91fce45bb61d44fd9dd8a41369

passaro (approved):

PR #1917, opened 2026-08-11
Use jemalloc as the global allocator

yerzhan7:
Use jemalloc with default settings as the global allocator in Rust.

Note: CRT still uses glibc malloc.

Benchmarks: https://github.com/awslabs/mountpoint-s3/actions/runs/32017340407
Stress tests: https://github.com/awslabs/mountpoint-s3/actions/runs/32017340069

N/A

N/A

---


package/package.py
```diff
@@ -140,6 +140,11 @@ def build_mountpoint_binary(metadata: BuildMetadata, args: argparse.Namespace) -
     for var in ["CC", "CXX", "LD_LIBRARY_PATH"]:
         if var in os.environ:
             env[var] = os.environ[var]
+    if metadata.arch == "aarch64":
+        # jemalloc fixes its page size at compile time and refuses to start when the system page is
+        # larger, so a 4KiB build aborts on 64KiB-page aarch64 kernels (RHEL 8, RHEL 9 `kernel-64k`).
+        # Build for 64KiB, arm64's largest page size: it is safe on smaller-page hosts.
+        env["JEMALLOC_SYS_WITH_LG_PAGE"] = "16"
```
yerzhan7: Could we add some kind of test for that? During release I don't think we test on 64KiB OS, do we?

package/package.py
```diff
@@ -140,6 +140,11 @@ def build_mountpoint_binary(metadata: BuildMetadata, args: argparse.Namespace) -
     for var in ["CC", "CXX", "LD_LIBRARY_PATH"]:
         if var in os.environ:
             env[var] = os.environ[var]
+    if metadata.arch == "aarch64":
+        # jemalloc fixes its page size at compile time and refuses to start when the system page is
+        # larger, so a 4KiB build aborts on 64KiB-page aarch64 kernels (RHEL 8, RHEL 9 `kernel-64k`).
+        # Build for 64KiB, arm64's largest page size: it is safe on smaller-page hosts.
```
yerzhan7: This means CI builds and release builds for aarch64 would be different. Can we also change CI builds for aarch64?

mountpoint-s3/src/main.rs
```diff
@@ -1,4 +1,20 @@
...
+///
+/// `background_thread:true` moves page purging off the freeing thread, where the `madvise` call
+/// would otherwise block the read path. `narenas:32` caps arenas well below the default (~4x CPUs)
+/// so freed part buffers concentrate in a shared dirty pool and can be reused across threads
+/// instead of each arena faulting in new pages.
+// SAFETY: overrides jemalloc's weak `_rjem_malloc_conf` symbol, which it reads as a NUL-terminated
+// options string during initialisation. The value below is NUL-terminated.
+#[allow(non_upper_case_globals)]
+#[unsafe(export_name = "_rjem_malloc_conf")]
+pub static malloc_conf: &[u8] = b"background_thread:true,narenas:32\0";
```
yerzhan7: Could we please add `abort_conf:true` as per recommendation in https://github.com/jemalloc/jemalloc/blob/dev/TUNING.md

mountpoint-s3-fs/tests/stress_tests.rs
```diff
@@ -3,5 +3,18 @@
...
+#[global_allocator]
+static GLOBAL: Jemalloc = Jemalloc;
+
+/// Matches the compiled-in jemalloc config of the `mount-s3` binary so the stress harness measures
+/// the allocator settings that ship. Overridable at runtime via `_RJEM_MALLOC_CONF`.
+// SAFETY: overrides jemalloc's weak `_rjem_malloc_conf` symbol, which it reads as a NUL-terminated
+// options string during initialisation. The value below is NUL-terminated.
+#[allow(non_upper_case_globals)]
+#[unsafe(export_name = "_rjem_malloc_conf")]
+pub static malloc_conf: &[u8] = b"background_thread:true,narenas:32\0";
```
yerzhan7: We might need to duplicate it everywhere we do `static GLOBAL: Jemalloc = Jemalloc;`

package/package.py
yerzhan7: Wait. Is this actually needed? It looks like it's already set by jemalloc by default?

https://github.com/jemalloc/jemalloc/blob/dev/ChangeLog#L108-L109

yerzhan7:
LGTM, but can't approve because i am co-author

passaro (approved):

renanmagagnin:
Investigation of `rand_read_4t` increase in Peak Memory Usage (S3 Standard, Memory-Limited) will be done as a follow-up.

PR #1921, opened 2026-08-12
Remove AL2 from various pieces of documentation

muddyfish:
Replace AL2 with AL2023 in various pieces of documentation. Removing because AL2 is now end of life

Remove instructions for building the packaging script without docker, as AL2023 already comes with a native release, and `dpkg` cannot be installed on it. Users who want to build custom releases can use docker.

Only documentation is changed. 

No

---

### DIFF
--- a/package/README.md
+++ b/package/README.md
@@ -21,33 +21,6 @@ You should pass the `--expected-version` to the build script to verify that the 
 
 The container will create an `out` folder in the root of the Git repository containing the build artifacts.
 
-## Building locally
-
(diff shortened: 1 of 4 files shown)


package/README.md
renanmagagnin, line 1: From the PR description:
> Users who want to build custom releases can use docker.

Should we add instructions related to this?

muddyfish: The instructions are above in the file

renanmagagnin (approved):

PR #1922, opened 2026-08-12
mem: Snapshot live cursors before iterating in the pruner

yerzhan7:
**Problem:** Maintenance thread could deadlock because it was dropping Cursors (which removes them from DashMap) while holding DashMap read shard lock.

**Solution:** Create helper method to collect live cursors into vector first.

N/A

N/A

---

### DIFF
--- a/mountpoint-s3-fs/src/memory/limiter.rs
+++ b/mountpoint-s3-fs/src/memory/limiter.rs
@@ -413,14 +413,26 @@ impl MemoryLimiter {
         &self.pruning_signal
     }
 
+    /// Collect strong handles to every live cursor, draining the `DashMap`
+    /// iterator into a `Vec` before returning.
+    ///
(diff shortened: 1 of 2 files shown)


passaro (commented):
The fix looks sound, but consider refactoring so that `limiter` only exposes "safe" access to `cursors`:
* expose `live_cursors` so it can be called by `run_pruning_round`,
* maintain the result for each round,
* refactor the 3 callers so they operate on the vector and move the into the `maintenance` module, since they are all called by `run_pruning_round`.

yerzhan7:
Refactored in https://github.com/awslabs/mountpoint-s3/pull/1922/commits/30487e51f19521e68b06dc59c160eb71bbe4fa9c

passaro (approved):
LGTM!

PR #1923, opened 2026-08-14
fix(stress): show metrics and invariants before stall panic

Priyankakarumuru1:
When the `many_readers_budget_part` stress test hit a stall, it panicked before printing any metrics, so we couldn't see what was happening. Also, the 20s stall timeout was too tight for this scenario where reads legitimately take 30-35s.
  
  Changes:
  - Print metrics and invariant results before the stall panic so they're always visible in CI.
  - Bump stall timeout and max latency ceiling to 60s for `many_readers_budget_part` only.

 Before (no metrics on failure): https://github.com/awslabs/mountpoint-s3/actions/runs/31726966422/job/94537685662
 After (metrics visible): https://github.com/awslabs/mountpoint-s3/actions/runs/31798312530/job/94760245418

N/A

N/A

---


mountpoint-s3-fs/tests/stress/scenarios/many_readers_budget_part.rs
```diff
@@ -18,12 +21,48 @@ const READ_PART_SIZE: usize = MINIMUM_MEM_LIMIT - 128 * 1024 * 1024; // 384 MiB
 const READ_CHUNK: usize = 8 * 1024 * 1024; // 8 MiB
 const NUM_WORKERS: usize = 4;
 
+/// Wrapper around SequentialReader that overrides max_idle to 60 seconds for this scenario.
+struct ReaderWith60sTimeout {
```
yerzhan7: To avoid this duplication, let's introduce `pub max_idle: fn(&dyn Worker) -> Duration,` trait into `Scenario` instead, and remove Worker's `fn max_idle(&self) -> Duration` method? (to match existing pattern with `max_latency`)

passaro: Why do we need this wrapper? Can't we just add `max_idle` to `SequentialReader` (with same default as now)?

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -145,6 +145,83 @@ fn collect_gauges_by_label(
     out
 }
 
+/// Log memory gauge peaks for diagnostic purposes without asserting.
+/// Useful when the test is already failing (e.g., due to a stall) but we still
+/// want to see the memory metrics to aid debugging.
+pub fn log_peak_memory_metrics(scenario_name: &str, mem_limit: f64, part_size: usize) {
```
yerzhan7: Is it possible to avoid duplication?

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -145,6 +145,83 @@ fn collect_gauges_by_label(
...
+            metric = %gauge.id(),
+            peak = %format_mib(peak),
+            budget = %format_mib(effective_budget),
+            ceiling = %format_mib(ceiling),
+            "stress: peak memory gauge"
+        );
+    }
+
+    // Log RSS
+    if let Some(metric) = recorder.get("process.memory_usage", &[]) {
```
yerzhan7: In https://github.com/awslabs/mountpoint-s3/blob/feature/memory-limit/mountpoint-s3-fs/tests/stress/harness/invariants.rs#L284 we also subtract `worker_io_buffer_bytes`.

mountpoint-s3-fs/tests/stress/harness/invariants.rs
```diff
@@ -145,6 +145,83 @@ fn collect_gauges_by_label(
     out
 }
 
+/// Log memory gauge peaks for diagnostic purposes without asserting.
```
passaro: how do all these new methods compare with `dump_summary`? isn't there any opportunity for reuse?

passaro (changes requested):

mountpoint-s3-fs/tests/stress/scenarios/many_readers_budget_part.rs
Priyankakarumuru1: I used the wrapper to avoid touching all scenario files, but agree that introducing `pub max_idle: fn(&dyn Worker) -> Duration` on `Scenario` is cleaner and matches the existing `max_latency` pattern. Will change.

mountpoint-s3-fs/tests/stress/harness/invariants.rs
Priyankakarumuru1: Yes, Removed this entirely. Instead, factored out `dump_metrics_snapshot` from the existing `dump_summary` and reused it.

mountpoint-s3-fs/tests/stress/harness/invariants.rs
Priyankakarumuru1: Removed it. Thanks.

yerzhan7 (approved):
LGTM, @passaro  suggested alternative approach though.

I will merge as-is for now, in the interest of time.

PR #1930, opened 2026-08-17
Promote a queued buffer request whose read turned active mid-push

yerzhan7:
**Problem:** When pushing to `low` priority allocation queue (prefetch, no longer active reader, in background thread) there is a race condition when FUSE thread on the same cursor may become active reader again (due to new FUSE `read()` request) first. 
That FUSE thread may do `upgrade(cursorId)` first before background thread pushes to low queue. This leaves that buffer allocation request incorrectly permanently in low priority queue.

**Solution:**  Make the lane decision under the allocation queue lock, so the window never opens.

**Testing**: `cache_hit_vs_miss_held_budget` stress scenario improvements:
- Before: https://github.com/awslabs/mountpoint-s3/actions/runs/32013026621/job/95336521015
- After: https://github.com/awslabs/mountpoint-s3/actions/runs/32044052918/job/95462218011

```
  ┌─────────────────────────────────────┬─────────────┬────────────┬────────┬────────┐
  │               Metric                │   Before    │   After    │   Δ    │ Better │
  ├─────────────────────────────────────┼─────────────┼────────────┼────────┼────────┤
  │ op=read p100 (harness read syscall) │ 19,431.4 ms │ 1,599.5 ms │ −91.8% │ After  │
  ├─────────────────────────────────────┼─────────────┼────────────┼────────┼────────┤
  │ fuse.request_latency{read} p100     │ 18.874 s    │ 0.467 s    │ −97.5% │ After  │
  ├─────────────────────────────────────┼─────────────┼────────────┼────────┼────────┤
  │ prefetch.part_queue_starved_us p100 │ 18.874 s    │ 0.460 s    │ −97.6% │ After  │
  ├─────────────────────────────────────┼─────────────┼────────────┼────────┼────────┤
  │ cache.get_latency{disk} p100        │ 0.597 s     │ 0.413 s    │ −30.8% │ After  │
  └─────────────────────────────────────┴─────────────┴────────────┴────────┴────────┘
```

N/A

N/A

---

### DIFF
--- a/mountpoint-s3-fs/src/memory/pool.rs
+++ b/mountpoint-s3-fs/src/memory/pool.rs
@@ -312,7 +312,16 @@ impl PagedPoolInner {
             Some(id) if self.limiter.has_active_read(id) => {
                 self.allocation_queue.push_high(cursor_id, size, kind, queued)
             }
-            Some(id) => self.allocation_queue.push_low(id, size, kind, queued),
+            Some(id) => {
+                let rx = self.allocation_queue.push_low(id, size, kind, queued);
+                // Re-check now that the entry is queued: `set_active_read` flips the cursor to
+                // `Active` before scanning the low queue, so an interleaving between the guard
+                // above and this push would strand the entry in `low` with nobody to promote it.
+                if self.limiter.has_active_read(id) {
+                    self.allocation_queue.upgrade(id);
+                }
+                rx
+            }
             None => self.allocation_queue.push_high(None, size, kind, queued), // uploads are urgent
         };
         // After pushing, try to wake immediately in case memory freed between the fast-path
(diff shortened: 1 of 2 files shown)


mountpoint-s3-fs/src/memory/pool.rs
```diff
@@ -312,7 +312,16 @@ impl PagedPoolInner {
             Some(id) if self.limiter.has_active_read(id) => {
```
passaro: Shouldn't we push this check inside the allocation queue lock to avoid the race in the first place?

yerzhan7: Done in https://github.com/awslabs/mountpoint-s3/pull/1930/commits/c368653d121b9822a1a87c3ee2c260bd88052d03

passaro (approved):

mountpoint-s3-fs/src/memory/pool.rs
passaro: Thanks. Please update the description.

PR #1933, opened 2026-08-18
Fix cgroup memory limit detection for inherited limits

Priyankakarumuru1:
mount-s3 doesn't detect `cgroup` memory limits set on parent slices. A customer has a 20GB limit on the parent, but mount-s3 doesn't see it, so it falls back to using the host's total RAM (200GB). It oversizes buffers which hits the limit and stalls completely. We use `System::cgroup_limits()` which only checks `/sys/fs/cgroup` (root). In cgroup v2, when a parent has a memory limit, children inherit it but if you only read the child's `memory.max` file, it shows "max" (meaning no limit set at this level). The actual limit is on the parent. We need to walk up the tree to find it. 

  Switched to `Process::cgroup_limits()`:
  - Reads `/proc/self/cgroup` to find the process's actual cgroup
  - Walks up ancestors checking each `memory.max`
  - Returns the smallest limit found

 ### Testing

Added a CI test that creates a cgroup hierarchy with `memory.max` set on a parent slice (1 GiB) and verifies the process detects the inherited limit rather than the container's own limit (2 GiB).
Without fix (fails):  https://github.com/awslabs/mountpoint-s3/actions/runs/32240451777/job/96029623759

With fix (passes): https://github.com/awslabs/mountpoint-s3/actions/runs/32267673710/job/96116060108
  
No breaking changes.

Yes - fixes customer issue.

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -431,8 +431,6 @@ jobs:
         cat /sys/fs/cgroup/memory/memory.limit_in_bytes 2>/dev/null || echo "not found"
         echo "=== Process cgroup location ==="
         cat /proc/self/cgroup
-        echo "=== Check if we can write to cgroups ==="
-        mkdir -p /sys/fs/cgroup/test-write 2>&1 || echo "Cannot create cgroups"
     - name: Run cgroup-aware memory detection test
       env:
         TEST_CGROUP_MEM_LIMIT_MB: "2048"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -393,6 +393,7 @@ jobs:
       options: >-
         --memory=2g
         --memory-swap=2g
+        --privileged
 
     steps:
     - name: Install prerequisites for container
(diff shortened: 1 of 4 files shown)


mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -160,14 +160,29 @@ impl MemoryLimiter {
...
+
+    // Try to get cgroup limits for this specific process. Process::cgroup_limits()
+    // walks the ancestor cgroup tree to find inherited limits.
+    let pid = Pid::from_u32(std::process::id());
+    sys.refresh_processes_specifics(
+        sysinfo::ProcessesToUpdate::Some(&[pid]),
+        false, // don't remove dead processes
+        sysinfo::ProcessRefreshKind::nothing(),
+    );
+    if let Some(cg) = sys.process(pid).and_then(|p| p.cgroup_limits()) {
```
yerzhan7, line 180: To be safe, let's keep the `sys.cgroup_limits()` (as a fallback when `Process::cgroup_limits()` returns `None`) because I suspect it might be a regression in some edge cases for cgroup v1.

.github/workflows/tests.yml
```diff
@@ -428,10 +429,49 @@ jobs:
         cat /sys/fs/cgroup/memory.max 2>/dev/null || echo "not found"
         echo "=== cgroup v1 ==="
         cat /sys/fs/cgroup/memory/memory.limit_in_bytes 2>/dev/null || echo "not found"
+        echo "=== Process cgroup location ==="
+        cat /proc/self/cgroup
+        echo "=== Check if we can write to cgroups ==="
+        mkdir -p /sys/fs/cgroup/test-write 2>&1 || echo "Cannot create cgroups"
```
yerzhan7: np: Let's remove this (as this is just diagnostic thing and we are not doing cleanup of this file)

Also, link https://github.com/awslabs/mountpoints3/actions/runs/32241385508/job/96032498941 from PR description is broken.

Priyankakarumuru1: Updated the link. Thanks!

mountpoint-s3-fs/src/mem_limiter.rs
```diff
@@ -160,14 +160,33 @@ impl MemoryLimiter {
...
     sys.refresh_memory();
-    sys.cgroup_limits()
-        .map(|cg| cg.total_memory)
-        .unwrap_or_else(|| sys.total_memory())
+
+    // Try to get cgroup limits for this specific process. Process::cgroup_limits()
+    // walks the ancestor cgroup tree to find inherited limits.
+    let pid = Pid::from_u32(std::process::id());
+    sys.refresh_processes_specifics(
+        sysinfo::ProcessesToUpdate::Some(&[pid]),
```
yerzhan7: np: Let's use `use` instead of fully qualified inline import.

yerzhan7 (dismissed):

yerzhan7 (dismissed):

yerzhan7 (approved):

PR #1935, opened 2026-08-18
Fail benchmarks when a memory limit is breached

renanmagagnin:
Memory-limited benchmark jobs already report peak RSS breaches in a summary table but still pass. This makes a breach fail the job, after the results have been saved and published.

Testing:
- `render-mem-summary.sh` over 5 input shapes locally (no results, all within limit, partial breach, full breach) — correct output and exit code in each.
- Gate wiring on a real Actions run, 4 cases: a breach fails the job; `false` and a skipped render step both pass; a breach still fails when an earlier step had already failed.
- Not covered: a real breaching benchmark run, as `--memory-target` has a 512 MiB floor.

Yes, intentionally: memory-limited benchmark jobs now fail on a breach. Results are still saved and published first, and other variants are unaffected.

No, this only changes CI benchmark workflows.

---

### DIFF
--- a/.github/actions/scripts/render-mem-summary.sh
+++ b/.github/actions/scripts/render-mem-summary.sh
@@ -5,7 +5,10 @@
 # (invoked with --mem-limit-mib) and appends a single Markdown table to
 # $GITHUB_STEP_SUMMARY (or prints to stdout if that variable is unset).
 #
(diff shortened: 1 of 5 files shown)


renanmagagnin:
Replaced by: https://github.com/awslabs/mountpoint-s3/pull/1940

PR #1939, opened 2026-08-21
Enforce `--read-only` in S3Filesystem

yerzhan7:
A read-only mount was enforced entirely by the kernel. That works when Mountpoint performs the mount syscall, but not for a FUSE file descriptor mount point, where the caller mounts and `--read-only` was rejected outright — discarding the only signal that could tell the file system the mount is read-only, since nothing about an fd exposes the caller's `MS_RDONLY`.

Changes:

- Enforce read-only in the file system: `open` requesting write access, `setattr`, `mknod`, `mkdir`, `rmdir`, `unlink` and `rename` now fail with the new `EROFS` instead of relying on the kernel to have refused them. This matches what the kernel does on a read-only mount.
- Accept `--read-only` with a FUSE file descriptor mount point
- `MountpointConfig::create_fuse_session` validates that `FuseOptions::read_only` and `S3FilesystemConfig::read_only` agree
- `cli.rs` sets `filesystem_config.read_only` from `--read-only`
- Unrelated drive-by in a file this touches: Fix `gid` in `examples/mount_from_config.rs` 

Yes. `--read-only` with a FUSE file descriptor mount point was previously rejected and now succeeds. Embedders using `MountpointConfig` must set `read_only` on both `FuseOptions` and `S3FilesystemConfig`. No existing mount-s3 invocation changes behavior.

Yes

---


mountpoint-s3-fs/src/config.rs
```diff
@@ -42,6 +41,19 @@ impl MountpointConfig {
         self
     }
 
+    /// Check that the configurations we were given agree with each other.
+    fn validate(&self) -> anyhow::Result<()> {
```
passaro: nit: [parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/). Can we move the check to the constructor?

mountpoint-s3-fs/src/fs.rs
```diff
@@ -320,6 +330,9 @@ where
             mtime,
             size
         );
+        if let Some(err) = self.read_only_error() {
+            return Err(err);
+        }
```
passaro, line 335: We could simplify with something like `self.read_only_error()?;` (and a better method name, I guess).

mountpoint-s3/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
...
+* `--read-only` may now be used with a FUSE file descriptor mount point (`/dev/fd/N`). Previously this combination was rejected, since Mountpoint does not perform the `mount` syscall in that case and so cannot apply the option itself. It is now accepted as a statement that the caller performed the mount read-only, which Mountpoint enforces for the operations it serves. ([#1939](https://github.com/awslabs/mountpoint-s3/pull/1939) by @yerzhan7)
+* Mountpoint now enforces `--read-only` itself, refusing file system operations that would modify the mount with `EROFS`. Previously a read-only mount relied entirely on the kernel to reject them, which it still does for a directory mount point; this makes the behaviour the same when the mount point is a FUSE file descriptor and Mountpoint did not perform the mount. ([#1939](https://github.com/awslabs/mountpoint-s3/pull/1939) by @yerzhan7)
```
passaro: I don't think we need the duplicate entry. At most we could mention the error code in the previous one.

mountpoint-s3/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
...
 * Change memory pool metrics in logs. `pool.reserved_bytes` is replaced by `pool.acquired_bytes` and `pool.bytes_in_use`, `pool.allocated_bytes` and `pool.allocate_latency_us` are new, and the `size` dimension on `pool.allocated_pages`, `pool.empty_pages`, `pool.slack_bytes` and `pool.trim_pages` is renamed to `buffer_size`. ([#1936](https://github.com/awslabs/mountpoint-s3/pull/1936))
+* `--read-only` may now be used with a FUSE file descriptor mount point (`/dev/fd/N`). Previously this combination was rejected, since Mountpoint does not perform the `mount` syscall in that case and so cannot apply the option itself. It is now accepted as a statement that the caller performed the mount read-only, which Mountpoint enforces for the operations it serves. ([#1939](https://github.com/awslabs/mountpoint-s3/pull/1939) by @yerzhan7)
```
passaro: I'd simplify by removing the whole sentence starting with "Previously..."

mountpoint-s3-fs/src/fs.rs
yerzhan7: This tripped `clippy::result_large_err` check previously. That's the reason I did it in this way.

mountpoint-s3-fs/src/config.rs
yerzhan7: Thanks, fixed

mountpoint-s3/CHANGELOG.md
yerzhan7: Done

mountpoint-s3/CHANGELOG.md
yerzhan7: Removed

mountpoint-s3-fs/src/fs.rs
passaro: That's surprising. I'd look into it. But not blocking for the PR.

passaro (approved):

PR #1947, opened 2026-08-26
Replace bincode with wincode in disk data cache

Priyankakarumuru1:
Replace unmaintained bincode serialization library with wincode (a bincode-compatible, actively maintained library) for disk cache header serialization/deserialization in `mountpoint-s3-fs`. bincode has been permanently marked as unmaintained ([RUSTSEC-2025-0141](https://osv.dev/vulnerability/RUSTSEC-2025-0141)). 
  
Implementation:
  
- Read: Header deserialized directly from the file reader via ReadAdapter. Data read directly into pool buffer.
- Write: Header serialized directly to file via WriteAdapter (no intermediate allocation, same pattern as bincode).
- Added `write_cache_block` and `write_file` benchmarks.

Benchmark results:
- `read_cache_block`: wincode 99.3 µs vs bincode 103.6 µs (faster ✅)
- `write_cache_block`: wincode 4.02 ms vs bincode 4.02 ms (identical)
  
No performance regression. Read path is slightly faster.

Note: `mountpoint-s3-fuser `still uses bincode 1.3.1 - will be addressed in a follow-up. The advisory suppression can be reverted once both are migrated.
 
No user-facing changes.

No

---

### DIFF
--- a/mountpoint-s3-fs/Cargo.toml
+++ b/mountpoint-s3-fs/Cargo.toml
@@ -17,7 +17,7 @@ async-lock = "3.4.2"
 async-stream = "0.3.6"
 async-trait = "0.1.89"
 base64ct = "1.8.3"
-wincode = { version = "0.6", features = ["derive"] }
+wincode = { version = "0.6.1", features = ["derive"] }
 bitflags = "2.13.0"
 bytes = { version = "1.12.0", features = ["serde"] }
 clap = { version = "4.6.1", features = ["derive"] }
--- a/mountpoint-s3-fs/Cargo.toml
+++ b/mountpoint-s3-fs/Cargo.toml
@@ -17,7 +17,7 @@ async-lock = "3.4.2"
 async-stream = "0.3.6"
 async-trait = "0.1.89"
 base64ct = "1.8.3"
-bincode = { version = "2.0.1", features = ["std"] }
+wincode = { version = "0.6", features = ["derive"] }
 bitflags = "2.13.0"
 bytes = { version = "1.12.0", features = ["serde"] }
 clap = { version = "4.6.1", features = ["derive"] }
--- a/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
+++ b/mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
@@ -7,9 +7,6 @@ use std::path::{Path, PathBuf};
 use std::time::Instant;
 
 use async_trait::async_trait;
-use bincode::config::{Configuration, Fixint, Limit, LittleEndian};
-use bincode::error::{DecodeError, EncodeError};
-use bincode::{Decode, Encode};
 use bytes::Bytes;
 use linked_hash_map::LinkedHashMap;
 use mountpoint_s3_client::checksums::crc32c::{self, Crc32c};
@@ -32,7 +29,7 @@ use crate::sync::Mutex;
 use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheResult};
 
 /// Disk and file-layout versioning.
-const CACHE_VERSION: &str = "V2";
+const CACHE_VERSION: &str = "V3";
 
 /// Index where hashed directory names for the cache are split to avoid FS-specific limits.
 const HASHED_DIR_SPLIT_INDEX: usize = 2;
@@ -79,7 +76,7 @@ impl Default for CacheLimit {
 ///
 /// It should be written alongside the block's data
 /// and used to verify it contains the correct contents to avoid blocks being mixed up.
-#[derive(Encode, Decode, Debug)]
+#[derive(wincode::SchemaWrite, wincode::SchemaRead, Debug)]
 struct DiskBlockHeader {
     block_idx: BlockIndex,
     block_offset: u64,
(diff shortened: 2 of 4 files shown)


mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -32,7 +29,7 @@ use crate::sync::Mutex;
 use super::{BlockIndex, ChecksummedBytes, DataCache, DataCacheResult};
 
 /// Disk and file-layout versioning.
-const CACHE_VERSION: &str = "V2";
+const CACHE_VERSION: &str = "V3";
```
passaro: Did the cache format actually change?

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -79,7 +76,7 @@ impl Default for CacheLimit {
 ///
 /// It should be written alongside the block's data
 /// and used to verify it contains the correct contents to avoid blocks being mixed up.
-#[derive(Encode, Decode, Debug)]
+#[derive(wincode::SchemaWrite, wincode::SchemaRead, Debug)]
```
passaro: nit: why not `use`?

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -90,18 +87,6 @@ struct DiskBlockHeader {
...
-/// - S3 key should always be less than or equal to 1024
-/// - Allow 1024 for ETag, even though its always much smaller
-/// - Integers should be up to 8 bytes each
-const BINCODE_HEADER_MAX_SIZE: usize = 10000;
-
-/// Binary encoding configuration for the block header.
-const BINCODE_CONFIG: Configuration<LittleEndian, Fixint, Limit<BINCODE_HEADER_MAX_SIZE>> = bincode::config::standard()
-    .with_fixed_int_encoding()
-    .with_limit::<BINCODE_HEADER_MAX_SIZE>();
-
```
passaro, line 104: Are these not needed with wincode? Why?

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -124,10 +109,8 @@ enum DiskBlockAccessError {
 enum DiskBlockReadWriteError {
     #[error("Invalid block length: {0}")]
     InvalidBlockLength(u64),
-    #[error("Error decoding the block: {0}")]
-    DecodeError(DecodeError),
-    #[error("Error encoding the block: {0}")]
-    EncodeError(EncodeError),
+    #[error("Error with serialization: {0}")]
+    SerializationError(wincode::Error),
```
passaro: "serialization" is a more generic and less user-friendly error. Can we reverse?

mountpoint-s3-fs/Cargo.toml
```diff
@@ -17,7 +17,7 @@ async-lock = "3.4.2"
 async-stream = "0.3.6"
 async-trait = "0.1.89"
 base64ct = "1.8.3"
-bincode = { version = "2.0.1", features = ["std"] }
+wincode = { version = "0.6", features = ["derive"] }
```
passaro: "0.6.1"?

passaro (changes requested):

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
Priyankakarumuru1: wincode has these built-in by default - fixed-width integers and size limits. [See](https://docs.rs/wincode/0.6.1/wincode/#compatibility) "will fail to serialize or deserialize large dynamic data structures by default, but this can be configured"
Happy to set an explicit tighter limit if you'd prefer to match the old 10KB.

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
Priyankakarumuru1: No actually. Will revert back to V2. Thanks.

mountpoint-s3-fs/Cargo.toml
Priyankakarumuru1: Yes good catch. Thanks!

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
passaro: Yes, we should be explicit. Also, if I read correctly here: [with_preallocation_size_limit](https://docs.rs/wincode/0.6.1/wincode/config/struct.Configuration.html#method.with_preallocation_size_limit), the limit is for each sequence, so it could be a smaller value.

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -265,43 +251,47 @@ impl DiskBlock {
         pool: &PagedPool,
         cursor_id: Option<CursorId>,
     ) -> Result<Self, DiskBlockReadWriteError> {
-        let header: DiskBlockHeader = bincode::decode_from_std_read(reader, BINCODE_CONFIG)?;
+        // Deserialize header directly from the file reader
+        let header: DiskBlockHeader = wincode::deserialize_from(wincode::io::std_read::ReadAdapter::new(&mut *reader))?;
```
passaro: Here and in other places: we typically import types rather than fully qualify them. Exceptions are mostly free-standing functions, just like `wincode::serialize_from`.

passaro (changes requested):

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
Priyankakarumuru1: Got it.

mountpoint-s3-fs/src/data_cache/disk_data_cache.rs
```diff
@@ -73,6 +76,17 @@ impl Default for CacheLimit {
     }
 }
 
+/// Per-sequence preallocation limit (in bytes) for dynamic fields (Strings) in
+/// the header. wincode's `with_preallocation_size_limit` applies to each
+/// sequence individually. The largest dynamic field is the S3 key, which AWS
+/// caps at 1024 bytes; ETags are much smaller. So 1024 is a tight, safe upper
+/// bound that rejects a corrupted length prefix as early as possible.
+const HEADER_STRING_SIZE_LIMIT: usize = 1024;
```
passaro, line 84: This makes sense, but we should test with a 1024 key. Could you add a test if we don't already?

Priyankakarumuru1: Yes, I'll add a test. Thanks.

passaro (approved):

PR #1948, opened 2026-08-26
Document overwrite decision timing and last-write-wins

renanmagagnin:
Improves clarity of two points in the "Writes" section of `doc/SEMANTICS.md`. For `--allow-overwrite`, the decision that a file may be overwritten is made at `open` and is not re-checked before the upload completes: the atomic upload path sends no conditional headers, so if another S3 client writes to the same key while the overwrite is in flight, the last upload to complete wins. For contrast, `--incremental-upload` appends are conditional on the object's ETag as of `open`, so a concurrent modification makes subsequent writes fail instead.

No.

No.

---

### DIFF
--- a/doc/SEMANTICS.md
+++ b/doc/SEMANTICS.md
@@ -207,6 +207,7 @@ but with some limitations:
   * The existing file must be opened in truncate mode (`O_TRUNC`).
   * You cannot overwrite files that are currently being read or renamed.
   * The upload to S3 starts as soon as Mountpoint receives the first `write` request and cannot be cancelled. The data is uploaded to S3 in fixed-size parts (controlled by `--write-part-size`).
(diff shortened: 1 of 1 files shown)


doc/SEMANTICS.md
```diff
@@ -207,13 +207,15 @@ but with some limitations:
   * The existing file must be opened in truncate mode (`O_TRUNC`).
   * You cannot overwrite files that are currently being read or renamed.
   * The upload to S3 starts as soon as Mountpoint receives the first `write` request and cannot be cancelled. The data is uploaded to S3 in fixed-size parts (controlled by `--write-part-size`).
+  * Whether the file can be overwritten is decided when the file is opened, and is not re-checked before the upload completes. If another client writes to the same key while the overwrite is in progress, the last upload to complete wins, replacing the data written by the other.
```
muddyfish, line 210: Might want to make it clear that 'another client' means another MP instance, and we locally block this. (we do, right?)

renanmagagnin: Discussed offline: will change to wording to make it clear that "another client" can be an external one or another Mountpoint instance

PR #1954, opened 2026-09-04
Fix directory lookup after empty S3 list pages

cayoub-oai:
S3 can return empty `ListObjectsV2` pages with a continuation token. Directory lookup currently treats these as evidence that the directory is missing, causing false `ENOENT` errors or exposing a file that should be shadowed by a directory.

Continue listing while both objects and common prefixes are empty and a continuation token is present. Stop once the directory is found or the listing is exhausted.

As part of testing, reproduced the failure with unpatched Mountpoint against real S3 and verified that the patched binary resolves the directory and reads a nested file matching its direct S3 download.

No breaking changes. Fixes incorrect lookup results after empty, truncated pages. Lookups may take longer because they now follow pagination instead of returning an incorrect result early.

Yes.

---

### DIFF
--- a/mountpoint-s3-fs/src/superblock.rs
+++ b/mountpoint-s3-fs/src/superblock.rs
@@ -1559,10 +1559,34 @@ impl<OC: ObjectClient + Send + Sync> SuperblockInner<OC> {
             .client
             .head_object(&self.s3_path.bucket, object_key, &head_object_params)
             .fuse();
-        let mut dir_lookup = self
-            .client
-            .list_objects(&self.s3_path.bucket, None, "/", 1, directory_prefix)
-            .fuse();
+        let dir_lookup = async {
+            let mut continuation_token = None;
+            loop {
+                let result = self
+                    .client
+                    .list_objects(
+                        &self.s3_path.bucket,
+                        continuation_token.as_deref(),
+                        "/",
+                        1,
+                        directory_prefix,
+                    )
+                    .await;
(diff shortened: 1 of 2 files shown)


mountpoint-s3-fs/src/superblock.rs
```diff
@@ -1559,10 +1559,34 @@ impl<OC: ObjectClient + Send + Sync> SuperblockInner<OC> {
             .client
             .head_object(&self.s3_path.bucket, object_key, &head_object_params)
             .fuse();
-        let mut dir_lookup = self
-            .client
-            .list_objects(&self.s3_path.bucket, None, "/", 1, directory_prefix)
-            .fuse();
+        let dir_lookup = async {
```
passaro, line 1562: I'd refactor the async block into a helper function on client, bucket, directory. The rustdoc on the function can explain why the loop is needed.

passaro (changes requested):
Thanks for opening the pull request. The change looks good, I've got only a minor suggestion. Could you also add an entry to the changelog in `mountpoint-s3-fs`?

mountpoint-s3-fs/src/superblock.rs
cayoub-oai: Updated!

passaro:
Rebased to pick up some CI changes.

passaro (approved):
Looks good, thanks!

